PR #3, opened 2023-07-24
Initial Splunk code donation

breedx-splk:
I merged from the original repo to retain full revision history.  Replaced all the references to splunk and picked a new package namespace: `io.opentelemetry.android`. Confirmed that an `.aar` is built, although the name is probably incorrect. 

There's definitely a lot more to do after this. 😁


LikeTheSalad (approved):
Nice! I didn't take a detailed look at everything because we're going to work on changing a lot after this is merged anyway so... ✅

breedx-splk:
Bumping to try and get status to clear.

breedx-splk:
once more with gusto

breedx-splk:
Ok, have opened https://github.com/open-telemetry/community/issues/1621 as requested by GC.

PR #14, opened 2023-08-02
Improve output artifact names

### LINKED ISSUE #4: Fix output artifact name
The current build outputs are:

* instrumentation-debug.aar
* instrumentation-release.aar

We probably want these to be named more consistently with the other projects. So these should be named:

* `opentelemetry-android-release.aar`
* `opentelemetry-android-debug.aar`

breedx-splk:
Resolves #4. 

I chose the filename prefix to be `opentelemetry-android` which ends up generating `opentelemetry-android-debug.aar` and `opentelemetry-android-release.aar`. I'm not sure why there are both debug and release variants, and if we'll want to publish both? For now, at least it's an improvement on `instrumentation-debug.aar` 🙃 

I also added a `CONTRIBUTING.md` and made a few `README.md` additions.

### DIFF
--- a/instrumentation/build.gradle.kts
+++ b/instrumentation/build.gradle.kts
@@ -28,18 +28,16 @@ android {
         }
     }
 
-    libraryVariants.all {
-        outputs
-            .map { it as com.android.build.gradle.internal.api.BaseVariantOutputImpl }
-            .all { variant ->
-                variant.outputs.all {
-                    if (variant.outputFileName.endsWith(".aar")) {
-                        variant.outputFileName = "opentelemetry-android-${variant.name}.aar"
-                    }
-                    false
+    androidComponents {
+        onVariants {
+            if (it.buildType == "release") { // The one we choose to release
+                project.tasks.register("createReleaseBuild", Copy::class) {
+                    from(it.artifacts.get(com.android.build.api.artifact.SingleArtifact.AAR))
+                    into(project.layout.buildDirectory.dir("outputs/aar"))
+                    rename(".+", "opentelemetry-android.aar")
                 }
-                false
             }
+        }
     }
 
     compileOptions {
--- a/instrumentation/build.gradle.kts
+++ b/instrumentation/build.gradle.kts
@@ -40,6 +40,12 @@ android {
         }
     }
 
+    project.afterEvaluate {
+        tasks.named("assembleRelease") {
+            finalizedBy("createReleaseBuild")
+        }
+    }
+
     compileOptions {
         isCoreLibraryDesugaringEnabled = true
 
--- a/instrumentation/build.gradle.kts
+++ b/instrumentation/build.gradle.kts
@@ -1,5 +1,4 @@
 plugins {
-    id("com.android.library")
     id("otel.android-library-conventions")
     id("otel.errorprone-conventions")
 }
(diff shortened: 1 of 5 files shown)


LikeTheSalad:
> I'm not sure why there are both debug and release variants, and if we'll want to publish both?

That's fine, the android plugin always creates those two build variants by default. We should only publish the `release` one though.

instrumentation/build.gradle.kts
```diff
@@ -29,6 +28,20 @@ android {
         }
     }
 
+    libraryVariants.all {
+        outputs
+            .map { it as com.android.build.gradle.internal.api.BaseVariantOutputImpl }
```
LikeTheSalad, line 33: While this works, I'm a bit concerned that it's using an internal type, which itself uses a deprecated API. So I think we should instead use the new `androidComponents` API, which has a few handy utilities for build-manipulation use cases.

For this case, I think we can even get an `aar` file without any suffixes (debug or release) to make it easier to tell what's the variant we plan to release. We could do it like so:

```kotlin
androidComponents {
    onVariants {
        if (it.buildType == "release") { // The one we choose to release
            project.tasks.register("createReleaseBuild", Copy::class) {
                from(it.artifacts.get(com.android.build.api.artifact.SingleArtifact.AAR))
                into(project.layout.buildDirectory.dir("outputs/aar"))

                rename(".+", "opentelemetry-android.aar")
            }
        }
    }
}
```

That way, whenever we need to create a release build, we can run the task `createReleaseBuild` (or any other name we come up with) which will generate the chosen variant and set the proper file name to it. We can also define any other folder to place it, it doesn't necessarily have to be "build/outputs/aar".

breedx-splk: Thanks, my approach was definitely a hack job and it feels gross. I definitely struggle with this gradle magic incantation stuff. I couldn't find any reasonable documentation whatsoever about the idiomatic way to rename/customize android library aars using the gradle plugin. 

I tried your approach and it seems to work just fine and I think we should use it. I wish it wasn't a new task name and I wish it was just part of the normal `aseemble` process.

LikeTheSalad: I agree, Gradle can be very unintuitive sometimes 😅 

> I wish it was just part of the normal aseemble process

Well, I just tried a couple of things and it seems like this could make it possible:

```kotlin
androidComponents {
// ...
}

project.afterEvaluate {
    tasks.named("assembleRelease") {
        finalizedBy("createReleaseBuild")
    }
}
```

Don't know if it's too "hacky" though... So anyway, I'll leave it here in case you want to add it 😬

LikeTheSalad (approved):

instrumentation/build.gradle.kts
```diff
@@ -29,6 +28,18 @@ android {
         }
     }
 
+    androidComponents {
+        onVariants {
+            if (it.buildType == "release") { // The one we choose to release
+                project.tasks.register("createReleaseBuild", Copy::class) {
```
surbhimaheshwari, line 34: Why can't we use ./gradlew assembleRelease to create the release variant of app instead of creating custom task

instrumentation/build.gradle.kts
breedx-splk: > I agree, Gradle can be very unintuitive sometimes 😅

You are being very generous. I'd usually say something much much worse. 🙃 

In any case, this is great. Doesn't feel hacky to me. Thank you!

instrumentation/build.gradle.kts
breedx-splk: Yeah, @LikeTheSalad added a snippet so we can tag onto `assembleRelease` and therefore `assemble`...so I've included that. Thanks!

PR #19, opened 2023-08-11
Configuring android lib publishing

LikeTheSalad:
Preparing for the first release!

These take care of configuring the project's release artifacts in order to:

- Generate the AAR file based on the "release" android variant using the latest [android plugin tools](https://developer.android.com/build/publish-library/configure-pub-variants#single-pub-var) to do so.
- Generate the javadoc and sources artifacts (only during a release).
- Signing all artifacts (only during a release).


buildSrc/src/main/java/io/opentelemetry/android/PublishPlugin.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.Locale;
+
+import kotlin.Unit;
+
+/**
+ * This plugin takes care of configuring a library project by creating a maven publication that
+ * contains the library artifact and (when the "-Prelease=true" param is present in the gradle command) it will also include
+ * the library's javadoc, sources artifacts and all the artifacts will be signed.
+ */
+public class PublishPlugin implements Plugin<Project> {
```
breedx-splk, line 21: How would you feel about sticking with kotlin for the gradle/build tooling? Might be a little unfortunate to have both java and kotlin hanging around in `buildSrc`.

buildSrc/src/main/java/io/opentelemetry/android/PublishPlugin.java
```diff
@@ -0,0 +1,112 @@
...
+            if (isARelease()) {
+                signPublication(publication);
+            }
+        });
+    }
+
+    private void configurePom(MavenPublication publication) {
+        publication.pom(mavenPom -> {
+            String repoUrl = "https://github.com/open-telemetry/opentelemetry-android";
+            mavenPom.getName().set(project.getGroup() + ":" + project.getName());
```
breedx-splk, line 67: In the [instrumentation repo](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/conventions/src/main/kotlin/otel.publish-conventions.gradle.kts#L42) the name is the human-readable name, which I'd prefer to just be something like "OpenTelemetry for Android" or just "OpenTelemetry Android".

buildSrc/src/main/kotlin/otel.android-library-conventions.gradle.kts
```diff
@@ -1,52 +0,0 @@
-import org.gradle.api.publish.maven.MavenPublication
-import java.net.URI
-
-plugins {
```
breedx-splk, line 4: I'm totally ok removing this for now, but when we end up having more than one project module we may want to bring it back. In the Splunk project it existed to help several modules not repeat themselves.

breedx-splk (approved):
Looking good to me, just hand a couple small ideas/comments. Thanks!

buildSrc/src/main/java/io/opentelemetry/android/PublishPlugin.java
LikeTheSalad: I don't have a strong opinion about it so I'm fine with using kotlin to keep consistency 👍 I'll add the changes.

buildSrc/src/main/java/io/opentelemetry/android/PublishPlugin.java
LikeTheSalad: Good call! I went with "OpenTelemetry Android".

buildSrc/src/main/kotlin/otel.android-library-conventions.gradle.kts
LikeTheSalad: That's a good point, at first I went with a plugin approach because to me it's a bit cleaner than using scripts and it's also more organized for when there's a lot of code needed for compilation and publishing processes that can be reused across different projects. However, since there are not too many things needed for now, and taking into account the consistency with previous gradle-related code as mentioned above, I changed it back to the script format and reintroduced the common android configs as well.

Probably in the future, if we need to add more compilation tools, or even to support other kinds of projects, not only android lib ones, we might want to consider reorganizing these tools in order to avoid adding too many different responsibilities into a single script... I think we can create one that takes care of configuring android-specific parameters and use the root `build.gradle.kts` script to automatically apply it to all android projects! That way we can still avoid repeating ourselves while keeping the scripts small and specific to a single task.

Issue #9, opened 2023-07-26
Publish versioned aar artifacts


breedx-splk:
The primary output of the build process is an `.aar` file. This should be versioned (initial version should be 0.1.0) with semver and a github release should be created. The github release should contain the artifact.

The artifact should be published to maven.

tonzhan2:
I'm inclined towards publish to maven

breedx-splk:
I hadn't thought if this until now, but we should probably start by getting snapshot builds publishing first. This would be a workflow that is triggered on merge to `main` branch and publishes with the SNAPSHOT qualifier, which we may not yet support.

@LikeTheSalad what do you think?

LikeTheSalad:
Creating snapshot builds automatically sounds good to me 👍 - after we have the maven central configuration set up we can start doing so right away.

LikeTheSalad:
I think this is done now

PR #27, opened 2023-08-11
Add workflow to run build when main is merged.

breedx-splk:
This should also automatically create an issue if/when the build step fails. 

This is also a likely candidate place to publish a snapshot build, so I added a TODO. Relates to #9.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -22,31 +22,10 @@ jobs:
       - name: run gradle
         run: ./gradlew check
 
-  issue:
-    name: Open issue on failure
-    needs: [ build ]
-    runs-on: ubuntu-20.04
-    permissions:
-      issues: write
+  workflow-notification:
+    needs:
+      - analyze
     if: always()
-    steps:
-      # run this action to get workflow conclusion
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -0,0 +1,52 @@
...
+    name: Open issue on failure
+    needs: [ build ]
+    runs-on: ubuntu-20.04
+    permissions:
+      issues: write
+    if: always()
+    steps:
+      # run this action to get workflow conclusion
+      # You can get conclusion by env (env.WORKFLOW_CONCLUSION)
+      - uses: technote-space/workflow-conclusion-action@v3.0.3
```
trask, line 35: have you seen https://github.com/open-telemetry/opentelemetry-java/blob/main/.github/workflows/reusable-workflow-notification.yml? (we use a copy of this across the other java repos)

LikeTheSalad (approved):

.github/workflows/build.yaml
breedx-splk: Oooh, I hadn't seen that. I'll copy it over for consistency then. Thank you!

PR #36, opened 2023-08-22
Fix build breakage

breedx-splk:
Included the reusable workflow but with the wrong `needs`.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -25,7 +25,7 @@ jobs:
   workflow-notification:
     needs:
       - build
-    if: always()
+    if: failure() && github.run_attempt == 1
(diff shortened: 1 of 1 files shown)


trask (approved):

breedx-splk:
I pushed a commit because I think we only want to notify on breakage.

trask:
it depends on what you want

java-instrumentation repo only uses this issue opening workflow on "daily" builds, and in that case the successes are useful to pass on to the workflow so that it will auto-close the issues on later success

java repo doesn't haev any "daily" builds and doesn't use the issue opening workflow at all (recent change)

breedx-splk:
Thanks, yeah we won't have daily builds any time soon in android, so I think just opening an issue when the main build fails is the move.

LikeTheSalad (approved):

PR #37, opened 2023-08-22
Adding @jack-berg and @trask as codeowners

breedx-splk:
We talked about this in the last SIG and @jack-berg and @trask agreed to be `CODEOWNERS` for a period to help get the project going and reduce time-to-merge across timezones.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,6 +2,4 @@
 # This file controls who is tagged for review for any given pull request.
 
  # For anything not explicitly taken by someone else:
-* @open-telemetry/android-maintainers
-* @trask
-* @jack-berg
+* @open-telemetry/android-approvers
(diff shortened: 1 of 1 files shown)


jack-berg:
I've went ahead and created @open-telemetry/android-approvers 

Skipped making it when I initialized the repo because we didn't have any approvers.

.github/CODEOWNERS
```diff
@@ -3,3 +3,5 @@
 
  # For anything not explicitly taken by someone else:
 * @open-telemetry/android-maintainers
+* @trask
+* @jack-berg
```
jack-berg, line 7:
```suggestion
* @open-telemetry/android-approvers
```

trask:
@breedx-splk probably worth adding something like https://github.com/open-telemetry/opentelemetry-java-instrumentation#contributing, which is what we PR against when adding/removing approvers/etc

breedx-splk:
> @breedx-splk probably worth adding something like https://github.com/open-telemetry/opentelemetry-java-instrumentation#contributing, which is what we PR against when adding/removing approvers/etc

Yeah agreed. I'll add that shortly! Thanks!

trask (approved):

jack-berg (approved):

PR #41, opened 2023-08-23
Add codeql to run nightly

breedx-splk:
Getting ahead of the security audit by adding nightly codeql like the other otel repos. Copied verbatim from the `opentelemetry-java` repo.

### DIFF
--- a/.github/workflows/codeql-daily.yml
+++ b/.github/workflows/codeql-daily.yml
@@ -39,5 +39,7 @@ jobs:
     # open an issue on failure because it can be easy to miss CI failure notifications
     needs:
       - analyze
-    if: failure() && github.run_attempt == 1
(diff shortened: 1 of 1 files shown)


trask:
I personally prefer the -instrumentation automation which won't spam you with opening a new issue every day if you get into failure state (and will auto-close the issue if it's a sporadic failure and the next one or a re-run succeeds)

breedx-splk:
> I personally prefer the -instrumentation automation which won't spam you with opening a new issue every day if you get into failure state (and will auto-close the issue if it's a sporadic failure and the next one or a re-run succeeds)

Agreed, that's nice....thanks for the tip. Updated it.

LikeTheSalad (approved):

PR #48, opened 2023-08-25
Corrections for nullaway and errorprone

breedx-splk:
I tried #46 locally and it already pointed out a few issues, so I tried to get ahead of that and clean those up here.


LikeTheSalad (approved):
Thank you for taking the time! It'll make my PR more straightforward 👍

PR #50, opened 2023-08-25
Exclude mangled package from gradle.properties

breedx-splk:
Over in #23 there is a lookup warning for a nonsensical maven package, which is really just a property sourced from gradle.properties and put into an android resource value. Let's try excluding it.

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -22,7 +22,6 @@
       // somehow renovate gets confused by the android property in gradle.properties,
       // so let's just exclude it and hopefully clean up the dashboard
       "matchPackageNames": ["string:rum.version"],
-      "matchUpdateTypes": ["major", "minor"],
       "enabled": false
     }
   ]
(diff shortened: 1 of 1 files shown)


renovate.json5
```diff
@@ -17,6 +17,13 @@
       "matchPackageNames": ["androidx.browser:browser"],
       "matchUpdateTypes": ["major", "minor"],
       "enabled": false
+    },
+    {
+      // somehow renovate gets confused by the android property in gradle.properties,
+      // so let's just exclude it and hopefully clean up the dashboard
+      "matchPackageNames": ["string:rum.version"],
+      "matchUpdateTypes": ["major", "minor"],
```
trask, line 25: don't need this if blanket disabling
```suggestion
```

trask (approved):

LikeTheSalad (approved):

renovate.json5
breedx-splk: Yeah clearly this is all very new to me. :) Thank you.

PR #55, opened 2023-08-29
Publish snapshots with every merge to main

breedx-splk:


.github/workflows/build.yaml
```diff
@@ -18,9 +18,16 @@ jobs:
         with:
           distribution: temurin
           java-version: 17.0.6
-        # TODO: Build and publish a snapshot instead!
-      - name: run gradle
+      - name: run gradle check
         run: ./gradlew check
+      - name: publish snapshot
+        # final=false here means SNAPSHOT
+        run: ./gradlew publishMavenPublicationToSonatypeRepository -Prelease=true -Pfinal=false
```
trask, line 25: is this different from other Java repos due to something android specific?

LikeTheSalad (approved):
LGTM

.github/workflows/build.yaml
LikeTheSalad: It's mostly the same, we also use the same [release plugin](https://github.com/gradle-nexus/publish-plugin). Though we added these 2 flags, The `release` one is disabled by default, which means that only the main, unsigned `.aar` file will be built for dev purposes, whereas when it's enabled, the sources and javadoc artifacts will also be built and everything will be signed which makes the compilation take a bit longer so I think it's better to avoid that during development. And the `final` one is there to append SNAPSHOT to the version when disabled which is the default behaviour.

I'm not sure if that answers your question, so please let me know if there are any other details you'd like to know about it!

breedx-splk: @trask In the interest of progress, I'm going to merge this. I would like to better understand what the differences are that you are concerned about, so maybe we should open an issue? We do want consistency with the other java repos where it makes sense....

trask: in the other repos we use `publishToSonatype`, i'm not sure what the differences is between that and `publishMavenPublicationToSonatypeRepository`

also we don't seem to need `-Pfinal=false` in the other repos

trask: opened #58

trask: > In the interest of progress, I'm going to merge this

(btw I always support this)

PR #69, opened 2023-09-05
Update gradle-build-action to mitigate CVE-2023-30853

breedx-splk:


trask (commented):
my understanding is that `v2` should be picking up the latest 2.x version already

LikeTheSalad:
> my understanding is that v2 should be picking up the latest 2.x version already

They seem to mention it [here](https://github.com/advisories/GHSA-h3qr-39j9-4r5v) as well. Though I'm not sure if we've gotten any recent build warnings/errors that suggest otherwise? @breedx-splk

breedx-splk:
going to clear the warning and close this now.

Issue #56, opened 2023-08-31
Can't configure ContextPropagators in OpenTelemetry instance


GerardPaligot:
**Context**
Because `OpenTelemetry` instance is created internally inside `OpenTelemetryRumBuilder` class, we can't add custom configuration in `OpenTelemetry` object but it could be useful for some cases.

e.g.: If I want to configure `ContextPropagators` to connect my traces with other external services, I can't with the actual codebase.

**Suggestions**
1. Expose OpenTelemetry builder to let the developer configure what he/she want before the Rum configuration.
2. or add a new function in `OpenTelemetryRumBuilder` to configure a `ContextPropagators`.

breedx-splk:
Hi @GerardPaligot. While yes, the traditional use case was to have OpenTelemetry RUM create the otel sdk for you (created internally, as you pointed out), we did add method to `OpenTelemetryRum` that you can pass your own SDK instance to. Check out [OpenTelemetryRum.builder(app, sdk)](https://github.com/open-telemetry/opentelemetry-android/blob/main/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRum.java#L49) and see if that can work for you.

Be warned, that's a fairly big hammer and you'll have to make sure that you get the rest of the sdk wired up for your needs...but it sounds that's the option 1 you proposed.

The design of this is still flexible, so please let us know how it goes for you so we can make informed decisions about the approach. Thanks!

GerardPaligot:
Yes, I used this function to add lifecycle instrumentation in my RUM OTeL object:

```kotlin
internal fun OpenTelemetryRumBuilder.addLifecycleInstrumentation(
    screenTracker: VisibleScreenTracker,
    startupTimer: AppStartupTimer = AppStartupTimer()
): OpenTelemetryRumBuilder {
    return this.addInstrumentation { app ->
        AndroidLifecycleInstrumentation.builder()
            .setVisibleScreenTracker(screenTracker)
            .setStartupTimer(startupTimer)
            .setTracerCustomizer { tracer ->
                return@setTracerCustomizer Tracer { spanName ->
                    val component = if (spanName == RumAttributeKeys.APP_START_SPAN_NAME) {
                        AttributeKeys.Features.COMPONENT_APPSTART
                    } else {
                        AttributeKeys.Features.COMPONENT_UI
                    }
                    tracer.spanBuilder(spanName)
                        .setAttribute(AttributeKeys.COMPONENT_KEY, component)
[...]

PR #71, opened 2023-09-06
Allow user to customize propagators

breedx-splk:
Relates to #56.

This adds two new methods to the `OpenTelemetryRumBuilder`: 

* `setPropagator(TextMapPropagator)` - replaces all configured propagators with the given one
* `addPropagator(TextMapPropagator)` - adds the propagator to the collection of configured propagators to use in the sdk

This also defaults the propagators to the same ones as the upstream sdk -- `W3CTraceContextPropagator` and `W3CBaggagePropagator`.


trask (commented):
does it make sense to align this with `AutoConfiguredOpenTelemetrySdkBuilder` and introduce `addPropagatorCustomizer` instead?

breedx-splk:
> does it make sense to align this with `AutoConfiguredOpenTelemetrySdkBuilder` and introduce `addPropagatorCustomizer` instead?

Oh yeah probably. I wish @GerardPaligot  would also chime in, but I'll take a stab at it to see how that looks.

breedx-splk:
Oh, I'm seeing now that  @GerardPaligot commented in the issue. Well, I hope the change to use a propagator customizer is also suitable. 🤞🏻

LikeTheSalad (approved):
I think both approaches are fine tbh, though I'm always up to keep consistency across similar/related tools whenever possible.

PR #75, opened 2023-09-12
Update to latest otel version

breedx-splk:
Might it be that renovate doesn't grok the `gradle.properties` approach here?


LikeTheSalad:
It seems like it doesn't 😞 https://github.com/renovatebot/renovate/issues/6220

Although it's quite strange because they say they do in their [docs](https://docs.renovatebot.com/modules/manager/gradle/). I'll take a deeper look into this this week to see if there's something we can do to make it work, otherwise, we'd probably have to define these versions somewhere else.

LikeTheSalad (approved):

Issue #58, opened 2023-08-31
in the other repos we use `publishToSonatype`, i'm not sure what the differences is between that and `publishMavenPublicationToSonatypeRepository`


trask:
in the other repos we use `publishToSonatype`, i'm not sure what the differences is between that and `publishMavenPublicationToSonatypeRepository`

also we don't seem to need `-Pfinal=false` in the other repos

_Originally posted by @trask in https://github.com/open-telemetry/opentelemetry-android/pull/55#discussion_r1312174063_

breedx-splk:
I'm also not entirely sure, but I suspect it might be a difference between `io.github.gradle-nexus.publish-plugin` (https://github.com/gradle-nexus/publish-plugin) and  `maven-publish` (https://docs.gradle.org/current/userguide/publishing_maven.html). 

It looks like core and instrumentation use both....I don't understand why you'd want or need both...

LikeTheSalad:
> I'm also not entirely sure, but I suspect it might be a difference between io.github.gradle-nexus.publish-plugin (https://github.com/gradle-nexus/publish-plugin) and maven-publish (https://docs.gradle.org/current/userguide/publishing_maven.html).
> 
> It looks like core and instrumentation use both....I don't understand why you'd want or need both...

That's right they're different but they complement each other, the `maven-publish` plugin is Gradle's official way to set up and publish maven publications from a project, whereas `io.github.gradle-nexus.publish-plugin` takes the publications created by `maven-publish` and uploads them to maven central, automating the whole process of creating sonatype's staging "repo", then closing and releasing it too. The automation of these processes is the value provided by the nexus publish plugin, as they mention in [maven central's doc](https://central.sonatype.org/publish/publish-gradle/#releasing-the-deployment-to-the-central-repository) where they talk about a plugin named "[Gradle Nexus Staging Plugin](https://github.com/Codearte/gradle-nexus-staging-plugin/)" which was later deprecated in favor of the nexus publish plugin.

PR #80, opened 2023-09-14
Update cicd

LikeTheSalad:
Related to #58

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -22,7 +22,7 @@ jobs:
         run: ./gradlew check
       - name: publish snapshot
         # final=false here means SNAPSHOT
-        run: ./gradlew publishToSonatype -Pfinal=false
+        run: ./gradlew publishToSonatype
         env:
           SONATYPE_USER: ${{ secrets.SONATYPE_USER }}
           SONATYPE_KEY: ${{ secrets.SONATYPE_KEY }}
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -21,7 +21,6 @@ jobs:
       - name: run gradle check
         run: ./gradlew check
       - name: publish snapshot
-        # final=false here means SNAPSHOT
         run: ./gradlew publishToSonatype
         env:
           SONATYPE_USER: ${{ secrets.SONATYPE_USER }}
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -22,7 +22,7 @@ jobs:
         run: ./gradlew check
       - name: publish snapshot
         # final=false here means SNAPSHOT
-        run: ./gradlew publishMavenPublicationToSonatypeRepository -Pfinal=false
+        run: ./gradlew publishToSonatype -Pfinal=false
         env:
           SONATYPE_USER: ${{ secrets.SONATYPE_USER }}
           SONATYPE_KEY: ${{ secrets.SONATYPE_KEY }}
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/kotlin/otel.publish-conventions.gradle.kts
```diff
@@ -5,7 +5,7 @@ plugins {
     id("signing")
 }
 
-val isARelease = project.hasProperty("release") && project.property("release") == "true"
+val isARelease = System.getenv("CI") != null
```
LikeTheSalad, line 8: @trask I noticed that this env var check is what is being used [in here](https://github.com/open-telemetry/opentelemetry-java/blob/96c895893c8fb8aeb7e786a050e772d087513f01/buildSrc/src/main/kotlin/otel.publish-conventions.gradle.kts#L60) to tell when the compilation is running in a GH action vs anywhere else, which is really cool and cleaner than having to pass a flag to the command, so we no longer need the `release` param 👍

.github/workflows/build.yaml
```diff
@@ -21,8 +21,7 @@ jobs:
       - name: run gradle check
         run: ./gradlew check
       - name: publish snapshot
-        # final=false here means SNAPSHOT
-        run: ./gradlew publishMavenPublicationToSonatypeRepository -Prelease=true -Pfinal=false
+        run: ./gradlew publishToSonatype
```
LikeTheSalad, line 24: @trask Regarding the differences between `publishMavenPublicationToSonatypeRepository` and `publishToSonatype` - The former is more specific, if you have different publications configured and you don't want to publish them all at once (which is the case in the Elastic repo) then you can be more specific and set the name of the publication (in this case [maven](https://github.com/open-telemetry/opentelemetry-android/blob/main/buildSrc/src/main/kotlin/otel.publish-conventions.gradle.kts#L35)) to publish, but if you are fine with publishing all the configured publications then you can just run `publishToSonatype` which ends up running all the specific-publication tasks.

It was set to the specific task name mostly due to a copy/paste from what we do at Elastic, however, since there's only one publication configured in this project, then there's no need to be specific, so I've just made the changes 👍

.github/workflows/release.yml
```diff
@@ -33,7 +33,7 @@ jobs:
           distribution: temurin
           java-version: 17.0.6
       - name: Build and publish artifacts
-        run: ./gradlew publishMavenPublicationToSonatypeRepository closeAndReleaseSonatypeStagingRepository -Prelease=true -Pfinal=${{ inputs.release_type == 'final' && 'true' || 'false' }}
+        run: ./gradlew publishToSonatype closeAndReleaseSonatypeStagingRepository -Pfinal=${{ inputs.release_type == 'final' && 'true' || 'false' }}
```
LikeTheSalad, line 36: There's just one flag (the `final` one) that I couldn't find a replacement for in the other repos. It essentially tells whether a release version is snapshot or not. When this flag is false or not provided, the version gets `-SNAPSHOT` appended before publishing the artifacts, otherwise the version remains untouched as defined in the `gradle.properties` file.

Having this flag is handy for controlling what type of release we want to publish from within the Gradle script, I couldn't find where in the other repos is the logic that appends (or removes) "-SNAPSHOT" from the version, I'm guessing it might be done in a bash script? @trask @breedx-splk

trask: check out https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/version.gradle.kts

this involves setting up the whole branching release workflow that the other repos use

I think it's fine to stay with what you have here for now, and we can discuss sync'ing release workflows in the future

trask (approved):

PR #81, opened 2023-09-14
Renovate otel version fix

LikeTheSalad:
This is an attempt to make Renovate recognize and update the OTel SDK version based on what was found here: https://github.com/open-telemetry/opentelemetry-android/pull/75


breedx-splk (approved):
Cool, here's hoping that works!

PR #87, opened 2023-09-22
Allow SpanExporters to be added.

### LINKED ISSUE #40: Allow to set custom exporters
I'm thinking about enabling users to provide their own exporters while keeping the rest of the "automatic setup" done in RumInitializer.

breedx-splk:
Resolves #40.

This introduces a new method called `addSpanExporter()` on the OpenTelemetryRumBuilder. This is a convenience method for users who are _not_ bringing their own preconfigured SDK and want to include their own preconfigured exporter for spans.

### DIFF
--- a/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
+++ b/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
@@ -21,6 +21,8 @@ import io.opentelemetry.sdk.metrics.SdkMeterProviderBuilder;
 import io.opentelemetry.sdk.resources.Resource;
 import io.opentelemetry.sdk.trace.SdkTracerProvider;
 import io.opentelemetry.sdk.trace.SdkTracerProviderBuilder;
+import io.opentelemetry.sdk.trace.export.BatchSpanProcessor;
+import io.opentelemetry.sdk.trace.export.SpanExporter;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.function.BiFunction;
@@ -44,6 +46,8 @@ public final class OpenTelemetryRumBuilder {
             meterProviderCustomizers = new ArrayList<>();
     private final List<BiFunction<SdkLoggerProviderBuilder, Application, SdkLoggerProviderBuilder>>
             loggerProviderCustomizers = new ArrayList<>();
+
+    private final List<SpanExporter> spanExporters = new ArrayList<>();
     private final List<Consumer<InstrumentedApplication>> instrumentationInstallers =
             new ArrayList<>();
 
@@ -179,6 +183,19 @@ public final class OpenTelemetryRumBuilder {
         return this;
     }
 
+    /**
+     * This is a convenience method for adding SpanExporters to the underlying OpenTelemetry SDK
+     * instance that is built when the OpenTelemetryRum instance is also built. This may be used as
+     * a simpler alternative to adding a TracerProviderCustomizer via addTracerProviderCustomizer().
+     *
+     * @param exporter An exporter to use when exporting span telemetry
+     * @return this
+     */
+    public OpenTelemetryRumBuilder addSpanExporter(SpanExporter exporter) {
+        spanExporters.add(exporter);
+        return this;
+    }
+
     public SessionId getSessionId() {
         return sessionId;
     }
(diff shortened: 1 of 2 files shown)


trask (commented):
I think there's a lot of benefit to following the pattern(s) already used in https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java where possible

breedx-splk:
> I think there's a lot of benefit to following the pattern(s) already used in https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java where possible

Ok, lemme see about changing away from `addSpanExporter()` to `addSpanExporterCustomizer()`. We clearly need more nouns in here. 🙃 I can't say it feels cleaner, but I guess it's more consistent.

instrumentation/src/test/java/io/opentelemetry/android/OpenTelemetryRumBuilderTest.java
```diff
@@ -132,4 +139,21 @@ void canSetPropagator() {
...
+        OpenTelemetryRum rum =
+                OpenTelemetryRum.builder(application).addSpanExporterCustomizer(customizer).build();
+        Span span = rum.getOpenTelemetry().getTracer("test").spanBuilder("foo").startSpan();
+        try (Scope scope = span.makeCurrent()) {
+            // no-op
+        } finally {
+            span.end();
+        }
+        // 5 sec is default
+        await().atMost(Duration.ofSeconds(30))
```
LikeTheSalad, line 156: If I understood correctly, this test would take at least 5 sec to complete? Because if that's the case I think we should find a way to use a simple span processor or something that exports the spans right away. Not necessarily has to be done in this PR though, but I'd like to consider it for a future one.

LikeTheSalad (approved):

instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
```diff
@@ -179,6 +184,24 @@ public OpenTelemetryRumBuilder addPropagatorCustomizer(
         return this;
     }
 
+    /**
+     * Adds a {@link Function} to invoke with the default {@link SpanExporter} to allow
+     * customization. The return value of the {@link Function} will replace the passed-in argument.
+     *
+     * <p>Multiple calls will execute the customizers in order.
+     */
+    public OpenTelemetryRumBuilder addSpanExporterCustomizer(
```
trask, line 193: ah, I think you want `addTracerProviderCustomizer()` instead for the use case you are targeting

`addSpanExporterCustomizer()` is for wrapping (customizing) an existing span exporter

breedx-splk: @trask - Apparently there is more than one way to skin this cat 🐱 and I'm sure Guido would disapprove. 😁 

For what it's worth, the `addTracerProviderCustomizer()` has already been in place, and we use it all over the place in our distro, including [setting up the zipkin-based exporter](https://github.com/signalfx/splunk-otel-android/blob/main/splunk-otel-android/src/main/java/com/splunk/rum/RumInitializer.java#L132).  For most users who just want to plug in their own (preconfigured) exporter, they probably don't want to deal with span processors -- they just want to tell the android sdk to use their exporter. Period.

In any case, there's now closer parity with the autoinstrumentation builder stuff, we have both `addTracerProviderCustomizer()` and `addExporterCustomizer()` in both the core autoinstrumentation and in android now.

trask: > For most users who just want to plug in their own (preconfigured) exporter, they probably don't want to deal with span processors -- they just want to tell the android sdk to use their exporter. Period.

outside of android we'd recommend `ConfigurableSpanExporterProvider` for this use case. I forget, does android not support SPI?

maybe we could have a new non-SPI way to register our normal SPI providers in RumBuilder, e.g. `RumBuilder.registerProvider()`?

LikeTheSalad: > does android not support SPI?

It does, but it was broken at some point and I'm also not sure if the common practice of [obfuscating the code](https://developer.android.com/build/shrink-code#obfuscate) might cause some issues too, issues that we could avoid by setting proguard consumer rules to prevent obfuscation for some classes, though in general it seems like using SPIs should be done carefully in android, so I would say it's good to know that we can use them if needed which is nice, but I would prefer to use them only when there's no better alternatives.

breedx-splk: Yeah, what he said. ^^ 😁 

We also don't currently take a dependency on the core autoconfigure sdk.

PR #88, opened 2023-09-22
Rename rum.session.id to session.id

breedx-splk:
Now that https://github.com/open-telemetry/semantic-conventions/pull/215 has landed, let's follow the convention.


trask (approved):
🥳

PR #91, opened 2023-09-25
Add dependencyManagement

### LINKED ISSUE #86: Consolidate dependency versions via module
Much like the other java projects, android instrumentation should be leveraging a topmost `dependencyManagement` module to consolidate versions into a singular place. See the[ core repo](https://github.com/open-telemetry/opentelemetry-java/blob/main/dependencyManagement/build.gradle.kts) or[ instrumentation repo ](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/dependencyManagement/build.gradle.kts)for reference examples.

breedx-splk:
This adds a dependencyManagement module to help consolidate dependencies across the project into one consistent place. This is consistent with the other java-based otel projects.

There were a few cases where I could not figure out how to propagate the constraint into `androidTestImplementation`, so those reference the version via root project variable.

This resolves #86

### DIFF
--- a/auto-instrumentation/okhttp/okhttp-3.0/library/build.gradle.kts
+++ b/auto-instrumentation/okhttp/okhttp-3.0/library/build.gradle.kts
@@ -4,7 +4,6 @@ plugins {
 }
 
 dependencies {
-    implementation(platform(project(":dependencyManagement")))
     // Pin at 3.0.0 for api compatibility
     api("com.squareup.okhttp3:okhttp:3.0.0")
     api("io.opentelemetry.instrumentation:opentelemetry-okhttp-3.0")
--- a/auto-instrumentation/okhttp/okhttp-3.0/library/build.gradle.kts
+++ b/auto-instrumentation/okhttp/okhttp-3.0/library/build.gradle.kts
@@ -5,7 +5,8 @@ plugins {
 
 dependencies {
     implementation(platform(project(":dependencyManagement")))
-    api("com.squareup.okhttp3:okhttp")
+    // Pin at 3.0.0 for api compatibility
+    api("com.squareup.okhttp3:okhttp:3.0.0")
     api("io.opentelemetry.instrumentation:opentelemetry-okhttp-3.0")
     implementation("io.opentelemetry.instrumentation:opentelemetry-instrumentation-api-semconv")
 }
--- a/auto-instrumentation/okhttp/okhttp-3.0/library/build.gradle.kts
+++ b/auto-instrumentation/okhttp/okhttp-3.0/library/build.gradle.kts
@@ -3,9 +3,9 @@ plugins {
     id("otel.publish-conventions")
 }
 
-val otelVersion = project.property("otel.sdk.version")
 dependencies {
-    api("com.squareup.okhttp3:okhttp:3.0.0")
-    api("io.opentelemetry.instrumentation:opentelemetry-okhttp-3.0:$otelVersion-alpha")
-    implementation("io.opentelemetry.instrumentation:opentelemetry-instrumentation-api-semconv:$otelVersion-alpha")
+    implementation(platform(project(":dependencyManagement")))
+    api("com.squareup.okhttp3:okhttp")
+    api("io.opentelemetry.instrumentation:opentelemetry-okhttp-3.0")
+    implementation("io.opentelemetry.instrumentation:opentelemetry-instrumentation-api-semconv")
 }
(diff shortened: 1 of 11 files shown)


auto-instrumentation/okhttp/okhttp-3.0/library/build.gradle.kts
```diff
@@ -3,9 +3,9 @@ plugins {
     id("otel.publish-conventions")
 }
 
-val otelVersion = project.property("otel.sdk.version")
 dependencies {
-    api("com.squareup.okhttp3:okhttp:3.0.0")
-    api("io.opentelemetry.instrumentation:opentelemetry-okhttp-3.0:$otelVersion-alpha")
-    implementation("io.opentelemetry.instrumentation:opentelemetry-instrumentation-api-semconv:$otelVersion-alpha")
+    implementation(platform(project(":dependencyManagement")))
+    api("com.squareup.okhttp3:okhttp")
```
LikeTheSalad, line 8: Only in this library, which will be used at runtime by a consumer project, I think we should leave the version to `3.0.0` because that's the minimum version supported by this auto instrumentation. If we set a higher version here, then we'd enforce our version into a project that might be using `3.0.0` as gradle would always pick the highest version when 2 or more are defined.

auto-instrumentation/okhttp/okhttp-3.0/testing/build.gradle.kts
```diff
@@ -4,8 +4,10 @@ plugins {
...
 
 dependencies {
+    implementation(platform(project(":dependencyManagement")))
+
     byteBuddy(project(":auto-instrumentation:okhttp:okhttp-3.0:agent"))
     implementation(project(":auto-instrumentation:okhttp:okhttp-3.0:library"))
-    implementation("com.squareup.okhttp3:okhttp:4.11.0")
-    androidTestImplementation("com.squareup.okhttp3:mockwebserver:4.11.0")
+    implementation("com.squareup.okhttp3:okhttp")
+    androidTestImplementation("com.squareup.okhttp3:mockwebserver:${rootProject.extra["okhttpVersion"]}")
```
LikeTheSalad, line 12: Isn't this artifact covered by the okhttp bom defined in `dependencyManagement`?

dependencyManagement/build.gradle.kts
```diff
@@ -0,0 +1,59 @@
...
+    allowDependencies()
+}
+
+dependencies {
+
+    // BOMs
+    api(enforcedPlatform("org.junit:junit-bom:$junitVersion"))
+    api(enforcedPlatform("io.opentelemetry:opentelemetry-bom-alpha:$otelSdkAlphaVersion"))
+    api(enforcedPlatform("io.opentelemetry.instrumentation:opentelemetry-instrumentation-bom-alpha:$otelInstrumentationAlphaVersion"))
+    api(enforcedPlatform("com.squareup.okhttp3:okhttp-bom:$okhttpVersion"))
```
LikeTheSalad, line 43: I think this instead of `enforcedPlatform` we should use `platorm` for this lib which I think would allow a subproject (such as the okhttp-library one) to override the version set in here.

auto-instrumentation/okhttp/okhttp-3.0/testing/build.gradle.kts
LikeTheSalad: Never mind! I just noticed you mentioned that these `androidTestImplementation` types of dependencies don't seem to work well with BOMs.

instrumentation/build.gradle.kts
```diff
@@ -58,11 +58,12 @@ android {
 }
 
 dependencies {
-    implementation("androidx.appcompat:appcompat:1.6.1")
-    implementation("androidx.core:core:1.12.0")
-    implementation("androidx.navigation:navigation-fragment:2.7.3")
+    implementation(platform(project(":dependencyManagement")))
```
LikeTheSalad, line 61: I'm not sure if it's going to work, but I was thinking that maybe we can just add this dependency into the java-library, android-app, and android-library convention files and maybe that'd help to not having to add it in each subproject like in here.

breedx-splk: If it works that would be cool. I'll give that a shot.

LikeTheSalad (approved):
Cool! Btw I noticed (a little late) [this comment](https://github.com/open-telemetry/opentelemetry-android/pull/90#issuecomment-1736199988) where @trask mentioned an alternative that might help us avoid having to pin old lib versions such as the okhttp one, I think it's the best solution for this case. We don't have to do it in this PR though just to let you know, I can create a follow-up one with those changes if you also like that other approach.

breedx-splk:
Makes sense. In the interest of timezones I'll get one out after merge. ☮️

PR #99, opened 2023-10-02
Pin okhttp dependency via compileOnly

breedx-splk:
This was mentioned as part of #91. It also addresses the renovatebot desire to update past 3.0.0.


breedx-splk:
Working on this. There's also something wrong with the published snapshots right now -- they're publishing the `dependencyManagement` module itself as a dependency. 🤦🏻 Obv. that can't be resolved, so it is breaking things.

breedx-splk:
Ok, due to this bug https://issuetracker.google.com/issues/248059128 I think it's best to just roll back #91 and do without consolidate dependencyManagement for the time being. If there are other more skilled gradle experts out there who can get that working without including the non-published platform dependency in the final .pom, please chime in.

PR #100, opened 2023-10-02
Add Event Emitter with Data to Android Repo

tonzhan2:


instrumentation/src/main/java/io/opentelemetry/android/emitter/EventEmitterWithData.java
```diff
@@ -0,0 +1,54 @@
...
+        delegate.emit(eventName, eventAttrs);
+    }
+
+    private Attributes convertRichDataObjectIntoAttributes(Map<String, Object> data) {
+        AttributesBuilder builder = Attributes.builder();
+        for (String key : data.keySet()) {
+            Object val = data.get(key);
+            if (val != null) {
+                if (val instanceof String) {
+                    builder.put(key, (String) Objects.requireNonNull(data.get(key)));
```
surbhiia, line 37: - As on line 35 it already checks that val != null, do we need Objects.requireNonNull here? 
- Objects.requireNonNull throws null pointer exception which might not be good.  I think just ignoring the attribute (map entry) if val is null is fine which is done on line 35. You can just use val here directly.

instrumentation/src/main/java/io/opentelemetry/android/emitter/EventEmitterWithData.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    public void emit(String eventName, Attributes attributes, Map<String, Object> data) {
+        Attributes dataAsAttribute = convertRichDataObjectIntoAttributes(data);
+        Attributes eventAttrs =
+                attributes.toBuilder().put(EVENT_DATA, dataAsAttribute.toString()).build();
+        delegate.emit(eventName, eventAttrs);
+    }
+
+    private Attributes convertRichDataObjectIntoAttributes(Map<String, Object> data) {
```
surbhiia, line 31: Let's double check this - Based on the data we expect in event.data for Android, are we sure we're okay to have these 4 types only - String, Long, Double, Boolean and all other types could be correctly represented by calling toString() [line 48] (As in it will not end up printing the memory address but will give actual content (value) of the attribute for any other types we expect)

instrumentation/src/main/java/io/opentelemetry/android/emitter/EventEmitterWithData.java
```diff
@@ -0,0 +1,54 @@
...
+
+package io.opentelemetry.android.emitter;
+
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.common.AttributesBuilder;
+import io.opentelemetry.api.events.EventEmitter;
+import java.util.Map;
+import java.util.Objects;
+
+class EventEmitterWithData {
```
surbhiia, line 14: It would be good to have a unit test for this class either in this or next PRs.

instrumentation/src/main/java/io/opentelemetry/android/emitter/EventEmitterWithData.java
tonzhan2: Not sure why I didn't use val directly here, my bad. Corrected

instrumentation/src/main/java/io/opentelemetry/android/emitter/EventEmitterWithData.java
tonzhan2: We could add other types yes, but at least when considering what this is currently planned to be used for (emitting crashes), I don't anticipate other data types unless I missed one and you can think of any? And if there are other object types, its possible that they'll have their own toString() method which will give an appropriate string representation of the object. 

And if not, it will resort to the default behaviour of toString() is to print class name, then @, then unsigned hexadecimal representation of the hash code of the object. Which I think is acceptable in an edge case. As if its not an object type we expect, whatever backend may not be able to parse it anyway. I'm definitely all ears and open to an alternative approach though if you have any ideas!

PR #101, opened 2023-10-02
Rollback dependency mgmt

breedx-splk:
Rolls back #91, supersedes #99.

There's a pretty wicked bug with AGP and `java-platform` dependencies. I think I've convinced myself this this is, at least in part, why the SNAPSHOT `.pom`s currently have a non-published dependency declared (`dependencyManagement`), which breaks our dependent users.

This also uses `compileOnly` for the okhttp api dependency and limits renovate.


trask (approved):

LikeTheSalad (approved):

LikeTheSalad:
This is unfortunate (and annoying). I was taking a look at what other options we have and I noticed that there's something called `version catalogs` which I haven't used so far, but at first glance, it seems like it might help in this case. [Here's the doc](https://docs.gradle.org/current/userguide/platforms.html#sub:central-declaration-of-dependencies) where they explain how to use it in case you want to take a look, I can take a look at it too when I'm back from pto on Thursday.

PR #106, opened 2023-10-05
Centralizing dependencies

LikeTheSalad:
Based on [this comment](https://github.com/open-telemetry/opentelemetry-android/pull/101#issuecomment-1744354656).

For posterity -- we were unable to use the `dependencyManagemetn` style `platform()` support (like in other projects) due to [this bug in AGP](https://issuetracker.google.com/issues/248059128).

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -8,6 +8,7 @@ opentelemetry-platform = { module = "io.opentelemetry.instrumentation:openteleme
 androidx-appcompat = "androidx.appcompat:appcompat:1.6.1"
 androidx-navigation-fragment = "androidx.navigation:navigation-fragment:2.7.4"
 androidx-core = "androidx.core:core:1.12.0"
+findbugs-jsr305 = "com.google.code.findbugs:jsr305:3.0.2"
 
 #Testing
 opentelemetry-sdk-testing = { module = "io.opentelemetry:opentelemetry-sdk-testing", version.ref = "opentelemetry" }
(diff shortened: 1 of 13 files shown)


gradle/libs.versions.toml
```diff
@@ -0,0 +1,53 @@
...
+opentelemetry = "1.29.0"
+opentelemetry-alpha = "1.29.0-alpha"
+mockito = "5.5.0"
+junit = "5.10.0"
+byteBuddy = "1.14.8"
+okhttp = "4.11.0"
+spotless = "6.22.0"
+
+[libraries]
+opentelemetry-platform = { module = "io.opentelemetry.instrumentation:opentelemetry-instrumentation-bom-alpha", version.ref = "opentelemetry-alpha" }
```
breedx-splk, line 11: This should be moved to `[bundles]` right?

LikeTheSalad: I gave it a try and moved this line below `[bundles]` though I got this error message:

```
* Exception is:
java.lang.RuntimeException: org.gradle.api.InvalidUserDataException: Invalid TOML catalog definition:
  - Unexpected type for bundle 'opentelemetry-platform'
    
    Reason: Expected an array but value of 'opentelemetry-platform' is a table.
    
    Possible solution: Use an array instead.
```

I'm not an expert using this feature of Gradle, however, based on [this part about bundles](https://docs.gradle.org/current/userguide/platforms.html#sec:dependency-bundles), it seems to me that bundles are kind of like buckets to place a bunch of individual dependencies declared under `[libraries]` that usually go together (such as mockito-core and mockito-junit-jupiter), so that you can later add a single dependency line with a bundle in a project's build.gradle, that would be the same as adding all the dependencies listed in that bundle (as it's done [here](https://github.com/LikeTheSalad/opentelemetry-android/blob/central-dependencies/instrumentation/build.gradle.kts#L74) for example), so I think that if we wanted to use bundles with OTel libs, we'd need to declare all of OTel's individual libs under `[libraries]` and then put them all into a bundle that we can add to projects that need all of those libs. However, if we're sticking to using this BOM, the only way I found to mix both, the "platform" dependency with the "catalogs", was to define the platform dependency as a regular one in the catalog, and then wrap it inside `platform()` [in here](https://github.com/LikeTheSalad/opentelemetry-android/blob/central-dependencies/instrumentation/build.gradle.kts#L65).

breedx-splk: Cool, thank you for giving it some consideration.

breedx-splk (approved):
Thanks for doing this, I think it's a great workaround since the platform approach wasn't working out. 🙏🏻

PR #110, opened 2023-10-10
Okhttp callback support

LikeTheSalad:
Adds support for instrumenting and properly propagating the Context when making calls using OkHttp callbacks (by calling [Call.enqueue](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-call/enqueue/))

### DIFF
--- a/auto-instrumentation/okhttp/okhttp-3.0/README.md
+++ b/auto-instrumentation/okhttp/okhttp-3.0/README.md
@@ -1,5 +1,7 @@
 # Android Instrumentation for OkHttp version 3.0 and higher
 
+# 🚧Work in progress 🚧
+
 Provides OpenTelemetry instrumentation for [okhttp3](https://square.github.io/okhttp/).
 
 ## Quickstart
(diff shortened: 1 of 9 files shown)


auto-instrumentation/okhttp/okhttp-3.0/README.md
```diff
@@ -1,5 +1,7 @@
 # Android Instrumentation for OkHttp version 3.0 and higher
 
+# 🚧Work in progress 🚧
```
LikeTheSalad, line 3: I added this line because I think we need to warn people not to use this tool until [this issue](https://github.com/open-telemetry/opentelemetry-java/issues/5886) is solved.

auto-instrumentation/okhttp/okhttp-3.0/testing/src/androidTest/java/io/opentelemetry/instrumentation/library/okhttp/v3_0/InstrumentationTest.java
```diff
@@ -34,27 +48,92 @@ public void setUp() throws IOException {
...
+                                    assertEquals(span, Span.current());
+                                    lock.countDown();
+                                }
+                            });
+        }
+
+        lock.await();
+        span.end();
+
+        assertEquals(2, spanExporter.getFinishedSpanItems().size());
```
breedx-splk, line 120: Love it!

breedx-splk (approved):
Looks good. Will be interesting to see how we can eliminate the infinite tracing of ourselves. The context key is an interesting approach....

PR #114, opened 2023-10-12
Update upstream otel deps

breedx-splk:
Supersedes #113.

In addition to the instrumentation update in #113, this also switches over to the newer semantic conventions that were pulled into the separate repo.

### DIFF
--- a/auto-instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/library/okhttp/v3_0/internal/OkHttp3Singletons.java
+++ b/auto-instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/library/okhttp/v3_0/internal/OkHttp3Singletons.java
@@ -11,7 +11,7 @@ import io.opentelemetry.api.GlobalOpenTelemetry;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.context.Scope;
 import io.opentelemetry.instrumentation.api.instrumenter.Instrumenter;
-import io.opentelemetry.instrumentation.api.instrumenter.http.HttpClientResend;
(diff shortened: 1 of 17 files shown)


gradle/libs.versions.toml
```diff
@@ -1,6 +1,7 @@
 [versions]
 opentelemetry = "1.31.0"
-opentelemetry-alpha = "1.29.0-alpha"
+opentelemetry-alpha = "1.31.0-alpha"
+opentelemetry-semconv-version = "1.21.0-alpha"
```
LikeTheSalad, line 4: Nit: I don't think we need to add the word `version` on these var names.

LikeTheSalad (approved):
This version bump was fun I see 😅

gradle/libs.versions.toml
breedx-splk: I only added it to distinguish it from the same-named thing `opentelemetry-semconv` in the `[libraries]` section, but now I'm realizing that it doesn't matter and `version` is redundant. I'll see about removing it.

auto-instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/library/okhttp/v3_0/internal/OkHttp3Singletons.java
```diff
@@ -38,15 +38,19 @@ public final class OkHttp3Singletons {
                                             OkHttpInstrumentationConfig
                                                     .getCapturedResponseHeaders())
                                     .setKnownMethods(OkHttpInstrumentationConfig.getKnownMethods()),
+                    spanNameExtractorConfigurer -> {
+                        // nop?
```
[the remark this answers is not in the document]
breedx-splk: Good call, I think you're right. I think this is necessary....thanks!

PR #117, opened 2023-10-17
Update dependencies list for consistency

breedx-splk:
There looked to be a few dependencies remaining in the instrumentation `build.gradle.kts`. This moves those over into the toml and provides one consistent place for coordinates and versions.


LikeTheSalad (approved):
Looks good!

PR #121, opened 2023-10-19
Introduce configuration for instrumentation components (phase 1)

breedx-splk:
Much of the existing instrumentation is somewhat burdensome for users to wire up correctly. The intent here is to introduce a configuration object that can be passed to the `OpenTelemetryRumBuilder` in order to more easily wire up various mobile components. This is intended to be first of several phases. This phase includes:

* Preconfigured global attributes. These are user-configured attributes that are intended to be included on every span and event.
* Network attributes (for things like connection type/subtype and carrier information)
* Screen attributes (the name of the currently visible screen)
* SDK initialization events - for tracking the startup of the sdk itself. NOTE: To keep this PR small and more easily reviewable, this implementation is left as a stub. Future PRs will fill in these events.

The behavior here should appear similar to Splunk's `RumInitializer` class, but is implemented within the builder to prevent confusion between building and initialization. This change should also still be backwards compatible -- users who want to configure their own components manually may continue doing so.

### DIFF
--- a/instrumentation/src/main/java/io/opentelemetry/android/OtelAndroidConfig.java
+++ b/instrumentation/src/main/java/io/opentelemetry/android/OtelAndroidConfig.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.android;
+
+import io.opentelemetry.android.instrumentation.network.CurrentNetworkProvider;
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.common.AttributesBuilder;
+
+/**
+ * Configuration object for OpenTelemetry Android. The configuration items in this class will be
+ * used in the OpenTelemetryRumBuilder to wire up and enable/disable various mobile instrumentation
+ * components.
+ */
+public class OtelAndroidConfig {
+
+    private AttributesBuilder globalAttributes = Attributes.builder();
+    private boolean includeNetworkAttributes = true;
+    private boolean generateSdkInitializationEvents = true;
+    private boolean includeScreenAttributes = true;
+
+    /**
+     * Configures the set of global attributes to emit with every span and event. Any existing
+     * configured attributes will be dropped. Default = none.
+     */
+    public OtelAndroidConfig setGlobalAttributes(Attributes attributes) {
+        globalAttributes = attributes.toBuilder();
+        return this;
+    }
+
+    boolean hasGlobalAttributes() {
+        return !globalAttributes.build().isEmpty();
+    }
+
+    Attributes getGlobalAttributes() {
+        return globalAttributes.build();
+    }
+
+    /**
+     * Disables the collection of runtime network attributes. See {@link CurrentNetworkProvider} for
+     * more information. Default = true.
+     *
+     * @return this
+     */
+    public OtelAndroidConfig disableNetworkAttributes() {
+        includeNetworkAttributes = false;
+        return this;
+    }
+
+    /** Returns true if runtime network attributes are enabled, false otherwise. */
+    public boolean shouldIncludeNetworkAttributes() {
+        return includeNetworkAttributes;
+    }
+
+    /**
+     * Disables the collection of events related to the initialization of the OTel Android SDK
+     * itself. Default = true.
+     *
+     * @return this
+     */
+    public OtelAndroidConfig disableSdkInitializationEvents() {
+        this.generateSdkInitializationEvents = false;
+        return this;
+    }
+
+    /** Returns true if the SDK is configured to generate initialization events, false otherwise. */
+    public boolean shouldGenerateSdkInitializationEvents() {
+        return generateSdkInitializationEvents;
+    }
+
+    /**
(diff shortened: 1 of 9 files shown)


instrumentation/src/main/java/io/opentelemetry/android/OtelAndroidConfig.java
```diff
@@ -0,0 +1,88 @@
...
+import io.opentelemetry.android.instrumentation.network.CurrentNetworkProvider;
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.common.AttributesBuilder;
+
+/**
+ * Configuration object for OpenTelemetry Android. The configuration items in this class will be
+ * used in the OpenTelemetryRumBuilder to wire up and enable/disable various mobile instrumentation
+ * components.
+ */
+public class OtelAndroidConfig {
```
LikeTheSalad, line 17: I believe that maybe the name of this class should somehow mention where it belongs, so maybe something like `OtelRumConfig` or `OpenTelemetryRumConfig` or maybe rename `OpenTelemetryRum` to `OpenTelemetryAndroid`? naming is hard 😅

instrumentation/src/main/java/io/opentelemetry/android/OtelAndroidConfig.java
```diff
@@ -0,0 +1,88 @@
...
+    private AttributesBuilder globalAttributes = Attributes.builder();
+    private boolean includeNetworkAttributes = true;
+    private boolean generateSdkInitializationEvents = true;
+    private boolean includeScreenAttributes = true;
+
+    /**
+     * Configures the set of global attributes to emit with every span and event. Any existing
+     * configured attributes will be dropped. Default = none.
+     */
+    public OtelAndroidConfig setGlobalAttributes(Attributes attributes) {
```
LikeTheSalad, line 28: Not necessarily for this PR, although I'm thinking that some of the use cases for global attrs might require those attrs to be dynamic, for example the session id one, so I'm thinking that it might be better to provide a supplier of attrs rather than the attrs right away.

instrumentation/src/main/java/io/opentelemetry/android/instrumentation/startup/InitializationEvents.java
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.android.instrumentation.startup;
+
+import io.opentelemetry.android.OtelAndroidConfig;
+
+public interface InitializationEvents {
```
LikeTheSalad, line 10: I love this!

instrumentation/src/main/java/io/opentelemetry/android/instrumentation/startup/InitializationEvents.java
```diff
@@ -0,0 +1,29 @@
...
+
+    void sdkInitializationStarted();
```
LikeTheSalad, line 12: Since we have this event, we should probably also include one for when the initialization has finished.

instrumentation/src/main/java/io/opentelemetry/android/OtelAndroidConfig.java
breedx-splk: I think OtelRumConfig is a good change. 👍🏻

instrumentation/src/main/java/io/opentelemetry/android/OtelAndroidConfig.java
breedx-splk: Let's tackle that as a separate effort. As it stands right now, the session id isn't lumped in with the other global attributes, it's a separate concern. Global attributes are currently changeable through the `GlobalAttributesSpanAppender.update()` method, but I'm not married to that. There is room for improvement for sure.

instrumentation/src/main/java/io/opentelemetry/android/instrumentation/startup/InitializationEvents.java
breedx-splk: Yeah totally....this thing is just a shell/stub right now.

instrumentation/src/main/java/io/opentelemetry/android/OtelAndroidConfig.java
LikeTheSalad: Sounds good, yeah it doesn't have to be for this PR though I took the opportunity to mention it. The session id is one example, connectivity info attrs would fit in here as well, maybe also user ids that would only become available after a login and would be removed/replaced after a logout, etc. The idea of having to manually call an update function to do so seems a bit easy to forget about, also people would have to spread telemetry-specific code across their project's classes in order to keep attrs updated with changes in their app's state, which might not be a huge problem depending on the app's architecture, but in general it seems handy/cleaner to have a single place to make these decisions in a reactive manner.

LikeTheSalad: I've created this issue to follow-up there: https://github.com/elastic/apm-agent-android/issues/216

LikeTheSalad (approved):

PR #122, opened 2023-10-19
Add Event Emitter with Data

tonzhan2:
Clean copy of this PR https://github.com/open-telemetry/opentelemetry-android/pull/100  with updated changes from master


instrumentation/src/test/java/io/opentelemetry/android/emitter/EventEmitterWithDataTest.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import org.junit.Test;
+
+public class EventEmitterWithDataTest {
+
+    @Test
+    public void testConvertRichDataObjectIntoAttributes() {
+        Map<String, Object> data = new HashMap<>();
+        data.put("stringKey", "key");
```
surbhiia, line 21: Can we use an actual android crash data here?

tonzhan2: Done, preserved the original test fields since they include double, boolean etc for other possible future cases

instrumentation/src/test/java/io/opentelemetry/android/emitter/EventEmitterWithDataTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+        data.put("stacktrace", stacktrace.toString());
+        data.put("thread", Thread.currentThread().toString());
+        data.put("throwableMessage", crashData.getMessage());
+
+        EventEmitterWithData emitter = new EventEmitterWithData(null);
+        Attributes attributes = emitter.convertRichDataObjectIntoAttributes(data);
+
+        assertEquals(attributes.size(), data.size());
+        assertEquals(data.get("stacktrace"), stacktrace.toString());
```
surbhiia, line 46: Should this be ->  
```suggestion
        assertEquals(data.get("stacktrace"), attributes.get(AttributeKey.stringKey("stacktrace"));
```
and similarly for other 2 lines.

tonzhan2: Just updated. In practice its the same thing but this looks better for consistency sake youre right

surbhiia: Just wanted to clarify for your understanding that it was not the same in practice as you were not checking what has been set in "attributes" variable instead earlier you were comparing "data" with the local variable "stacktrace" which you had set in the starting that fed data to "data"

breedx-splk:
Hi @tonzhan2 . Just wanted to touch base and let you know we haven't dropped this -- it's very much just in flux while the special events working group figures out some additional basics that might directly impact this PR. I appreciate your patience around this effort.

If you want to keep track of the work over there, here is the public tracking notes doc: https://docs.google.com/document/d/1BKjQWP32FXL9g1cGbyj7DMXV1Uq_RL8_78rWaMBhN0A/edit#heading=h.1nz72gv5evo2

Thanks!

instrumentation/src/main/java/io/opentelemetry/android/emitter/EventEmitterWithData.java
```diff
@@ -0,0 +1,51 @@
...
+                if (val instanceof String) {
+                    builder.put(AttributeKey.stringKey(key), (String) val);
+                } else if (val instanceof Long) {
+                    builder.put(AttributeKey.longKey(key), (Long) val);
+                } else if (val instanceof Double) {
+                    builder.put(AttributeKey.doubleKey(key), (Double) val);
+                } else if (val instanceof Boolean) {
+                    builder.put(AttributeKey.booleanKey(key), (Boolean) val);
+                } else {
+                    builder.put(AttributeKey.stringKey(key), val.toString());
```
marandaneto, line 45: Do we need to consider the list types? https://github.com/open-telemetry/opentelemetry-java/blob/48d41d3a5aa27ff8e5e4033ef0c8c3ba0e53f683/api/all/src/main/java/io/opentelemetry/api/common/AttributeKey.java#L49-L67

breedx-splk:
With all respect... Given that:

* this PR is stale and the repo has been reorganized from underneath it
* [core #6318](https://github.com/open-telemetry/opentelemetry-java/pull/6318) is likely to merge soon 
* ...and we want consistency/reuse of the upstream/standard otel APIs

I'm going to close this PR with the expectation that we will soon be able to leverage the new "data" (complex types) API from core. 

Feel free to reopen if this seems off-base. Thanks again for the submission and for blazing the trail. 😁

PR #133, opened 2023-10-30
Http/sURLConnection auto instrumentation

surbhiia:
### What's covered in this PR
As the Android SDK classes can't be instrumented with ByteBuddy Android Gradle Plugin, this PR instruments all calls to relevant URLConnection/HttpURLConnection/HttpsURLConnection APIs in android applications source code. 

**APIs Instrumented:**

**APIs that connect _but do not_ read bytes:**
- connect()
- getOutputStream()

**APIs that connect and read bytes:**
- getContent()
- getContent(Class[])
- getContentEncoding()
- getContentLength()
- getContentType()
- getDate()
- getExpiration()
- getLastModified()
- getHeaderFieldDate(name, Default)
- getHeaderFields()
- getHeaderFieldKey(int)
- getHeaderField(int)
- getHeaderField(name)
- getHeaderFieldInt(name, Default)
- getResponseMessage()
- getResponseCode()
- getInputStream()
- getErrorStream()
- getContentLengthLong() ( added in API level 24)
- getHeaderFieldLong() ( added in API level 24)

**Approach Taken**
- Request header can only be added before connecting. So, trace context header should be added before any of the above API calls are made.
- All above API calls are replaced with our own replacement methods which do the following:
1. Add connection to our ongoing URLConnections map if not added already. If it is the first time connecting, trace context is added as a header to the outgoing request
2. Original API call is made. If method can throw an exception, we wrap the method around a try/catch, so we can close the ongoing spans with exception reporting before re-throwing exception back to client. 
3. We update the last seen timestamp for the connection. If API is amongst the ones that read bytes also, we mark this URLConnection as harvestable.

Harvestable meaning here - if harvestable and if not already reported and if left idle for more than CONNECTION_KEEP_ALIVE_INTERVAL ms, connection is reported by a connection killer thread. So, once connection is made and any data is read from the connection, we mark it as harvestable because we don't anticipate connection to be sitting idle for longer after this.

As noted above, We keep track of all ongoing URLConnections and three other properties for it - last seen time, harvestable (boolean), reported (boolean) 

**When are spans closed and reported:** 
- If any exception is thrown anywhere, spans are closed and reported. 
- As a general behavior, either getInputStream or getErrorStream is always called and called in the end. We close spans and report when these two methods are called. 
- A connection reporter is scheduled which periodically check if any tracked connection, which is marked harvestable and not reported yet, was last seen more than CONNECTION_KEEP_ALIVE_INTERVAL ms before, then it closes the spans and reports the connection. [Currently, couldn't think of a way to proactively schedule this for clients, customers can manually call our API to schedule this runnable for now.]

**Sample App to test this PR**
Added an android sample app with many test scenarios for this PR [here](https://github.com/surbhiia/HttpURLConnection). 

If you would like to test this PR using the above sample app:
- Pull this PR in your local repo.
- Update the return value to some dummy value for the API you would like to test in HttpUrlReplacements.java
- Update gradle.properties with relevant version number
- Use ./gradlew publishToMavenLocal

In sample app:
- Update the version of httpURLConnection agent and lib dependencies to the version number you just published
- Build and run the app. Click the "ClickMe" button. 
- See output in logcat.


auto-instrumentation/httpurlconnection/agent/src/main/java/io/opentelemetry/instrumentation/agent/httpurlconnection/HttpUrlConnectionPlugin.java
```diff
@@ -0,0 +1,195 @@
...
+    public void close() throws IOException {
+        // No operation.
+    }
+
+    @Override
+    public boolean matches(TypeDescription target) {
+        // TODO: Add configuration for package name
+        return target.isSamePackage(
+                new TypeDescription.ForPackageDescription(
+                        new PackageDescription.Simple("com.example.httpurlconnection")));
```
surbhiia, line 193: @breedx-splk @LikeTheSalad Is there a better way to specify the packages to instrument here? 

1. I thought of using Context.getPackageName() but that could be different from the actual package names in the application code. 
2.  Just returning true for all types, to let byte buddy look at all .classes is very inefficient and it also looks into attached android libraries and was throwing errors for some kotlin libraries. 
3.  Best way I could think of was to allow customers to specify a list of package names from their application or may be the one topmost package name. This helps avoid any unnecessary instrumentations as well from any added third party libraries ( as we do not want to instrument other http client libraries that might internally use HttpURLConnection). May be a property like "httpURLConnectionPackage" specified somewhere in their application code property files rather than having them to make an API call to our APIs to set this?

auto-instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlReplacements.java
```diff
@@ -0,0 +1,421 @@
...
+            public void run() {
+                endOngoingConnectionsOlderThan(CONNECTION_KEEP_ALIVE_INTERVAL);
+            }
+
+            @Override
+            public String toString() {
+                return "EndOngoingConnectionsRunnable";
+            }
+        };
+    }
```
surbhiia, line 395: Is it okay for it to be optional and client's who think they need this can configure it (they will need manual code change to call our APIs)? If not, I can look for the main application activity and use byte buddy to do BCI to schedule this always. Is there any other better ways, to schedule thread runnable using existing open-telemetry codebases?

auto-instrumentation/httpurlconnection/agent/src/main/java/io/opentelemetry/instrumentation/agent/httpurlconnection/HttpUrlConnectionPlugin.java
LikeTheSalad: For targeting only classes that belong to the host project itself (and ignoring classes from libraries) the best option here would be to use the `AndroidDescriptor` tool. This is an example of how to do so: https://github.com/elastic/apm-agent-android/blob/main/android-instrumentation/src/main/java/co/elastic/apm/android/instrumentation/ui/activities/ActivityLifecyclePlugin.java#L50 

That way it's not needed to check for package names since the ByteBuddy gradle plugin already can tell you which classes are local, this is one of the benefits of instrumenting bytecode at compile-time.

auto-instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlReplacements.java
LikeTheSalad: I'm not sure I fully understand what is it that you suggest to make configurable, but generally speaking we can make anything configurable statically, as we did for the [OkHttp instrumentation](https://github.com/open-telemetry/opentelemetry-android/blob/main/auto-instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/library/okhttp/v3_0/OkHttpInstrumentationConfig.java) for example. Users would have to manually call those functions, setting their config params, and then we'll need to read those values later on.

LikeTheSalad:
Thanks @surbhiia I think it's a good start, although I noticed that there are some comments and todos in the wrapper class that seem like we should discuss further, probably in a call if needed.

I've also created [this issue](https://github.com/open-telemetry/opentelemetry-android/issues/138) to prevent API level restrictions on this module. I think that if we can verify that no runtime issues would happen when the wrapper class is loaded in envs where some methods aren't available, then it should be fine to ignore API restrictions for these cases.

auto-instrumentation/httpurlconnection/agent/src/main/java/io/opentelemetry/instrumentation/agent/httpurlconnection/HttpUrlConnectionPlugin.java
surbhiia: Thanks for suggesting this! It worked :)

surbhiia:
> Thanks @surbhiia I think it's a good start, although I noticed that there are some comments and todos in the wrapper class that seem like we should discuss further, probably in a call if needed.
> 
> I've also created [this issue](https://github.com/open-telemetry/opentelemetry-android/issues/138) to prevent API level restrictions on this module. I think that if we can verify that no runtime issues would happen when the wrapper class is loaded in envs where some methods aren't available, then it should be fine to ignore API restrictions for these cases.

surbhiia:
Oops, mistakenly closed, re-opening again!

auto-instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlReplacements.java
surbhiia: Actually, I wanted to discuss the approach we should take here - Manual vs Automatic scheduling of this thread (which reports connections that are not already reported and idle for given time threshold). 

Ideally, we should proactively **automatically** schedule this thread always to cover all possible scenario and to ensure connections are reported in the end always. 

Without the thread, right now, connections are reported if any exceptions occur and for getInputStream
/getErrorStream method calls assuming this would be called in the end. ( This can be extended to also report connections when getResponseCode method is called). This should mostly cover all scenarios but we can't 100% ensure that these methods are always certainly called. So, thread becomes useful. 

The[ java HttpURLConnection instrumentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/11cac29ba7c0e9e6eeb2a882912a684e3d348fb0/instrumentation/http-url-connection/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpurlconnection/HttpUrlConnectionInstrumentation.java#L140) as well only reports when exceptions occur or getInputStream is called, which I think doesn't cover all scenarios. 

I did not find a clean way to define this auto-instrumentation for thread scheduling. Android Apps don't have a method like java main method which would be called always in the starting. We can perhaps target the "onCreate" methods of starting activity or fragment or service but that might not cover all scenarios. Any suggestions for handling this?

breedx-splk:
Thanks @surbhiia for the contribution!
Please add a `README.md` for the instrumentation. At a minimum it should contain:

* Status: Experimental
* What a user needs to do to opt into this instrumentation
  * Build changes
  * Starting the harvester 
* A very brief description of what this does (bytecode instrumentation wrapping user calls to the http call sites)

breedx-splk:
As discussed, I've created #148 so that we can defer adding test to this (large!) PR. I'll assign it to you @surbhiia if you think you'd like to also take that on after this goes in. 🙏🏻 Thanks.

auto-instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlReplacements.java
surbhiia: We've decided to have this as a manual configuration for now and we can improve on it in future.

surbhiia:
Thanks @breedx-splk! I've added a readme now! 

Thanks for creating the issue. I'll add the tests in next PR. :)

breedx-splk:
@surbhiia @LikeTheSalad We chatted about some edge cases, specifically related to use cases where the client just pumps data (like a POST) and doesn't bother to read the response. I have put together a test gist that does exactly this: https://gist.github.com/breedx-splk/246962f7d1c180f924a530630a966c9f

In my testing with that, I run the collector locally and then I used netcat (`nc`) to see the request payload.  I then ran that main method with the java agent attached and observed the following:

* No traces sent to the collector (debug logger enabled in the collector)
* No input seen by `nc`.

This indicates to me that the upstream instrumentation is behaving correctly -- it doesn't create a span because the request/response has not taken place. I think the `HttpURLConnection` "apis" (if you can even call them that!) expect the user to read the response in some way -- it is very very lazy.

If you add `int responseCode = conn.getResponseCode();` after the `flush()` for example, the request is actually sent.

With that in mind, I definitely don't think it's something we should worry about covering this early in the instrumentation development.

breedx-splk:
> Thanks for creating the issue. Do assign it to me. I'll add the tests in next PR. :)

😎 You have to comment on the issue before I can assign. Just a simple "assign me please" will do. 👍🏻

surbhiia:
> @surbhiia @LikeTheSalad We chatted about some edge cases, specifically related to use cases where the client just pumps data (like a POST) and doesn't bother to read the response. I have put together a test gist that does exactly this: https://gist.github.com/breedx-splk/246962f7d1c180f924a530630a966c9f
> ...

@breedx-splk Thanks for testing this! I created [this](https://github.com/open-telemetry/opentelemetry-java-instrumentation/discussions/9869) issue about it in that repo yesterday and wanted to test this. 

Other than getInputStream/getErrorStream, getResponseCode and getResponseMessage are other two methods generally always called. Java instrumentation is closing spans for the latter two as well, as those methods inherently call getInputStream. We're instrumenting the call sites itself and we've not ended spans for the latter two right now. We can perhaps do that but that isn't a good idea as then we will close early in many cases and miss out on any exceptions occurring after that as usually first getResponseCode is checked and based on that either getInputStream/getErrorStream is called and we want to capture any IOExceptions there. Sort of a trade-off between waiting to close in order to capture any possible exceptions vs closing early in order to have a span ended for more possible scenarios. I still think we should go with the former as we can have the correct telemetry reported. 

Right now in the readme we've mentioned scheduling of the harvester as a required configuration. We can perhaps make it optional and call out how it affects the behavior.

trask:
> We can perhaps do that but that isn't a good idea as then we will close early in many cases and miss out on any exceptions occurring after that

is this any worse than the Java agent instrumentation? (I'm guessing we're probably closing the span early in these cases too?)

surbhiia:
> is this any worse than the Java agent instrumentation? (I'm guessing we're probably closing the span early in these cases too?)

@trask No, we're not closing spans early, instead waiting till the end and we're scheduling a thread to periodically look for idle connections and close spans.

breedx-splk:
> is this any worse than the Java agent instrumentation? (I'm guessing we're probably closing the span early in these cases too?)

I _think_ we're all in agreement here -- we're doing the best we can with a problematic legacy classes. 😅  I don't think this is any worse than what is happening in the agent instrumentation. This is acknowledging that, under some circumstances, we may not correctly close an open context, so there is a periodic harvester for those edge cases.

auto-instrumentation/httpurlconnection/README.md
```diff
@@ -0,0 +1,60 @@
...
+- [HttpsURLConnection](https://developer.android.com/reference/javax/net/ssl/HttpsURLConnection)
+
+## Quickstart
+
+### Overview
+
+This plugin instruments calls to all relevant APIs (APIs that cause a connection to be established) in client's android application code. It wraps calls to these APIs to ensure the following:
+- Context is added for distributed tracing before actual API is called (i.e before connection is established)
+- Any exceptions thrown are recorded and spans are ended.
+- If it's getInputStream()/getErrorStream() API, span is ended right after calling the API (as these are usually the last APIs to be called).
```
breedx-splk, line 17:
```suggestion
- If the getInputStream()/getErrorStream() APIs are used, the span is ended right after calling the API (as these are usually the last APIs to be called).
```

auto-instrumentation/httpurlconnection/README.md
```diff
@@ -0,0 +1,60 @@
...
+implementation("io.opentelemetry.android:httpurlconnection-library:AUTO_HTTP_URL_INSTRUMENTATION_VERSION")
+byteBuddy("io.opentelemetry.android:httpurlconnection-agent:AUTO_HTTP_URL_INSTRUMENTATION_VERSION")
+```
+
+### Configurations
+You can configure the automatic instrumentation by using the setters in [HttpUrlInstrumentationConfig](library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlInstrumentationConfig.java)).
+
+#### Required Configuration
+It is required to **manually** schedule the following runnable to periodically run at the below given fixed interval to end any open spans if connection is left idle >10s.
+- API to call to get the runnable: HttpUrlInstrumentationConfig.getReportIdleConnectionRunnable()
```
breedx-splk, line 49:
```suggestion
- API to call to get the runnable: `HttpUrlInstrumentationConfig.getReportIdleConnectionRunnable()`
```

auto-instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlReplacements.java
```diff
@@ -0,0 +1,466 @@
...
+        HttpURLConnectionInfo info = activeURLConnections.get(c);
+        if (info != null && !info.reported) {
+            Context context = info.context;
+            instrumenter().end(context, c, responseCode, error);
+            info.reported = true;
+        }
+    }
+
+    @SuppressWarnings("MustBeClosedChecker")
+    private static synchronized void startTracingAtFirstConnection(URLConnection c) {
```
breedx-splk, line 399: Do we _need_ `synchronized` on the private methods of this class? I think the entire public surface is already covered with `synchronized`, so I think an extra lock step is unnecessary?

auto-instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlReplacements.java
```diff
@@ -0,0 +1,466 @@
...
+
+        updateLastSeenTime(c);
+        markHarvestable(c);
+
+        // return headerFieldLong;
+        return 1L;
+    }
+
+    public static synchronized long replacementForHeaderFieldDate(
+            URLConnection c, String name, long Default) {
```
breedx-splk, line 213:
```suggestion
            URLConnection c, String name, long default) {
```

auto-instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlReplacements.java
```diff
@@ -0,0 +1,466 @@
...
+        // HttpURLConnection also overrides this and that is covered in
+        // replacementForHttpHeaderFieldDate method.
+        startTracingAtFirstConnection(c);
+
+        long headerFieldDate = c.getHeaderFieldDate(name, Default);
```
breedx-splk, line 218:
```suggestion
        long headerFieldDate = c.getHeaderFieldDate(name, default);
```

auto-instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlReplacements.java
```diff
@@ -0,0 +1,466 @@
...
+
+        updateLastSeenTime(c);
+        markHarvestable(c);
+
+        return headerFieldDate;
+    }
+
+    public static synchronized long replacementForHttpHeaderFieldDate(
+            HttpURLConnection c, String name, long Default) {
```
breedx-splk, line 227:
```suggestion
            HttpURLConnection c, String name, long default) {
```

auto-instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlReplacements.java
```diff
@@ -0,0 +1,466 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.net.HttpURLConnection;
+import java.net.URLConnection;
+import java.util.List;
+import java.util.Map;
+import java.util.WeakHashMap;
+
+public class HttpUrlReplacements {
```
breedx-splk, line 22: This class is pretty huge and has a lot of internal duplication. Is it possible to do something like https://gist.github.com/breedx-splk/28af92b4d26421a0f0e2014fc76aa28b to help mitigate some of this? 

_Note: The inner interfaces are only required to maintain API 21 compat due to `Supplier` being API 24+._

auto-instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlReplacements.java
surbhiia: I looked into it more. The functions which read/write the weak hash map need to be synchronized. We can instead remove synchronization from all the public replacement apis. I will do that change, test and submit the commit.

auto-instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlReplacements.java
surbhiia: That looks great! Thanks for suggesting this! :) I will make the change, test and submit a commit shortly.

trask:
my recommendation would be to match the Java agent instrumentation, which since instrumenting call sites here, would mean making a list of all the the HttpURLConnection methods that end up calling one of `connect`, `getOutputStream`, or `getInputStream`, and ending the span if any of those methods are called

this would eliminate the need for a harvester thread (which we don't have in the Java agent instrumentation)

surbhiia:
> end up calling one of `connect`, `getOutputStream`, or `getInputStream`

Actually ending spans at `connect` might not be a good idea as that's the first function being called from all APIs before anything can be read from the server we're connecting to. In that case, we don't even need to open/close spans in getInputStream/getOutputStream or any other method.

trask:
> I'm more leaning towards keeping the harvester thread as a recommended configuration

should we be adding harvester thread to the Java agent instrumentation too?

breedx-splk:
> java agent might be missing out on ending spans in all possible scenarios

I'm still looking for a repro with one of these scenarios. :) 

I tried (in the above gist) to make that happen with the POST without reading the response, but it turns out that a span doesn't get created because the POST doesn't even happen unless the response is read. @surbhiia do you know of another possible scenario where this might could happen?

surbhiia:
> @surbhiia do you know of another possible scenario where this might could happen

@breedx-splk Can you please try something like below:
//Example 1: No exception occurs and getInputStream is not called
URL url = new URL("<Your URL>");
URLConnection urlConnection = url.openConnection();
Map<String, List> map = urlConnection.getHeaderFields();
// Or any other similar api’s other than getInputStream/getResponseCode/getResponseMessage that read from connection are called like getContentType, getHeaderField(index)

Also, the example you mentioned is a reproduction of the issue right, as no span is created? POST should happen without reading anything, that's strange. Are you setting `connection.setDoOutput(true);`  Also, I think it does a PUT, and to do POST you need to add this too - `connection.setRequestMethod("POST");` This should be done before calling `getOutputStream`

surbhiia:
> I tried (in the above gist) to make that happen with the POST without reading the response,

You mean you tried with our codebase or java-agent to reproduce this? (in our codebase we've not marked request to be harvestable if they just call getOutputStream but I think we should mark it harvestable even after getOutputStream. Just connect should not mark request harvestable. )

trask:
I would love to see the Android and Java agent HttpURLConnection instrumentations produce the same telemetry (even though one is call-site based).

@surbhiia maybe you can remove the harvester thread from the initial Android instrumentation, and we can discuss pros and cons of adding that to both the Java agent and Android instrumentation in a follow-up?

bidetofevil:
@surbhiia @breedx-splk Looking at the [UrlConnection implementation](https://cs.android.com/android/platform/superproject/+/master:external/okhttp/repackaged/okhttp-urlconnection/src/main/java/com/android/okhttp/internal/huc/HttpURLConnectionImpl.java) that is used on Android API 19+, I can confirm that the request body written to the output stream isn't sent until you try to read a response or call `connect` on it afterwards. 

This is bit confusing because `getOutputStream()` already calls `connect()` first, but that [first call](https://cs.android.com/android/platform/superproject/+/master:external/okhttp/repackaged/okhttp/src/main/java/com/android/okhttp/internal/http/HttpEngine.java;l=207) only establishes the connection given there's no request body, and you would have to call it again to actually send the request (or it's called implicitly if you try to get something that requires the response).

Given that, I think the existing Java instrumentation is behaving correctly like Jason said - that a request was actually not sent, so there should be no automatic instrumentation - and I don't think you need the harvester thread.

surbhiia:
Summarizing where we are at right now based on all discussions so far (here, in sig meeting and slack)  - I got a chance to try out java agent instrumentation and can confirm that getOutputStream call doesn't lead to anything being posted to the connected server until any of the read apis are called (This behavior is same in android too). So, there is no need to start or end span for just getOutputStream api call scenario. 

I tested out other scenarios too where someone just **calls** any of the other read apis (getContent, getHeader... etc) and **not calls** any of these 4 (getInput/ErrorStream, getResponseCode/Message). In java agent, we're ending spans for all of these read APIs also. And java agent doesn't need to do that explicitly, it just defines the span ending logic for getInputStream API and seems like all these other read APIs internally call getInputStream, thereby leading to span ending for all.  

As android is instrumenting call sites and not the actual HttpURLConnection API, and we have added the logic to end spans for only getInput/ErrorStream calls, we're not ending spans for all the read apis right now. 

The only **open question** now is : Should android also define the span ending logic for all the read apis. My only concern with that is - we will end up ending spans too early and not capture any IOExceptions that can occur on any other read api calls after this first read api call. Ending at getInput/ErrorStream is a little bit safe as it's generally called in the end. And if getInput/ErrorStream is not called, harvester will be able to end spans.

LikeTheSalad:
> we will end up ending spans too early and not capture any IOExceptions that can occur on any other read api calls

Do you know if this is a problem for the Java Agent too?

> Ending at getInput/ErrorStream is a little bit safe as it's generally called in the end. And if getInput/ErrorStream is not called, harvester will be able to end spans.

I usually don't work with this tool so I'm not sure how often IOExceptions can happen when using it, if it's a common thing, then I agree we should cover that scenario, not only here but in the Java agent as well, so this really seems like a great topic for the Java SIG.

For now, without knowing whether it happens on the Java Agent too, or if it's a common issue or an edge case, this is what I think that we could do:

- Making the harverster solution configurable, whether its default is enabled or disabled I think would depend on what's the current behavior of the Java Agent regarding IOExceptions.
- If the Java Agent doesn't handle IOExceptions but they are quite common, then I think we can enable the harvester by default and create an issue for the Java Agent to make it exception-aware too.

The reasons why I think it should be configurable are:

- I know some people could be quite reluctant about having [too many threads](https://developer.android.com/topic/performance/threads) running in their apps. This might no longer be an issue with newer phones though, but I remember it used to be a performance concern some years ago and probably still is for some apps.
- The main reason we need a harvesting feature seems to be due to bad usage of the httpurlconnection "API" in the host app (because people wouldn't call `getInput/ErrorStream` in the end). However, if I were a pro user of this API and I knew I wouldn't make that mistake in my project, then I would prefer to disable the harvester option to reduce the amount of background work going on in my app.

An alternative I can see, to make sure users would always call `getInput/ErrorStream` in the end, could be to create our own lint rule that would raise a compilation warning whenever it spots that `getInput/ErrorStream` aren't called, and [publish it](https://googlesamples.github.io/android-custom-lint-rules/api-guide/publishing.md.html) into this agent's library artifact.

surbhiia:
> Do you know if this is a problem for the Java Agent too?
Yes, it is a problem with java agent too as it closes spans on first read api and no new spans are created on the same connection object afterwards, exception or no exception. 

> if it's a common thing
I do not know in practice how common these issues are, but some examples of these scenarios could be: 

- One example of exception occurring after a few read apis have already been successful - Connection being lost in between (which could happen for a lot reasons - timeouts, server temporarily unavailable / overloaded, internet connection lost etc)  which would result into an Exception. 

- One example of getInputStream not being called in the end could be an "OPTIONS" request method. Where you just want to look at headers to figure out what methods the server supports. Another example could be a HEAD request. 

> If the Java Agent doesn't handle IOExceptions but they are quite common, then I think we can enable the harvester by default and create an issue for the Java Agent to make it exception-aware too.

I think we should by default cover any possible exception scenarios -  enable harverster by default. And making it configurable allows those with issues with multiple threads running can disable it at the cost of losing some spans - when they don't call getInput/errorStream and we use the lint rule as you suggested to point out this when harvester is disabled. 

> An alternative I can see, to make sure users would always call `getInput/ErrorStream` in the end, could be to create our own lint rule that would raise a compilation warning whenever it spots that `getInput/ErrorStream` aren't called, and [publish it](https://googlesamples.github.io/android-custom-lint-rules/api-guide/publishing.md.html) into this agent's library artifact.

bidetofevil:
> The only open question now is : Should android also define the span ending logic for all the read apis. My only concern with that is - we will end up ending spans too early and not capture any IOExceptions that can occur on any other read api calls after this first read api call. Ending at getInput/ErrorStream is a little bit safe as it's generally called in the end. And if getInput/ErrorStream is not called, harvester will be able to end spans.

I think whatever we do in Android should be the same as what the Java agent does. If this is an issue there as well, I think we should ping those folks if we can find a repo case that.

> I do not know in practice how common these issues are, but some examples of these scenarios could be:
>
> One example of exception occurring after a few read apis have already been successful - Connection being lost in between (which could happen for a lot reasons - timeouts, server temporarily unavailable / overloaded, internet connection lost etc) which would result into an Exception.

I believe that most server/network errors like timeouts occur during when the bytes are streamed across the the wire to populate the input stream, and none of the read APIs will return until this is done. The errors you can from reading the input stream is happening client side, so like corrupt data or data that doesn't fit the stream's expectation (e.g. if you try to push uncompressed data through a `GzipInputStream`.)

I think we should come up with a repro test case to see if any of read APIs will actually return before the server finishes streaming all the bytes.

> One example of getInputStream not being called in the end could be an "OPTIONS" request method. Where you just want to look at headers to figure out what methods the server supports. Another example could be a HEAD request.

Do you think you can reproduce these cases on a test as well? I think it'd be good to have something concrete first before implementing the harvester in case it's not necessary. The Android implementations of Http[s]UrlConnection may surprise you in how they behave, so it would be good if we can see one of these scenarios happen before implementing fixes for them.

surbhiia:
I've not had time to look into it fully yet. I was able to repro the exception scenario while reading from input stream and server goes down. Next, I will repro the HEAD/OPTIONS scenario where same connection object is used but disconnected and connected again for another full GET request. I will get back on this as soon as possible and update you all about what I find.

breedx-splk:
In the spirit of incremental changes and continuous improvement, what if we started with a non-reaper version and assume that the edge-cases are minimal enough for now. If/when the problem is demonstrated we can expand the try/catch/close parts and/or introduce a reaper? 

That might get this unblocked and move it forward while keeping closer parity with upstream instrumentation.

Thoughts?

marandaneto:
I wonder if it's worth the effort to maintain the Http/sURLConnection instrumentation, I mean, the PR is done so LGTM, but I don't see apps using Http/sURLConnection for a long time, it is always OkHttp.

surbhiia:
Sorry about the delays in getting back on this. I was caught up with other work. I finally got a chance to try out possible scenarios where we might get exceptions later in the flow and might want to wait for ending spans and not end them at first read API call. 

I tried an found the following:

1. Calling any read API, gets everything at once from the server so you do not have to go to server again for consequent read APIs ( like getHeader would also get the inputStream object. Note - reading from input stream requires you to go back to the server - covered in point below). So, even if I set caching to false, or set a low connection timeout, or shut the server before the consequent read API instruction is called, it doesn't lead to IOException.
2. IOExceptions **do occur** when I'm **reading** from the inputStream. I tried shutting down the server after an initial read API was successful and some `inputStream.read` was successful but before the entire input stream is read. This does cause exception. I've added a new commit to this PR, to look into inputStream/errorStream read operation as well and capture any exceptions on that due to reasons like "server no longer available" or other connectivity issues. This is also not a case of bad implementation of HttpURL logic in your code but a genuine possible error at run-time. 
3. I tried creating HEAD/OPTIONS request, and then disconnecting and trying to use the same connection object for a GET request. To my surprise, HttpURLConnections handles it well. It doesn't allow same connection object to be re-used. You always have to open new connection for each request (internally it takes care of connection pooling, and you might save resources by using the same connection but the instance of connection in customer code has to be different always). 

So, the only scenario to justify not ending span at the first read API is -- IOExceptions during inputStream/errorStream read. And harvester will help close any un-ended spans if getInputStream/getErrorStream is not called at all OR if connection was stuck for too long (>10s) between the last inputStream.read and the next. (Please refer the latest commit where this logic is implemented) 

I think if we can answer the following question, we can conclude our discussion - Do we need to capture the above scenario? If not, we do not need the harvester. If we're still divided on whether to capture this scenario or not, we can for now remove the harvester for the first phase and re-visit it later.

surbhiia:
> I wonder if it's worth the effort to maintain the Http/sURLConnection instrumentation, I mean, the PR is done so LGTM, > but I don't see apps using Http/sURLConnection for a long time, it is always OkHttp.

I did look into it before starting development on this. A lot of apps still do use Http/sURLConnection (I guess due to it being a part of the Android SDK and light-weight). Amongst our customers, there are more usages of Http/sURLConnection (1st) than Okhttp (2nd highly used client).

auto-instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlReplacements.java
```diff
@@ -151,21 +151,97 @@ public static String replacementForResponseMessage(URLConnection c) throws IOExc
...
+                return res;
+            }
+
+            @Override
+            public int read(byte[] b) throws IOException {
+                int res;
+                try {
+                    res = inputStream.read(b);
+                } catch (IOException e) {
+                    reportWithThrowable(c, e);
```
surbhiia, line 209: Here is where we're reporting any exceptions during input/error stream read. And similar for other read methods covered below.

surbhiia:
I know some of us are already on vacation and others are gearing up for same, and we will pick this up when we come back, I wanted to add one more approach proposal which I think would be more acceptable - 

We can instrument the` disconnect` call to end any un-ended spans before disconnecting ( covers OPTION/HEAD requests and any other requests when getInput/ErrorStream is not called in the end). After this, the only scenarios requiring a harvester thread would be:

1. Not so good implementations where they do not call disconnect in the end. 
2. Edge cases like connection is stuck for 10s for some reason (like while reading from stream etc) ( I do not know how common these scenarios are)

Assuming above scenarios are rare - we can make harvester optional configuration and call out the scenarios in documentation that require it. And customers can decide based on their own implementations of HttpURLConnection. 
  
And with this approach, we do not have to end spans early too on the first read API and we can capture any exceptions on stream reads as well and ensure we close any un-ended spans before disconnecting.

breedx-splk:
> We can instrument the` disconnect` call to end any un-ended spans before disconnecting ( covers OPTION/HEAD requests and any other requests when getInput/ErrorStream is not called in the end). 

I think this is a great proposal and I'd like to see it in action. As you suggested, making the harvester opt-in makes the default usage of this instrumentation more lightweight and should cover the vast majority of use cases. 🤞🏻 
And of course if not, we can always improve/change it in the future.

breedx-splk:
It looks like this has gone stale. What do we want to do with this? @surbhiia @LikeTheSalad ?

surbhiia:
> It looks like this has gone stale. What do we want to do with this? @surbhiia @LikeTheSalad ?

Sorry about the delays on this. Have been focusing on other high priority items at work. I will be able to pick it up again in mid-May.

surbhiia:
I've picked up this PR again now! Will soon start creating commits and make it ready for review.

breedx-splk:
> I've picked up this PR again now! Will soon start creating commits and make it ready for review.

Yay! Thanks @surbhiia !

If the repo structure has changed so much that this PR is hard to wrangle, feel free to open another one. Whichever makes the most sense for you. 🙏🏻

gradle/libs.versions.toml
```diff
@@ -20,15 +20,16 @@ androidx-core = "androidx.core:core:1.13.1"
...
-opentelemetry-instrumentation-apiSemconv = { module = "io.opentelemetry.instrumentation:opentelemetry-instrumentation-api-incubator" }
+opentelemetry-instrumentation-api = { module = "io.opentelemetry.instrumentation:opentelemetry-instrumentation-api", version.ref = "opentelemetry-instrumentation" }
+opentelemetry-instrumentation-apiSemconv = { module = "io.opentelemetry.instrumentation:opentelemetry-instrumentation-api-incubator", version.ref = "opentelemetry-instrumentation-alpha" }
 opentelemetry-instrumentation-okhttp = { module = "io.opentelemetry.instrumentation:opentelemetry-okhttp-3.0", version.ref = "opentelemetry-instrumentation-alpha" }
 opentelemetry-semconv = { module = "io.opentelemetry.semconv:opentelemetry-semconv", version.ref = "opentelemetry-semconv" }
 opentelemetry-semconv-incubating = { module = "io.opentelemetry.semconv:opentelemetry-semconv-incubating", version.ref = "opentelemetry-semconv" }
-opentelemetry-api = { module = "io.opentelemetry:opentelemetry-api" }
+opentelemetry-api = { module = "io.opentelemetry:opentelemetry-api", version.ref = "opentelemetry"}
```
breedx-splk, line 28: I think the BOM was handling this already? Was there a reason for adding the explicit version?

instrumentation/httpurlconnection/README.md
```diff
@@ -0,0 +1,75 @@
...
+plugins {
+    id 'net.bytebuddy.byte-buddy-gradle-plugin' version 'BYTEBUDDY_VERSION'
+}
+```
+
+#### Project dependencies
+
+```kotlin
+implementation("io.opentelemetry.android:httpurlconnection-library:AUTO_HTTP_URL_INSTRUMENTATION_VERSION")
+byteBuddy("io.opentelemetry.android:httpurlconnection-agent:AUTO_HTTP_URL_INSTRUMENTATION_VERSION")
```
breedx-splk, line 58: We have release automation now that can update these versions with the current/latest (see `prepare-release-branch.yml`). It doesn't have to be part of this PR, but we should probably open an issue to automate this as part of the release.

Doing that would also allow us to remove the statement above that tells the reader to replace a thing.

gradle/libs.versions.toml
surbhiia: I was confused a bit earlier. I've now updated the library module to depend on BOM for version resolution. Just need "io.opentelemetry.instrumentation:opentelemetry-instrumentation-api-incubator" version to be specified explicitly as it is not defined by the BOM.

instrumentation/httpurlconnection/README.md
surbhiia: Thanks for sharing that! That would be better. Created this issue to handle that - https://github.com/open-telemetry/opentelemetry-android/issues/417.

instrumentation/httpurlconnection/library/lint-baseline.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
breedx-splk, line 1: What is this file? Is it needed?

instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlInstrumentationConfig.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.Map;
+import java.util.Set;
+
+/** Configuration for automatic instrumentation of HttpURLConnection/HttpsURLConnection requests. */
+public final class HttpUrlInstrumentationConfig {
+    private static List<String> capturedRequestHeaders = new ArrayList<>();
+    private static List<String> capturedResponseHeaders = new ArrayList<>();
+    private static Set<String> knownMethods = HttpConstants.KNOWN_METHODS;
+    private static Map<String, String> peerServiceMapping = new HashMap<>();
+    private static boolean emitExperimentalHttpClientMetrics;
```
breedx-splk, line 23: Normally we would want configuration classes like this to be normal POJO or data style classes, rather than a pile of grouped globals (statics). We don't have to make that change in this PR, but ideally the config can be created and passed to the instrumentation that cares about it.

Since you're focused on build-time instrumentation, I can see how this came to be, though. For that case, a static block or other static initializer should create a single instance of the config from whatever the mechanism is (gradle properties, config file, whatever), and then that instance can be used by the instrumentation's static singleton stuff. Ensuring the order is also tricky. :)

breedx-splk: I made #418 to follow-up on this so we don't block this PR.

instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlReplacements.java
```diff
@@ -0,0 +1,402 @@
...
+public class HttpUrlReplacements {
+
+    private static final Map<URLConnection, HttpURLConnectionInfo> activeURLConnections;
+    private static final Logger logger;
+    public static final int UNKNOWN_RESPONSE_CODE = -1;
+
+    static {
+        activeURLConnections = new ConcurrentHashMap<>();
+        logger = Logger.getLogger("HttpUrlReplacements");
+    }
```
breedx-splk, line 34: Please just move these to the corresponding declaration lines.

instrumentation/httpurlconnection/library/lint-baseline.xml
surbhiia: As our minCompileSDK is 21 in otel.android-app-conventions.gradle.kts and in HttpUrlReplacements.java we use some API 24 functions like getContentLengthLong(), we need to define a lint baseline file like this to prevent compile errors.

instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlReplacements.java
```diff
@@ -0,0 +1,402 @@
...
+
+    public static void replacementForDisconnect(HttpURLConnection c) {
```
breedx-splk, line 36: We generally don't favor single-letter parameter/variable names. Let's call them `connection` or at worst `conn`. :)

breedx-splk: Applies to all the other methods as well.

instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlReplacements.java
```diff
@@ -0,0 +1,402 @@
...
+        if (errorStream == null) {
+            return errorStream;
+        }
+
+        return getInstrumentedInputStream(c, errorStream);
+    }
+
+    private static InputStream getInstrumentedInputStream(
+            URLConnection c, InputStream inputStream) {
+        return new InputStream() {
```
breedx-splk, line 199: Might be slightly more readable to not have this inline, but pulled out to an inner class.

instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/internal/HttpUrlConnectionSingletons.java
```diff
@@ -0,0 +1,75 @@
...
+import io.opentelemetry.instrumentation.api.semconv.http.HttpSpanNameExtractor;
+import io.opentelemetry.instrumentation.api.semconv.http.HttpSpanNameExtractorBuilder;
+import io.opentelemetry.instrumentation.api.semconv.http.HttpSpanStatusExtractor;
+import io.opentelemetry.instrumentation.library.httpurlconnection.HttpUrlInstrumentationConfig;
+import java.net.URLConnection;
+
+public final class HttpUrlConnectionSingletons {
+
+    private static final Instrumenter<URLConnection, Integer> INSTRUMENTER;
+    private static final String INSTRUMENTATION_NAME = "io.opentelemetry.auto-http-url-connection";
```
breedx-splk, line 26: I think we probably want android in the name.

```suggestion
    private static final String INSTRUMENTATION_NAME = "io.opentelemetry.android.http-url-connection";
```

instrumentation/httpurlconnection/agent/src/main/java/io/opentelemetry/instrumentation/agent/httpurlconnection/HttpUrlConnectionPlugin.java
```diff
@@ -0,0 +1,226 @@
...
+                            .replaceWith(
+                                    HttpUrlReplacements.class.getDeclaredMethod(
+                                            "replacementForErrorStream", HttpURLConnection.class))
+                            .method(is(HttpURLConnection.class.getDeclaredMethod("disconnect")))
+                            .replaceWith(
+                                    HttpUrlReplacements.class.getDeclaredMethod(
+                                            "replacementForDisconnect", HttpURLConnection.class))
+                            .on(isMethod()));
+        } catch (NoSuchMethodException e) {
+            throw new RuntimeException(e);
```
breedx-splk, line 210: Will this crash the build at build-time? Seems like a good way to know if something is broken!

breedx-splk:
Thanks again @surbhiia for circling back on this. I think it's looking like a great start. I had a few comments, but I think it's close.

instrumentation/httpurlconnection/agent/src/main/java/io/opentelemetry/instrumentation/agent/httpurlconnection/HttpUrlConnectionPlugin.java
surbhiia: Yes! It breaks host application build if the method to replace a given method with is not found. Like this:
Screenshot 2024-06-07 at 10 52 11 AM

If the customer somehow doesn't add the library dependency which contains HttpUrlReplacements.java class, they will see this error at build time. Otherwise, for our well tested code, it won't happen. :)

surbhiia:
Thanks @breedx-splk for such a detailed review! I've addressed your comments now. Do take a look and let me know if anything seems missing. :)

breedx-splk (approved):

breedx-splk:
@open-telemetry/android-approvers Please have a look!

instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlInstrumentationConfig.java
```diff
@@ -0,0 +1,157 @@
...
+    // Time (ms) to wait before assuming that an idle connection is no longer
+    // in use and should be reported.
+    private static final long CONNECTION_INACTIVITY_TIMEOUT = 10000;
+
+    private HttpUrlInstrumentationConfig() {}
+
+    /**
+     * Configures the HTTP request headers that will be captured as span attributes as described in
+     * <a
+     * href="https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-request-and-response-headers">HTTP
```
marandaneto, line 34: better to use permlinks (check all the other links too)

instrumentation/httpurlconnection/library/lint-baseline.xml
marandaneto: can we use either`@SuppressLint("NewApi")` or androidx `RequiresApi` method annotations or even do a runtime check for the SDK version? I think they will make a lint pass and you don't need this extra file that is a bit uglier to read.
also, this file has to be regenerated when the AGP version changes or the line number of the suppressed issues changes as well, so more maintenance.

instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlInstrumentationConfig.java
surbhiia: Updated! :)

instrumentation/httpurlconnection/library/lint-baseline.xml
surbhiia: That looks better! Updated to add suppress annotation. We do not need a sdk check or requiresApi annotation as it is a replacement function, will only be replaced for original API 24 calls.

marandaneto (approved):

bidetofevil (commented):
Are there tests forthcoming?

bidetofevil:
Hi @surbhiia - sorry for taking so long for adding my comments. My main question is regarding the timeout value - 10 seconds doesn't seem sufficient if it's for the duration of the entire request-response cycle for an Android app. What would you think about a higher value, or maybe make it configurable?

surbhiia:
> Are there tests forthcoming?

Yes! I have the test module ready already in my local in the same branch. As soon as this PR is merged, will put up another PR for that.

instrumentation/httpurlconnection/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+A span associated with a given request is concluded in the following scenarios:
+- When the getInputStream()/getErrorStream() APIs are called, the span concludes after the stream is fully read, an IOException is encountered, or the stream is closed.
+- When the disconnect API is called.
+
+Spans won't be automatically ended and reported otherwise. If any of your URLConnection requests do not call the span concluding APIs mentioned above, refer the section entitled ["Scheduling Harvester Thread"](#scheduling-harvester-thread). This section provides guidance on setting up a recurring thread that identifies unreported, idle connections (those that have been read from but have been inactive for more than 10 seconds) and concludes any open spans associated with them.
+
+> The minimum supported Android SDK version is 21, though it will also instrument APIs added in the Android SDK version 24 when running on devices with API level 24 and above.
+
+> If your project's minSdk is lower than 26, then you must enable
```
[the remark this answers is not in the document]
surbhiia: Maybe we can pull out the similar test mentioned in main readme to a separate section in main readme itself,  something like "Critical Notes"(there is a slight more addition to that we can make - for AGP < 8.3.0, the property to be used for correct dexing is different). And link to that section from this readme. As this module can be used separately to the main module, it might be important to have that information linked from this readme as well. Does that sound good to you as well?

instrumentation/httpurlconnection/library/build.gradle.kts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+description = "OpenTelemetry HttpURLConnection library instrumentation for Android"
+
+android {
+    namespace = "io.opentelemetry.android.httpurlconnection.library"
+}
+
+dependencies {
+    api(platform(libs.opentelemetry.platform))
```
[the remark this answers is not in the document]
surbhiia: As the versioning for the api dependencies exposed by the module like io.opentelemetry:opentelemetry-api/context comes from this BOM, it needs to api dependency as well.

instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlInstrumentationConfig.java
```diff
@@ -0,0 +1,153 @@
...
+public final class HttpUrlInstrumentationConfig {
+    private static List<String> capturedRequestHeaders = new ArrayList<>();
+    private static List<String> capturedResponseHeaders = new ArrayList<>();
+    private static Set<String> knownMethods = HttpConstants.KNOWN_METHODS;
+    private static Map<String, String> peerServiceMapping = new HashMap<>();
+    private static boolean emitExperimentalHttpClientMetrics;
+
+    // Time (ms) to wait before assuming that an idle connection is no longer
+    // in use and should be reported.
+    private static final long CONNECTION_INACTIVITY_TIMEOUT = 10000;
```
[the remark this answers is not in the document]
surbhiia: Actually, 10s is the time for which connection has been found to be idle after becoming harvestable (last time it successfully read from the server). Every time any successful read API (getContent etc or even the inputStream individual read APIs) is triggered the last seen time is updated. So, the idea was - only those connections that get stuck in between - not reading anything from server for 10s, no throwable, not disconnected - would be picked up after 10s of being idle. 

But I see your point, there could be cases when there could be longer than 10s delay between two read APIs  (specially the consecutive inputStream.read calls ) due to various reasons.  

I did plan on discussing about making it configurable, so consumers can set it closer to their connection_timeout setting or based on their application expected behavior. Will bring it up in Android sig meeting, we can perhaps add that in next PRs along with making the span name extractor function configurable. 

We can still keep 10s as the default or increase it if you think it's very less. The downside of going with default for the consumer right now, would be that they will miss capturing any throwable that would have occurred from consecutive reads from the connection, as the connection span will get reported as successful by the harvester after 10s of being idle. There's always option for consumer to not schedule this thread if they have good implementation where they finally call disconnect on the connection. That will always make sure the span gets reported. The only span that will not be closed for them then would be for cases when the connection is legit stuck. If they're handling such scenarios too in their application code nicely, by disconnecting, they don't have to worry about the harvester.

surbhiia:
> Hi @surbhiia - sorry for taking so long for adding my comments. My main question is regarding the timeout value - 10 seconds doesn't seem sufficient if it's for the duration of the entire request-response cycle for an Android app. What would you think about a higher value, or maybe make it configurable?

No problem at all! I'm glad it's getting such good reviews. :) I've answered your question about timeout in previous comment. We can continue the discussion further in our Android Sig this week.

instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlInstrumentationConfig.java
```diff
@@ -5,15 +5,16 @@
...
-import io.opentelemetry.instrumentation.api.incubator.semconv.net.PeerServiceResolver;
-import io.opentelemetry.instrumentation.api.internal.HttpConstants;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import io.opentelemetry.instrumentation.api.incubator.semconv.net.PeerServiceResolver;
```
surbhiia, line 15: Question - Any idea why spotlessApply felt the need to move the imports? They were not touched in this change.

instrumentation/httpurlconnection/README.md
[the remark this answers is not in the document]
bidetofevil: I did a similar duplication in the Java sem conv repo earlier but was advised to put the details in one place so if we need to update, we didn't have to do it multiple places. But I think both ways have advantages, so it's up to you which one you prefer. I think merging as is and figuring it out later is perfectly reasonable.

instrumentation/httpurlconnection/library/build.gradle.kts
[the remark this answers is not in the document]
bidetofevil: Got it. I didn't realize that consuming the BOM requires exporting the dependency as well. TIL!

instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlInstrumentationConfig.java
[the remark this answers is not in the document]
bidetofevil: Per offline discussion, aligning this with the default would be nice, and making it configurable would be even nicer. That said, it's not blocking, so feel free to take care of this as you please.

bidetofevil (approved):
LGTM. Comments added for potential future work, but nothing blocking.

LikeTheSalad (approved):
Thank you!

Issue #132, opened 2023-10-30
Add disk buffering support


LikeTheSalad:
Using the [disk buffering](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/disk-buffering) lib.

breedx-splk:
I think this is complete in the above PRs and in #194 and #221. Reopen if not the case.

Issue #129, opened 2023-10-27
Create internal tools to interact with Android APIs


LikeTheSalad:
Android APIs/tools tend to [change a lot](https://github.com/open-telemetry/opentelemetry-android/issues/49), sometimes multiple ones need to be used for the same purpose due to API level restrictions, which causes a lot of checks to the `android.os.Build.VERSION.SDK_INT` value in order to decide at runtime which one to use. Some other times, the APIs themselves might have not changed, but the [way to access them](https://stackoverflow.com/questions/70275782/smsmanager-getdefault-deprecation-what-to-replace) has. In some cases, the existing tools are not deprecated but a new one has been launched and [its usage is encouraged by Google](https://developer.android.com/training/data-storage/shared-preferences) until it makes sense to deprecate the older one, which seems like it could happen to [SharedPreferences](https://proandroiddev.com/welcome-datastore-good-bye-sharedpreferences-4bf68e70efdb) for example. Other times, the APIs themselves haven't changed, but their [permission requirements have](https://developer.android.com/about/versions/11/privacy/storage#permissions), changing the way the API must be used by having to wrap calls with [permission checks](https://developer.android.com/reference/androidx/core/content/ContextCompat#checkSelfPermission(android.content.Context,java.lang.String)). Also, just to make my point on the overall stability of Android APIs/tools, I've even come across instances in which they propose multiple tools that allow doing the [exact same](https://stackoverflow.com/questions/69042012/livedata-vs-stateflow-should-we-switch-from-live-data-to-state-flow) things, [without encouraging](https://developer.android.com/kotlin/flow/stateflow-and-sharedflow#livedata) the usage of any of them above the other one, but just leaving the choice to the devs, which then causes people to debate for a long time ([ref1](https://www.youtube.com/watch?v=fIiGyfNinEk), [ref2](https://stackoverflow.com/a/70223407/7031224)) until a favorite is picked which then creates tech debt issues for projects using the non-favorite alternative and so on.

On top of all that, all of them would require in some way or another an Android [Context](https://developer.android.com/reference/android/content/Context) to work, which means that it's necessary to keep on passing around the Context instance to every place where any Android API usage is needed, which tends to be annoying for unit testing and the overall project's architecture.

[...]

PR #145, opened 2023-11-13
Adding disk buffering, part 1

LikeTheSalad:
Related to #132 

Setting up the configuration and utilities needed for adding the disk buffering/cache support in a follow-up PR.


instrumentation/src/main/java/io/opentelemetry/android/internal/services/Service.java
LikeTheSalad, line 1: This is related to #129 

At the moment is pretty similar to what we have at Elastic, though I'm open to suggestions/alternatives.

instrumentation/src/main/java/io/opentelemetry/android/internal/services/AppInfoService.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Utility to get information about the host app.
+ *
+ * <p>This class is internal and not for public use. Its APIs are unstable and can change at any
+ * time.
+ */
+public class AppInfoService implements Service {
```
breedx-splk, line 25: Looking at the public surface/outline of this class, it doesn't _really_ seem to deal with `AppInfo` very much....it deals more with the cache. Can we call it `CacheStorageService` instead?

instrumentation/src/main/java/io/opentelemetry/android/internal/services/AppInfoService.java
```diff
@@ -0,0 +1,93 @@
...
+    public AppInfoService(Context appContext) {
+        this.appContext = appContext;
+    }
+
+    public File getCacheDir() {
+        return appContext.getCacheDir();
+    }
+
+    @WorkerThread
+    public long getAvailableCacheSpace(long maxSpaceNeeded) {
```
breedx-splk, line 38: It can be frustrating to troubleshoot a problem and then later discover that a seemingly harmless "getter" method has side effects (like allocating cache memory here). I think a rename like `ensureCacheSpaceAvailable(long)` helps with this -- that way the caller is more aware. 

A javadoc is also warranted. Right now the reader has no clue what the return value means.

instrumentation/src/main/java/io/opentelemetry/android/internal/services/ServiceManager.java
```diff
@@ -0,0 +1,77 @@
...
+package io.opentelemetry.android.internal.services;
+
+import android.content.Context;
+import androidx.annotation.Nullable;
+import androidx.annotation.RestrictTo;
+import java.util.HashMap;
+
+public final class ServiceManager implements Lifecycle {
+
+    private final HashMap<Service.Type, Service> services = new HashMap<>();
```
breedx-splk, line 15:
```suggestion
    private final Map<Service.Type, Service> services = new HashMap<>();
```

Is concurrency important to this class?

instrumentation/src/main/java/io/opentelemetry/android/internal/services/Service.java
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.android.internal.services;
+
+public interface Service extends Lifecycle {
```
breedx-splk, line 8: Can we default the `start()` and `stop()` methods to no-op?

instrumentation/src/main/java/io/opentelemetry/android/internal/services/PreferencesService.java
```diff
@@ -0,0 +1,49 @@
...
+ * time.
+ */
+public class PreferencesService implements Service {
+    private final SharedPreferences preferences;
+
+    public PreferencesService(Context context) {
+        preferences =
+                context.getSharedPreferences(
+                        BuildConfig.LIBRARY_PACKAGE_NAME + ".prefs", Context.MODE_PRIVATE);
+    }
```
breedx-splk, line 25: Code smell -- wrong dependency passed to constructor. I'd make this a factory method `create(context)` and pass the `SharedPreferences` instance into a new constructor.

instrumentation/src/main/java/io/opentelemetry/android/internal/features/persistence/DiskManager.java
```diff
@@ -0,0 +1,104 @@
...
+                throw new IOException("Could not create dir " + dir);
+            }
+        }
+        deleteFiles(dir);
+        return dir;
+    }
+
+    public int getMaxFolderSize() {
+        int storedSize = preferencesService.retrieveInt(MAX_FOLDER_SIZE_KEY, -1);
+        if (storedSize != -1) {
```
breedx-splk, line 70: nit: if somehow another negative number ended up in there we would return it. How about `if (storedSize > 0)`?

instrumentation/src/main/java/io/opentelemetry/android/internal/features/persistence/DiskManager.java
```diff
@@ -0,0 +1,104 @@
...
+            logger.log(
+                    Level.FINER,
+                    String.format("Returning max folder size from preferences: %s", storedSize));
+            return storedSize;
+        }
+        int requestedSize = persistenceConfiguration.maxCacheSize;
+        int availableCacheSize = (int) appInfoService.getAvailableCacheSpace(requestedSize);
+        int calculatedSize = (availableCacheSize / 3) - MAX_FILE_SIZE;
```
breedx-splk, line 78: I think we should guard against this going negative. Maybe it needs to be ``>= 2*MAX_FILE_SIZE` as well?

breedx-splk (commented):
This looks like a good start, thanks!

instrumentation/src/main/java/io/opentelemetry/android/internal/services/ServiceManager.java
LikeTheSalad: > Is concurrency important to this class?

Not for now, ideally the writing to this map should happen once, during the rum initialization process.

instrumentation/src/main/java/io/opentelemetry/android/internal/features/persistence/DiskManager.java
```diff
@@ -0,0 +1,104 @@
...
+            PreferencesService preferencesService,
+            PersistenceConfiguration persistenceConfiguration) {
+        this.appInfoService = appInfoService;
+        this.preferencesService = preferencesService;
+        this.persistenceConfiguration = persistenceConfiguration;
+    }
+
+    public File getSignalsCacheDir() throws IOException {
+        File dir = new File(appInfoService.getCacheDir(), "opentelemetry/signals");
+        if (!dir.exists()) {
```
marandaneto, line 49: `mkdirs` already calls `exists` internally so you are doing IO twice.
To consider is that `mkdirs` returns `false` if `exists` returns true, so you cannot `throw IOException` in this case which is also suboptinal.

marandaneto: `Files.deleteIfExists` does that better but its probably not available under older Android versions.

instrumentation/src/main/java/io/opentelemetry/android/internal/services/PreferencesService.java
marandaneto: Ideally this is also lazily loaded (or off of the main thread) instead of in the ctor since this does IO and will be warned by `StrictMode`.

instrumentation/src/main/java/io/opentelemetry/android/internal/features/persistence/DiskManager.java
LikeTheSalad: > `Files.deleteIfExists` does that better but its probably not available under older Android versions.

Yeah, it seems like it's available from API 26 😅

gradle/libs.versions.toml
```diff
@@ -25,7 +25,8 @@ opentelemetry-api = { module = "io.opentelemetry:opentelemetry-api" }
 opentelemetry-sdk = { module = "io.opentelemetry:opentelemetry-sdk" }
 opentelemetry-exporter-zipkin = { module = "io.opentelemetry:opentelemetry-exporter-zipkin" }
 opentelemetry-exporter-logging = { module = "io.opentelemetry:opentelemetry-exporter-logging" }
-zipkin-sender-okhttp3 = { module = "io.zipkin.reporter2:zipkin-sender-okhttp3", version.ref = "zipkin-reporter"}
+zipkin-sender-okhttp3 = { module = "io.zipkin.reporter2:zipkin-sender-okhttp3", version.ref = "zipkin-reporter" }
+opentelemetry-diskBuffering = "io.opentelemetry.contrib:opentelemetry-disk-buffering:1.31.0-alpha"
```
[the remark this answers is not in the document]
LikeTheSalad: Thanks, it's updated now.

instrumentation/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
```diff
@@ -90,4 +93,16 @@ public OtelRumConfig disableScreenAttributes() {
     public boolean shouldIncludeScreenAttributes() {
         return includeScreenAttributes;
     }
+
+    public PersistenceConfiguration getPersistenceConfiguration() {
+        if (persistenceConfiguration == null) {
+            return PersistenceConfiguration.builder().build();
```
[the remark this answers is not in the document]
LikeTheSalad: jsyk, this was deleted after making the config field not nullable.

instrumentation/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
```diff
@@ -20,6 +22,7 @@ public class OtelRumConfig {
     private boolean includeNetworkAttributes = true;
     private boolean generateSdkInitializationEvents = true;
     private boolean includeScreenAttributes = true;
+    @Nullable private PersistenceConfiguration persistenceConfiguration = null;
```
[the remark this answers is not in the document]
LikeTheSalad: Sounds good, I've made it not nullable and set it to a new, default instance in order to reduce memory allocation.

instrumentation/src/main/java/io/opentelemetry/android/config/PersistenceConfiguration.java
```diff
@@ -0,0 +1,51 @@
...
+    public final boolean enabled;
+    public final int maxCacheSize;
+
+    private PersistenceConfiguration(Builder builder) {
+        this.enabled = builder.enabled;
+        this.maxCacheSize = builder.maxCacheSize;
+    }
+
+    public static Builder builder() {
+        return new Builder(false, 60 * 1024 * 1024);
```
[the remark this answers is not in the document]
LikeTheSalad: I agree! And I usually complain about magic numbers... 🤦‍♂️ somehow I didn't notice it. Thanks, it's updated now.

instrumentation/src/main/java/io/opentelemetry/android/config/PersistenceConfiguration.java
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.android.config;
+
+/** Configuration for disk buffering. */
+public final class PersistenceConfiguration {
```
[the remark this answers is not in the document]
LikeTheSalad: Sounds good. The "persistence" term comes from the name they gave this feature in the Swift SDK and from there on it _persisted_ in my mind 😄 But yeah, without that context is just confusing to be calling it multiple names in here. I've updated it, thanks!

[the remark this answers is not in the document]
LikeTheSalad: I don't really have a strong opinion on having all the params straight into the main config or grouping them within "subconfigs" if you will, though I'm a bit concerned about the intuitiveness/discoverability factor for someone who's not familiarized with the lib and wants to know all there is to configure about a specific feature. Now, this doesn't necessarily have to be the only way to address that issue, we could use a common prefix for all the config names related to a single feature, but that brings other concerns about having to enforce naming rules and also probably code placement rules (having all the related config methods together in one part of the file and so on) for expanding and maintaining the config code.

Now, the reason why it's not a strong opinion, at least not for now anyway, is because this is really only useful at a larger scale, in case there are multiple cases like this one, where several params are needed for a feature. And at the moment I can't think of another feature that would need this in the future (but this is a new project so I'm guessing that several features should come soon) and also we're talking about only 2 params for the disk buffering case (although we can technically expose more config options available in the [disk buffering lib](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/disk-buffering#configuration), it's just that I tried to keep it simple for now).

Generally speaking, I try to make things as scalable as possible from the beginning, though I must admit that sometimes it backfires and I end up adding unnecessary things so, I like having these kinds of conversations because it helps making the line between _useful-for-scalability_ and _too-early/unnecessary-optimization_ a bit clearer.

instrumentation/src/main/java/io/opentelemetry/android/internal/features/persistence/DiskManager.java
```diff
@@ -0,0 +1,104 @@
...
+
+    public static DiskManager create(OtelRumConfig config) {
+        ServiceManager serviceManager = ServiceManager.get();
+        return new DiskManager(
+                serviceManager.getService(Service.Type.APPLICATION_INFO),
+                serviceManager.getService(Service.Type.PREFERENCES),
+                config.getPersistenceConfiguration());
+    }
+
+    DiskManager(
```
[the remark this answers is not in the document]
LikeTheSalad: Yeah, I just double-checked why is it like that and I remembered is because I needed it visible for tests (this code comes from the Elastic repo) so, depending on the test strategy we want to go with here, it might not be needed to have it package-private. Which brings me to the fact that I should update this PR to include tests since the overall approach was well received.

instrumentation/src/main/java/io/opentelemetry/android/internal/features/persistence/DiskManager.java
```diff
@@ -0,0 +1,104 @@
...
+            AppInfoService appInfoService,
+            PreferencesService preferencesService,
+            PersistenceConfiguration persistenceConfiguration) {
+        this.appInfoService = appInfoService;
+        this.preferencesService = preferencesService;
+        this.persistenceConfiguration = persistenceConfiguration;
+    }
+
+    public File getSignalsCacheDir() throws IOException {
```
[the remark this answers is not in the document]
LikeTheSalad: Sure, it's updated now.

instrumentation/src/main/java/io/opentelemetry/android/internal/services/AppInfoService.java
LikeTheSalad: Sure, makes sense. It's updated now.

instrumentation/src/main/java/io/opentelemetry/android/internal/services/Service.java
LikeTheSalad: Sure, sometimes I forget we can do that... It's updated now.

instrumentation/src/main/java/io/opentelemetry/android/internal/services/PreferencesService.java
LikeTheSalad: > I'm assuming this will be created once the SDK starts on the main thread

You're right. The idea is to initialize it in `OpenTelemetryRumBuilder` which should be called in the App.onCreate method. We're going to have to take a look at what parts of the initialization need to happen in a worker thread.

instrumentation/src/main/java/io/opentelemetry/android/internal/features/persistence/DiskManager.java
LikeTheSalad: Good point! it's updated now.

instrumentation/src/main/java/io/opentelemetry/android/internal/features/persistence/DiskManager.java
LikeTheSalad: You're right, I've added some javadoc and a validation for it. It might need some polish in the future though.

instrumentation/src/main/java/io/opentelemetry/android/internal/services/PreferencesService.java
breedx-splk: Good call @marandaneto. Would you mind opening an issue to track this so that we don't have to force it into this sizable PR? Thanks.

instrumentation/src/main/java/io/opentelemetry/android/internal/services/AppInfoService.java
LikeTheSalad: Thanks, it's updated now.

LikeTheSalad:
Thanks for the feedback! @breedx-splk @marandaneto 

I've addressed the comments and also added a couple of tests that were pending. I think it should be ready for another check.

instrumentation/src/main/java/io/opentelemetry/android/internal/features/persistence/DiskManager.java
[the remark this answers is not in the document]
LikeTheSalad: It's done now 👍 - I've added tests while also making the constructor private.

instrumentation/src/main/java/io/opentelemetry/android/internal/services/PreferencesService.java
breedx-splk: Ok, I've went ahead and logged #149 for this.

breedx-splk (approved):
Good stuff!

PR #147, opened 2023-11-14
Making android min sdk configurable per auto-instrumentation subproject

### LINKED ISSUE #138: Remove animalsniffer from auto-instrumentation subprojects
The automatic instrumentation tools in this project are opt-in, so users would have to manually add them on top of the core lib in order to use them in their projects. While the core lib must remain compatible with our minimum API version, it might not be the case for auto-instrumentation ones, since those could target code that became available in newer API versions, such as the http URLConnection methods added in API 24 that are mentioned [here](https://github.com/open-telemetry/opentelemetry-android/pull/133).

In [this](https://github.com/open-telemetry/opentelemetry-android/pull/133) example, the host project's code is inspected to look for usages of URLConnection methods in order to replace them with wrappers, so if a host project with a min API lower than 24 (21 for example) gets this auto-instrumentation added, then the instrumentation would never have to replace methods that aren't available on API 21, even though it could if it found them. So in that sense, making the httpurlconnection auto-instrumentation unaware of API limitations shouldn't cause issues, although we should validate in tests that for projects with lower APIs there won't be any runtime exceptions due to no methods found when the wrapper class is loaded.

LikeTheSalad:
Closes #138 

This should allow to set the minimum Android SDK supported by a specific `auto-instrumentation` subproject like so:

```kotlin
// library build.gradle.kts file

plugins {
    id("otel.java-library-conventions")
    //...
}

otelAndroid.minSdk = 26
```

This will prevent animal sniffer from complaining about code references used by the auto-instrumentation project that aren't available neither in the Android SDK 21 (the current core lib min SDK supported) nor in the desugar lib.

The README of the auto-instrumentation subproject that changes this value should mention what's the minimum supported Android API level for the auto-instrumentation to work properly, this is because some auto-instrumentations might target code found on versions above 21 while still providing some functionality and being safe to run on devices with API level 21 (since the code from higher SDK versions would just be ignored as it wouldn't be found in devices with SDK version 21). In some other cases, the auto-instrumentation could cause issues when running on devices with the SDK version 21, so that's why it should be specified in the README what's the minimum version supported.

### DIFF
--- a/buildSrc/src/main/kotlin/otel.java-library-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.java-library-conventions.gradle.kts
@@ -6,6 +6,15 @@ plugins {
     id("ru.vyarus.animalsniffer")
 }
 
+// Extension to configure android parameters for non-android projects.
+interface OtelAndroidExtension {
+    val minSdk: Property<Int>
+}
+
+val otelAndroidExtension =
+    project.extensions.create("otelAndroid", OtelAndroidExtension::class.java)
+otelAndroidExtension.minSdk.convention((project.property("android.minSdk") as String).toInt())
+
 java {
     val javaVersion = rootProject.extra["java_version"] as JavaVersion
     sourceCompatibility = javaVersion
@@ -14,7 +23,6 @@ java {
 
 val libs = extensions.getByType<VersionCatalogsExtension>().named("libs")
 dependencies {
-    signature("com.toasttab.android:gummy-bears-api-${project.property("android.minSdk")}:0.5.1:coreLib@signature")
     implementation(libs.findLibrary("findbugs-jsr305").get())
 }
 
@@ -30,4 +38,10 @@ tasks.withType<AnimalSniffer> {
 // Attaching animalsniffer check to the compilation process.
 tasks.named("classes").configure {
     finalizedBy("animalsnifferMain")
+}
+
+afterEvaluate {
+    dependencies {
+        signature("com.toasttab.android:gummy-bears-api-${otelAndroidExtension.minSdk.get()}:0.5.1:coreLib@signature")
+    }
 }


surbhiia:
Thanks @LikeTheSalad! I also tried it out in httpurlconnection module, it works perfectly!

Just had a question about what we should call out in readme - 
The "otelAndroid.minSdk = 24" we specify in the build file is what's used to compile the module. And as you also pointed out, and I tested it before - a module compiled with higher sdk version can RUN on all prior versions (android offers this backward compatibility). So, If I were to mention min sdk compatible to be run on, it would be API level 1. We can perhaps skip that? But we will definitely make sure we test our module runs perfectly on one of the prior versions to the compile sdk version we mention in build script.

surbhiia (approved):

LikeTheSalad:
> Just had a question about what we should call out in readme - The "otelAndroid.minSdk = 24" we specify in the build file is what's used to compile the module. And as you also pointed out, and I tested it before - a module compiled with higher sdk version can RUN on all prior versions (android offers this backward compatibility). So, If I were to mention min sdk compatible to be run on, it would be API level 1 (or whatever Android provides backward compatibility upto if that's different). We can perhaps skip that? But we will definitely make sure we test our module runs perfectly on one of the prior versions to the compile sdk version we mention in build script.

It's probably the case that it works back to API 1 as you mentioned, however, I think it might be a little confusing to use that value since this tool relies on the OTel Java lib which has min API 21. So I would mention something like "The minimum supported Android SDK version is 21, though it will also instrument tools added in the Android SDK version 24 when running on devices with API level 24 and above". There's probably a better way to phrase it, but generally speaking, that would be my approach.

surbhiia:
>> It's probably the case that it works back to API 1 as you mentioned, however, I think it might be a little confusing to use that value since this tool relies on the OTel Java lib which has min API 21. So I would mention something like "The minimum supported Android SDK version is 21, though it will also instrument tools added in the Android SDK version 24 when running on devices with API level 24 and above". There's probably a better way to phrase it, but generally speaking, that would be my approach.

Oh, I see. That makes more sense! Will add that in the readme :)

marandaneto:
> > > It's probably the case that it works back to API 1 as you mentioned, however, I think it might be a little confusing to use that value since this tool relies on the OTel Java lib which has min API 21. So I would mention something like "The minimum supported Android SDK version is 21, though it will also instrument tools added in the Android SDK version 24 when running on devices with API level 24 and above". There's probably a better way to phrase it, but generally speaking, that would be my approach.
> 
> Oh, I see. That makes sense! Will add that in the readme :)

Another option is to throw an exception with this message when setting `minSdk` < 21, so it's not failing at runtime in case it compiles somehow.

breedx-splk (approved):
This is cool. I am not expecting to see a huge number of wildly differing instrumentations in the short term, but I do think about end-user ergonomics, and I wonder if having different minSdk versions for different components could cause user confusion (or at least burden).

Not a showstopper, but wanted to be thinking about it for later.

LikeTheSalad:
> This is cool. I am not expecting to see a huge number of wildly differing instrumentations in the short term, but I do think about end-user ergonomics, and I wonder if having different minSdk versions for different components could cause user confusion (or at least burden).
> 
> Not a showstopper, but wanted to be thinking about it for later.

Yeah, ideally, and most likely, we wouldn't have to create too many of these add-ons with higher SDK version compatibility, especially because that should only happen when addressing Android OS's classes and I'm guessing third-party libs would be more often the targets of instrumentation. However, I cannot say we will never encounter a situation where we need to instrument OS's classes-related code added in newer API levels, for those cases, I think we should make sure that their READMEs explain this quite well to avoid confusion. The burden is still there though, but since these tools are optional, I think it's better to have that burden rather than enforcing our core lib api limitations on end-user's projects.

[This case](https://github.com/open-telemetry/opentelemetry-android/pull/133) is kinda like a hybrid one since it doesn't break the min SDK compatibility for 21 as @surbhiia explained, but it still supports code added past that version.

PR #156, opened 2023-11-20
Add VERSIONING.md

### LINKED ISSUE #143: Create VERSIONING.md
We should follow the convnetions in the other repos by making a VERSIONING.md.

This should follow the documentation standards being set up in https://github.com/open-telemetry/opentelemetry-java/issues/5936. This includes making a clear statement about release schedule (following upstreams) and the min version of android SDK that we support, and that updating that is not considered a breaking change.

breedx-splk:
Resolves #143 .

### DIFF
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -5,20 +5,20 @@ the OpenTelemetry Android project.
 
 ## Versioning scheme
 
-This codebase uses [Semantic Versioning](https://semver.org/) (semver) for its version numbers. 
+This codebase uses [Semantic Versioning](https://semver.org/) (semver) for its version numbers.
 All modules in this codebase are released at the same time and, as such, will
 be versioned together. All modules in this repo are released with the same version number.
 
 Until 1.0.0 stability has been achieved, regular releases will only typically increment
 the minor version (second number in the semver triplet). Patch releases are considered
-exceptional, and will only be created if a critical issue needs to be addressed shortly after 
+exceptional, and will only be created if a critical issue needs to be addressed shortly after
 a regular release.
 
 ## Snapshot builds
 
-Every commit to the `main` branch will cause a 
+Every commit to the `main` branch will cause a
 [snapshot build](https://oss.sonatype.org/content/repositories/snapshots/io/opentelemetry/android/)
-to be published to Sonatype. Users may choose to build and test and file issues against SNAPSHOT 
+to be published to Sonatype. Users may choose to build and test and file issues against SNAPSHOT
 builds, but their use in production is strongly discouraged.
 
 ## Android API Compatibility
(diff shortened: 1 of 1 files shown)


VERSIONING.md
```diff
@@ -0,0 +1,58 @@
...
+
+Every commit to the `main` branch will cause a
+[snapshot build](https://oss.sonatype.org/content/repositories/snapshots/io/opentelemetry/android/)
+to be published to Sonatype. Users may choose to build and test and file issues against SNAPSHOT
+builds, but their use in production is strongly discouraged.
+
+## Android API Compatibility
+
+The core Android instrumentation provided here is designed to be used with
+Android API Level 21 and above. API levels 21 to 25 require desugaring of the core library.
```
LikeTheSalad, line 27:
```suggestion
Android API Level 21 and above. API levels 21 to 25 require [desugaring of the core library](https://developer.android.com/studio/write/java8-support#library-desugaring).
```

I think having the link might help someone who might not be familiar with the desugaring functionality.

LikeTheSalad (approved):
LGTM!

VERSIONING.md
marandaneto: Since desugaring is mandatory and the`desugarJdkLibs` is always up to date (2.0.4 right now?), it might be important to write down the AGP min. version as well, for example, `desugarJdkLibs` >= 2.0.3 requires AGP >= 7.4

marandaneto: And probably Gradle as well since there's a compat. Matrix between Gradle <-> AGP

LikeTheSalad: > (2.0.4 right now?)

Yes, good point, we're currently relying on `2.0.4` so people should use at least that version in their projects.

> it might be important to write down the AGP min. version as well

Good point. Ideally by explaining that they need to use at least version `2.0.4` of the desugaring lib, it should be enough for people to realize that they need to use AGP 7.4+ as well, since it's mentioned in the [linked docs](https://developer.android.com/studio/write/java8-support#library-desugaring), however, I know that usually what's "ideal" is not what ends up happening 😅 so it does seem like a good idea to also mention the AGP min version requirement 👍 

> And probably Gradle as well since there's a compat. Matrix between Gradle <-> AGP

True, although if people set their AGP version to 7.4+, the AGP itself will make sure to complain by refusing to compile if the project's Gradle version is not compatible, so I'd say that we leave the Gradle min version requirement up to whatever the AGP used in the project demands for, otherwise we'd essentially have to duplicate info shown in the matrix from the [android docs](https://developer.android.com/build/releases/gradle-plugin#updating-gradle) which requires some extra maintenance work that doesn't seem necessary.

marandaneto: Yeah I agree, the dupe bits aren't ideal, maybe just mention the Gradle x AGP matrix compat. [docs](https://developer.android.com/build/releases/gradle-plugin#updating-gradle).
Not mentioning at all has the downside of error messages aren't always clear, so:

> The AGP itself will make sure to complain by refusing to compile if the project's Gradle version is not compatible

Isn't always the case and/or crystal clear.

breedx-splk: Ok I've mentioned both the desugar lib version and the agp version and have linked to the compat matrix docs. Please let me know what you think now.

marandaneto (approved):

PR #159, opened 2023-11-22
Okhttp autoinstrumentation stabilization

LikeTheSalad:
The OkHttp auto-instrumentation tool had an issue where it would create spans for OkHttp requests created by OTel's exporters, causing an infinite amount of spans to be created since each export would mean a new OkHttp call. Said issue was addressed as part of these PRs:

- https://github.com/open-telemetry/opentelemetry-java/pull/5918
- https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/9739

This PR does the following:
* Adds tests to validate that the infinite spans issue is solved. 
* Adds proguard rules to avoid issues with a reflection call needed to identify OTel's okhttp requests, more info [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/9739#discussion_r1394507569). 
* Fixes an initialization racing issue where the global OpenTelemetry instance was being set statically [here](https://github.com/open-telemetry/opentelemetry-android/blob/3c5ca844e8b7925fa3ca0de8bfd9599b02e5694b/auto-instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/library/okhttp/v3_0/internal/OkHttp3Singletons.java#L33) when creating an [OtlpHttpSpanExporter instance](https://github.com/open-telemetry/opentelemetry-java/blob/1e82b72f36d847e743a525ae8767a6badd8deb81/exporters/otlp/all/src/main/java/io/opentelemetry/exporter/otlp/http/trace/OtlpHttpSpanExporterBuilder.java#L163) which ends up creating a [new OkHttpClient](https://github.com/open-telemetry/opentelemetry-java/blob/1e82b72f36d847e743a525ae8767a6badd8deb81/exporters/sender/okhttp/src/main/java/io/opentelemetry/exporter/sender/okhttp/internal/OkHttpHttpSender.java#L60) instance before users could create and set their own GlobalOpenTelemetry instance.

### DIFF
--- a/auto-instrumentation/okhttp/okhttp-3.0/README.md
+++ b/auto-instrumentation/okhttp/okhttp-3.0/README.md
@@ -1,5 +1,7 @@
 # Android Instrumentation for OkHttp version 3.0 and higher
 
+## Status: Experimental
+
 Provides OpenTelemetry instrumentation for [okhttp3](https://square.github.io/okhttp/).
 
 ## Quickstart
--- a/auto-instrumentation/okhttp/okhttp-3.0/README.md
+++ b/auto-instrumentation/okhttp/okhttp-3.0/README.md
@@ -1,7 +1,5 @@
 # Android Instrumentation for OkHttp version 3.0 and higher
 
-# 🚧Work in progress 🚧
-
 Provides OpenTelemetry instrumentation for [okhttp3](https://square.github.io/okhttp/).
 
 ## Quickstart
--- a/auto-instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/library/okhttp/v3_0/internal/CachedSupplier.java
+++ b/auto-instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/library/okhttp/v3_0/internal/CachedSupplier.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.library.okhttp.v3_0.internal;
+
+import java.util.function.Supplier;
+
+/**
+ * This class is internal and is hence not for public use. Its APIs are unstable and can change at
+ * any time.
+ */
+public class CachedSupplier<T> implements Supplier<T> {
+    private Supplier<T> supplier;
+    private T instance;
+
+    public static <T> CachedSupplier<T> create(Supplier<T> instance) {
+        return new CachedSupplier<>(instance);
+    }
+
+    private CachedSupplier(Supplier<T> supplier) {
+        this.supplier = supplier;
+    }
+
+    @Override
+    public T get() {
+        synchronized (this) {
+            if (instance == null) {
+                instance = supplier.get();
+                supplier = null;
+            }
+            return instance;
+        }
+    }
+}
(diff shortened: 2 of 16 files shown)


auto-instrumentation/okhttp/okhttp-3.0/README.md
```diff
@@ -1,7 +1,5 @@
 # Android Instrumentation for OkHttp version 3.0 and higher
 
-# 🚧Work in progress 🚧
```
breedx-splk, line 3: Do you think it might also make sense to put a note about stability? Like experimental?

instrumentation/build.gradle.kts
```diff
@@ -79,8 +75,6 @@ dependencies {
     testImplementation(libs.androidx.test.core)
     testImplementation(libs.assertj.core)
     testImplementation(libs.awaitility)
-
-    coreLibraryDesugaring(libs.desugarJdkLibs)
```
breedx-splk, line 83: Why remove this at the top level?

auto-instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/library/okhttp/v3_0/internal/CachedSupplier.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    private CachedSupplier(Supplier<T> supplier) {
+        this.supplier = supplier;
+    }
+
+    @Override
+    public T get() {
+        synchronized (this) {
+            if (instance == null) {
```
breedx-splk, line 29: Two things stand out with this code:

1. I feel pretty strongly that you shouldn't `synchronzied(this)`. This uses a lock that is potentially exposed to other users of this instance, and because you can't control it, it opens this class up to locking problems (eg. deadlocks). Better to use a dedicated private lock instance.
2. If the supplier returns null, you will null out the supplier and then on next invocation you will get an NPE. Maybe the right move is to just keep the supplier and not null it out...or at least check to see that instance is not-null before doing so.

breedx-splk: There should also be unit test coverage for a new class like this one.

breedx-splk (changes requested):
This is cool, and it's nice to see it come together because there were several moving parts across repos. 🥂

auto-instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/library/okhttp/v3_0/internal/CachedSupplier.java
LikeTheSalad: > I feel pretty strongly that you shouldn't synchronzied(this). This uses a lock that is potentially exposed to other users of this instance, and because you can't control it, it opens this class up to locking problems (eg. deadlocks). Better to use a dedicated private lock instance.

Thanks, it's updated now.

> If the supplier returns null, you will null out the supplier and then on next invocation you will get an NPE. Maybe the right move is to just keep the supplier and not null it out...or at least check to see that instance is not-null before doing so.

Good point, though I really think the supplier must never return null, so I've just added a runtime exception in case that happens. I wouldn't like to keep the supplier if it's not needed to avoid wasting memory.

LikeTheSalad: > There should also be unit test coverage for a new class like this one.

Thanks, it's added now.

instrumentation/build.gradle.kts
LikeTheSalad: It was moved over to `otel.android-library-conventions.gradle.kts` to ensure that all android libs created in this repo are corelib-aware.

breedx-splk (approved):
Nice! Thanks thanks.

Issue #149, opened 2023-11-16
Defer PreferencesService to lazy/later and/or worker thread


breedx-splk:
As part of the first disk buffering PR #145, it was raised by @marandaneto that  the `PreferencesService` is loading shared preferences from the context, which does perform IO and will result in a warning when StrictMode is used. This suggests that this work should either be deferred until needed or moved to a worker thread.

See: 
https://github.com/open-telemetry/opentelemetry-android/pull/145#discussion_r1392352275

LikeTheSalad:
I think the solution should be to run all the initialization in a background thread and provide callbacks for the user to know when OTel is ready to be used.

The reason I mention this is because `PreferencesService` isn't the only blocking work needed during initialization. The calls for [checking available cache disk space](https://github.com/open-telemetry/opentelemetry-android/blob/50206e59f9eb7dccbf6b49d46b929f7ecdbd5c26/instrumentation/src/main/java/io/opentelemetry/android/internal/services/CacheStorageService.java#L46) should also be done out of the main thread, and it'll be needed to create signals exporters for initializing the OpenTelemetry instance. Plus (probably won't be needed over here) though at least in the Elastic agent, we make a request to an NTP server during initialization to get the real time for signals' timestamps, since mobile devices' clock isn't too reliable for accuracy. There might be other solutions that wouldn't involve an NTP server though, so that's why I'm not sure if it'll be added here, however, my point is that initializing OTel in the main thread seems like it could cause troubles in the long run if we need to add features that take some time to set up.

marandaneto:
> I think the solution should be to run all the initialization in a background thread and provide callbacks for the user to know when OTel is ready to be used.

In theory, it's ok but in practice, this has downsides, for example:
* Some features to be installed require to be in the main thread, which is fine, we can always post an execution to the main thread.
* For tracing instrumentation, the SDK must be init. as soon as possible, and we can only guarantee that if the SDK is init in the main thread, executing all the code sync.
* For crash reporting, it's also important that the SDK is init. as soon as possible, for example, to capture errors during app start.

Maybe an option is to init. the "core" parts of the SDK in the main thread and defer everything else to a worker.

Another option is just to defer the blocking bits to a worker (preferred IMO), and await the worker whenever needed even if that blocks for a tiny bit, an example of that is the storage init., it's gonna be init. in a worker, but if a trace or error has to be cached in storage and the worker is still ongoing, it's ok to `await` until it's finished.

[...]

Issue #142, opened 2023-11-06
Feature: Android OTEL Logging exporter


Archish27:
To Create logging solution where logs can be pushed using Log exporter. Like `Logger.d("Hello World")`

breedx-splk:
Thanks. Technically this is already available today -- you can configure logging and a log exporter when you (manually) configure your otel sdk instance. I do think that there is some room to improve the usability/ergonomics of this.

I would love to hear some input from users with ideas around this.

GerardPaligot:
@breedx-splk I guess that if a sample with Timber (https://github.com/JakeWharton/timber) could be great to know how to do this kind of things.

PR #160, opened 2023-11-25
Adding disk buffering, part 2

LikeTheSalad:
Related to #132 

This part connects the tools added in the previous part to the OpenTelemetryRumBuilder class.

#### For the next part, "Fun with Threads":

* Make sure the initialization happens in a secondary thread, more info: https://github.com/open-telemetry/opentelemetry-android/issues/149.
* Set up configurable, periodic calls to export cached data as background work.

### DIFF
--- a/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
+++ b/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
@@ -324,7 +324,7 @@ public final class OpenTelemetryRumBuilder {
             try {
                 spanExporter = createDiskExporter(defaultExporter, diskBufferingConfiguration);
             } catch (IOException e) {
-                LOGGER.log(Level.WARNING, "Could not create span disk exporter", e);
+                LOGGER.log(Level.WARNING, "Could not create span disk exporter.", e);
             }
         }
         return spanExporterCustomizer.apply(spanExporter);
--- a/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
+++ b/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
@@ -8,6 +8,7 @@ package io.opentelemetry.android;
 import static java.util.Objects.requireNonNull;
 
 import android.app.Application;
+import android.util.Log;
 import io.opentelemetry.android.config.DiskBufferingConfiguration;
 import io.opentelemetry.android.config.OtelRumConfig;
 import io.opentelemetry.android.instrumentation.InstrumentedApplication;
@@ -42,8 +43,6 @@ import java.util.List;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.Function;
-import java.util.logging.Level;
-import java.util.logging.Logger;
 
 /**
  * A builder of {@link OpenTelemetryRum}. It enabled configuring the OpenTelemetry SDK and disabling
@@ -64,7 +63,6 @@ public final class OpenTelemetryRumBuilder {
             loggerProviderCustomizers = new ArrayList<>();
     private final OtelRumConfig config;
     private final VisibleScreenTracker visibleScreenTracker = new VisibleScreenTracker();
-    private static final Logger LOGGER = Logger.getLogger("OpenTelemetryRumBuilder");
 
     private Function<? super SpanExporter, ? extends SpanExporter> spanExporterCustomizer = a -> a;
     private final List<Consumer<InstrumentedApplication>> instrumentationInstallers =
@@ -324,7 +322,7 @@ public final class OpenTelemetryRumBuilder {
             try {
                 spanExporter = createDiskExporter(defaultExporter, diskBufferingConfiguration);
             } catch (IOException e) {
-                LOGGER.log(Level.WARNING, "Could not create span disk exporter.", e);
+                Log.w(RumConstants.OTEL_RUM_LOG_TAG, "Could not create span disk exporter.", e);
             }
         }
         return spanExporterCustomizer.apply(spanExporter);
--- a/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
+++ b/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
@@ -42,6 +42,8 @@ import java.util.List;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * A builder of {@link OpenTelemetryRum}. It enabled configuring the OpenTelemetry SDK and disabling
@@ -62,6 +64,7 @@ public final class OpenTelemetryRumBuilder {
             loggerProviderCustomizers = new ArrayList<>();
     private final OtelRumConfig config;
     private final VisibleScreenTracker visibleScreenTracker = new VisibleScreenTracker();
+    private static final Logger LOGGER = Logger.getLogger("OpenTelemetryRumBuilder");
 
     private Function<? super SpanExporter, ? extends SpanExporter> spanExporterCustomizer = a -> a;
     private final List<Consumer<InstrumentedApplication>> instrumentationInstallers =
@@ -314,17 +317,15 @@ public final class OpenTelemetryRumBuilder {
     private SpanExporter buildSpanExporter() {
         // TODO: Default to otlp...but how can we make endpoint and auth mandatory?
         SpanExporter defaultExporter = LoggingSpanExporter.create();
-        SpanExporter spanExporter;
+        SpanExporter spanExporter = defaultExporter;
         DiskBufferingConfiguration diskBufferingConfiguration =
                 config.getDiskBufferingConfiguration();
         if (diskBufferingConfiguration.isEnabled()) {
             try {
                 spanExporter = createDiskExporter(defaultExporter, diskBufferingConfiguration);
             } catch (IOException e) {
-                spanExporter = defaultExporter;
+                LOGGER.log(Level.WARNING, "Could not create span disk exporter", e);
             }
-        } else {
-            spanExporter = defaultExporter;
         }
         return spanExporterCustomizer.apply(spanExporter);
     }
--- a/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
+++ b/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
@@ -8,6 +8,7 @@ package io.opentelemetry.android;
 import static java.util.Objects.requireNonNull;
 
 import android.app.Application;
+import io.opentelemetry.android.config.DiskBufferingConfiguration;
 import io.opentelemetry.android.config.OtelRumConfig;
 import io.opentelemetry.android.instrumentation.InstrumentedApplication;
 import io.opentelemetry.android.instrumentation.activity.VisibleScreenTracker;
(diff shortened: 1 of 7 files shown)


instrumentation/src/main/java/io/opentelemetry/android/internal/features/persistence/SimpleTemporaryFileProvider.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+public final class SimpleTemporaryFileProvider implements TemporaryFileProvider {
+    private final File tempDir;
+
+    public SimpleTemporaryFileProvider(File tempDir) {
+        this.tempDir = tempDir;
+    }
+
+    @Override
+    public File createTemporaryFile(String prefix) {
```
breedx-splk, line 22: The name suggests that maybe the file is actually _created_ on disk. Worth either a clarifying comment or rename to `getTemporaryFile()`? I guess it is an interface method...hmmmm... 🤔

instrumentation/src/main/java/io/opentelemetry/android/internal/features/persistence/SimpleTemporaryFileProvider.java
```diff
@@ -0,0 +1,25 @@
...
+        return new File(tempDir, prefix + "_" + System.currentTimeMillis() + ".tmp");
```
breedx-splk, line 23: Missing a unit test for this class.

instrumentation/src/test/java/io/opentelemetry/android/OpenTelemetryRumBuilderTest.java
```diff
@@ -158,6 +168,69 @@ void setSpanExporterCustomizer() {
...
+        doReturn(60 * 1024 * 1024L).when(cacheStorage).ensureCacheSpaceAvailable(anyLong());
+        setUpServiceManager(preferences, cacheStorage);
+        OtelRumConfig config = buildConfig();
+        config.setDiskBufferingConfiguration(
+                DiskBufferingConfiguration.builder().setEnabled(true).build());
+
+        OpenTelemetryRum.builder(application, config)
+                .addSpanExporterCustomizer(
+                        spanExporter -> {
+                            assertThat(spanExporter).isInstanceOf(SpanDiskExporter.class);
```
breedx-splk, line 184: Prefer not to have asserts nested in callbacks that might not have ever been called. This can lead to false sense of confidence. 😁  I think it's better to save the param in an `AtomicReference` in the callback and then do the assert after `build()` is called.

instrumentation/src/test/java/io/opentelemetry/android/OpenTelemetryRumBuilderTest.java
```diff
@@ -158,6 +168,69 @@ void setSpanExporterCustomizer() {
...
+    void diskBufferingEnabled_when_exception_thrown() {
+        PreferencesService preferences = mock();
+        CacheStorageService cacheStorage = mock();
+        doReturn(60 * 1024 * 1024L).when(cacheStorage).ensureCacheSpaceAvailable(anyLong());
+        doAnswer(
+                        invocation -> {
+                            throw new IOException();
+                        })
+                .when(cacheStorage)
+                .getCacheDir();
```
breedx-splk, line 200: this formatting looks weird...do we not have spotless working?

instrumentation/src/test/java/io/opentelemetry/android/OpenTelemetryRumBuilderTest.java
```diff
@@ -158,6 +168,69 @@ void setSpanExporterCustomizer() {
...
+        setUpServiceManager(preferences, cacheStorage);
+        OtelRumConfig config = buildConfig();
+        config.setDiskBufferingConfiguration(
+                DiskBufferingConfiguration.builder().setEnabled(true).build());
+
+        OpenTelemetryRum.builder(application, config)
+                .addSpanExporterCustomizer(
+                        spanExporter -> {
+                            assertThat(spanExporter).isNotInstanceOf(SpanDiskExporter.class);
```
breedx-splk, line 209: Same comment as above about the location of the assert.

instrumentation/src/test/java/io/opentelemetry/android/OpenTelemetryRumBuilderTest.java
```diff
@@ -158,6 +168,69 @@ void setSpanExporterCustomizer() {
...
+                        })
+                .build();
+    }
+
+    @Test
+    void diskBufferingDisabled() {
+        makeBuilder()
+                .addSpanExporterCustomizer(
+                        spanExporter -> {
+                            assertThat(spanExporter).isNotInstanceOf(SpanDiskExporter.class);
```
breedx-splk, line 220: same comment about assert location.

instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
```diff
@@ -308,7 +314,34 @@ private SdkTracerProvider buildTracerProvider(SessionId sessionId, Application a
...
         SpanExporter defaultExporter = LoggingSpanExporter.create();
-        return spanExporterCustomizer.apply(defaultExporter);
+        SpanExporter spanExporter;
+        DiskBufferingConfiguration diskBufferingConfiguration =
+                config.getDiskBufferingConfiguration();
+        if (diskBufferingConfiguration.isEnabled()) {
+            try {
+                spanExporter = createDiskExporter(defaultExporter, diskBufferingConfiguration);
+            } catch (IOException e) {
+                spanExporter = defaultExporter;
```
breedx-splk, line 324: Probably worth logging this exception so that there's a chance of seeing the problem. It's only once at startup time, so hopefully not considered too verbose.

instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
```diff
@@ -308,7 +314,34 @@ private SdkTracerProvider buildTracerProvider(SessionId sessionId, Application a
     private SpanExporter buildSpanExporter() {
         // TODO: Default to otlp...but how can we make endpoint and auth mandatory?
         SpanExporter defaultExporter = LoggingSpanExporter.create();
-        return spanExporterCustomizer.apply(defaultExporter);
+        SpanExporter spanExporter;
```
breedx-splk, line 317:
```suggestion
        SpanExporter spanExporter = defaultExporter;
```

If you do this assignment early here, you can remove the assignment in the `catch` block and the entire `else` block.

breedx-splk (changes requested):

instrumentation/src/main/java/io/opentelemetry/android/internal/features/persistence/SimpleTemporaryFileProvider.java
LikeTheSalad: Thanks, it's updated now.

instrumentation/src/test/java/io/opentelemetry/android/OpenTelemetryRumBuilderTest.java
LikeTheSalad: Good point, it's updated now.

instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
LikeTheSalad: Sounds good, I've added a log there.

instrumentation/src/main/java/io/opentelemetry/android/internal/features/persistence/SimpleTemporaryFileProvider.java
breedx-splk: Yeah fair enough, I had overlooked that the interface lived upstream. It's fine for now, but still a comment could be nice.

instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
```diff
@@ -308,7 +317,32 @@ private SdkTracerProvider buildTracerProvider(SessionId sessionId, Application a
...
         SpanExporter defaultExporter = LoggingSpanExporter.create();
-        return spanExporterCustomizer.apply(defaultExporter);
+        SpanExporter spanExporter = defaultExporter;
+        DiskBufferingConfiguration diskBufferingConfiguration =
+                config.getDiskBufferingConfiguration();
+        if (diskBufferingConfiguration.isEnabled()) {
+            try {
+                spanExporter = createDiskExporter(defaultExporter, diskBufferingConfiguration);
+            } catch (IOException e) {
+                LOGGER.log(Level.WARNING, "Could not create span disk exporter.", e);
```
breedx-splk, line 327: Actually, this makes me wonder if this shouldn't be using `android.util.Log` like the rest of the library. I'm noticing that the only places using `jul` are the `CacheStorageService` and the `DiskManager`. Is this intentional?

LikeTheSalad: This is a good point! Tbh I didn't notice the usage of Android Log until now, so I thought we'd configure some other framework to delegate to the Android Log at some point, but it seems like it'd be better to do so with slf4j or timber rather than jul now that I checked better. In any case, I'm fine with using the Android Log class directly, I've just replaced jul usages by Android Log ones across the project 👍

instrumentation/src/test/java/io/opentelemetry/android/internal/features/persistence/SimpleTemporaryFileProviderTest.java
```diff
@@ -0,0 +1,30 @@
...
+        SimpleTemporaryFileProvider provider = new SimpleTemporaryFileProvider(tempDir);
+        File first = provider.createTemporaryFile("a");
+        File second = provider.createTemporaryFile("b");
+        Thread.sleep(1);
+        File third = provider.createTemporaryFile("a");
+
+        assertThat(first.getName()).startsWith("a").endsWith(".tmp");
+        assertThat(second.getName()).startsWith("b").endsWith(".tmp");
+        assertThat(third.getName()).startsWith("a").endsWith(".tmp");
+        assertThat(first).isNotEqualTo(third);
```
breedx-splk, line 28: Isn't it important to test that the temp files have a path that's located within the temp dir provided to the provider? As it stands right now, you could change the impl:

```
new File(tempDir, prefix + "_" + System.currentTimeMillis() + ".tmp");
```
to
```
new File("hard-coded-path", prefix + "_" + System.currentTimeMillis() + ".tmp");
```

and these tests would still pass I think.

instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
breedx-splk: Cool, yeah I figure we'll have to pick this up eventually when we look at #142.

breedx-splk (approved):

instrumentation/src/main/java/io/opentelemetry/android/internal/features/persistence/SimpleTemporaryFileProvider.java
```diff
@@ -0,0 +1,26 @@
...
+    private final File tempDir;
+
+    public SimpleTemporaryFileProvider(File tempDir) {
+        this.tempDir = tempDir;
+    }
+
+    /** Creates a unique file instance using the provided prefix and the current time in millis. */
+    @Override
+    public File createTemporaryFile(String prefix) {
+        return new File(tempDir, prefix + "_" + System.currentTimeMillis() + ".tmp");
```
marandaneto, line 24: What if the file name is a UUID? it could happen that 2 files would have the same name and it conflicts, likely hard to happen though.

marandaneto: You can still sort files by timestamp if that's a concern, no need to rely on the file name.

instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
```diff
@@ -308,7 +315,32 @@ private SdkTracerProvider buildTracerProvider(SessionId sessionId, Application a
...
+            } catch (IOException e) {
+                Log.w(RumConstants.OTEL_RUM_LOG_TAG, "Could not create span disk exporter.", e);
+            }
+        }
+        return spanExporterCustomizer.apply(spanExporter);
+    }
+
+    private static SpanExporter createDiskExporter(
+            SpanExporter defaultExporter, DiskBufferingConfiguration diskBufferingConfiguration)
+            throws IOException {
```
marandaneto, line 333: Is it only IOEx or also securityexception?

instrumentation/src/test/java/io/opentelemetry/android/internal/features/persistence/SimpleTemporaryFileProviderTest.java
LikeTheSalad: This is a great point! I've just improved the tests to avoid this issue. Cheers!

instrumentation/src/main/java/io/opentelemetry/android/internal/features/persistence/SimpleTemporaryFileProvider.java
LikeTheSalad: I think it's really difficult for that case to happen, although I also can't say it's impossible... So I've just made the change to use UUID just to be extra safe 👍 cheers.

instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
LikeTheSalad: At the moment it's only IOExceptions thrown in case the dirs cannot be created for whatever reason.

PR #173, opened 2023-12-06
Improve version resource handling

### LINKED ISSUE #7: Rework rum_version in AndroidResource
This is leftover from the Splunk Android sdk. The idea is that there would be a string created at build time and put into the android resource which could be propagated to the OpenTelemetry Resource at runtime.

It is probably not working, or maybe not working as it should. We should consider these things:

* determine if we can just use [instrumentation scope](https://opentelemetry.io/docs/specs/otel/glossary/#instrumentation-scope) instead
* consider changing `rum_version` to another name like `android.instrumentation.version` or similar
  * pair this with a change to the semantic conventions
* tests to make sure this works as intended.

And when we're happy with this we can remove the TODO in `AndroidResource`.

breedx-splk:
Resolves #7

This started as addressing a warning about a deprecated property in the `gradle.properties` file. After scratching deeper, it seemed like a good time to align with the [upstream java version handling](https://github.com/open-telemetry/opentelemetry-java/blob/main/buildSrc/src/main/kotlin/io/opentelemetry/gradle/OtelVersionClassPlugin.kt) by generating a class at build time. 

This now allows us to decouple from the application when determining the otel android version number at runtime. Overall should be more robust, and lets us put away the dragging #7 issue. 🍻


marandaneto:
An option is to just use a plugin to generate such things eg https://github.com/gmazzo/gradle-buildconfig-plugin
Look at the `APP_VERSION` example

buildSrc/src/main/kotlin/io/opentelemetry/gradle/OtelAndroidVersionClassPlugin.kt
LikeTheSalad, line 1: This is a pretty clever solution, also, the tool mentioned by @marandaneto is very useful for non-android projects to generate static fields at build time, I've used it in the past for my non-android projects and so far so good.

However, for this case I think it's worth trying Android's built-in compile-time [field generator tool](https://developer.android.com/build/gradle-tips#share-custom-fields-and-resource-values-with-your-app-code) which should provide the same benefits but without having to write custom compile-time logic nor having to include extra dependencies.

We could make it work in this project like so:

```kt
// instrumentation/build.gradle.kts

android {
  // ...
  defaultConfig {
    // ...
    buildConfigField("String", "OTEL_ANDROID_VERSION", "\"$version\"")
  }
  // ...
}
```

And after compiling the project we could access it in the code like so:

```java
import io.opentelemetry.android.BuildConfig;

String theVersion = BuildConfig.OTEL_ANDROID_VERSION;
```

marandaneto: yep, the [gradle-buildconfig-plugin](https://github.com/gmazzo/gradle-buildconfig-plugin) plugin is a non-Android version of the `buildConfigField` method, so you can use the very same `BuildConfig` approach for non-Android projects, on Android, just use `buildConfigField` of course.

breedx-splk:
@marandaneto @LikeTheSalad Thanks for taking a look at this. I do agree that the built-in buildconfig yields a much simpler and more maintainable solution. Lemme close this and submit a new one with that approach. 🥂

Issue #49, opened 2023-08-25
SimpleNetworkDetector uses deprecated APIs


breedx-splk:
There is a build warning about the `SimpleNetworkDetector` using deprecated APIs. I believe that this is specifically related to the use of `NetworkInfo` and some of the methods (see the comments in `SimpleNetworkDetector`).

The [Elastic android instrumentation](https://github.com/elastic/apm-agent-android/blob/8c8a5d4dc420b5f3e5df29f3f2b0d15b20671ff1/android-sdk/src/main/java/co/elastic/apm/android/sdk/internal/services/network/NetworkService.java#L40) seems to agree with [a long-lived stackoverflow post](https://stackoverflow.com/questions/32547006/connectivitymanager-getnetworkinfoint-deprecated) about how to mitigate this. 

Although things are probably still working, it would be nice to migrate to a more sophisticated/robust solution going forward.

breedx-splk:
Then again, because the `SimpleNetworkDetector` is really only wired up in older SDKs, suppressing the warning might be appropriate.

LikeTheSalad:
There was an important API addition in the way Android provides a "default" network connectivity info on API 24 that makes this a lot easier, which is the one we use in the Elastic agent. I think it's worth waiting on the outcome of https://github.com/open-telemetry/opentelemetry-java/issues/5936 before adding any further changes to `SimpleNetworkDetector` or to the overall network detection mechanism.

LikeTheSalad:
As a side note, it seems like the overall network detection process is started [here](https://github.com/open-telemetry/opentelemetry-android/blob/5db424c7caf1a55f0ef9623cc50b60a04dbd8511/instrumentation/src/main/java/io/opentelemetry/android/instrumentation/network/CurrentNetworkProvider.java#L40), although I can't find where this function is called, so I'm wondering if this is currently dead code @breedx-splk?

breedx-splk:
> As a side note, it seems like the overall network detection process is started [here](https://github.com/open-telemetry/opentelemetry-android/blob/5db424c7caf1a55f0ef9623cc50b60a04dbd8511/instrumentation/src/main/java/io/opentelemetry/android/instrumentation/network/CurrentNetworkProvider.java#L40), although I can't find where this function is called, so I'm wondering if this is currently dead code @breedx-splk?

[...]

PR #176, opened 2023-12-08
Add network change monitoring

breedx-splk:
The instrumentation that reports network changes has not yet been wired up. This adds it to the configuration and enables the instrumentation in the builder's `build()`.

One additional aspect is that the `CurrentNetworkProvider` now has a setter on the `OpenTelemetryRumBuilder`. This allows existing components that may rely on this functionality to share the instance.  The creation is lazy, so that if the builder user didn't create one, it is created at build time on demand.

I expect this `CurrentNetworkProvider` sharing to hopefully not be necessary in the long term, but right now there are existing disk buffering implementations (eg. Splunk's distro) that also need to determine information about the current network availability before attempting export.


LikeTheSalad (approved):
Cool! I'm planning to add some changes in the future to the way the network is monitored due to the [deprecated Google APIs issue](https://github.com/open-telemetry/opentelemetry-android/issues/49)... But in the meantime, I think it should be fine to use it as it is now.

breedx-splk:
Yeah, cool, long term I think we'll still need to:

* Consider hiding the `CurrentNetworkProvider` and provide a more elegant way for users to get notified on otel-compatible network change events.
* Consider if we need to check current network status before attempting to send from disk-buffered telemetry. Could cut down on failures/exceptions/unnecessary attempts and in worse case maybe throwing buffered data away.

PR #181, opened 2023-12-11
Enable core desugaring for instrumentation

breedx-splk:
This is needed to fix the currently broken build.


marandaneto:
I guess this is being addressed here https://github.com/open-telemetry/opentelemetry-android/pull/180 ?

breedx-splk:
Yeah, sorry I had missed that it was being addressed in the prior PR (#180).

PR #182, opened 2023-12-11
Use the BuildConfig for version property

### LINKED ISSUE #7: Rework rum_version in AndroidResource

breedx-splk:
Resolves #7. This uses an AGP built-in to support for generating BuildConfig and adds the version number as a constant field. It then updates the AndroidResource to use this field instead of the resource string.

This also removes the `android.defaults.buildfeatures.buildconfig` deprecated gradle property.

### DIFF
--- a/instrumentation/build.gradle.kts
+++ b/instrumentation/build.gradle.kts
@@ -63,7 +63,6 @@ dependencies {
     implementation(libs.androidx.appcompat)
     implementation(libs.androidx.core)
     implementation(libs.androidx.navigation.fragment)
-    coreLibraryDesugaring(libs.desugarJdkLibs)
 
     api(platform(libs.opentelemetry.platform))
     api(libs.opentelemetry.api)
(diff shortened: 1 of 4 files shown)


marandaneto (approved):

instrumentation/build.gradle.kts
```diff
@@ -51,12 +53,17 @@ android {
...
+    buildFeatures {
+        buildConfig = true
+    }
 }
 
 dependencies {
     implementation(libs.androidx.appcompat)
     implementation(libs.androidx.core)
     implementation(libs.androidx.navigation.fragment)
+    coreLibraryDesugaring(libs.desugarJdkLibs)
```
breedx-splk, line 66:
```suggestion
```

Issue #157, opened 2023-11-21
Update versions in release process


breedx-splk:
We should have a process to semi/automatically update the version number(s) in the README and elsewhere whenever a new release is made. This can be a simple as mentioning it in the `RELEASING.md` doc, but other repos have some scripted automation to help with this (and reduce the burden and chance of error).

fractalwrench:
Closing this issue as it seems obsolete given that #1428 automatically bumped the versions

PR #188, opened 2023-12-14
Automatically create and push tag when doing release builds

### LINKED ISSUE #187: Create tag or branch automatically as part of the release process
So far, releases have had a fair amount of manual work. The creation of a tag from which the release is created is manual and error prone. Upstream creates release branches [like this](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/.github/workflows/prepare-release-branch.yml#L69). We haven't done a patch release, but having a branch in place might be nice to have, as well as being consistent.

As part of the release workflow, we should automatically create a release branch.

breedx-splk:
Resolves #187

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -39,4 +39,13 @@ jobs:
           SONATYPE_KEY: ${{ secrets.SONATYPE_KEY }}
           GPG_PRIVATE_KEY: ${{ secrets.GPG_PRIVATE_KEY }}
           GPG_PASSWORD: ${{ secrets.GPG_PASSWORD }}
+      - name: Create release tag for release builds
+        env:
+          GH_TOKEN: ${{ secrets.OPENTELEMETRYBOT_GITHUB_TOKEN }}
+        if: "${{ inputs.release_type == 'final' }}"
+        run: |
+          VERSION_NUMBER=$(grep ^version= gradle.properties  | sed -e 's/^.*=//')
+          VERSION="v${VERSION_NUMBER}"
+          git tag ${VERSION}
+          git push origin ${VERSION}
 


LikeTheSalad (approved):
Nice!

.github/workflows/release.yml
```diff
@@ -39,4 +39,13 @@ jobs:
...
           GPG_PASSWORD: ${{ secrets.GPG_PASSWORD }}
+      - name: Create release tag for release builds
+        env:
+          GH_TOKEN: ${{ secrets.OPENTELEMETRYBOT_GITHUB_TOKEN }}
+        if: "${{ inputs.release_type == 'final' }}"
+        run: |
+          VERSION_NUMBER=$(grep ^version= gradle.properties  | sed -e 's/^.*=//')
+          VERSION="v${VERSION_NUMBER}"
+          git tag ${VERSION}
+          git push origin ${VERSION}
```
marandaneto, line 50: to push tags you need `--tags` IIRC

marandaneto (approved):

.github/workflows/release.yml
```diff
@@ -39,4 +39,13 @@ jobs:
           SONATYPE_KEY: ${{ secrets.SONATYPE_KEY }}
           GPG_PRIVATE_KEY: ${{ secrets.GPG_PRIVATE_KEY }}
           GPG_PASSWORD: ${{ secrets.GPG_PASSWORD }}
+      - name: Create release tag for release builds
+        env:
+          GH_TOKEN: ${{ secrets.OPENTELEMETRYBOT_GITHUB_TOKEN }}
+        if: "${{ inputs.release_type == 'final' }}"
+        run: |
+          VERSION_NUMBER=$(grep ^version= gradle.properties  | sed -e 's/^.*=//')
```
marandaneto, line 47: Not for this PR but just adding a suggestion, upgrading the `gradle.properties` and creating the GH release could be part of the release automation as well.

`gradle.properties` could be upgraded with a bash script called in this workflow, and the version would be gotten from the `workflow_dispatch` as well.

The GH release could be created with an Action, eg [action-gh-release](https://github.com/softprops/action-gh-release) or similar.

breedx-splk: Yeah definitely! I think there's a ton of room for incremental improvements to the release process. The upstream repos have gotten quite sophisticated -- including generating the contributor list etc. 👍🏻

breedx-splk: See #157 as well. We're open for contributions. :)

Issue #82, opened 2023-09-15
Migrating this repo codebase to Kotlin


surbhiia:
As Kotlin is being increasingly used for development of android applications and is now the [preferred language by Google](https://techcrunch.com/2019/05/07/kotlin-is-now-googles-preferred-language-for-android-app-development/), it might be worth migrating this repo codebase to Kotlin. [This](https://insights.stackoverflow.com/survey/2019#most-loved-dreaded-and-wanted) 2019 stack overflow survey shows kotlin is in the top 5 most loved languages.

yogurtearl:
Writing your Android SDK in Java and not Kotlin is akin to writing your iOS SDK in objective-C and not Swift.

> Even though Kotlin is the indisputable winner when choosing a language for Android projects, it is technically possible to create an Android project using Java only, in those cases, if we decide to use Kotlin as part of the core of this library, we'd be enforcing Kotlin's transitive dependencies on the host project, such as Kotlin's standard lib

As noted above, this is a non-issue and Google imposes this dependency for apps using `androidx`, which is all but required for a modern app. 

I can't imagine an app being modern enough to use otel-android while not using `androidx` 

From https://developer.android.com/kotlin/stories

> Read more to see why over 95% of the top 1000 Android apps use Kotlin.

> Android devs might not be the only ones interested in reading and modifying code related to observability, maybe people from DevOps teams for example might want to have a look and create a PR, and in those cases, since Java has been around longer than Kotlin, it might be easier for someone without an Android background to understand it rather than Kotlin, making a wider range of people able to quickly contribute some changes.

Alot of DevOps teams have golang experience, and I would imagine Kotlin is easier to digest for a golang developer than Java.

yogurtearl:
Also, this repo is already using kotlin for the build scripts (`*.build.kts`), so it has a nice synergy there.

bherbst:
In addition to the direct benefits to the opentelemetry-android codebase and the benefits to consumers in using the APIs, I suspect that opentelemetry-android will soon reach a point where shipping the Kotlin stdlib is inevitable and trying to avoid doing so would limit the utility the library can provide.

In particular, many AndroidX libraries such as androidx-lifecycle are already written in Kotlin, and we could benefit from leveraging those APIs. 

[...]

PR #189, opened 2023-12-14
Adding Kotlin compile support

LikeTheSalad:
Related to [this discussion](https://github.com/open-telemetry/opentelemetry-android/issues/82).

### DIFF
--- a/buildSrc/src/main/kotlin/otel.android-app-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.android-app-conventions.gradle.kts
@@ -1,8 +1,14 @@
+import org.jetbrains.kotlin.gradle.dsl.JvmTarget
+import org.jetbrains.kotlin.gradle.dsl.KotlinVersion
+
 plugins {
     id("com.android.application")
+    id("org.jetbrains.kotlin.android")
     id("otel.errorprone-conventions")
 }
 
+val javaVersion = rootProject.extra["java_version"] as JavaVersion
+val minKotlinVersion = rootProject.extra["kotlin_min_supported_version"] as KotlinVersion
 android {
     namespace = "io.opentelemetry.android"
     compileSdk = (property("android.compileSdk") as String).toInt()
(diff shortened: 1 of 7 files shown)


buildSrc/src/main/kotlin/otel.android-library-conventions.gradle.kts
```diff
@@ -18,13 +24,18 @@ android {
...
         sourceCompatibility(javaVersion)
         targetCompatibility(javaVersion)
         isCoreLibraryDesugaringEnabled = true
     }
 }
 
+kotlin {
+    compilerOptions {
+        jvmTarget.set(JvmTarget.fromTarget(javaVersion.toString()))
+        languageVersion.set(minKotlinVersion)
```
marandaneto, line 36:
```suggestion
        languageVersion.set(minKotlinVersion)
        apiVersion.set(minKotlinVersion)
```

marandaneto: Probably has to change in the other places as well

LikeTheSalad: Cool, I'll make the changes 👍 tbh the docs for that option weren't too clear for me so I wasn't sure if we needed it. Though I guess is better to have it to avoid unexpected issues. We should probably add it in the core repo too, as it's only got [languageVersion set](https://github.com/open-telemetry/opentelemetry-java/pull/5910).

buildSrc/src/main/kotlin/otel.android-app-conventions.gradle.kts
```diff
@@ -14,12 +20,17 @@ android {
     }
 
     compileOptions {
-        val javaVersion = rootProject.extra["java_version"] as JavaVersion
         sourceCompatibility(javaVersion)
         targetCompatibility(javaVersion)
         isCoreLibraryDesugaringEnabled = true
     }
 }
+kotlin {
```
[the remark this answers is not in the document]
LikeTheSalad: It seems like the android.kotlinOptions is just a wrapper of the other config, though yeah it's usually the way Android suggests to set Kotlin stuff so I've made the changes 👍

breedx-splk (approved):
LGTM. One step closer! Thanks.

Issue #130, opened 2023-10-27
Set OTLP exporters as default


LikeTheSalad:

LikeTheSalad:
@breedx-splk - regarding:

https://github.com/open-telemetry/opentelemetry-android/blob/5db424c7caf1a55f0ef9623cc50b60a04dbd8511/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java#L251

Maybe they could be part of the new config object if we decide to allow custom exporters in the config in which case I think they wouldn't need to be mandatory.

breedx-splk:
If you use the initializer, you get OTLP by default.

PR #191, opened 2023-12-15
Make span debug logging a first-class config component

breedx-splk:
This is baby steps on #130.

It does include the otlp http exporter, but just uses the defaults (which is localhost, not what most android users will want 😄 ).

This allows us to first pull the logging exporter into a configurable setting that is disabled by default. When enabled, it wires up another separate span processor to do the logging. The logging exporter also has the customizer applied.

### DIFF
--- a/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
+++ b/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
@@ -27,6 +27,7 @@ import io.opentelemetry.context.propagation.TextMapPropagator;
 import io.opentelemetry.contrib.disk.buffering.SpanDiskExporter;
 import io.opentelemetry.contrib.disk.buffering.internal.StorageConfiguration;
 import io.opentelemetry.exporter.logging.LoggingSpanExporter;
+import io.opentelemetry.exporter.otlp.http.trace.OtlpHttpSpanExporter;
 import io.opentelemetry.sdk.OpenTelemetrySdk;
 import io.opentelemetry.sdk.logs.SdkLoggerProvider;
 import io.opentelemetry.sdk.logs.SdkLoggerProviderBuilder;
@@ -37,6 +38,7 @@ import io.opentelemetry.sdk.trace.SdkTracerProvider;
 import io.opentelemetry.sdk.trace.SdkTracerProviderBuilder;
 import io.opentelemetry.sdk.trace.SpanProcessor;
 import io.opentelemetry.sdk.trace.export.BatchSpanProcessor;
+import io.opentelemetry.sdk.trace.export.SimpleSpanProcessor;
 import io.opentelemetry.sdk.trace.export.SpanExporter;
 import java.io.IOException;
 import java.util.ArrayList;
@@ -75,6 +77,7 @@ public final class OpenTelemetryRumBuilder {
 
     private Resource resource;
     @Nullable private CurrentNetworkProvider currentNetworkProvider = null;
+    @Nullable private SpanExporter spanExporter = null;
     private InitializationEvents initializationEvents = InitializationEvents.NO_OP;
 
     private static TextMapPropagator buildDefaultPropagator() {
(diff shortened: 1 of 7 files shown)


instrumentation/src/main/java/io/opentelemetry/android/export/SpanDataModifier.java
```diff
@@ -127,7 +118,7 @@ public <T> SpanDataModifier replaceSpanAttribute(
         return this;
     }
 
-    public SpanExporter build() {
+    public SpanExporter wrap(SpanExporter delegate) {
```
breedx-splk, line 121: Sorry that this is kinda unrelated to the main change, but I noticed this and decided that `wrap()` was a more fitting semantic than a build method not on a builder that returns an unrelated type.

instrumentation/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
```diff
@@ -24,6 +24,7 @@ public class OtelRumConfig {
     private DiskBufferingConfiguration diskBufferingConfiguration =
             DiskBufferingConfiguration.builder().build();
     private boolean networkChangeMonitoringEnabled = true;
+    private boolean debugLogEnabled = false;
```
LikeTheSalad, line 27: If I look at this option name without knowing anything about this repo, my first impression of it would be that it's related to turning on/off the internal logs that are produced by this lib. Maybe if I wasn't too knowledgeable about OTel in general, I might even think that it might have something to do with OTel log signals... In any case, I think the name doesn't convey that it's related to logging Spans at the time of exporting.

instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
```diff
@@ -317,6 +320,21 @@ private void applyConfiguration() {
                         return tracerProviderBuilder.addSpanProcessor(screenAttributesAppender);
                     });
         }
+
+        // Wire up the logging exporter, if enabled.
```
LikeTheSalad, line 324: My understanding about the use of `LoggingSpanExporter` was as a temporary replacement for Zipkin's until we implemented the OTLP exporters. So I'm a bit confused by these changes because we're now essentially adding the logger exporter as a first-class citizen feature but I'm not sure that the demand for this kind of feature is high enough to provide official support for it, or to justify the [extra dependency](https://github.com/open-telemetry/opentelemetry-java/tree/main/exporters/logging) needed to make it work. Especially because it seems like the logs are printed using JUL, which doesn't seem too useful for Android in general, and also because I think it's easy to add it (or any other exporter that logs the signals) if needed by using the existing customizer apis.

instrumentation/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
breedx-splk: Yeah that's a good call out. I think a more precise name is in order -- maybe something like `debugSpanLoggingEnabled` or similar.

instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
breedx-splk: It's been my experience that developers want to see their generated app telemetry well in advance of actually sending it somewhere meaningful, and that's where the logging exporter comes into play.

You can definitely see the JUL logs with `logcat` when running in the emulator. I think you might also be able to see them with adb?

In any case, that's fine if you think this is premature. We can wait until people ask for it.

LikeTheSalad: I think the idea that people might want to see what they're sending in logs is reasonable, what I'm not sure about is if it needs to be an officially supported feature here, especially with how easy it is to add that functionality by using the existing customizers and the flexibility that it can provide when it comes to choosing how to print those logs too, as opposed to enforcing JUL an also an extra dependency, for a feature that seems mostly useful for development purposes.

PR #192, opened 2023-12-15
Additional instrumentation configs

breedx-splk:
Add configs and automatically wire up instrumentation for:

* ANR (application not responding) detector
* Crash reporting
* Slow rendering detection

All of these instrumentations will be enabled by default, and there are config methods to disable them. There is also config to specify the polling interval for slow rendering.

Furthermore, the `OpenTelemetryRumBuilder` also allows users to provide customizers for the `CrashReporterBuilder` and `AnrDetectorBuilder`. These can be used by users or 3rd party distributions in order to provide custom/specific attribute extraction.

### DIFF
--- a/instrumentation/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
+++ b/instrumentation/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
@@ -8,6 +8,7 @@ package io.opentelemetry.android.config;
 import io.opentelemetry.android.ScreenAttributesSpanProcessor;
 import io.opentelemetry.android.instrumentation.network.CurrentNetworkProvider;
 import io.opentelemetry.api.common.Attributes;
+import java.time.Duration;
 import java.util.function.Supplier;
 
 /**
@@ -17,6 +18,9 @@ import java.util.function.Supplier;
  */
 public class OtelRumConfig {
 
+    private static final Duration DEFAULT_SLOW_RENDERING_DETECTION_POLL_INTERVAL =
+            Duration.ofSeconds(1);
+
     private Supplier<Attributes> globalAttributesSupplier = Attributes::empty;
     private boolean includeNetworkAttributes = true;
     private boolean generateSdkInitializationEvents = true;
@@ -25,6 +29,9 @@ public class OtelRumConfig {
             DiskBufferingConfiguration.builder().build();
     private boolean networkChangeMonitoringEnabled = true;
     private boolean anrDetectionEnabled = true;
+    private boolean slowRenderingDetectionEnabled = true;
+    private Duration slowRenderingDetectionPollInterval =
+            DEFAULT_SLOW_RENDERING_DETECTION_POLL_INTERVAL;
 
     /**
      * Configures the set of global attributes to emit with every span and event. Any existing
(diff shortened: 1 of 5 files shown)


marandaneto (approved):

instrumentation/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
```diff
@@ -17,13 +18,21 @@
...
     private Supplier<Attributes> globalAttributesSupplier = Attributes::empty;
     private boolean includeNetworkAttributes = true;
     private boolean generateSdkInitializationEvents = true;
     private boolean includeScreenAttributes = true;
     private DiskBufferingConfiguration diskBufferingConfiguration =
             DiskBufferingConfiguration.builder().build();
     private boolean networkChangeMonitoringEnabled = true;
+    private boolean anrDetectionEnabled = true;
+    private boolean slowRenderingDetectionEnabled = true;
+    private Duration slowRenderingDetectionPollInterval =
```
LikeTheSalad, line 33: What do you think about making feature configs, that require more parameters than just a flag for enabling/disabling it, to have their own subconfig object where we wouldn't have to repeat the same prefixes for different fields/setters? Similar to what's done with the disk buffering one. I think it's worth considering because it could help to keep things more organized and probably avoid making the OtelRumConfig file too big as well.

LikeTheSalad (approved):
LGTM! I've left one comment related to future usability and maintainability. I'd like to see what do you think about it. Though other than that, it's looking good.

instrumentation/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
breedx-splk: Thanks. Yeah, as a general principle, I agree with this...and I think we will need to show care that the config doesn't balloon out of control. In the above, though, you highlighted 3 lines, but ANR and slow rendering detection are two separate features.

I think that pulling out those two into their own config is a bit premature? Maybe we can use the rule of threes?

LikeTheSalad: It looks like I highlighted the three lines though I just added the comment on line 33 😅 

Though I think I get what you're saying... It could get too verbose having to create a configuration object composed of other configuration objects and their builders and so on just to set a flag and maybe an extra param... However, I do believe that there's room for a lot of features and I'm really concerned about having to maintain the kind of "god object" that this could become. 

At the same time, I'd like to keep consistency across the overall usage of the OTel Android API, so for example, if we mix config "plain" params in `OtelRumConfig` along with config "objects" we could end up with an ugly and thus probably cumbersome to use API.

Regarding:

> Maybe we can use the rule of threes?

It sounds fair enough, although I'm not sure it could work well for scalability though, what happens if we add more config params for the same feature in future releases? I don't think that should make us introduce breaking changes just to comply with the rule of threes, so we'll end up adding the extra params straight into `OtelRumConfig` and then we'd have an even more mixed up API.

Now don't get me wrong, I'm not saying that having config objects inside `OtelRumConfig` is the right solution either, though what I can say for sure is that, from previous experiences with API configs, it's usually quite challenging to come up with something that's flexible enough to allow almost any use case as well as clean enough to be predictable in how it should be extended in future iterations while keeping it simple enough to get it up and running for simple use cases but not too opinionated to prevent more complex use cases to work smoothly as well. So in other words, the config part is a big deal for me 😄 

While I was writing this comment I thought of an alternative that might tick all those boxes without having to add config objects inside `OtelRumConfig` by using something I saw that is already there in the current codebase that allows to "install" instrumentations which seems really handy for these kinds of scenarios. I'll create an issue explaining the details to discuss it over there.

PR #194, opened 2023-12-18
Adding disk buffering, part 3

LikeTheSalad:
It should take just one more PR after this one!

This PR is for setting the scaffolding that will allow to periodically read and export previously stored data from the disk. This is needed because the previous PRs related to disk buffering have taken care of the writing part only, so now we're going to set up the reading part which has to happen periodically (and in a background thread) to make sure new data is exported at some point.

### Goals for these changes

* Providing an API for scheduling periodic reads/exports of previously stored data.
* The provided API allows for custom implementations so that projects can decide the mechanism they want to use to trigger a read/export action periodically.
* A default implementation for the API is provided with a **very basic** functionality that ensures the following:
  * Polls and exports the data in disk periodically in the background while the app is running.
  * Doesn't rely on extra dependencies to ensure it won't negatively affect the lib size.
  * Doesn't create permanent running secondary threads if it's not needed to avoid consuming the device's resources unnecessarily.

### For the next part

Which seems like it will be the last one, we should take care of:

* Making the default scheduler to read and export the available data on disk.
* Use the disk buffering configuration, when the agent is initialized, to start the scheduling process with the provided scheduler or the default one if no custom scheduler is provided in the configs.


instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/scheduler/DefaultExportScheduleHandler.kt
```diff
@@ -0,0 +1,25 @@
...
+
+    override fun enable() {
+        if (!enabled.getAndSet(true)) {
+            ServiceManager.get().getService(PeriodicWorkService::class.java)
+                .enqueue(DefaultExportScheduler())
+        }
+    }
+
+    override fun disable() {
+        // No operation.
```
marandaneto, line 23: Should we flip `enabled` here and stop the `ServiceManager` as well?

instrumentation/src/main/java/io/opentelemetry/android/internal/services/periodicwork/PeriodicWorkService.kt
```diff
@@ -0,0 +1,76 @@
...
+                queue.add(runnable)
+            }
+
+        override fun run() {
+            delegateToWorkerThread()
+            scheduleNextLookUp()
+        }
+
+        private fun delegateToWorkerThread() =
+            synchronized(this) {
```
marandaneto, line 66: This is causing contention on the `enqueue` method since it synchronizes the whole thing until the queue is empty (which might take a bit of time, I guess?).
Maybe an option is to use a `ConcurrentLinkedQueue` instead and just poll its items (as a copy) and run the while loop on the copy of the queue, so the queue is released and ready to be used again.

instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/scheduler/ExportScheduleHandler.kt
```diff
@@ -0,0 +1,23 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.android.features.diskbuffering.scheduler
+
+/**
+ * Sets up a scheduling mechanism to read and export previously stored signals in disk.
+ */
+interface ExportScheduleHandler {
```
marandaneto, line 11: Not for this PR:
If using the Kotlin `explicitApi` mode, all those interfaces have to be made `public`, then you can use the `org.jetbrains.kotlinx.binary-compatibility-validator` plugin to generate a public API file its quite nice.

instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/scheduler/DefaultExportScheduleHandler.kt
LikeTheSalad: We could flip `enabled`, although `ServiceManager` shouldn't be affected by the disk buffering behavior. Probably the name is a bit misleading, but `ServiceManager` is just a set of "services" (not all of them are constantly running, so wouldn't need to be "stopped") where we can access tools globally, especially tools related to Android OS APIs which tend to require an android `Context` and also different ways to use them depending on the device's API level... Which is annoying and difficult to maintain if we made those calls all over the project, so `ServiceManager` is a convenient place to have those tools available everywhere and protecting the rest of the code from future API breaking changes. It's also nice to have tools like `PeriodicWorkService` which allows to centralize any background work needed across multiple features, such as the disk buffering exports in this case, and in the future it might help with NTP calls as well. There's more info on ServiceManager [here](https://github.com/open-telemetry/opentelemetry-android/issues/129).

marandaneto: Gotcha, `disable` can be a default method then (NoOp during the declaration), and it's only implemented if necessary.

instrumentation/src/main/java/io/opentelemetry/android/internal/services/periodicwork/PeriodicWorkService.kt
LikeTheSalad: This is a good point, since the `executor.execute()` function should just move the item to the executor's internal queue, I was thinking that the overhead shouldn't be too much. But it's definitely better to go extra safe for these kinds of scenarios, let me try it out 👍

instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/scheduler/ExportScheduleHandler.kt
LikeTheSalad: That sounds really nice 👍 I think we should create an issue to address it later.

instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/scheduler/DefaultExportScheduleHandler.kt
LikeTheSalad: Thanks, good point. I've just made the changes.

instrumentation/src/main/java/io/opentelemetry/android/internal/services/periodicwork/PeriodicWorkService.kt
LikeTheSalad: Thank you, I've made the changes and it all seems to be working well 👍 It's definitely better to avoid blocking ops whenever possible.

marandaneto (approved):
noice!

breedx-splk (commented):
There do appear to be a lot of moving parts here...There's a `PeriodicWorkService` that delegates to the `WorkDelegator` that, when run, delegates all of its enqueued items to be run later in its `ExecutorService` and also tells a `Handler` to post a callback run on the `MainLooper`, which will repeat the process (every 10 seconds).

There's also this `PeriodicRunnable` which is a `Runnable` that when `run()`s checks to see if enough time has passed and yes then it actually runs via the `onRun()` template method, but it always reenqueues itself back into the `PeriodicWorkService` global service instance, which it finds though the global `ServiceManager` instance.

Phew. That's a lot to keep track of for something that we just want to run every 10 seconds. 

Oh yeah, and I haven't even mentioned yet that there's the `DefaultExportScheduler` which is an implementation of `PeriodicRunnable`.

There's something that's eating at me about the period in the `PeriodicRunnable` and the scheduled time in the `PeriodicWorkService` and how they might be at odds with each other. If you have a `PeriodicRunnable` that wants to be run every second, but the `PeriodicWorkService` only can schedule work every 10 seconds, then doesn't that effectively make the `PeriodicRunnable` period 10 seconds as well? In other words, maybe it can only run as fast as its service?

In any case, there is probably something in the merits of using the `Handler`/`Looper` mechanism that I need to read up on...but I wonder why we can't just use a plain old `ScheduledExecutorService` for all of this?

instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfiguration.java
```diff
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package io.opentelemetry.android.config;
+package io.opentelemetry.android.features.diskbuffering;
```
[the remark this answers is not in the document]
LikeTheSalad: It's definitely not needed to reorganize it this way, it's more of a personal preference of mine to have all related things into a single package because I've had issues in the past tracking down related classes that were spread across a project by type. For example, projects with a package named "activities" had all the activities of the app in it, each one representing a feature that needed other classes to work properly but that couldn't be in the same package because they were not of type Activity, and so whenever we needed to do some refactoring or removing some feature we had to go hunting for all the related classes in the project and it wasn't fun 😩 

However, I'm open to suggestions and/or debating what kind of structure could fit better this project! These are the kinds of things that make me want to have a dedicated Android SIG tbh, it seems like the communication would flow a bit faster having that option.

instrumentation/src/main/java/io/opentelemetry/android/internal/tools/time/SystemTime.kt
```diff
@@ -0,0 +1,28 @@
...
+
+package io.opentelemetry.android.internal.tools.time
+
+/**
+ * Utility to be able to mock the current system time for testing purposes.
+ *
+ * <p>This class is internal and not for public use. Its APIs are unstable and can change at any
+ * time.
+ */
+internal interface SystemTime {
```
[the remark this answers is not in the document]
LikeTheSalad: True, I still like organizing stuff as much as possible because it'd make it easier for us to find things later if the project becomes too big. But from the user's perspective you're right, just by using this modifier we won't have to be worried about the misuse of these classes (as long as they're used in Kotlin code since you can still access kotlin internal stuff from Java code, though that's probably not something to be too worried about for an Android lib).

instrumentation/src/main/java/io/opentelemetry/android/internal/tools/time/SystemTime.kt
```diff
@@ -0,0 +1,28 @@
...
+    companion object {
+        private var instance: SystemTime = DefaultSystemTime()
```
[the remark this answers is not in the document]
LikeTheSalad: Yeah it makes sense, I'll add the changes.

LikeTheSalad:
@breedx-splk regarding:

> There do appear to be a lot of moving parts here...

Yeah it's a lot 😅 - I'll try to go over them one by one in here.

> There's a PeriodicWorkService that delegates to the WorkDelegator that, when run, delegates all of its enqueued items to be run later in its ExecutorService

Correct, the `PeriodicWorkService` tool is supposed to be used for any kind of periodic background work needed by the RUM agent, I think that, by having the background work centralized there, we get more control over how many threads the OTel Android spawns and how many processes are running at a given time, which is important for an Android app because resources are limited, the amount of threads spawned means more memory usage, and the amount of work being done at a time consumes both memory and battery. I know in the near future of other features that might need to run periodically in the background, such as NTP requests and ANR detection, so I thought of adding the scaffolding right away instead of piling up some tech debt and refactorings for later.

The `WorkDelegator` is just an inner class that implements Runnable and is passed to the Handler. It's not really needed to make `PeriodicWorkService` well though, it can all be part of `PeriodicWorkService`'s class, is just that I was worried about accidental calls to `PeriodicWorkService.run` that could bypass the timer.

> also tells a Handler to post a callback run on the MainLooper, which will repeat the process (every 10 seconds).

This is related to not abusing of Android app's resources, I believe that whatever work is done by this lib will always be secondary in terms of the host app's priorities, so from a user's perspective, having a lot of background secondary work happening at any time, probably continuously running doesn't seem like a behavior I'd like to get in my app from one of its libs, so that's why I believe there should be a small window of time where it does all it needs to do in the background and then waits for another window. It's common to have some kind of minimum delay between background jobs in Android due to proper resource consumption, like in [WorkManager](https://developer.android.com/develop/background-work/background-tasks/persistent/getting-started/define-work#schedule_periodic_work) for example.

10 seconds doesn't mean anything special to me though, if we think we need shorter delays I think we can reduce that number.

> There's also this PeriodicRunnable

This one's a utility for background tasks that need to happen more than once. But if some task only needs to happen once it can just implement Runnable directly and get enqueued once in the `PeriodicWorkService` object.

> there's the DefaultExportScheduler which is an implementation of PeriodicRunnable.

Yeah the `DefaultExportScheduler` is an intentionally very simple implementation of `ExportScheduler` that runs periodically and triggers the disk buffering exporting from the disk while the app is running. Users should be able to create their own `ExportScheduler` implementation to cover more scenarios if they like, such as triggering periodic exports even when their app isn't running, but I don't think we should add that behavior in here since it would mean adding more dependencies and also we [couldn't guarantee](https://dontkillmyapp.com/) that it will always work as expected on every flavor of Android.

> If you have a PeriodicRunnable that wants to be run every second, but the PeriodicWorkService only can schedule work every 10 seconds, then doesn't that effectively make the PeriodicRunnable period 10 seconds as well?

Yes, we can lower that number, although I think we should also be cautious about how often we need to do work in the background for the reasons I mentioned earlier.

> In any case, there is probably something in the merits of using the Handler/Looper mechanism that I need to read up on...but I wonder why we can't just use a plain old ScheduledExecutorService for all of this?

I think we could also use `ScheduledExecutorService`, the reason I went with a Handler is to take advantage of an existing thread to do the scheduling to avoid spawning unnecessary ones. It also seems to me that `ScheduledExecutorService`s don't have a limit on how many threads they can spawn, based on the code inside `Executors.newScheduledThreadPool`, which is concerning to me due to the same device's resources handling issue, and threads don't seem to be cheap [when it comes to memory](https://developer.android.com/topic/performance/threads#how-many-threads-should-you-create). Although I could have missed something about `ScheduledExecutorService` that won't cause more than one thread to be spawned at a time, or at least limit the max amount of parallel ones?

I think it all can be summarized to trying to make this lib a good android lib citizen 😅 - there might be cases where we might need to do more work or take more resources but I'd like to keep those as exceptions and try to do most of this lib's work by using as little as possible.

instrumentation/src/main/java/io/opentelemetry/android/internal/tools/time/SystemTime.kt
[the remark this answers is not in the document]
LikeTheSalad: Thanks, it's updated now.

instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfiguration.java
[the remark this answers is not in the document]
breedx-splk: Ok, I mean all hierarchal categories are going to fail at some point. :) I don't think this should be blocker here, just mostly wanted to pick your brain.

We could bring up the idea of splitting out a SIG for mobile or even more specifically for Android, but I'd want to be a little cautious and make sure that there's more momentum than just you and me....but really, if we get Hanson and Santosh and Surbhi and @marandaneto interested in a SIG, it could have some value. I do agree that it would help information flow a little faster, especially around larger initiatives and things that are cross-cutting concerns (like package structure, etc).

Let's bring it up at the next client sig.

breedx-splk:
Oh that's a good point about the `ScheduledExecutorService`, it does indeed look like the `Executors.newScheduledThreadPool()` can't leverage an upper limit. What about `Executors.newSingleThreadScheduledExecutor()` here then? We probably don't need a lot of concurrency, especially if one of the goals is to reduce resource usage. I dunno.

I do wonder, when it comes to reducing resource usage, if having a recurring task firing when there may not be work to be done is a good idea. 

In any case, I don't think any of this is a showstopper. Thanks for putting in all this work. I'll do another pass through and see if I have anything specific, but I remember it looks good overall.

breedx-splk (approved):
Progress! Looking forward to seeing it all come together in the final episode. :)

instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/scheduler/ExportScheduleHandler.kt
marandaneto: Sure, https://github.com/open-telemetry/opentelemetry-android/issues/217
It's quite simple, the plugin does all the magic, all you need to do is install the plugin and make the commands part of the CI pipeline, so the CI generates the API dump in every CI run if any, the check is already done automatically part of the builtin gradle `check`.

instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfiguration.java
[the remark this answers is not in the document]
marandaneto: I'm down, my timezone is CET (UTC+1), if I have no collision with the date and time, let's do this.

instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/scheduler/DefaultExportScheduleHandler.kt
```diff
@@ -0,0 +1,21 @@
...
+import io.opentelemetry.android.internal.services.periodicwork.PeriodicWorkService
+import java.util.concurrent.atomic.AtomicBoolean
+
+class DefaultExportScheduleHandler : ExportScheduleHandler {
+    private val enabled = AtomicBoolean(false)
+
+    override fun enable() {
+        if (!enabled.getAndSet(true)) {
+            ServiceManager.get().getService(PeriodicWorkService::class.java)
+                .enqueue(DefaultExportScheduler())
```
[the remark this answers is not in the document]
LikeTheSalad: You're right, that sounds way better! I'll add the changes.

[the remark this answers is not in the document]
LikeTheSalad: Thanks, it's updated now.

Issue #195, opened 2023-12-19
Support initialization hooks


LikeTheSalad:
Some tools might require to do some work as part of the OTel RUM initialization process, this could be done in different ways such as by creating wrapper initialization APIs for example, although it seems to me that it could be easier/more convenient for the users of this SDK to always do the initialization process the same way regardless of any external tools they might need to use from any vendor that might need to add vendor-specific capabilities top.

Based on that, I believe we should provide some "hooks" for extensions to be able to get initialized as part of this SDK's RUM initialization process.

LikeTheSalad:
This has been addressed here: https://github.com/open-telemetry/opentelemetry-android/pull/397

PR #196, opened 2023-12-19
Initialization hooks

LikeTheSalad:
Related to #195

### DIFF
--- a/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
+++ b/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
@@ -17,6 +17,7 @@ import io.opentelemetry.android.instrumentation.network.CurrentNetworkProvider;
 import io.opentelemetry.android.instrumentation.network.NetworkAttributesSpanAppender;
 import io.opentelemetry.android.instrumentation.network.NetworkChangeMonitor;
 import io.opentelemetry.android.instrumentation.startup.InitializationEvents;
+import io.opentelemetry.android.instrumentation.startup.InitializationListener;
 import io.opentelemetry.android.instrumentation.startup.SdkInitializationEvents;
 import io.opentelemetry.android.internal.features.persistence.DiskManager;
 import io.opentelemetry.android.internal.features.persistence.SimpleTemporaryFileProvider;
@@ -249,6 +250,7 @@ public final class OpenTelemetryRumBuilder {
      * @return A new {@link OpenTelemetryRum} instance.
      */
     public OpenTelemetryRum build() {
+        notifyInitializationStart();
 
         applyConfiguration();
 
@@ -263,7 +265,21 @@ public final class OpenTelemetryRumBuilder {
         SdkPreconfiguredRumBuilder delegate =
                 new SdkPreconfiguredRumBuilder(application, sdk, sessionId);
         instrumentationInstallers.forEach(delegate::addInstrumentation);
+
+        notifyInitializationEnd();
         return delegate.build();
+    }
+
+    private void notifyInitializationStart() {
+        for (InitializationListener listener : config.getInitializationListeners()) {
+            listener.onStart();
+        }
+    }
+
+    private void notifyInitializationEnd() {
+        for (InitializationListener listener : config.getInitializationListeners()) {
+            listener.onEnd();
+        }
     }
 
     /** Leverage the configuration to wire up various instrumentation components. */
--- a/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
+++ b/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
@@ -21,6 +21,7 @@ import io.opentelemetry.android.instrumentation.startup.InitializationListener;
 import io.opentelemetry.android.instrumentation.startup.SdkInitializationEvents;
 import io.opentelemetry.android.internal.features.persistence.DiskManager;
 import io.opentelemetry.android.internal.features.persistence.SimpleTemporaryFileProvider;
+import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.api.baggage.propagation.W3CBaggagePropagator;
 import io.opentelemetry.api.trace.propagation.W3CTraceContextPropagator;
 import io.opentelemetry.context.propagation.ContextPropagators;
@@ -266,7 +267,7 @@ public final class OpenTelemetryRumBuilder {
                 new SdkPreconfiguredRumBuilder(application, sdk, sessionId);
         instrumentationInstallers.forEach(delegate::addInstrumentation);
 
-        notifyInitializationEnd();
+        notifyInitializationEnd(sdk);
         return delegate.build();
     }
 
@@ -276,9 +277,9 @@ public final class OpenTelemetryRumBuilder {
         }
     }
 
-    private void notifyInitializationEnd() {
+    private void notifyInitializationEnd(OpenTelemetry openTelemetry) {
         for (InitializationListener listener : config.getInitializationListeners()) {
-            listener.onEnd();
+            listener.onEnd(openTelemetry);
         }
     }
 
--- a/instrumentation/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
+++ b/instrumentation/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
@@ -7,7 +7,11 @@ package io.opentelemetry.android.config;
 
 import io.opentelemetry.android.ScreenAttributesSpanProcessor;
 import io.opentelemetry.android.instrumentation.network.CurrentNetworkProvider;
+import io.opentelemetry.android.instrumentation.startup.InitializationListener;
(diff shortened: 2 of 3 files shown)


instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
```diff
@@ -263,9 +265,23 @@ public OpenTelemetryRum build() {
         SdkPreconfiguredRumBuilder delegate =
                 new SdkPreconfiguredRumBuilder(application, sdk, sessionId);
         instrumentationInstallers.forEach(delegate::addInstrumentation);
+
+        notifyInitializationEnd();
```
marandaneto, line 269: Can one of the methods above throw an exception? if not all good, otherwise maybe the `notifyInitializationEnd` call should be under a `finally` block, or even having a `error` hook.

instrumentation/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
```diff
@@ -23,7 +27,9 @@ public class OtelRumConfig {
     private boolean includeScreenAttributes = true;
     private DiskBufferingConfiguration diskBufferingConfiguration =
             DiskBufferingConfiguration.builder().build();
+
     private boolean networkChangeMonitoringEnabled = true;
+    private final List<InitializationListener> initializationListeners = new ArrayList<>();
```
marandaneto, line 32: Should this list be thread-safe or not expected/allowed to be called from multiple threads?

instrumentation/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
```diff
@@ -119,4 +125,15 @@ public void disableNetworkChangeMonitoring() {
...
     }
+
+    /** Adds a new initialization listener. It will be called during the RUM initialization. */
+    public void addInitializationListener(InitializationListener listener) {
+        if (!initializationListeners.contains(listener)) {
+            initializationListeners.add(listener);
+        }
+    }
+
+    public List<InitializationListener> getInitializationListeners() {
```
marandaneto, line 136: A comment that is an unmodifiable list would be nice, maybe people will try to mutate directly from the list instead of calling `add`, etc.
Maybe adding a `remove` method as well? not sure if really needed.

instrumentation/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
```diff
@@ -119,4 +125,15 @@ public void disableNetworkChangeMonitoring() {
     public boolean isNetworkChangeMonitoringEnabled() {
         return this.networkChangeMonitoringEnabled;
     }
+
+    /** Adds a new initialization listener. It will be called during the RUM initialization. */
+    public void addInitializationListener(InitializationListener listener) {
+        if (!initializationListeners.contains(listener)) {
```
marandaneto, line 131: It could be a `Set` if the goal is to not allow dupes anyway.

instrumentation/src/main/java/io/opentelemetry/android/instrumentation/startup/InitializationListener.kt
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+package io.opentelemetry.android.instrumentation.startup
+
+/**
+ * Provides callbacks to know the sate of the initialization.
+ */
+interface InitializationListener {
+    companion object {
+    }
```
marandaneto, line 13:
```suggestion
```

marandaneto (approved):

instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
LikeTheSalad: That's a good point, @breedx-splk has a better understanding of all the processes that happen during the RUM initialization, so I think we should wait for his input on this. However, atm what I can say is that I don't think the RUM init should throw any exceptions or should silently fail either, maybe some components might fail to initialize (such as the disk buffering for example) but so far that's not a reason to stop the overall RUM initialization. Though it's definitely a good conversation to have on what (and how) should we tell our users in case something fails.

instrumentation/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
LikeTheSalad: I can't think of a valid use case where it would be needed to access it from multiple threads atm. The writing should only happen before initializing RUM, and the reading should only be done inside our RUM init code which is currently happening in a single thread. Though there's a valid case to be made on how far we should take "defensive programming" methodologies. I wonder if OTel has some rules on it.

instrumentation/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
LikeTheSalad: I like the idea of adding a comment to explain it. I think we could add a `remove` method in the future if needed, but so far I don't have a use case in mind for it.

instrumentation/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
LikeTheSalad: Good point, I've added the changes 👍

instrumentation/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
marandaneto: Yeah, most likely what could happen is someone iterating over the list and trying to mutate the list (removing by a specific type or something), it'd throw an exception, it'd not if its thread-safe (no iterable needed).
I don't see the need if not this use case as well since the single thread only during the init.

instrumentation/src/main/java/io/opentelemetry/android/instrumentation/startup/InitializationListener.kt
```diff
@@ -0,0 +1,24 @@
...
+import io.opentelemetry.api.OpenTelemetry
+
+/**
+ * Provides callbacks to know the sate of the initialization.
+ */
+interface InitializationListener {
+    /**
+     * Called when the RUM initialization starts.
+     */
+    fun onStart()
```
breedx-splk, line 17: We have found it helpful in upstream to have default no-op bodies for some of these, so that implementations only need to override the parts they actually care about. What do you think?

instrumentation/src/main/java/io/opentelemetry/android/instrumentation/startup/InitializationListener.kt
```diff
@@ -0,0 +1,24 @@
...
+
+    /**
+     * Called when the RUM initialization ends.
+     * @param openTelemetry - The initialized OpenTelemetry instance.
+     */
+    fun onEnd(openTelemetry: OpenTelemetry)
```
breedx-splk, line 23: What if this was `onEnd(rum: OpenTelemetryRum)` instead? The otel instance (like what you have here) as well as the session id are both obtainable from the `OpenTelemetryRum` instance.

breedx-splk (commented):
Had a couple of small ideas here.

instrumentation/src/main/java/io/opentelemetry/android/instrumentation/startup/InitializationListener.kt
LikeTheSalad: Makes sense, though now that I think about the whole thing after [the proposal](https://github.com/open-telemetry/opentelemetry-android/issues/214) I wrote yesterday, it seems to me that these hooks might overlap with [instrumentation installers](https://github.com/open-telemetry/opentelemetry-android/blob/d7d5045c8178b0a6069ae93c1cd0b11dc6641fe0/instrumentation/src/main/java/io/opentelemetry/android/instrumentation/InstrumentedApplication.java#L19). So maybe it's not needed to add another, similar mechanism if we decide to make that one public?

breedx-splk:
@LikeTheSalad What do you want to do with this one?

LikeTheSalad:
> @LikeTheSalad What do you want to do with this one?

It seems like it might no longer be needed after the changes we discussed [here](https://github.com/open-telemetry/opentelemetry-android/issues/214#issuecomment-1928209007). Since in order to add instrumentations from outside I'm guessing we should expose some sort of API that would allow registered instrumentations to get notified when the OpenTelemetry instance is ready to be used, such as the existing [InstrumentedApplication](https://github.com/open-telemetry/opentelemetry-android/blob/d7d5045c8178b0a6069ae93c1cd0b11dc6641fe0/instrumentation/src/main/java/io/opentelemetry/android/instrumentation/InstrumentedApplication.java#L19) one (which is essentially what this PR was trying to do). What do you think?

breedx-splk:
Ok I think I'm ok either sitting on this or closing it until we see how init events might fall out in the new mode. Agreed that there's probably going to be an api needed. :) Just wanted to check and see what your opinion on it was. 👍🏻

LikeTheSalad:
Closing as this is no longer needed.

PR #216, opened 2024-01-10
Remove zipkin dependencies

breedx-splk:
Because the zipkin exporter was removed some time ago, these are no longer needed and should be safely removed.


trask (approved):

marandaneto (approved):
Removing deps is always a good feeling :D

Issue #214, opened 2024-01-09
Proposal: RUM Configuration


LikeTheSalad:
The OTel Android library is essentially an OTel Java SDK wrapper that adds very useful features for Android apps on top. This provides huge value when it comes to adopting OpenTelemetry in Android apps as some of the features can be really difficult and/or time-consuming to implement manually, and some might even need specialized compilation tools that allow for bytecode instrumentation, which is accomplished very differently than doing the same for regular Java applications due to differences at runtime. So having Android-specific tools handy can provide an inestimable value when using OTel in an Android app, possibly comparable to creating an Android app with Android [Jetpack tools](https://developer.android.com/jetpack) vs without using those tools.

## The problem

Providing several tools comes with an important challenge though, which is configuration. The challenge comes when we ask ourselves questions such as “What features are meant to be added out of the box?”, and “How should the ‘X’ feature behave by default?”, or “What aspects of the ‘X’ feature should be configurable?” And so on, and the reason why these are difficult questions is because all apps are different and can therefore have different needs, some of which might not be possible to foresee at the time of creating the feature.

There is a risk of making the library too opinionated if there aren’t enough config options available, but there’s also a risk of making it cumbersome to use if there’s a long list of flags and different params associated with several features, sometimes with several params associated to a single feature but only valid when the feature flag is enabled and so on, within the [RUM config](https://github.com/open-telemetry/opentelemetry-android/blob/d7d5045c8178b0a6069ae93c1cd0b11dc6641fe0/instrumentation/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java#L19) that would require making different combinations of flags and other params to make the overall agent work as expected.

## The proposal

Because of the above, I’d like to propose dividing (and conquering) the overall functionality of the OTel Android RUM agent into 2 categories:

- **Initialization**: Covers what’s needed to initialize the OpenTelemetry instance. This may include the app’s info (`service.*` params) as well as signal processors/exporters customizers, dynamic global attrs, resources, and clock configurations.
- **Instrumentation**: Covers features that are **not** needed to initialize the OpenTelemetry instance. This includes auto-instrumentations though not necessarily limited to bytecode-instrumentation-related instrumentations, it could also cover things like crash reporting, ANR detection, and Activity/Fragment lifecycle instrumentations, which are arguably also auto-instrumentations, is just that they don’t rely on bytecode instrumentation to work.

[...]

PR #221, opened 2024-01-12
Adding disk buffering, part 4

LikeTheSalad:
This is the final PR to get the first iteration of the disk buffering feature fully running.

The changes added here do the following:

- Providing config option to set custom `ExportScheduleHandler`s.
- Using `DefaultExportScheduleHandler` if none is set.
- Wrapping the final version (after customizers) of the `SpanExporter` into a `SpanDiskExporter` when disk buffering is enabled.
- Created `SignalDiskExporter` to handle the reading and exporting processes of all signals.
- Exporting all signals inside `DefaultExportScheduler`.

### DIFF
--- a/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
+++ b/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
@@ -84,11 +84,10 @@ public final class OpenTelemetryRumBuilder {
 
     private Resource resource;
     @Nullable private CurrentNetworkProvider currentNetworkProvider = null;
-
+    @Nullable private SignalDiskExporter.Builder signalDiskExporterBuilder = null;
     private InitializationEvents initializationEvents = InitializationEvents.NO_OP;
     private Consumer<AnrDetectorBuilder> anrCustomizer = x -> {};
     private Consumer<CrashReporterBuilder> crashReporterCustomizer = x -> {};
-    @Nullable private SignalDiskExporter.Builder signalDiskExporterBuilder;
 
     private static TextMapPropagator buildDefaultPropagator() {
         return TextMapPropagator.composite(
--- a/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
+++ b/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
@@ -83,6 +83,7 @@ public final class OpenTelemetryRumBuilder {
 
     private Resource resource;
     @Nullable private CurrentNetworkProvider currentNetworkProvider = null;
+
     private InitializationEvents initializationEvents = InitializationEvents.NO_OP;
     private Consumer<AnrDetectorBuilder> anrCustomizer = x -> {};
     private Consumer<CrashReporterBuilder> crashReporterCustomizer = x -> {};
@@ -311,7 +312,9 @@ public final class OpenTelemetryRumBuilder {
     /** Leverage the configuration to wire up various instrumentation components. */
     private void applyConfiguration() {
         if (config.shouldGenerateSdkInitializationEvents()) {
-            initializationEvents = new SdkInitializationEvents();
+            if (initializationEvents == InitializationEvents.NO_OP) {
+                initializationEvents = new SdkInitializationEvents();
+            }
             initializationEvents.recordConfiguration(config);
         }
         initializationEvents.sdkInitializationStarted();
@@ -418,6 +421,7 @@ public final class OpenTelemetryRumBuilder {
                         .addSpanProcessor(new SessionIdSpanAppender(sessionId));
 
         SpanExporter spanExporter = buildSpanExporter();
+        initializationEvents.spanExporterInitialized(spanExporter);
         BatchSpanProcessor batchSpanProcessor = BatchSpanProcessor.builder(spanExporter).build();
         tracerProviderBuilder.addSpanProcessor(batchSpanProcessor);
 
@@ -489,5 +493,10 @@ public final class OpenTelemetryRumBuilder {
     private ContextPropagators buildFinalPropagators() {
         TextMapPropagator defaultPropagator = buildDefaultPropagator();
         return ContextPropagators.create(propagatorCustomizer.apply(defaultPropagator));
+    }
+
+    OpenTelemetryRumBuilder setInitializationEvents(InitializationEvents initializationEvents) {
+        this.initializationEvents = initializationEvents;
+        return this;
     }
 }
--- a/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
+++ b/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
@@ -14,6 +14,7 @@ import android.util.Log;
 import io.opentelemetry.android.config.OtelRumConfig;
 import io.opentelemetry.android.features.diskbuffering.DiskBufferingConfiguration;
 import io.opentelemetry.android.features.diskbuffering.SignalDiskExporter;
+import io.opentelemetry.android.features.diskbuffering.scheduler.ExportScheduleHandler;
 import io.opentelemetry.android.instrumentation.InstrumentedApplication;
 import io.opentelemetry.android.instrumentation.activity.VisibleScreenTracker;
 import io.opentelemetry.android.instrumentation.anr.AnrDetector;
@@ -299,14 +300,29 @@ public final class OpenTelemetryRumBuilder {
                         .setLoggerProvider(buildLoggerProvider(application))
                         .setPropagators(buildFinalPropagators())
                         .build();
-        if (signalDiskExporterBuilder != null) {
-            SignalDiskExporter.set(signalDiskExporterBuilder.build());
-        }
+
+        setUpDiskBuffering();
 
         SdkPreconfiguredRumBuilder delegate =
                 new SdkPreconfiguredRumBuilder(application, sdk, sessionId);
         instrumentationInstallers.forEach(delegate::addInstrumentation);
         return delegate.build();
+    }
+
+    private void setUpDiskBuffering() {
+        ExportScheduleHandler exportScheduleHandler =
+                config.getDiskBufferingConfiguration().getExportScheduleHandler();
+        if (signalDiskExporterBuilder != null) {
+            // Not null means that disk buffering is enabled and disk exporters are successfully
+            // initialized.
+            SignalDiskExporter.set(signalDiskExporterBuilder.build());
+            exportScheduleHandler.enable();
+        } else {
+            // Disabling here allows to cancel previously scheduled exports using tools that
+            // can run even after the app has been terminated (such as WorkManager).
+            // But for in-memory only schedulers, nothing should need to be disabled.
+            exportScheduleHandler.disable();
+        }
     }
 
     /** Leverage the configuration to wire up various instrumentation components. */
--- a/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
+++ b/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
@@ -13,6 +13,7 @@ import android.os.Looper;
 import android.util.Log;
 import io.opentelemetry.android.config.OtelRumConfig;
 import io.opentelemetry.android.features.diskbuffering.DiskBufferingConfiguration;
+import io.opentelemetry.android.features.diskbuffering.SignalDiskExporter;
 import io.opentelemetry.android.instrumentation.InstrumentedApplication;
 import io.opentelemetry.android.instrumentation.activity.VisibleScreenTracker;
 import io.opentelemetry.android.instrumentation.anr.AnrDetector;
@@ -85,6 +86,7 @@ public final class OpenTelemetryRumBuilder {
     private InitializationEvents initializationEvents = InitializationEvents.NO_OP;
     private Consumer<AnrDetectorBuilder> anrCustomizer = x -> {};
     private Consumer<CrashReporterBuilder> crashReporterCustomizer = x -> {};
+    @Nullable private SignalDiskExporter.Builder signalDiskExporterBuilder;
 
     private static TextMapPropagator buildDefaultPropagator() {
         return TextMapPropagator.composite(
@@ -296,6 +298,9 @@ public final class OpenTelemetryRumBuilder {
                         .setLoggerProvider(buildLoggerProvider(application))
                         .setPropagators(buildFinalPropagators())
                         .build();
+        if (signalDiskExporterBuilder != null) {
+            SignalDiskExporter.set(signalDiskExporterBuilder.build());
+        }
 
         SdkPreconfiguredRumBuilder delegate =
                 new SdkPreconfiguredRumBuilder(application, sdk, sessionId);
@@ -426,20 +431,27 @@ public final class OpenTelemetryRumBuilder {
     private SpanExporter buildSpanExporter() {
         // TODO: Default to otlp...but how can we make endpoint and auth mandatory?
         SpanExporter defaultExporter = LoggingSpanExporter.create();
-        SpanExporter spanExporter = defaultExporter;
+        SpanExporter spanExporter = spanExporterCustomizer.apply(defaultExporter);
+
         DiskBufferingConfiguration diskBufferingConfiguration =
                 config.getDiskBufferingConfiguration();
         if (diskBufferingConfiguration.isEnabled()) {
             try {
-                spanExporter = createDiskExporter(defaultExporter, diskBufferingConfiguration);
+                SpanDiskExporter diskExporter =
+                        createDiskExporter(defaultExporter, diskBufferingConfiguration);
+                if (signalDiskExporterBuilder == null) {
+                    signalDiskExporterBuilder = SignalDiskExporter.builder();
+                }
+                signalDiskExporterBuilder.setSpanDiskExporter(diskExporter);
+                spanExporter = diskExporter;
             } catch (IOException e) {
                 Log.w(RumConstants.OTEL_RUM_LOG_TAG, "Could not create span disk exporter.", e);
             }
         }
-        return spanExporterCustomizer.apply(spanExporter);
+        return spanExporter;
     }
 
-    private static SpanExporter createDiskExporter(
+    private static SpanDiskExporter createDiskExporter(
             SpanExporter defaultExporter, DiskBufferingConfiguration diskBufferingConfiguration)
             throws IOException {
         DiskManager diskManager = DiskManager.create(diskBufferingConfiguration);
--- a/instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfiguration.java
+++ b/instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfiguration.java
@@ -5,16 +5,22 @@
 
 package io.opentelemetry.android.features.diskbuffering;
 
+import io.opentelemetry.android.features.diskbuffering.scheduler.DefaultExportScheduleHandler;
+import io.opentelemetry.android.features.diskbuffering.scheduler.DefaultExportScheduler;
+import io.opentelemetry.android.features.diskbuffering.scheduler.ExportScheduleHandler;
+
 /** Configuration for disk buffering. */
 public final class DiskBufferingConfiguration {
     private final boolean enabled;
     private final int maxCacheSize;
+    private final ExportScheduleHandler exportScheduleHandler;
     private static final int DEFAULT_MAX_CACHE_SIZE = 60 * 1024 * 1024;
     private static final int MAX_FILE_SIZE = 1024 * 1024;
 
     private DiskBufferingConfiguration(Builder builder) {
         this.enabled = builder.enabled;
         this.maxCacheSize = builder.maxCacheSize;
+        this.exportScheduleHandler = builder.exportScheduleHandler;
     }
 
     public static Builder builder() {
@@ -33,9 +39,15 @@ public final class DiskBufferingConfiguration {
         return MAX_FILE_SIZE;
     }
 
+    public ExportScheduleHandler getExportScheduleHandler() {
+        return exportScheduleHandler;
+    }
+
     public static final class Builder {
         private boolean enabled;
         private int maxCacheSize;
+        private ExportScheduleHandler exportScheduleHandler =
+                new DefaultExportScheduleHandler(new DefaultExportScheduler());
 
         private Builder(boolean enabled, int maxCacheSize) {
             this.enabled = enabled;
@@ -55,6 +67,15 @@ public final class DiskBufferingConfiguration {
          */
         public Builder setMaxCacheSize(int maxCacheSize) {
             this.maxCacheSize = maxCacheSize;
+            return this;
+        }
+
+        /**
+         * Sets a scheduler that will take care of periodically read data stored in disk and export
+         * it.
+         */
+        public Builder setExportScheduleHandler(ExportScheduleHandler exportScheduleHandler) {
+            this.exportScheduleHandler = exportScheduleHandler;
             return this;
         }
 
--- a/instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/SignalDiskExporter.kt
+++ b/instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/SignalDiskExporter.kt
@@ -1,0 +1,125 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.android.features.diskbuffering
+
+import androidx.annotation.WorkerThread
+import io.opentelemetry.contrib.disk.buffering.LogRecordDiskExporter
+import io.opentelemetry.contrib.disk.buffering.MetricDiskExporter
+import io.opentelemetry.contrib.disk.buffering.SpanDiskExporter
+import java.io.IOException
+import java.util.concurrent.TimeUnit
+
+/**
+ * Entrypoint to read and export previously cached signals.
+ */
+class SignalDiskExporter internal constructor(
+    private val spanDiskExporter: SpanDiskExporter?,
+    private val metricDiskExporter: MetricDiskExporter?,
+    private val logRecordDiskExporter: LogRecordDiskExporter?,
+    private val exportTimeoutInMillis: Long,
+) {
+    @WorkerThread
+    @Throws(IOException::class)
+    fun exportBatchOfSpans(): Boolean {
+        return spanDiskExporter?.exportStoredBatch(
+            exportTimeoutInMillis,
+            TimeUnit.MILLISECONDS,
+        ) ?: false
+    }
+
+    @WorkerThread
+    @Throws(IOException::class)
+    fun exportBatchOfMetrics(): Boolean {
+        return metricDiskExporter?.exportStoredBatch(
+            exportTimeoutInMillis,
+            TimeUnit.MILLISECONDS,
+        ) ?: false
+    }
+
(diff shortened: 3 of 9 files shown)


instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
```diff
@@ -413,6 +436,7 @@ private SdkTracerProvider buildTracerProvider(SessionId sessionId, Application a
                         .addSpanProcessor(new SessionIdSpanAppender(sessionId));
 
         SpanExporter spanExporter = buildSpanExporter();
+        initializationEvents.spanExporterInitialized(spanExporter);
```
LikeTheSalad, line 439: I was having issues to verify that the right type of exporter (the disk exporter) was properly set in the unit tests, so one of the least ugly ways I could manage to do so was by adding this new function to the InitializationEvents interface and also a package-private setter for it below in this class (to set a mock impl from the tests). I'm open to better alternatives I might have missed 🙂

instrumentation/src/main/java/io/opentelemetry/android/instrumentation/startup/SdkInitializationEvents.java
```diff
@@ -43,4 +44,9 @@ public void slowRenderingDetectorInitialized() {
     public void crashReportingInitialized() {
         // TODO: Build me "crashReportingInitialized"
     }
+
+    @Override
+    public void spanExporterInitialized(SpanExporter spanExporter) {
```
LikeTheSalad, line 49: Not related to this PR though when I was changing this class I thought that maybe we don't need this implementation of `InitializationEvents` and instead we can just add a public setter [in this class](https://github.com/open-telemetry/opentelemetry-android/blob/6de29672a2bf44cc36b651bdb12b7f5228856333/instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java#L85) so that, if a user needs to enable initialization events, they can just set their own implementation in there? That way we wouldn't need to define [this config flag](https://github.com/open-telemetry/opentelemetry-android/blob/6de29672a2bf44cc36b651bdb12b7f5228856333/instrumentation/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java#L27) either, as the default impl would be the noop one.

instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
marandaneto: Maybe `initializationEvents` takes `spanExporter` in the ctor? (L331) so if you can provide/mock the `spanExporter`, you don't need a setter.

instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/SignalDiskExporter.kt
```diff
@@ -0,0 +1,155 @@
...
+import java.util.concurrent.TimeUnit
+
+/**
+ * Entrypoint to read and export previously cached signals.
+ */
+class SignalDiskExporter internal constructor(
+    private val spanDiskExporter: SpanDiskExporter?,
+    private val metricDiskExporter: MetricDiskExporter?,
+    private val logRecordDiskExporter: LogRecordDiskExporter?,
+    private val exportTimeoutInMillis: Long,
```
marandaneto, line 22:
```suggestion
    private val exportTimeoutInMillis: Long,
    private val spanDiskExporter: SpanDiskExporter?,
    private val metricDiskExporter: MetricDiskExporter?,
    private val logRecordDiskExporter: LogRecordDiskExporter?,
```

Just personal preference, non optional params go first

instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/SignalDiskExporter.kt
```diff
@@ -0,0 +1,155 @@
...
+        if (exportBatchOfMetrics()) {
+            atLeastOneWorked = true
+        }
+        if (exportBatchOfLogs()) {
+            atLeastOneWorked = true
+        }
+        return atLeastOneWorked
+    }
+
+    class Builder {
```
marandaneto, line 94: Usually, Builders in Kotlin are a bit odd (and have downsides) since the language provides named and default parameters.
https://www.baeldung.com/kotlin/builder-pattern for context

instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/scheduler/DefaultExportScheduler.kt
```diff
@@ -14,11 +18,20 @@ class DefaultExportScheduler : PeriodicRunnable() {
...
 
     override fun onRun() {
-        // TODO for next PR.
+        val exporter = SignalDiskExporter.get() ?: return
+
+        try {
+            var didExport = exporter.exportBatchOfEach()
+            while (didExport) {
+                didExport = exporter.exportBatchOfEach()
+            }
```
marandaneto, line 27: I wonder if this can ever be an infinite loop due to a bug, not a big fan of such a while (`possibleAlwaysTrue`) but if there are tests in place, all good.

instrumentation/src/main/java/io/opentelemetry/android/instrumentation/startup/SdkInitializationEvents.java
marandaneto: That sounds like a good idea

instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/SignalDiskExporter.kt
```diff
@@ -0,0 +1,155 @@
...
+import io.opentelemetry.contrib.disk.buffering.LogRecordDiskExporter
+import io.opentelemetry.contrib.disk.buffering.MetricDiskExporter
+import io.opentelemetry.contrib.disk.buffering.SpanDiskExporter
+import java.io.IOException
+import java.util.concurrent.TimeUnit
+
+/**
+ * Entrypoint to read and export previously cached signals.
+ */
+class SignalDiskExporter internal constructor(
```
breedx-splk, line 18: Something is bugging me about the name of this class -- it makes the reader think that this is responsible for exporting a single signal to disk...but really, it's just a container (not really an entrypoint(?)) for 3 type-specific `@WorkerThread` methods that read a batch from disk and export to a signal type-specific delegate.

I think part of the problem might stem from something in the upstream contrib repo, so I think it's probably fine to continue with this approach until we can address it there. However, I think we should rename this...and of course naming is hard.

How about something like `StoredBatchReadingExporters` or `BatchExportingDiskReaders` or similar?

instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
```diff
@@ -297,16 +300,36 @@ public OpenTelemetryRum build() {
                         .setPropagators(buildFinalPropagators())
                         .build();
 
+        setUpDiskBuffering();
```
breedx-splk, line 303: So I think this name is wrong. "Buffering" implies sending something into a buffer. From where I stand, this is exactly the opposite of that -- it's reading something that has already been buffered (on disk). A better name might be something like `schedulePeriodicDiskReadingForBufferedTelemetry()` or maybe just `scheduleDiskTelemetryReader()`.

instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
```diff
@@ -426,20 +450,27 @@ private SdkTracerProvider buildTracerProvider(SessionId sessionId, Application a
...
         if (diskBufferingConfiguration.isEnabled()) {
             try {
-                spanExporter = createDiskExporter(defaultExporter, diskBufferingConfiguration);
+                SpanDiskExporter diskExporter =
+                        createDiskExporter(defaultExporter, diskBufferingConfiguration);
+                if (signalDiskExporterBuilder == null) {
+                    signalDiskExporterBuilder = SignalDiskExporter.builder();
+                }
+                signalDiskExporterBuilder.setSpanDiskExporter(diskExporter);
+                spanExporter = diskExporter;
```
breedx-splk, line 465: This seems overly complicated. It introduces a new field so that we can optionally create it in order to set the `diskExporter` here, which is really just a side effect that gets leveraged inside of the `setUpDiskBuffering()` method.

I think you should be able to get away with removing the new field and then factor out this chunk into a new method called `scheduleDiskReader()` or something, which does what the `setUpDiskBuffering()` method basically does. I think it will simplify some things, and this part will read more like a story:

* If the disk buffering is turned on
  * create the disk exporter
  * schedule the disk reader

instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
breedx-splk: What if, instead of creating an init event for the span exporter, there was an init event for the disk buffering being initialized? That's really the feature here, and then you can assert that it was called I guess?

instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfiguration.java
```diff
@@ -5,16 +5,22 @@
...
 
+import io.opentelemetry.android.features.diskbuffering.scheduler.DefaultExportScheduleHandler;
+import io.opentelemetry.android.features.diskbuffering.scheduler.DefaultExportScheduler;
+import io.opentelemetry.android.features.diskbuffering.scheduler.ExportScheduleHandler;
+
 /** Configuration for disk buffering. */
 public final class DiskBufferingConfiguration {
     private final boolean enabled;
     private final int maxCacheSize;
+    private final ExportScheduleHandler exportScheduleHandler;
```
breedx-splk, line 16: I think this is problematic because it introduces an action class as a field of a configuration data class. It mixes up dumb/simple config with things that do work....and I think that's a troublesome path. 

What are the use cases that you can think of for allowing a user to pass in their own `ExportScheduleHandler` instance to the `DiskBufferingConfiguration`? I thought that maybe it was to specify their own schedule/period, but the way the code is structured now I'm not even sure that's possible.

If that was the main intent, I'd rather that the period be the configurable thing, and the builders for the action classes can do the building by using the config, not by having the config do the building itself. Make sense?

instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/scheduler/DefaultExportScheduler.kt
breedx-splk, line 27:
```suggestion
            do {
              val didExport = exporter.exportBatchOfEach()
            } while (didExport)
```

instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/scheduler/DefaultExportScheduler.kt
```diff
@@ -14,11 +18,20 @@ class DefaultExportScheduler : PeriodicRunnable() {
     }
 
     override fun onRun() {
-        // TODO for next PR.
+        val exporter = SignalDiskExporter.get() ?: return
```
breedx-splk, line 21: Can we just pass an instance to this class instead of reaching out to a global static singleton?

breedx-splk (commented):
It's exciting to have things getting so close....have some questions along the way tho.

instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
LikeTheSalad: > What if, instead of creating an init event for the span exporter, there was an init event for the disk buffering being initialized? That's really the feature here, and then you can assert that it was called I guess?

I see what you mean, although the initialization of the disk buffering tool doesn't say anything about the feature actually being used in the OpenTelemetry instance that has been built here, which I believe should be what we verify to guarantee that the RUM config is honored, and all those pieces are out of the scope of the disk buffering feature.

instrumentation/src/main/java/io/opentelemetry/android/instrumentation/startup/SdkInitializationEvents.java
LikeTheSalad: > It probably doesn't help that the existing implementation is just a stub. I think I would _not_ want 20 different users to have 20 different ways of formulating these various events. In other words, ideally one day in the future we will have semconv for these events, and I'd hate for `opentelemetry-android` users to have to go read the semconv in order to build an implementation themselves. We should provide that implementation for them, even if we do decide to default these to "off" (no-op). Users that want them shouldn't have to build it.

Definitely having a default impl should be the way to go, although it's quite pointless having a stub for now that also cannot be overridden (since it can also be a great tool for unit tests). I would prefer to provide a default implementation returned by a `InitializationEvents.getDefault()` static function and a setter in `OpenTelemetryRumBuilder` that users (and ourselves from unit tests) could call to enable this feature, since it would be set to the `noop` impl by default, which essentially disables it. Rather than having to create a config flag for it, generally speaking I don't like flags, they can be very useful for small things but are not good for scalability. I also haven't seen them used in OTel Java for example, is mostly interfaces, setters for their implementations, and implementations already created in the same repo that are useful for most use cases and ready to be used so that users can take advantage of them easily.

instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/SignalDiskExporter.kt
LikeTheSalad: This seems related to [this comment](https://github.com/open-telemetry/opentelemetry-android/pull/221#discussion_r1454126269) from @breedx-splk, I haven't taken a deep look yet but probably what he suggests might avoid the need to use a builder altogether.

instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/scheduler/DefaultExportScheduler.kt
LikeTheSalad: The one reason I see it might run for a long time is if there are a lot of signals in the disk and the internet connectivity is very slow, so each loop might take some time which would cause the whole data to take quite long to finish getting exported, but it shouldn't run forever as `exportBatchOfEach` only returns true if there's data available and it also gets exported successfully, so it's actually quite difficult for this loop to run uninterrupted.

instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
```diff
@@ -297,16 +300,36 @@ public OpenTelemetryRum build() {
...
         SdkPreconfiguredRumBuilder delegate =
                 new SdkPreconfiguredRumBuilder(application, sdk, sessionId);
         instrumentationInstallers.forEach(delegate::addInstrumentation);
         return delegate.build();
     }
 
+    private void setUpDiskBuffering() {
+        ExportScheduleHandler exportScheduleHandler =
+                config.getDiskBufferingConfiguration().getExportScheduleHandler();
+        if (signalDiskExporterBuilder != null) {
```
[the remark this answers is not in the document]
LikeTheSalad: Thanks, it's done now.

instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/SignalDiskExporter.kt
LikeTheSalad: > I think part of the problem might stem from something in the upstream contrib repo, so I think it's probably fine to continue with this approach until we can address it there. However, I think we should rename this...and of course naming is hard.

After taking a look at [your PR](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1161) I think we definitely can simplify this further once that's merged! In the meantime, yeah the name isn't ideal, it's always a struggle for me to come up with good names X_x so thanks for the suggestions. This object is really just a utility to trigger "`FromDisk`-exports", if that term makes sense, of all types of signals, so it's supposed to be kinda handy to avoid having to manually keeping track of all the `ToDisk`-exporters, which are set internally during the RUM initialization, so it could be awkward to get them from there, so instead this utility serves as a container for them all, as well as providing `FromDisk`-exporting functions in a singleton that's accessible from anywhere, even from custom scheduler impls.

Based on that, maybe a good name could be something like `StoredFromDiskExporters` or maybe `FromDiskExporters`, which would make sense after [your PR](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1161) gets merged, though I guess it depends on how quick we need to merge this one (I wouldn't mind waiting for the contrib one).

instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
LikeTheSalad: Makes sense 👍  it's updated now.

instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
LikeTheSalad: I see, the thing is that, by the time `setUpDiskBuffering` is called (now renamed to `scheduleDiskTelemetryReader`), the OpenTelemetry instance has already been created, so we'd lose the opportunity to set this disk exporter during the OTel initialization because it's immutable once it's all done (afaik), so the reason why I'm currently setting it up here is so that we can intercept the final, "normal" exporter before it's set into the span processor.

Maybe after [this PR](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1161) gets merged it won't need to be so complicated though. However, there's one more thing to be aware of regarding this part: `If the disk buffering is turned on` which is that even though the flag is enabled there's no guarantee that the buffering will actually work because there might be an issue when initializing the disk exporters (due to IO errors), causing the feature to get disabled even if it was requested to get enabled. I think @marandaneto mentioned something about this in a previous PR where we were pondering what would happen if some feature fails to get initialized during the RUM setup.

I guess a workaround that would allow to initialize the disk buffering exporters after the OpenTelemetry instance is created could be to create an exporter implementation that always wrapped the final exporter and would dynamically delegate to either the disk exporter or the "normal" exporter provided during the RUM setup depending on the config.

instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfiguration.java
LikeTheSalad: > What are the use cases that you can think of for allowing a user to pass in their own `ExportScheduleHandler` instance to the `DiskBufferingConfiguration`?

The default "scheduler" only works while the app is running and I'm guessing some users would like to attempt exporting data even if their app isn't running by using  [WorkManager](https://developer.android.com/develop/background-work/background-tasks/persistent) for example.

Though I agree that the config should generally be made of data instead of functions or the like, maybe since the disk buffering feature is needed to initialize RUM, we could categorize it as an "initialization" feature, which will allow us to get rid of `DiskBufferingConfiguration` altogether as we would only use `OtelRumConfig` to set all initialization params, and then we could move the ExportScheduleHandler setter over to `OpenTelemetryRumBuilder` if we decided to go with the proposed structure in [here](https://github.com/open-telemetry/opentelemetry-android/issues/214).

instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/scheduler/DefaultExportScheduler.kt
LikeTheSalad: Yeah I think we could, though there are still benefits of having it as a singleton which I mentioned [here](https://github.com/open-telemetry/opentelemetry-android/pull/221#discussion_r1457393546). I think we should address that comment first before deciding how to handle this scenario.

instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
marandaneto: ok, I see the issue, since GitHub collapsed the code diff, I thought that these changes and the changes on L331 were a single method, so you could have used a ctor, my bad.

instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/SignalDiskExporter.kt
marandaneto: Looking at `createInstance`, all you need is a data class with optional parameters, so you create the instance with all the parameters at once, if they are null or not it's not important, the final object is created at the very end and immutable, it's only important to construct all the parameters beforehand which you have to do with the builder anyway.

```kotlin
val instance =
            createInstance(
                spanDiskExporter,
                metricDiskExporter,
                logRecordDiskExporter,
                timeoutInMillis,
            )
```

becomes

```kotlin
val instance = SignalDiskExporter(
  spanDiskExporter = spanDiskExporter,
  metricDiskExporter = metricDiskExporter,
  ..
)
```

marandaneto: If the Java OTel SDK has builders everywhere, maybe still using builders makes sense, so it's not different everywhere, your call here, just wanted to share the Builders is not a Kotlin-friendly pattern

LikeTheSalad: > Looking at `createInstance`, all you need is a data class with optional parameters, so you create the instance with all the parameters at once, if they are null or not it's not important, the final object is created at the very end and immutable, it's only important to construct all the parameters beforehand which you have to do with the builder anyway.

Are you referring to the function `createInstance` within `SignalDiskExporterTest`? Because if that's the case, I agree, a builder is unnecessary there. My concern is with `OpenTelemetryRumBuilder` where the span exporter is created in `buildSpanExporter` which is called from `buildTracerProvider`, which is called from `build` where the OpenTelemetry instance is created, and the setup of `SignalDiskExporter` is done right after. The way the OpenTelemetry instance is built is by setting each signal up independently, which involves a couple of steps, so for the spans one, we need to set a tracer provider with a span processor with the span exporter, and the last one in that chain (the exporter setup) is where we can attach a disk exporter. There are many functions involved in setting the span signal up, and I'm guessing the same will happen with the other 2 signals, so in this scenario is handy to have a place to collect each disk exporter until they're all ready to create the `SignalDiskExporter` instance. It doesn't necessarily need to be called a builder though, but it's essentially what it is so it made sense. I'm not sure if Kotlin has another name for it or a more Kotlin-y way to handle these cases. I'm thinking that maybe after merging [this PR](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1161) we might get other options to address this without a builder though.

marandaneto: gotcha, yeah I was looking at the `createInstance` as an example, just that you could replicate the same approach but got your point now, thanks.
I don't know the OTel codebase at all, just getting used to it by doing reviews and skimming the code once in a while so feel free to ignore me when I tell nonsense :D

LikeTheSalad: No worries, I might miss a lot of things when creating these PRs too, especially recent features added to Kotlin so I appreciate any feedback, in this case, we could actually get rid of the builder is just that there might be different ways to create this instance and there's not a clear "best one" to me, plus it might change after this [PR gets](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1161) merged. I like these kinds of conversations, although I'm looking forward to be able to go through them in an Android SIG otherwise it's a lot of typing 😅

marandaneto: Indeed. Haha

instrumentation/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
```diff
@@ -297,16 +300,36 @@ public OpenTelemetryRum build() {
...
+            // But for in-memory only schedulers, nothing should need to be disabled.
+            exportScheduleHandler.disable();
+        }
+    }
+
     /** Leverage the configuration to wire up various instrumentation components. */
     private void applyConfiguration() {
         if (config.shouldGenerateSdkInitializationEvents()) {
-            initializationEvents = new SdkInitializationEvents();
+            if (initializationEvents == InitializationEvents.NO_OP) {
```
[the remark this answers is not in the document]
breedx-splk: Ah ok, I think I understand.

instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/SignalDiskExporter.kt
breedx-splk: Yeah let's definitely punt on this until the upstream changes land here. 🤘🏻 Totally fine with it being this way for now.

instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfiguration.java
breedx-splk: I'm ok with moving this around after this PR, possibly as part of dealing with the larger config structure questions.

LikeTheSalad:
> What if, instead of creating an init event for the span exporter, there was an init event for the disk buffering being initialized? That's really the feature here, and then you can assert that it was called I guess?

This has been answered [here](https://github.com/open-telemetry/opentelemetry-android/pull/221#discussion_r1457268362) (not sure why GH shows it duplicated).

LikeTheSalad:
> So I set this up initially with the expectation that most users will want the InitializationEvents to be emitted by default. This can help o11y practitioners to understand how slowly/quickly the otel android sdk starts up and is also super important in understanding which features were enabled for a user session after the fact.
> 
> It probably doesn't help that the existing implementation is just a stub. I think I would not want 20 different users to have 20 different ways of formulating these various events. In other words, ideally one day in the future we will have semconv for these events, and I'd hate for opentelemetry-android users to have to go read the semconv in order to build an implementation themselves. We should provide that implementation for them, even if we do decide to default these to "off" (no-op). Users that want them shouldn't have to build it.

Answered [here](https://github.com/open-telemetry/opentelemetry-android/pull/221#discussion_r1457297356).

breedx-splk:
@LikeTheSalad let's get the build passing!

LikeTheSalad:
> @LikeTheSalad let's get the build passing!

✅

breedx-splk:
So the disk buffering changes are now available in contrib https://github.com/open-telemetry/opentelemetry-java-contrib/releases/tag/v1.33.0. @LikeTheSalad are you going to pull that in here now?

instrumentation/src/main/java/io/opentelemetry/android/features/diskbuffering/SignalFromDiskExporter.kt
```diff
@@ -0,0 +1,115 @@
...
+        @WorkerThread
+        @Throws(IOException::class)
+        fun exportBatchOfEach(): Boolean {
+            var atLeastOneWorked = exportBatchOfSpans()
+            if (exportBatchOfMetrics()) {
+                atLeastOneWorked = true
+            }
+            if (exportBatchOfLogs()) {
+                atLeastOneWorked = true
+            }
```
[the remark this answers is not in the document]
breedx-splk: Then again, this way makes me have to think about the possibility for short circuiting....so let's skip it. Ignore me. :)

breedx-splk (approved):
I think we made it! Thanks for all the good work on this @LikeTheSalad and for sticking with it.

marandaneto (approved):

breedx-splk:
@LikeTheSalad needs a rebase. 🙏🏻

PR #228, opened 2024-01-24
public CurrentNetworkAttributesExtractor

Archish27:
Exposing CurrentNetworkAttributesExtractor can be used if required.

### DIFF
--- a/instrumentation/src/main/java/io/opentelemetry/android/instrumentation/network/CurrentNetworkAttributesExtractor.java
+++ b/instrumentation/src/main/java/io/opentelemetry/android/instrumentation/network/CurrentNetworkAttributesExtractor.java
@@ -17,7 +17,7 @@ import io.opentelemetry.api.common.AttributeKey;
 import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.api.common.AttributesBuilder;
 
-final class CurrentNetworkAttributesExtractor {
+public final class CurrentNetworkAttributesExtractor {
 
     Attributes extract(CurrentNetwork network) {
         AttributesBuilder builder =


LikeTheSalad:
Hi @Archish27 - Do you have any specific use-cases where you'd need to use `CurrentNetworkAttributesExtractor`? I'm curious for two reasons, the first one is that I think it's a good opportunity to see if there's a missing feature that we could provide in a more generic way so that it could help with other attrs too, and the second one is that we have plans to do some changes regarding [the way we handle network](https://github.com/open-telemetry/opentelemetry-android/issues/49) changes in general, and this class is related to them, so I'm not sure if this part of the code would be affected, but I wouldn't discard that possibility, so I'd like to get a better understanding of what's needed so that we can avoid future breaking changes.

breedx-splk:
+1 to what @LikeTheSalad has said. The intent was never for users to use this class directly, but more to allow the instrumentation to leverage this class to emit telemetry that contains network information for you.

Archish27:
Goal for this change, currently for react native android, I am trying to leverage the full potential of opentelemetery android it seems, there are potential classes which do the similar thing trying to leverage this class for setting up the attributes in span. If attributes extractor are public it would be great to leverage to set attributes. Otherwise, it needs a rewriting for such classes in native modules.

Archish27:
> > Goal for this change, currently for react native android, I am trying to leverage the full potential of opentelemetery android it seems, there are potential classes which do the similar thing trying to leverage this class for setting up the attributes in span. If attributes extractor are public it would be great to leverage to set attributes. Otherwise, it needs a rewriting for such classes in native modules.
> 
> 
> 
> If I understood correctly, it seems like you're looking for a way to set attributes to all spans? Is that correct?

@LikeTheSalad  Yes correct if such extractors are not exposed the same logic I have to rewrite, tomorrow if new attributes or extractors are added we can make use of it. Rather duplicating again.

breedx-splk:
@Archish27 I'm not sure that this change will do what you think it does. Even if this class was public, the method to extract the attributes is not public...and the means of creating the `CurrentNetwork` parameter is also intentionally hidden from public use.

By default, this library includes the `NetworkAttributesSpanAppender` which will do what you're expecting -- add the network attributes to all exported spans. I'm guessing that this isn't working for you down in react native land then?

Can you help explain how were you expecting to leverage this with react native?

Archish27:
```
'NetworkAttributesSpanAppender(io.opentelemetry.android.instrumentation.network.CurrentNetworkProvider)' is not public in 'io.opentelemetry.android.instrumentation.network.NetworkAttributesSpanAppender'. Cannot be accessed from outside package
```

I tried using NetworkAttributesSpanAppender doesn't seem to work, but correct use case is to set network attributes on custom span.

breedx-splk:
> ```
> 'NetworkAttributesSpanAppender(io.opentelemetry.android.instrumentation.network.CurrentNetworkProvider)' is not public in 'io.opentelemetry.android.instrumentation.network.NetworkAttributesSpanAppender'. Cannot be accessed from outside package
> ```

Right, that was my point. It's intentionally kept private.

> I tried using NetworkAttributesSpanAppender doesn't seem to work, but correct use case is to set network attributes on custom span.

So why do you need to set network attributes down in react native land when there are already facilities to do this build into this sdk? In other words, why are you not able to let `opentelemetry-android` set these attributes for you? Are there some cases/conditions that are missing or ?

Issue #212, opened 2024-01-08
Using logs to send crashes


LikeTheSalad:
The current implementation of the crash reporter feature relies on the use of [immediately-ended spans](https://github.com/open-telemetry/opentelemetry-android/blob/d7d5045c8178b0a6069ae93c1cd0b11dc6641fe0/instrumentation/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashReportingExceptionHandler.java#L44) to report crash events, this is most likely due to historical reasons since spans were available first in the Java SDK before log signals, so now that we have logs available, we should send standalone exception events using [log signals](https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-logs/).

breedx-splk:
Do you think that eventually this should just be an event? I think I do. Does it make sense to try and send the crash using logs while the events SIG is still getting the events signal sorted out?

LikeTheSalad:
> Do you think that eventually this should just be an event?

Yes, it makes sense as an event.

> Does it make sense to try and send the crash using logs while the events SIG is still getting the events signal sorted out?

I think it depends on how far away we are from getting events sorted out, if we had to wait for long enough that we'd get a considerable size of users in the meantime, I think logs would make more sense during that time to make the whole process (of both, adapting analytics tools to display errors in the meantime as well as migrating to events when they are finally sorted) smoother, for a couple of reasons:

- Spans are designed to cover a period of time and some vendors' analytics tools might have a dedicated UI to show them based on that assumption, if we go with spans in the meantime, this could mean having to add special validations for both, not displaying those types of spans in the "spans UI", and also to display those specific spans as errors (where some vendors might already have used logs in the past to report errors even before the OTel spec mentioned them).
- Regardless of the outcome of how OTel events should look like, their underlying signal will still be logs, so using logs in the meantime is taking a step further in the event's direction, which could considerably make the migration to the events spec smoother when it's ready.

Do you have an estimated ETA for when the events should be fully sorted?

breedx-splk:
> Do you have an estimated ETA for when the events should be fully sorted?

Unfortunately no, and I think you do make a good point about logs and adoption. A span is definitely a more confusing and inappropriate signal in comparison.

LikeTheSalad:
> Using spans for errors (exceptions) is odd, logs make much more sense to me, and it's semantically closer to just a raw exception.

[...]

PR #237, opened 2024-02-01
Crashes as logs

LikeTheSalad:
Related to #212


instrumentation/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashReporter.java
```diff
@@ -38,16 +45,47 @@ public void installOn(InstrumentedApplication instrumentedApplication) {
...
+    }
+
+    private String stackTraceToString(Throwable throwable) {
+        StringWriter sw = new StringWriter();
+        PrintWriter pw = new PrintWriter(sw);
+
+        throwable.printStackTrace(pw);
+        pw.flush();
+
+        return sw.toString();
```
marandaneto, line 82: for some reason Timber allocates only 256 and flush manually, see https://github.com/JakeWharton/timber/blob/dfc4e70bd1a4b4e1e136647b96fc9c9dbcb41f65/timber/src/main/java/timber/log/Timber.kt#L175-L181
not sure there's any performance gain.

instrumentation/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashReporter.java
```diff
@@ -38,16 +45,47 @@ public void installOn(InstrumentedApplication instrumentedApplication) {
...
+        }
+
+        crashReporter.logRecordBuilder().setAllAttributes(attributesBuilder.build()).emit();
+    }
+
+    private String stackTraceToString(Throwable throwable) {
+        StringWriter sw = new StringWriter();
+        PrintWriter pw = new PrintWriter(sw);
+
+        throwable.printStackTrace(pw);
```
marandaneto, line 79: does the `printStackTrace` also stringify all the [inner exceptions](https://docs.oracle.com/javase/8/docs/api/java/lang/Throwable.html#getCause--)? if not, we'd need to iterate the cause recursively and append it in a string builder maybe.

instrumentation/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashReporter.java
```diff
@@ -38,16 +45,47 @@ public void installOn(InstrumentedApplication instrumentedApplication) {
...
-                .addAttributesExtractors(additionalExtractors)
-                .buildInstrumenter();
+    private void emitCrashEvent(Logger crashReporter, CrashDetails crashDetails) {
+        Throwable throwable = crashDetails.getCause();
+        Thread thread = crashDetails.getThread();
+        AttributesBuilder attributesBuilder =
+                Attributes.builder()
+                        .put(SemanticAttributes.EXCEPTION_ESCAPED, true)
+                        .put(SemanticAttributes.THREAD_ID, thread.getId())
+                        .put(SemanticAttributes.THREAD_NAME, thread.getName())
```
marandaneto, line 63: Maybe worth considering adding a new attribute if it is the main thread or not

instrumentation/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashReporter.java
LikeTheSalad: That's quite interesting, the 256 initial buffer size seems related to performance indeed. I'm usually not keen on doing premature optimizations, though I think, based on how popular Timber is, I guess it's worth following their approach.

Regarding the `and flush manually` part I'm a bit confused because it seems like that's the default behavior of `PrintWriter`, so I'm not sure why they had to use that constructor if they didn't need to change the defaults.

instrumentation/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashReporter.java
LikeTheSalad: Yes, it prints the "Caused by: ..." parts too.

instrumentation/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashReporter.java
LikeTheSalad: Hmm, I'm actually not sure if the concept of "main thread" is as relevant in every other platform as it is in Android in order to have it as a general thread attr, but I guess we can always try and check what people think about it. If you have some use cases in mind that would benefit from this attr, I think you could mention them in an issue in [this repo](https://github.com/open-telemetry/semantic-conventions), or maybe just create a draft adding the new attr [in this file](https://github.com/open-telemetry/semantic-conventions/blob/0b437c0edf3221786cb516c7f24c526d2a39a6a2/docs/attributes-registry/thread.md), is usually easier to convey ideas with code changes rather than only with a description in an issue.

instrumentation/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashReporter.java
LikeTheSalad: I've added the 256 initial size.

marandaneto: probably because you call `flush` manually anyway? so makes sense that the autoFlush=false, unless the default is false already

instrumentation/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashReporter.java
marandaneto: good point, will do, just for context, exception triage, IMO main thread errors are most likely more important than background errors (if the thread is a daemon) maybe the app won't even crash, but not sure if in this case the default signal handler would be called anyway.

instrumentation/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashReporter.java
LikeTheSalad: Yeah, the default is false, that's why I'm a little confused about the code in Timber 😅

instrumentation/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashReporter.java
```diff
@@ -38,16 +45,47 @@ public void installOn(InstrumentedApplication instrumentedApplication) {
...
+                        .put(SemanticAttributes.EXCEPTION_MESSAGE, throwable.getMessage())
+                        .put(SemanticAttributes.EXCEPTION_STACKTRACE, stackTraceToString(throwable))
+                        .put(SemanticAttributes.EXCEPTION_TYPE, throwable.getClass().getName());
+
+        for (AttributesExtractor<CrashDetails, Void> extractor : additionalExtractors) {
+            extractor.onStart(attributesBuilder, Context.current(), crashDetails);
+        }
+
+        crashReporter.logRecordBuilder().setAllAttributes(attributesBuilder.build()).emit();
```
breedx-splk, line 72: One bit that's non-obvious with this switch of signal types is that the attributes that might currently be applied from the `GlobalAttributesSpanAppender` are no longer applicable, so those attributes will be missing. I think that sort of breaking change is fine now, but I don't want us to lose sight of that. 

Would you mind opening an issue to track it and brainstorm solutions?

instrumentation/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashReporter.java
```diff
@@ -38,16 +45,47 @@ public void installOn(InstrumentedApplication instrumentedApplication) {
...
+        throwable.printStackTrace(pw);
+        pw.flush();
+
+        return sw.toString();
+    }
+
+    private Consumer<CrashDetails> buildInstrumenter(LoggerProvider loggerProvider) {
+        return crashDetails ->
+                emitCrashEvent(
+                        loggerProvider.loggerBuilder("io.opentelemetry.crash").build(),
```
breedx-splk, line 88: While I do understand that "crashes" aren't something that happens all the time, it's possible that the consumer here gets called multiple times? It reads a little oddly to me that each time we want to emit a crash event we create a new logger. 

I think instead you should create the logger above the return and then reference it so that it will both front-load the object creation, which should make the consumer more lightweight, and also allows for reuse without having to create an entirely new logger each time.

instrumentation/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashReporter.java
breedx-splk: Oh, and from a RUM standpoint, most importantly, I think this change, as it exists right now, prevents the session id from being attached to these crash LogRecords, right? I think that's pretty important.

instrumentation/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashReporter.java
LikeTheSalad: Thanks, that makes sense. I've updated it.

instrumentation/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashReporter.java
LikeTheSalad: Good point, we should definitely allow for global attrs to be present in all types of signals. I've created this issue to work on it: https://github.com/open-telemetry/opentelemetry-android/issues/256

marandaneto (approved):

instrumentation/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashReporter.java
LikeTheSalad: I've created this PR to support global attrs in logs: https://github.com/open-telemetry/opentelemetry-android/pull/266

breedx-splk (approved):

PR #249, opened 2024-02-07
Add issue automation

### LINKED ISSUE #243: Wire up stale issue workflow automation
We want this to look a lot like the upstream autoinstrumentation workflow here: 

* https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/.github/workflows/issue-management-stale-action.yml
* https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/.github/workflows/issue-management-feedback-label.yml
* not sure where the close is, but it's there. :)

* Mark stale after 3 weeks and `needs author feedback` label
* Remove label when OP responds
* Close after 1 week of stale

breedx-splk:
Resolves #243.

These are copied over from upstream instrumentation repo as linked in #243. Because this repo isn't yet as active, I decided to leave stale around for 2 weeks before closing.

### DIFF
--- a/.github/workflows/add-stale-label.yaml
+++ b/.github/workflows/add-stale-label.yaml
@@ -1,0 +1,35 @@
+name: Issue management - run stale action
+
+on:
+  schedule:
+    # hourly at minute 23
+    - cron: "23 * * * *"
+
+permissions:
+  contents: read
+
+jobs:
+  stale:
+    permissions:
+      issues: write         # for actions/stale to close stale issues
+      pull-requests: write  # for actions/stale to close stale PRs
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/add-stale-label.yaml
```diff
@@ -0,0 +1,35 @@
+name: Issue management - run stale action
+
+on:
+  schedule:
+    # hourly at minute 41
+    - cron: "41 * * * *"
```
LikeTheSalad, line 6: A bit specific but still works for me 😄

marandaneto: https://affinitynumerology.com/number-meanings/number-41-meaning.php

breedx-splk: LOL I just didn't want it to be the same as the other repos, and I didn't want it to be at the top of the hour. 😁

PR #269, opened 2024-03-07
Restructure modules phase 2

breedx-splk:
Continuation of #267.

This creates the `annotations` and `common-api` modules that other instrumentations will depend on in phase 3. 

Once this goes it, I will move over the instrumentations to look more like the [experimental branch](https://github.com/breedx-splk/opentelemetry-android/tree/restructure_repo_modules/instrumentation) we discussed previously.


instrumentation/annotations/src/main/java/io/opentelemetry/android/instrumentation/annotations/RumScreenName.java
LikeTheSalad, line 1: It seems like the "annotations" module only has this class which is only used in "common-api", is there a reason we should create a separate module for annotations for this use case, instead of having it in "common-api"?

breedx-splk: Right, it's currently only used by the `ScreenNameExtractor`. I thought we might introduce more in the future, and it would be convenient for users to have them in a single package and artifact. If you think that grouping is strange or clumsy or something, I can move it next to the place where it's used for now and we can evaluate in the future if/when we have more annotations.

Please let me know your preference and I'm happy to change.

breedx-splk: Tons of ideas about future annotations, but we don't need to future proof. I'll combine and we can split apart later when the desire arises.

LikeTheSalad: Thank you!

LikeTheSalad (approved):

PR #272, opened 2024-03-12
Initializing services

LikeTheSalad:


buildSrc/src/main/kotlin/otel.publish-conventions.gradle.kts
```diff
@@ -79,7 +79,7 @@ afterEvaluate {
 
 fun computeArtifactId(): String {
     val path = project.path
-    if (!path.contains("auto-instrumentation")) {
+    if (!path.contains("instrumentation")) {
```
LikeTheSalad, line 82: This change is not needed to initialize services, although I had to do some local tests and noticed that the instrumentation artifacts were not found due to their recent name change. So this change makes the instrumentation artifact's names stay the same from before the "auto-instrumentation" module was renamed to "instrumentation".

marandaneto: I'd probably make this type-safe, a list of modules with their types, so you can check if the artifact here is part of the list (if possible), maybe you don't have access to this info here, just throwing an idea.

marandaneto (approved):

android-agent/src/test/java/io/opentelemetry/android/OpenTelemetryRumBuilderTest.java
```diff
@@ -308,6 +309,17 @@ void verifyGlobalAttrsForLogs() {
...
 
+    @Test
+    void verifyServicesAreInitialized() {
+        makeBuilder().build();
+
+        try {
+            ServiceManager.get();
+        } catch (IllegalStateException e) {
+            fail("Services were not initialized.");
+        }
```
breedx-splk, line 320: Prefer to just let the exception leave the method and fail the test. I might also add a non-null on the result of `get()`.

LikeTheSalad: Thanks, it's updated now.

buildSrc/src/main/kotlin/otel.publish-conventions.gradle.kts
breedx-splk: @LikeTheSalad Great catch! Thank you.

@marandaneto I appreciate where you are coming from. I have opened this issue to track this: https://github.com/open-telemetry/opentelemetry-android/issues/273

LikeTheSalad: > I'd probably make this type-safe, a list of modules with their types, so you can check if the artifact here is part of the list (if possible), maybe you don't have access to this info here, just throwing an idea.

I'm not sure if we have that info available but I'm open to ideas! I was planning to talk about this in today's Android SIG to get more details on it 😅

android-agent/src/main/java/io/opentelemetry/android/internal/services/ServiceManager.java
```diff
@@ -25,9 +26,14 @@ public final class ServiceManager implements Lifecycle {
...
     @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP_PREFIX)
     public static void initialize(Context appContext) {
+        if (instance != null) {
+            // Already initialized.
+            return;
+        }
         instance = new ServiceManager();
         instance.addService(PreferencesService.create(appContext));
         instance.addService(new CacheStorageService(appContext));
+        instance.addService(new PeriodicWorkService());
```
breedx-splk, line 36: This is technically a functional change, do we want a test around this?

breedx-splk (approved):
Thanks, we need this! 😁

buildSrc/src/main/kotlin/otel.publish-conventions.gradle.kts
breedx-splk: We commented at the same time, heh. I made an issue to track this.

android-agent/src/main/java/io/opentelemetry/android/internal/services/ServiceManager.java
LikeTheSalad: I've added some validations for the registered services. It was easier by making it a robolectric test though so I had to change it to junit4 to make it work.

PR #276, opened 2024-03-13
Restructure modules -  phase 3

breedx-splk:
This is the continuation of #267 and #269. 

This migrates the subdirs in instrumentation to be their own instrumentation modules, as discussed previously.


android-agent/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
```diff
@@ -370,7 +372,9 @@ private void applyConfiguration() {
             if (initializationEvents == InitializationEvents.NO_OP) {
                 initializationEvents = new SdkInitializationEvents();
             }
-            initializationEvents.recordConfiguration(config);
+            Map<String, String> configMap = new HashMap<>();
+            // TODO: Convert config to map
```
breedx-splk, line 376: I opened https://github.com/open-telemetry/opentelemetry-android/issues/277 to track this.

instrumentation/activity/src/test/java/io/opentelemetry/android/instrumentation/activity/VisibleScreenTrackerTest.java
```diff
@@ -22,17 +22,17 @@ void activityLifecycle() {
         VisibleScreenTracker visibleScreenTracker = new VisibleScreenTracker();
         Activity activity = mock(Activity.class);
 
-        assertEquals("unknown", visibleScreenTracker.getCurrentlyVisibleScreen());
+        Assertions.assertEquals("unknown", visibleScreenTracker.getCurrentlyVisibleScreen());
```
LikeTheSalad, line 25: I've seen a lot of these changes which seem to make the PR bigger than it should be, is it needed to change the imports for these tests?

instrumentation/startup/src/main/java/io/opentelemetry/android/instrumentation/startup/InitializationEvents.java
```diff
@@ -5,14 +5,14 @@
...
-import io.opentelemetry.android.config.OtelRumConfig;
 import io.opentelemetry.sdk.trace.export.SpanExporter;
+import java.util.Map;
 
 public interface InitializationEvents {
 
     void sdkInitializationStarted();
 
-    void recordConfiguration(OtelRumConfig config);
+    void recordConfiguration(Map<String, String> config);
```
LikeTheSalad, line 15: If I understand correctly, this is needed to avoid a circular dependency with the `android-agent` module, right?

android-agent/build.gradle.kts
LikeTheSalad, line 1: Do you think we should add the fragment module too?

instrumentation/activity/src/test/java/io/opentelemetry/android/instrumentation/activity/VisibleScreenTrackerTest.java
breedx-splk: No, this was unintentional. I'll remove them. Thanks!

instrumentation/startup/src/main/java/io/opentelemetry/android/instrumentation/startup/InitializationEvents.java
breedx-splk: Yeah exactly.

android-agent/build.gradle.kts
breedx-splk: It's used indirectly through the Activity instrumentation. If we want to change that, can we do it in a separate PR plz?

LikeTheSalad: Ah got it, that's fine then, I thought it was not being added. Cheers!

android-agent/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
LikeTheSalad: I wish there could be a way to pass the config object without having to transform it into a map though, it feels awkward. But I guess we can leave it as is for now. I'll follow up on the ticket on this.

LikeTheSalad (approved):

PR #291, opened 2024-04-01
Add Volley HTTP client instrumentation.

breedx-splk:
This is a taken from the Splunk Volley http client instrumentation, which was not included in the initial donation.

https://github.com/signalfx/splunk-otel-android/tree/main/splunk-otel-android-volley

There are 2 primary differences:

1. Updated to use the newest/latest http semantic conventions
2. Does not attempt to understand or parse the `Server-Timing` header


instrumentation/volley/library/build.gradle.kts
```diff
@@ -0,0 +1,72 @@
+import java.time.Duration
+
+plugins {
+    id("com.android.library")
+    id("otel.android-library-conventions")
+    id("otel.errorprone-conventions")
```
LikeTheSalad, line 6: Both plugins, `otel.errorprone-conventions` and `com.android.library`, are added by `otel.android-library-conventions`, so it's not needed to explicitly declare them here ;)

instrumentation/volley/library/build.gradle.kts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+android {
+    namespace = "io.opentelemetry.android.volley"
+
+    compileSdk = 34
+    buildToolsVersion = "34.0.0"
+
+    defaultConfig {
+        minSdk = 21
```
LikeTheSalad, line 16: The values `minSdk` and `compileSdk` are defined in the conventions so it's better to leave them there only to try and keep consistency across all android libs in this repo (unless it's needed to override them depending on the case).

And regarding `buildToolsVersion`, I think it's safe to remove it since the AGP infers it from elsewhere (don't remember exactly where from though) but it's no longer mandatory to define it, which is nice IMO (a version less to take care of) 😌

instrumentation/volley/library/build.gradle.kts
```diff
@@ -0,0 +1,72 @@
...
+        consumerProguardFiles("consumer-rules.pro")
+    }
+
+    buildTypes {
+        release {
+            isMinifyEnabled = false
+            proguardFiles(getDefaultProguardFile("proguard-android-optimize.txt"), "proguard-rules.pro")
+        }
+    }
+    compileOptions {
```
LikeTheSalad, line 28: All this `compileOptions` block is already defined in the conventions so it's not needed to define it here too.

instrumentation/volley/library/build.gradle.kts
```diff
@@ -0,0 +1,72 @@
...
+                events("started", "passed", "failed")
+            }
+        }
+    }
+}
+
+val otelVersion = "1.32.1"
+val otelSdkVersion = "1.35.0"
+val otelAlphaVersion = otelVersion.replaceFirst("(-SNAPSHOT)?$".toRegex(), "-alpha$1")
+val otelSemconvVersion = "1.23.1-alpha"
```
LikeTheSalad, line 50: It seems like we can delete these vars.

instrumentation/volley/library/src/main/java/io/opentelemetry/android/instrumentation/volley/UrlParser.java
LikeTheSalad, line 1: In case it helps (not needed for this PR) but it seems like the tools provided in this class might be provided by an Android SDK one already: https://developer.android.com/reference/android/net/Uri - So if that's the case, maybe we can take advantage of it in the future.

instrumentation/volley/library/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
LikeTheSalad, line 1: Do we need this file?

LikeTheSalad (commented):
I have mostly Gradle-related comments, but apart from that, I don't see any issues with the code. Just to confirm tho, as part of this PR, there are no plans to add automatic instrumentation support for Volley?

instrumentation/volley/library/build.gradle.kts
breedx-splk: I bet this was just carry-over from the old way. Will fix. Thanks!

instrumentation/volley/library/build.gradle.kts
breedx-splk: Yeah, another carryover. Thanks!

instrumentation/volley/library/src/main/java/io/opentelemetry/android/instrumentation/volley/UrlParser.java
breedx-splk: Oh I agree that we shouldn't be reinventing this. I can't remember the exact history, but let's definitely try and go forward with the `android.net.Uri implementation`. I'll make that change.

instrumentation/volley/library/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
breedx-splk: Probably not anymore? At some point in the past it was required to mock certain types (final? I can't remember) but it looks like the tests pass now without it so I'll remove it.

instrumentation/volley/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Add to your project
+
+To use this library instrumentation, you will need to add
+the dependency in your `build.gradle.kts`:
+
+```gradle
+dependencies {
+    ...
+    implementation("io.opentelemetry.android:instrumentation-volley:{version}")
```
LikeTheSalad, line 18: I'm not sure if this would be the uri, it might be `io.opentelemetry.android:volley-library:{version}` based on the dirs structure though I'm not fully sure.

Btw we could change the logic to set the artifact names of instrumentation modules in case the existing one isn't intuitive.

LikeTheSalad (approved):

Issue #275, opened 2024-03-12
SNAPSHOT artifacts are missing `alpha` in the name.


breedx-splk:
Maybe we don't care too much since you already have to jump through hoops to use the SNAPSHOT builds, but there is definitely an inconsistency. I think we should include `alpha` in the SNAPSHOT filenames as well.

Here's an example:
https://oss.sonatype.org/content/repositories/snapshots/io/opentelemetry/android/agent/maven-metadata.xml

```xml

<metadata>
  <groupId>io.opentelemetry.android</groupId>
  <artifactId>agent</artifactId>
  <versioning>
    <latest>0.5.0-SNAPSHOT</latest>
    <versions>
      <version>0.5.0-SNAPSHOT</version>
    </versions>
    <lastUpdated>20240312081534</lastUpdated>
  </versioning>
</metadata>
```

breedx-splk:
Hmmm, maybe this isn't still an issue, just an artifact of the multi-step project reorg?
https://oss.sonatype.org/content/repositories/snapshots/io/opentelemetry/android/android-agent/

marandaneto:
sonatype should not have SNAPSHOT artifacts, they have a different repository for that.
https://central.sonatype.org/publish/publish-guide/#accessing-repositories

what I usually do/did is that I don't publish snapshots in the snapshot repository either but I publish semver builds in the stable repository with suffixes (alpha, beta, RC, stable).
Also, check this recommendation for the suffixes https://publicobject.com/2019/12/18/naming-versions/ otherwise you might get the wrong version as the latest stable version (sorting issue).

LikeTheSalad:
I think this only happens to all the modules except for the `android-agent` one, because the logic to append "-alpha" is added only there: https://github.com/open-telemetry/opentelemetry-android/blob/783ef1b8ae48ecbf4d0c4226c0fa35fe220da2b4/android-agent/build.gradle.kts#L8 

So I think it should be a matter of moving that logic to the convention scripts to cover all other modules as well.

breedx-splk:
Fixed in #297

PR #297, opened 2024-04-09
Adding '-alpha' to all modules' versions

LikeTheSalad:
Right now the version of the released artifacts from this project don't match. The `android-agent` one has the format: `0.x.x-alpha`, while all the other modules have this other one: `0.x.x`.

These changes aim to change the version formatting of all the artifacts in this project to match the one used by `android-agent`, i.e. appending `-alpha` to it.

### DIFF
--- a/android-agent/build.gradle.kts
+++ b/android-agent/build.gradle.kts
@@ -1,11 +1,7 @@
-
 plugins {
     id("otel.android-library-conventions")
     id("otel.publish-conventions")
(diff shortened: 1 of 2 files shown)


breedx-splk:
Resolves #275

breedx-splk (approved):
Thanks for taking this on, it's been nagging at me since the project reorg. 👍🏻

PR #300, opened 2024-04-10
Update Release process

### LINKED ISSUE #262: Update release process to use tags
The current release process performs a release from a branch, typically `main`. This is risky, because there is a window of time during which unwanted changes could be merged. 

I think we should create and push a tag _first_, and the release build can then target that tag. That should give us significantly more safety around locking in what we intend to publish.

breedx-splk:
Resolves #262.

This borrows HEAVILY from the java instrumentation ([upstream](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/.github/workflows)) [release process](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/RELEASING.md#preparing-a-new-major-or-minor-release). 
Would definitely like to try this out on the next release build, because there are bound to be some rough edges.

There's also still some room for improvement, like we don't yet have anything to help generate the CHANGELOG entries, and we're only adding the topmost agent `aar` to the release page....but we can improve things later.


breedx-splk:
@trask since you know this stuff really well it would be great to have your eyes on this too. ❤️

LikeTheSalad (approved):

trask:
> @trask since you know this stuff really well it would be great to have your eyes on this too. ❤️

will review today 👍

.github/workflows/update-version.sh
trask, line 1: I think you want this file under `.github/scripts/` directory

breedx-splk: Thanks, that would have been frustrating. 😅

PR #304, opened 2024-04-15
Update dependency versions.

breedx-splk:
* Supersedes #303
* Supersedes #279 

The instrumentation bom is not alpha, so this updates that while updating us to the latest. It also updates the gradle version wrapper to the latest name.

### DIFF
--- a/.github/workflows/gradle-wrapper-validation.yaml
+++ b/.github/workflows/gradle-wrapper-validation.yaml
@@ -12,4 +12,4 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: gradle/wrapper-validation-action@v2.1.3
+      - uses: gradle/actions/wrapper-validation@v3
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,6 +1,7 @@
 [versions]
 opentelemetry = "1.37.0"
-opentelemetry-alpha = "2.2.0-alpha"
+opentelemetry-inst = "2.3.0"
+opentelemetry-inst-alpha = "2.3.0-alpha"
 opentelemetry-semconv = "1.25.0-alpha"
 opentelemetry-contrib = "1.34.0-alpha"
 mockito = "5.11.0"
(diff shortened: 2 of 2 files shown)


gradle/libs.versions.toml
```diff
@@ -1,6 +1,7 @@
 [versions]
 opentelemetry = "1.37.0"
-opentelemetry-alpha = "2.2.0-alpha"
+opentelemetry-inst = "2.3.0"
```
breedx-splk, line 3: I think it's clearer with `inst` in the name, to know when you have the instrumentation version (vs. core version).

.github/workflows/gradle-wrapper-validation.yaml
```diff
@@ -12,4 +12,4 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: gradle/wrapper-validation-action@v2.1.3
\ No newline at end of file
+      - uses: gradle/actions/wrapper-validation@v3
```
marandaneto, line 15: I usually prefer pinned versions and using [dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates) to open PRs upgrading it.
So I know before merging whether the CI is happy or not.
Using `@v3` will always use the latest patch or minor and might break things requiring immediate action.

marandaneto: Renovate does that already btw eg https://github.com/open-telemetry/opentelemetry-android/pull/303

breedx-splk: Yeah, we pin exact versions in our distro. I guess I can do that here as well. Thanks!

breedx-splk, line 15:
```suggestion
      - uses: gradle/actions/wrapper-validation@v3.3.0
```

gradle/libs.versions.toml
LikeTheSalad: I like the idea of replacing `alpha` with something more meaningful. I would prefer to avoid abbreviations though, I know `opentelemetry-instrumentation` could be a lot, though the first time I saw `inst`, for some reason "instrumentation" wasn't the first word that came to my mind 😅

breedx-splk: Ok, fine, I'll spell out `inst` -> `instrumentation`...but I'm not going to change semconv. :)

LikeTheSalad (approved):

PR #314, opened 2024-04-23
Release branch settings

breedx-splk:
So we need to set up branch protection on the `release/*` branches, and it will need to require `required-status-check` like the other java-based builds.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -40,3 +40,13 @@ jobs:
     uses: ./.github/workflows/reusable-workflow-notification.yml
     with:
       success: ${{ needs.build.result == 'success' }}
+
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -39,4 +39,14 @@ jobs:
...
+      success: ${{ needs.build.result == 'success' }}
+
+  required-status-check:
+    needs:
+      - build
+    runs-on: ubuntu-latest
+    if: always()
+    steps:
+      - if: |
+          needs.build.result != 'success' ||
```
breedx-splk, line 51:
```suggestion
          needs.build.result != 'success'
```

trask (approved):

PR #320, opened 2024-04-23
Fix how next version is calculated during pre-release

breedx-splk:
We append -SNAPSHOT in all non -Pfinal=true builds, so this is not needed.

### DIFF
--- a/.github/workflows/prepare-release-branch.yml
+++ b/.github/workflows/prepare-release-branch.yml
@@ -89,7 +89,7 @@ jobs:
             echo "unexpected version: $version"
             exit 1
           fi
-          echo "NEXT_VERSION=${next_version}-SNAPSHOT" >> $GITHUB_ENV
+          echo "NEXT_VERSION=${next_version}" >> $GITHUB_ENV
           echo "VERSION=$version" >> $GITHUB_ENV
 
       - name: Update version


trask (approved):
another option is to adopt the version management strategy (and automation) from the other java repos, which also append `-SNAPSHOT` to the version, but do it more explicitly, e.g.

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/version.gradle.kts

breedx-splk:
> another option is to adopt the version management strategy (and automation) from the other java repos, which also append `-SNAPSHOT` to the version, but do it more explicitly, e.g.
> 
> https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/version.gradle.kts

Yeah I appreciate that. I can't remember the details about it right now, but I thought there was something Android-specific (AGP?) that preferred having the version in settings, and maybe having suffixes there caused problems. I can't remember. 

@LikeTheSalad do you remember?

I any case, I think that's a much bigger change at this point, and I'm giving preference for keeping it small right now.

LikeTheSalad:
> Thanks, yeah I appreciate that. I can't remember the details about it right now, but I thought there was something Android-specific (AGP?) that preferred having the version in settings, and maybe having suffixes there caused problems. I can't remember.
> 
> @LikeTheSalad do you remember?
> 
> I any case, I think that's a much bigger change at this point, and I'm giving preference for keeping it small right now.

I think the current logic to append `-SNAPSHOT` predates OTel Android so I'm not sure about the reason behind it, however, I don't think it's a bad approach, and I agree that there are already a couple of parts built around that logic so... The benefit that I see of using the same approach as the other repos is WYSIWYG for the build version, however, I think it shouldn't be too strange to get `-SNAPSHOT` appended out of the blue in a non-final build so I'm not sure if it's worth making the changes for that reason. Though now that we're on the topic, I do believe it's quite unexpected to get `-alpha` appended out of the blue, which is something we're currently doing as well 😅 so maybe we should consider removing that logic and just set the version inside `gradle.properties` to `x.x.x-alpha`.

PR #321, opened 2024-04-23
Simplify the release process branch choice.

breedx-splk:
The release process should just use (via checkout) whatever is chosen in the "workflow_dispatch" selection. No additional branch name needed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,12 +3,6 @@ name: Release
 
 on:
   workflow_dispatch:
-    inputs:
-      branch_specifier:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -21,8 +15,6 @@ jobs:
             exit 1
           fi
       - uses: actions/checkout@v4
-        with:
-          ref: ${{ inputs.branch_specifier }}
```
breedx-splk, line 25: Choosing the branch from which to run the workflow is a built-in part of github. No need to duplicate that.

trask (approved):

Issue #329, opened 2024-04-29
Configuring session handler timeout time


Archish27:
Currently, as per the [code ](https://github.com/open-telemetry/opentelemetry-android/blob/66de8ae9a2d993e107765248a75a8183b9e35204/android-agent/src/main/java/io/opentelemetry/android/SessionIdTimeoutHandler.java#L12) We have default timeout of 15mins.

Need a way for configuring this timeout during initialization.

LikeTheSalad:
It seems like a config param [here](https://github.com/open-telemetry/opentelemetry-android/blob/66de8ae9a2d993e107765248a75a8183b9e35204/android-agent/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java#L20) should be ideal for this use case.

PR #330, opened 2024-04-29
chore: added session timeout minutes config in OtelConfig

Archish27:
Fixes : #329

### DIFF
--- a/android-agent/src/main/java/io/opentelemetry/android/SessionIdTimeoutHandler.java
+++ b/android-agent/src/main/java/io/opentelemetry/android/SessionIdTimeoutHandler.java
@@ -7,7 +7,6 @@ package io.opentelemetry.android;
 
 import io.opentelemetry.android.instrumentation.common.ApplicationStateListener;
 import io.opentelemetry.sdk.common.Clock;
-
 import java.util.concurrent.TimeUnit;
 
 /**
@@ -77,9 +76,7 @@ final class SessionIdTimeoutHandler implements ApplicationStateListener {
     private enum State {
         FOREGROUND,
         BACKGROUND,
-        /**
-         * A temporary state representing the first event after the app has been brought back.
-         */
+        /** A temporary state representing the first event after the app has been brought back. */
         TRANSITIONING_TO_FOREGROUND
     }
 }
--- a/android-agent/src/main/java/io/opentelemetry/android/SessionIdTimeoutHandler.java
+++ b/android-agent/src/main/java/io/opentelemetry/android/SessionIdTimeoutHandler.java
@@ -7,6 +7,7 @@ package io.opentelemetry.android;
 
 import io.opentelemetry.android.instrumentation.common.ApplicationStateListener;
 import io.opentelemetry.sdk.common.Clock;
+
 import java.util.concurrent.TimeUnit;
 
 /**
@@ -25,23 +26,24 @@ import java.util.concurrent.TimeUnit;
  */
 final class SessionIdTimeoutHandler implements ApplicationStateListener {
 
-    private long SESSION_TIMEOUT_NANOS = TimeUnit.MINUTES.toNanos(15);
+    private final long sessionTimeoutNanos;
 
     private final Clock clock;
     private volatile long timeoutStartNanos;
     private volatile State state = State.FOREGROUND;
 
     SessionIdTimeoutHandler() {
-        this(Clock.getDefault());
+        this(Clock.getDefault(), TimeUnit.MINUTES.toNanos(15));
+    }
+
+    SessionIdTimeoutHandler(long sessionTimeoutMinutes) {
+        this(Clock.getDefault(), TimeUnit.MINUTES.toNanos(sessionTimeoutMinutes));
     }
 
     // for testing
-    SessionIdTimeoutHandler(Clock clock) {
+    SessionIdTimeoutHandler(Clock clock, long sessionTimeoutNanos) {
         this.clock = clock;
-    }
-
-    public void setSessionTimeout(final int sessionTimeout) {
-        this.SESSION_TIMEOUT_NANOS = TimeUnit.MINUTES.toNanos(sessionTimeout);
+        this.sessionTimeoutNanos = sessionTimeoutNanos;
     }
 
     @Override
@@ -60,7 +62,7 @@ final class SessionIdTimeoutHandler implements ApplicationStateListener {
             return false;
         }
         long elapsedTime = clock.nanoTime() - timeoutStartNanos;
-        return elapsedTime >= SESSION_TIMEOUT_NANOS;
+        return elapsedTime >= sessionTimeoutNanos;
     }
 
     void bump() {
--- a/android-agent/src/main/java/io/opentelemetry/android/SessionIdTimeoutHandler.java
+++ b/android-agent/src/main/java/io/opentelemetry/android/SessionIdTimeoutHandler.java
@@ -25,7 +25,7 @@ import java.util.concurrent.TimeUnit;
  */
 final class SessionIdTimeoutHandler implements ApplicationStateListener {
 
-    private static final long SESSION_TIMEOUT_NANOS = TimeUnit.MINUTES.toNanos(15);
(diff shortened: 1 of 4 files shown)


breedx-splk:
Hey thanks @Archish27. Take note that [the build is currently broken](https://github.com/open-telemetry/opentelemetry-android/actions/runs/8880916607/job/24382060962?pr=330#step:4:544) and you should run `./gradlew spotlessApply` to reformat the code and then push. Thanks!

android-agent/src/main/java/io/opentelemetry/android/SessionIdTimeoutHandler.java
```diff
@@ -40,6 +40,10 @@ final class SessionIdTimeoutHandler implements ApplicationStateListener {
         this.clock = clock;
     }
 
+    public void setSessionTimeout(final int sessionTimeout) {
+        this.SESSION_TIMEOUT_NANOS = TimeUnit.MINUTES.toNanos(sessionTimeout);
+    }
```
breedx-splk, line 45: We prefer immutable objects. This field should be made final and passed into the constructor.

android-agent/src/main/java/io/opentelemetry/android/SessionIdTimeoutHandler.java
```diff
@@ -25,7 +25,7 @@
  */
 final class SessionIdTimeoutHandler implements ApplicationStateListener {
 
-    private static final long SESSION_TIMEOUT_NANOS = TimeUnit.MINUTES.toNanos(15);
+    private long SESSION_TIMEOUT_NANOS = TimeUnit.MINUTES.toNanos(15);
```
breedx-splk, line 28: Fields should be camel case, and we want to favor immutability.

```suggestion
    private final long sessionTimeoutNanos;
```

android-agent/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
```diff
@@ -179,14 +194,33 @@ public OtelRumConfig setSlowRenderingDetectionPollInterval(Duration duration) {
...
+        this.sessionTimeoutMinutes = sessionTimeoutMinutes;
+        return this;
+    }
+
+    /**
+     * Call this method to retrieve session timeout
+     */
+    public int getSessionTimeoutMinutes() {
+        return sessionTimeoutMinutes;
+    }
```
breedx-splk, line 225: Prefer to use the general purpose `Duration` instead of explicit minutes.

breedx-splk (changes requested):
Thanks for the contribution @Archish27...I think this approach is solid, but I have noted a few changes that I think need to be made. Furthermore, the `SdkPreconfiguredRumBuilder` will also need to be modified to take the config and pass the configured timeout to the `SessionIdTimeoutHandler` that it creates. 

Both of these cases are violations of the dependency injection principle, so maybe a better change is to have the `SessionId` passed into the constructors.

Thanks again!

android-agent/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
LikeTheSalad: ➕

LikeTheSalad:
> Both of these cases are violations of the dependency injection principle, so maybe a better change is to have the `SessionId` passed into the constructors.

I agree, just to provide a bit more details, I think the steps to address this could be:

- Removing [this constructor](https://github.com/open-telemetry/opentelemetry-android/blob/66de8ae9a2d993e107765248a75a8183b9e35204/android-agent/src/main/java/io/opentelemetry/android/SdkPreconfiguredRumBuilder.java#L24).
- Adding a new SdkPreconfiguredRumBuilder `builder` function into `OpenTelemetryRum`, similar to [this one](https://github.com/open-telemetry/opentelemetry-android/blob/66de8ae9a2d993e107765248a75a8183b9e35204/android-agent/src/main/java/io/opentelemetry/android/OpenTelemetryRum.java#L62), though with a `SessionId` param.

Archish27:
> Adding a new SdkPreconfiguredRumBuilder builder function into OpenTelemetryRum, similar to [this one](https://github.com/open-telemetry/opentelemetry-android/blob/66de8ae9a2d993e107765248a75a8183b9e35204/android-agent/src/main/java/io/opentelemetry/android/OpenTelemetryRum.java#L62), though with a SessionId param.

Hey @LikeTheSalad SessionId seems to be getting exposed outside classes is the expected behavior? Also, along with SessionTimeoutHandler will get exposed too. Can you suggest what can we do here?

image

android-agent/src/main/java/io/opentelemetry/android/SessionIdTimeoutHandler.java
Archish27: Done fixed ✅

android-agent/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
Archish27: Done fixed ✅

LikeTheSalad:
> > Adding a new SdkPreconfiguredRumBuilder builder function into OpenTelemetryRum, similar to [this one](https://github.com/open-telemetry/opentelemetry-android/blob/66de8ae9a2d993e107765248a75a8183b9e35204/android-agent/src/main/java/io/opentelemetry/android/OpenTelemetryRum.java#L62), though with a SessionId param.
> 
> Hey @LikeTheSalad SessionId seems to be getting exposed outside classes is the expected behavior? Also, along with SessionTimeoutHandler will get exposed too. Can you suggest what can we do here?
> 
> image

Good point, we still haven't fully figured out what will be part of the public and/or internal classes so I don't have a clear answer here, although I think we can just expose them for now, taking into account that this project is not stable yet. And then we can re-evaluate this later once we have all the features we need for the first stable version and see if this makes sense as a public class or not within that state of the project. Otherwise, I think that slowing down development over these situations while the project is not stable would defeat the purpose of having it marked as not-stable right now. Any objections/thoughts on this @breedx-splk ?

android-agent/src/main/java/io/opentelemetry/android/SessionIdTimeoutHandler.java
```diff
@@ -25,19 +25,24 @@
...
     private final Clock clock;
     private volatile long timeoutStartNanos;
     private volatile State state = State.FOREGROUND;
 
     SessionIdTimeoutHandler() {
-        this(Clock.getDefault());
+        this(Clock.getDefault(), TimeUnit.MINUTES.toNanos(15));
+    }
+
+    SessionIdTimeoutHandler(long sessionTimeoutMinutes) {
```
bidetofevil, line 38: Any reason we don't use `Duration` here as well to make as more flexible? A granularity of minutes seems a bit coarse for general usage. Doing that will also mean we don't have to add a separate method just for testing, reducing the API surface.

breedx-splk: Agreed, cleaner to use Duration here as well.

breedx-splk:
> Good point, we still haven't fully figured out what will be part of the public and/or internal classes so I don't have a clear answer here, although I think we can just expose them for now, taking into account that this project is not stable yet. 

Yeah, I was surprised that `SessionId` wasn't already public!

> And then we can re-evaluate this later once we have all the features we need for the first stable version and see if this makes sense as a public class or not within that state of the project. Otherwise, I think that slowing down development over these situations while the project is not stable would defeat the purpose of having it marked as not-stable right now. Any objections/thoughts on this @breedx-splk ?

No objections, I think I agree. Fixing this is definitely not in scope for this PR, which I think is getting pretty close to being ready. Long-term, I think the right approach is to make `SessionId` an interface whose only no-op implementation exists in the agent sdk. This allows us to have public api surface area while keeping some of the implementation details private/internal.

breedx-splk (approved):
I'm fine with us changing to Duration in a follow-up PR. I think that's the last piece of feedback. Thanks!

PR #334, opened 2024-05-01
Migrate DiskManager to Kotlin

breedx-splk:
So this started as a change to limit the coupling to the global ServiceManager. Initially it was just going to be a refactor, but it seemed to make sense to leverage some of the kotlin language features in the process. Because the `SimpleTemporaryFileProvider` was right next door, I picked that up as well.


android-agent/src/main/java/io/opentelemetry/android/internal/features/persistence/DiskManager.kt
```diff
@@ -0,0 +1,114 @@
...
+import java.io.IOException
+
+internal class DiskManager(
+    private val cacheStorageService: CacheStorageService,
+    private val preferencesService: PreferencesService,
+    private val diskBufferingConfiguration: DiskBufferingConfiguration,
+) {
+    @get:Throws(IOException::class)
+    val signalsBufferDir: File
+        get() {
```
breedx-splk, line 23: I _think_ this is fairly idiomatic these days....but I'm happy to be talked out of this style/syntax if it is important to anyone.

LikeTheSalad: That's fair enough, considering that Kotlin allows you to access getter methods as if they were class fields later on, it seems like the question of defining them like this vs regular getters boils down to personal preferences really. I've seen custom accessors widely used for cases where the `get()` or `set()` body is quite small, like this one, although it does seem a bit strange to me to see them as big as the `get()` in `maxFolderSize` for example.

Regarding my personal preferences, I find it easier to read and quickly set apart in my mind what parts in a class hold business logic vs what others only store things in memory by using regular methods and fields, although it's probably a matter of getting used to different styles, so it's not the end of the world for me to use either approach.

LikeTheSalad (approved):

android-agent/src/main/java/io/opentelemetry/android/internal/features/persistence/DiskManager.kt
LikeTheSalad: That's fair enough, considering that Kotlin allows you to access getter methods as if they were class fields later on, it seems like the question of defining them like this vs regular getters boils down to personal preferences really. I've seen custom accessors widely used for cases where the `get()` or `set()` body is quite small, like this one, although it does seem a bit strange to me to see them as big as the `get()` in `maxFolderSize` for example.

Regarding my personal preferences, I find it easier to read and quickly set apart in my mind what parts in a class hold business logic vs what others only store things in memory by using regular methods and fields, although it's probably a matter of getting used to different styles, so it's not the end of the world for me to use either approach.

PR #338, opened 2024-05-03
Spike demo app

### LINKED ISSUE #8: Placeholder for sample app development
The sample application in the splunk repo was not brought over as part of the initial donation because there is a lot of vendor-specific behavior in there. It is super valuable, however, and we should really consider boostrapping a sample application in this repo as well.

The idea is to provide new users a concrete, hands-on example of how to wire up the Android SDK in an application and how to configure the underlying components, like the opentelemetry java sdk and exporters.

We should start small and incrementally add features.

breedx-splk:
Resolves #8.

This lays a foundation on which we can continue building a sample app, which I'm now calling a demo app. Right now it looks like this:
image

Clicking the telescope logo will generate a contrived span. Clicking the "Click to begin" button will navigate to the next Activity:

image

So yeah, it's just a skeleton/foundation. If we like this approach, I will file a series of enhancement issues along the lines of:
* populate the session id on the main screen
* add a copy button for the session id 
* make it easy to run a collector+jaeger or something via docker/compose so that we can actually see the data
* in the second activity, actually implement the AstronomyShop, similar to what's done in the main otel demo
* add special places to demonstrate features
* etc. etc.


.github/workflows/pr-check.yaml
```diff
@@ -13,6 +13,8 @@ jobs:
         with:
           distribution: temurin
           java-version: 17
+      - name: touch local props
+        run: touch local.properties
```
LikeTheSalad, line 17: Do we need this file when running in a Github action?

demo-app/build.gradle.kts
```diff
@@ -0,0 +1,88 @@
+import java.io.FileInputStream
+import java.util.Properties
+
+plugins {
+    id("otel.android-app-conventions")
+}
+
+val localProperties = Properties()
+localProperties.load(FileInputStream(rootProject.file("local.properties")))
```
LikeTheSalad, line 9: If I'm not mistaken, I think we don't need to manually load preferences from the `local.preferences` file, I remember it being done automatically in android projects.

demo-app/build.gradle.kts
```diff
@@ -0,0 +1,88 @@
...
+
+android {
+    namespace = "io.opentelemetry.android.demo"
+    compileSdk = 34
+
+    defaultConfig {
+        applicationId = "io.opentelemetry.android.demo"
+        minSdk = 21
+        targetSdk = 34
```
LikeTheSalad, line 18: These are set in the conventions so we don't need to do so here (unless we want to use different values only here).

demo-app/build.gradle.kts
```diff
@@ -0,0 +1,88 @@
...
+plugins {
+    id("otel.android-app-conventions")
+}
+
+val localProperties = Properties()
+localProperties.load(FileInputStream(rootProject.file("local.properties")))
+
+android {
+    namespace = "io.opentelemetry.android.demo"
+    compileSdk = 34
```
LikeTheSalad, line 13: This is added as part of the conventions.

demo-app/build.gradle.kts
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+
+    buildTypes {
+        all {
+            val accessToken = localProperties["rum.access.token"] as String?
+            resValue("string", "rum_access_token", accessToken ?: "fakebroken")
+        }
+        release {
+            isMinifyEnabled = false
```
LikeTheSalad, line 34: I think it would make sense to have it enabled for the release build as it could allow us to test some real-world scenarios.

demo-app/build.gradle.kts
```diff
@@ -0,0 +1,88 @@
...
+            proguardFiles(
+                getDefaultProguardFile("proguard-android-optimize.txt"),
+                "proguard-rules.pro",
+            )
+        }
+    }
+    compileOptions {
```
LikeTheSalad, line 41: All this block is already provided in the conventions.

demo-app/build.gradle.kts
```diff
@@ -0,0 +1,88 @@
...
+        sourceCompatibility = JavaVersion.VERSION_1_8
+        targetCompatibility = JavaVersion.VERSION_1_8
+        isCoreLibraryDesugaringEnabled = true
+    }
+    kotlinOptions {
+        jvmTarget = "1.8"
```
LikeTheSalad, line 47: This is already provided in the conventions.

demo-app/build.gradle.kts
```diff
@@ -0,0 +1,88 @@
...
+    buildFeatures {
+        compose = true
+        viewBinding = true
+    }
+    composeOptions {
+        kotlinCompilerExtensionVersion = "1.5.13"
+    }
+    packaging {
+        resources {
+            excludes += "/META-INF/{AL2.0,LGPL2.1}"
```
LikeTheSalad, line 58: Do we need this line?

gradle/libs.versions.toml
```diff
@@ -59,10 +60,29 @@ android-plugin = "com.android.tools.build:gradle:8.4.0"
...
+androidx-navigation-ui-ktx = { group = "androidx.navigation", name = "navigation-ui-ktx", version = "2.7.7" }
+
 [bundles]
 mocking = ["mockito-core", "mockito-junit-jupiter", "mockk"]
 junit = ["junit-jupiter-api", "junit-jupiter-engine", "junit-vintage-engine"]
 
 [plugins]
 publishPlugin = { id = "io.github.gradle-nexus.publish-plugin", version = "2.0.0" }
 spotless = { id = "com.diffplug.spotless", version.ref = "spotless" }
+jetbrainsKotlinAndroid = { id = "org.jetbrains.kotlin.android", version.ref = "kotlin" }
```
LikeTheSalad, line 88: Nit: I think just naming it "kotlinAndroid" seems enough.

demo-app/build.gradle.kts
breedx-splk: Thanks, yeah, total copypasta. I'll clean that up.

demo-app/build.gradle.kts
breedx-splk: Not sure, probably not, it was auto-generated I'm sure. I'll try removing it.

.github/workflows/pr-check.yaml
breedx-splk: The [build was definitely failing](https://github.com/open-telemetry/opentelemetry-android/actions/runs/8942277094/job/24564512546) without this, which is why I added it. Your point below, however, about `local.properties` being automatic might be relevant here....but I'll see if I can remove the explicit handling. 

The local properties are primarily used to get a `rum.access.token` into the resource, so that it can be configured at runtime in user code. This is a hold-over from the splunk sample app -- but I think it's a nice thing to have. We might be able to just remove it and assume users have their own ways of getting an ingest token into the app. I dunno.

demo-app/build.gradle.kts
breedx-splk: I think we need this in order to reference those properties in the build script itself (like this does, by putting an access token into the resource values). If we just want to remove that for now we can do too.

LikeTheSalad: I see, what I meant is that I think this should work out of the box: `rootProject.property("rum.access.token")` (or maybe `rootProject.extra["rum.access.token"]`, not too sure anymore after the change from groovy to kts 😅) because Android projects _should_ be aware of properties inside a `local.properties` file. However, I haven't checked that functionality in a while so maybe it's not the same anymore... Or maybe I'm mixing things up, anyway I think we can leave it like this too, it's not a blocker. 👍

LikeTheSalad (approved):
Thank you!

PR #352, opened 2024-05-08
Break conventions for the demo app

breedx-splk:
This allows the new version of `core-ktx` without breaking the build. This does this by breaking the conventions for the demo-app -- basically letting it do its own thing and duplicating some things from the conventions.


breedx-splk:
(the image was only updated to reduce its size after seeing the app crash once)

gradle/libs.versions.toml
```diff
@@ -9,7 +9,7 @@ junit = "5.10.2"
 byteBuddy = "1.14.14"
 okhttp = "4.12.0"
 spotless = "6.25.0"
-kotlin = "1.9.24"
+kotlin = "1.9.23"
```
LikeTheSalad, line 12: I'm fine with breaking the conventions within the demo app, however, I really don't like the idea of having to downgrade Kotlin (or any library for that matter), for the whole project, just to comply with a demo app's needs. It doesn't sound right, and it leaves me with questions such as, how can we automate bumping up Kotlin's version only after Compose has given its thumbs up to it? Or, what do we do if Kotlin messes up something in a release and then publishes a patch version that's critical to upgrade to but we can't because Compose hasn't tested it yet? And then after imagining all the work that it would take in the short and long run to comply with these special needs from a library that's not even part of the artifacts that this project publishes, I can't help but ask myself why is this worth it?

If the option of having independent Gradle projects that I mentioned [here](https://github.com/open-telemetry/opentelemetry-android/pull/342#issuecomment-2100040009) is not ideal either, then I'd suggest just disabling compose's check for the "right" kotlin version, which seems to be possible by using the `suppressKotlinVersionCompatibilityCheck` config.

breedx-splk: > I really don't like the idea of having to downgrade Kotlin (or any library for that matter), for the whole project, just to comply with a demo app's needs. 

Oh, I totally agree...and whoops, I did not mean to leave that downgraded in the root! Let me see if I can get it working with the latest kotlin in the root `.toml` and overridden for the demo app.

I would be happy to manually bump kotlin for the demo app on-demand (eg. have renovate ignore it), if possible. Let me try.

breedx-splk: Well, unsurprisingly, you can only have one version of `id("org.jetbrains.kotlin.jvm")` in the classpath at runtime.

```
Error resolving plugin [id: 'org.jetbrains.kotlin.jvm', version: '1.9.23']
> The request for this plugin could not be satisfied because the plugin is already on the classpath with an unknown version, so compatibility cannot be checked.
```

So yeah, I think we're back to making the demo app a separate project and including it somehow as you described.

breedx-splk:
Bah, closing for now.

PR #354, opened 2024-05-09
Service manager refactor

LikeTheSalad:
Based on what was discussed in this week's SIG


android-agent/src/main/java/io/opentelemetry/android/internal/services/ServiceManager.kt
```diff
@@ -0,0 +1,91 @@
...
+
+import android.content.Context
+import androidx.annotation.VisibleForTesting
+import com.squareup.wire.internal.toUnmodifiableMap
+import io.opentelemetry.android.internal.services.periodicwork.PeriodicWorkService
+
+/**
+ * This class is internal and not for public use. Its APIs are unstable and can change at any time.
+ */
+class ServiceManager
```
bidetofevil, line 16: Another alternative to organize this is to expose a public interface for `ServiceManager` that contains all the methods folks will call (e.g. `getPreferencesService()`), make the implementation `internal`, and then just create an instance of the object and mark that as `@JvmStatic` so that the singleton is accessible.

That way, you don't need to change the visible of the constructor or mark as being visible for testing, or have methods that allows you to replace the instance in the companion object just for testing - you can just create a new instance in tests instead of relying on the singleton used at runtime. To do that, you can pass in an instance of this in via the constructor of objects that depend on it instead of referencing the singleton directly. You can even create a `FakeServiceManager` for testing objects that require this, but the tests don't actually need a real instance.

This way of organizing this will reduce the amount of test-only scaffolding you need, and generally make your tests easier to write.

LikeTheSalad: Sounds good! I'll take a look into this approach 👍 thanks.

LikeTheSalad: I think it should now match what you've described @bidetofevil

android-agent/src/main/java/io/opentelemetry/android/internal/services/PreferencesService.java
```diff
@@ -15,7 +15,7 @@
  * <p>This class is internal and not for public use. Its APIs are unstable and can change at any
  * time.
  */
-public class PreferencesService implements Service {
+public class PreferencesService {
```
breedx-splk, line 18: Could consider renaming to just `Preferences`.

android-agent/src/test/java/io/opentelemetry/android/internal/services/ServiceManagerImplTest.kt
```diff
@@ -0,0 +1,52 @@
...
+
+        assertThat(serviceManager.getPeriodicWorkService()).isInstanceOf(PeriodicWorkService::class.java)
+        assertThat(serviceManager.getCacheStorageService()).isInstanceOf(CacheStorageService::class.java)
+        assertThat(serviceManager.getPreferencesService()).isInstanceOf(PreferencesService::class.java)
+    }
+
+    @Test
+    fun delegatingStartCall() {
+        val firstService = Mockito.mock<FirstService>()
+        val secondService = Mockito.mock<SecondService>()
```
breedx-splk, line 36: Probably worth a comment saying something like "SecondService is not Startable".

breedx-splk (approved):
This is a nice improvement! There is meaningfully less coupling to the static `ServiceManager` instance now, and greatly improved injection of dependencies. Thanks.

android-agent/src/main/java/io/opentelemetry/android/internal/services/PreferencesService.java
LikeTheSalad: Thanks, I've just made the changes.

android-agent/src/main/java/io/opentelemetry/android/internal/services/CacheStorageService.java
```diff
@@ -22,7 +22,7 @@
  * <p>This class is internal and not for public use. Its APIs are unstable and can change at any
  * time.
  */
-public class CacheStorageService implements Service {
+public class CacheStorageService {
```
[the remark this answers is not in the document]
LikeTheSalad: Thanks, I've just made the changes.

android-agent/src/main/java/io/opentelemetry/android/internal/services/ServiceManager.kt
```diff
@@ -0,0 +1,45 @@
...
+            if (instance != null) {
+                return
+            }
+            instance =
+                ServiceManagerImpl(
+                    listOf(
+                        PreferencesService.create(appContext),
+                        CacheStorageService(appContext),
+                        PeriodicWorkService(),
+                    ),
```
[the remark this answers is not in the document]
LikeTheSalad: I'm not sure I'm following, do you mean if they need to use something from `OtelRumConfig`?

android-agent/src/test/java/io/opentelemetry/android/internal/services/ServiceManagerImplTest.kt
LikeTheSalad: Thanks, I've added the comment.

bidetofevil (changes requested):
LGTM

android-agent/src/main/java/io/opentelemetry/android/internal/services/ServiceManager.kt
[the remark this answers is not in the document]
breedx-splk: Yes, exactly.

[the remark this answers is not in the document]
LikeTheSalad: Ah, gotcha. I don't see a use case that would require it now, or in the near future, though I'm open to discussing it later if needed 👍

PR #355, opened 2024-05-10
Making demo app independent

LikeTheSalad:
Option to avoid [these](https://github.com/open-telemetry/opentelemetry-android/pull/342#issuecomment-2100040009) kinds of issues.

This option extracts the demo app into its own project and adds a dependency onto the agent project by using the composite builds config. This option seems to be more stable for the future, considering that the other option proposed below uses a configuration param that's not properly documented, and also adds a risk of Compose not being able to properly compile due to Kotlin version issues.

Some things to consider with this option are:
- There's a duplication of all the gradlew-related files for the independent project.
- Android Studio has to explicitly open the project `demo-app` in order to be able to compile the app with it.

Another option is: https://github.com/open-telemetry/opentelemetry-android/pull/356


demo-app/build.gradle.kts
```diff
@@ -42,6 +45,16 @@ android {
     composeOptions {
         kotlinCompilerExtensionVersion = "1.5.13"
     }
+    val javaVersion = JavaVersion.VERSION_11
```
bidetofevil, line 48: Question: do we require (desugared) Java 11 features to use the android extension? For a demo app, it'd be nice to just use Kotlin and not having to worry about Java

demo-app/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,8 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionSha256Sum=544c35d6bd849ae8a5ed0bcea39ba677dc40f49df7d1835561582da2009b961d
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.7-bin.zip
+networkTimeout=10000
```
bidetofevil, line 5: Is this deliberately low so it can show errors more easily?

bidetofevil (approved):
LGTM. Just have question about Java usage but non-blocking

demo-app/build.gradle.kts
```diff
@@ -2,17 +2,20 @@ import java.io.FileInputStream
...
 }
 
 val localProperties = Properties()
 localProperties.load(FileInputStream(rootProject.file("local.properties")))
 
 android {
     namespace = "io.opentelemetry.android.demo"
+    compileSdk = 34
 
     defaultConfig {
```
bidetofevil, line 16: In notice the default targetSdk for the extension project is 23 - I'll ask more about that in Slack, but for the app, I think we should target 34 as well. Not that we'll ever submit this to the Play Store, but it requires the targetSdk to be at least your compileSdk if your target is less than 34, so the current combination won't be legal under Play Store rules.

demo-app/build.gradle.kts
breedx-splk: I think you're right -- at least right now it looks like the demo app is all kotlin, so I would think we could remove this and see if something else in the build requires it?

demo-app/build.gradle.kts
breedx-splk: I also commented on your other PR @bidetofevil but now that I see this, it makes more sense. I think with demo-app being its own stand-alone project (this PR) we can target 34. Seems reasonable.

demo-app/build.gradle.kts
```diff
@@ -2,17 +2,20 @@ import java.io.FileInputStream
...
+    id("com.android.application") version "8.4.0"
+    id("org.jetbrains.kotlin.android") version "1.9.23"
 }
 
 val localProperties = Properties()
 localProperties.load(FileInputStream(rootProject.file("local.properties")))
 
 android {
     namespace = "io.opentelemetry.android.demo"
+    compileSdk = 34
```
breedx-splk, line 14: Based on the discussion below: 
```suggestion
    compileSdk = 34
    targetSdk = 34
```

demo-app/gradle/wrapper/gradle-wrapper.properties
breedx-splk: I'm sure this is autogenerated boilerplate from running gradle wrapper.

breedx-splk (approved):
Thanks @LikeTheSalad! This is straightforward and moves us out of the current kotlin version mismatch hell and soup of build breakages. 

When I took an attempt at this, I wanted to try and avoid having to help the demo-app as another IntelliJ window, and that didn't work out at all. I think this is fine tho, if not just a little awkward for otel-android devs.

demo-app/gradle/wrapper/gradle-wrapper.properties
bidetofevil: Ah. It's a demo app, so this is more good than bad. Hope no one takes this and uses it in production lol

demo-app/build.gradle.kts
```diff
@@ -42,6 +46,16 @@ android {
     composeOptions {
         kotlinCompilerExtensionVersion = "1.5.13"
     }
+    val javaVersion = JavaVersion.VERSION_11
```
breedx-splk, line 49: We're going to try without this for now.
```suggestion
```

PR #356, opened 2024-05-10
Disabling compose kotlin checks

LikeTheSalad:
Option to avoid [these](https://github.com/open-telemetry/opentelemetry-android/pull/342#issuecomment-2100040009) kinds of issues.

This option disables compose's kotlin checks, although the config to do so is not well documented (it's mostly properly "officially" mentioned [here](https://github.com/JetBrains/compose-multiplatform/issues/2459#issuecomment-1312185906)) - Although it seems like it might change in the future. The good thing about this approach though is that it's the one that requires the least amount of changes to work.

Another option is: https://github.com/open-telemetry/opentelemetry-android/pull/355

### DIFF
--- a/buildSrc/src/main/kotlin/otel.errorprone-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.errorprone-conventions.gradle.kts
@@ -43,6 +43,7 @@ tasks {
                     isEnabled.set(true)
                     isCompilingTestOnlyCode.set(false)
                 }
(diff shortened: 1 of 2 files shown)


bidetofevil:
#355 seems like the way to go for me. I'll add my reasoning on that PR.

breedx-splk:
I think we're aligned on this...so I'm closing. Reopen if we want to continue discussing.

Thanks @LikeTheSalad for putting together both options to compare/contrast. ❤️

PR #359, opened 2024-05-13
Remove targetSdk declaration

bidetofevil:
Setting the `targetSdkVersion` is the domain of apps, not libraries. Given the project's `minSdkVersion` is 21, apps that target 21 should be able use it, so adding this is probably too restrictive as the manifest merging process will use the highest one, which would be 23, if we left this in, which would override what the app's setting stipuates.

### DIFF
--- a/buildSrc/src/main/kotlin/otel.android-app-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.android-app-conventions.gradle.kts
@@ -14,7 +14,6 @@ android {
 
     defaultConfig {
         minSdk = (property("android.minSdk") as String).toInt()
-        targetSdk = (property("android.targetSdk") as String).toInt()
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
     }
 
--- a/gradle.properties
+++ b/gradle.properties
@@ -17,8 +17,6 @@ org.gradle.jvmargs=-Xmx2048m -Dfile.encoding=UTF-8
 android.useAndroidX=true
 
 android.minSdk=21
-android.targetSdk=23
 android.compileSdk=34
-
 version=0.6.0
 group=io.opentelemetry.android


breedx-splk:
I think I understand better now after reading [the comment](https://github.com/open-telemetry/opentelemetry-android/pull/355/files#r1597680140) in the other (separate demo-app) PR. We should specify it explicitly in the demo-app build and not here.

LikeTheSalad (approved):
If this is possible without causing any hidden/strange behaviors then I'm all for it, IIRC the stuff that's added to the conventions was the bare minimum requirements for libs to compile, so for some reason I was under the impression that the compiler would complain if this value was not set. Although, if that's not the case and it still works without it (and it doesn't add any value by default that's too high either) then by all means I'd like to remove it. To me, the least amount of versions to keep track of, the better! 😁

LikeTheSalad:
> If this is possible without causing any hidden/strange behaviors then I'm all for it, IIRC the stuff that's added to the conventions was the bare minimum requirements for libs to compile, so for some reason I was under the impression that the compiler would complain if this value was not set. Although, if that's not the case and it still works without it (and it doesn't add any value by default that's too high either) then by all means I'd like to remove it. To me, the least amount of versions to keep track of, the better! 😁

Btw I just checked and it seems like we didn't even use this for libs so most likely it wouldn't cause any issues to remove it. Maybe it was there just as a convenience for tests that involve creating a full app, such as the [okhttp one](https://github.com/open-telemetry/opentelemetry-android/blob/main/instrumentation/okhttp/okhttp-3.0/testing/build.gradle.kts#L2) for example. So based on that, if we removed this from the app conventions, we'd have to manually add this value for all of those kinds of tests, @bidetofevil?

bidetofevil:
Ugh, I messed up my branch. If I can't fix the PR I'll just blow it up and start it again. It's a two line change

bidetofevil:
Man, I suck at git. I set up my repo once and forget all my settings. I'm using Graphite to simplify my workflow but it's not great at coming in the middle of PRs. Anyway, rebased. Feel free to merge when ready.

bidetofevil:
> Btw I just checked and it seems like we didn't even use this for libs so most likely it wouldn't cause any issues to remove it. Maybe it was there just as a convenience for tests that involve creating a full app, such as the [okhttp one](https://github.com/open-telemetry/opentelemetry-android/blob/main/instrumentation/okhttp/okhttp-3.0/testing/build.gradle.kts#L2) for example. So based on that, if we removed this from the app conventions, we'd have to manually add this value for all of those kinds of tests, @bidetofevil?

By default, the `minSdkVersion` set on the library will be used when a test requires an app to be built - or whatever the highest min or target set on the other libraries included in the test will be. Of course if the test is testing specific logic in a particular Android version, you can always override that on a test by test level, or set it on a global basis.

If we want a higher targetSdk to be used for tests (which we probably should?), I can find the global setting and do it there.

Issue #364, opened 2024-05-14
Need to include demo-app in CI and main builds


breedx-splk:
Now that the demo-app is a separate gradle project, we need to account for this in our PR and other build processes. Without this, it's possible that we could make API changes or other breaking changes that impact the demo-app without knowing about it.

The CI automation should help us catch this stuff early.

PR #365, opened 2024-05-14
Add demo app to CI builds

breedx-splk:
Part of the reason for having the demo-app is to show features, but those features are meaningless if the usage of those doesn't work. We need to build the demo app with every PR and merge to main, in order to ensure compatibility.

I guess I should also mention that this provides the `demo-app` with its own version catalog. I think this is going to be required in order for us to version it separately (via renovate) without impacting the main build. It's some additional duplication, unfortunately, but I think it's needed given the multi-project approach we've chosen.


breedx-splk:
Resolves #364

demo-app/proguard-rules.pro
```diff
@@ -0,0 +1,5 @@
+-dontwarn com.fasterxml.jackson.core.JsonFactory
+-dontwarn com.fasterxml.jackson.core.JsonGenerator
+-dontwarn com.google.auto.value.AutoValue$Builder
+-dontwarn com.google.auto.value.AutoValue
+-dontwarn com.google.auto.value.extension.memoized.Memoized
```
breedx-splk, line 5: Minification was failing without this.

.github/workflows/build.yaml
```diff
@@ -18,8 +18,15 @@ jobs:
         with:
           distribution: temurin
           java-version: 17
+      - name: touch local props
+        run: |
+          touch local.properties
```
LikeTheSalad, line 23: Do we still need to create this file for the root project? I'm asking because it seems like it's only needed for `demo-app`.

demo-app/build.gradle.kts
```diff
@@ -12,7 +12,6 @@ localProperties.load(FileInputStream(rootProject.file("local.properties")))
 android {
     namespace = "io.opentelemetry.android.demo"
     compileSdk = 34
-    targetSdk = 34
```
LikeTheSalad, line 15: If I understood correctly, what @bidetofevil mentioned in [this other PR](https://github.com/open-telemetry/opentelemetry-android/pull/359) is that libs shouldn't define this value, but apps should. So maybe we should keep it here?

demo-app/settings.gradle.kts
```diff
@@ -12,8 +14,8 @@ dependencyResolutionManagement {
         google()
     }
     versionCatalogs {
-        create("libs") {
```
LikeTheSalad, line 15: This block was only needed to reuse the same toml file from the root project. Since that's no longer the case, I think we can just remove the whole `versionCatalogs` block from this file, as the new catalog file `demo-app/gradle/libs.versions.toml` should be used automatically within the `demo-app` gradle files as `libs`.

demo-app/build.gradle.kts
```diff
@@ -58,28 +58,30 @@ android {
...
 dependencies {
-    implementation(libs.androidx.appcompat)
-    implementation(libs.androidx.constraintlayout)
-    implementation(libs.material)
-    implementation(libs.androidx.lifecycle.livedata.ktx)
-    implementation(libs.androidx.lifecycle.viewmodel.ktx)
-    implementation(libs.androidx.navigation.fragment.ktx)
-    implementation(libs.androidx.navigation.ui.ktx)
-    coreLibraryDesugaring(libs.desugarJdkLibs)
+    implementation(demoLibs.androidx.appcompat)
```
LikeTheSalad, line 61: I think there's no need to change the prefix to `demoLibs` after doing what I mentioned in this [other comment](https://github.com/open-telemetry/opentelemetry-android/pull/365/files#r1602909961).

.github/workflows/build.yaml
breedx-splk: Ok let's try it without!

demo-app/build.gradle.kts
breedx-splk: Ok I added it back to the `defaultConfig` block.

demo-app/settings.gradle.kts
breedx-splk: Ok, I'll remove this and then change `buildLibs` back to `libs` primarily because it's less code....but I wonder if we'll get ourselves confused not knowing which catalog we're looking at sometimes. I know I struggled with that once or twice while getting this going.

LikeTheSalad: Ah, gotcha. I didn't think it could be confusing. That's fair enough, I'm in favor of changing it if that makes it easier to read 👍

LikeTheSalad (approved):
Cheers!

bidetofevil:
A little late to this but yeah this seems like the right move. The demo app can also pull in new versions of Android libraries that the agent doesn't need, so I think we'll get here eventually anyway

PR #374, opened 2024-05-20
Enable desugaring in debug for demo app

bidetofevil:
Make building the demo app in debug work by enabling the Gradle setting that makes desugaring work when minSdkVersion is set to < 24


LikeTheSalad (approved):
Cheers!

PR #385, opened 2024-05-22
Don't add artifacts to the GH release page

breedx-splk:
We don't need them, and since we're so modular now it doesn't make as much since to just publish the agent/sdk thing alone.


bidetofevil (approved):
LGTM. This doesn't seem like it's worth the time. Android folks use maven to grab their other dependencies, so why not this?

LikeTheSalad (approved):

PR #387, opened 2024-05-29
Release docs update

LikeTheSalad:
This is to provide a GH action to run the changelog draft script as well as updating the release instructions to mention it.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -11,7 +11,8 @@ We first need to prepare the release. This creates a versioned release branch, T
 - Make sure that the `gradle.properties` version property is set to the value you want to release.
   This must be different than the most recent release number (typically one minor version increase).
 - Merge a pull request to `main` branch that updates the `CHANGELOG.md`.
-  - The heading for the unreleased entries must be `## Unreleased`.
+    - The heading for the unreleased entries must be `## Unreleased`.
+    - Use [this action](https://github.com/open-telemetry/opentelemetry-android/actions/workflows/draft-change-log-entries.yaml) as a starting point for writing the change log. It will print a draft in the console that you can copy to create your PR for updating the `CHANGELOG.md` file.
 - Go to the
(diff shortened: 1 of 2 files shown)


marandaneto (approved):

RELEASING.md
```diff
@@ -11,7 +11,8 @@ We first need to prepare the release. This creates a versioned release branch, T
 - Make sure that the `gradle.properties` version property is set to the value you want to release.
   This must be different than the most recent release number (typically one minor version increase).
 - Merge a pull request to `main` branch that updates the `CHANGELOG.md`.
-  - The heading for the unreleased entries must be `## Unreleased`.
+    - The heading for the unreleased entries must be `## Unreleased`.
+    - Use [this action](https://github.com/open-telemetry/opentelemetry-android/actions/workflows/draft-change-log-entries.yaml) as a starting point for writing the change log entries. It will print a draft in the console that you can copy to create your PR.
```
breedx-splk, line 15: Will this work if this runs before the release branch is prepared? I'm a little skeptical. It basically does a `git log --reverse [...] v$prior_major.$prior_minor.0..HEAD | sed [...]` and I am not sure that the vX.Y.0 will be there yet.

LikeTheSalad: I'm not sure I'm following. Do you mean that if it runs before the release branch is prepared, then it would use the wrong version range to get the git logs? (I didn't write this script) though my understanding is that it will compute the previously released version (to use as the range start) based on the existing one in `gradle.properties`.

breedx-splk: Right. So let's say for the sake of discussion that we're releasing 7.4.0 and the previous version was 7.3.0. The `draft-change-log-entries.sh` will call `get-version.sh` to get `7.4.0` from the `gradle.properties`, and it will compute the previous version to be `7.3.0`, and it will the compare `7.3.0` with HEAD, which is what I think we want.

Just had to talk it out. I think we're good. 👍🏻

breedx-splk (approved):

PR #388, opened 2024-05-29
Upgrading Kotlin to 2.0.0

LikeTheSalad:
Supersedes #380 

Using the new compose compiler plugin to fix compose version compatibilities with Kotlin 2.0.0. Also, the `1.6` min kotlin `languageVersion`/`apiVersion` has been marked as deprecated, so I've bumped it to the next minimum `1.7`.


marandaneto (approved):

breedx-splk (approved):
Oh yeah this rules, thanks!

PR #396, opened 2024-05-30
Instrumentation API - Part 1

LikeTheSalad:
First part - Creating the API


android-agent/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentationRegistryImpl.kt
```diff
@@ -0,0 +1,32 @@
...
+
+package io.opentelemetry.android.instrumentation
+
+import java.util.Collections
+import java.util.ServiceLoader
+
+internal class AndroidInstrumentationRegistryImpl : AndroidInstrumentationRegistry {
+    private val instrumentations: MutableMap<Class<out AndroidInstrumentation>, AndroidInstrumentation> by lazy {
+        ServiceLoader.load(AndroidInstrumentation::class.java).associateBy { it.javaClass }
+            .toMutableMap()
```
breedx-splk, line 14: So this means that when an instance of `AndroidInstrumentationRegistryImpl` is created, the map is not yet created, right? As soon as `get()` or `getAll()` or `register()` is called, the map will be created lazily through the `ServiceLoader` mechanism?

Just wanted to think it through. Seems fine, I think, and I like that the user of the class doesn't have to do any work. Pretty elegant.

android-agent/src/test/java/io/opentelemetry/android/instrumentation/AndroidInstrumentationRegistryImplTest.kt
```diff
@@ -0,0 +1,72 @@
...
+    @BeforeEach
+    fun setUp() {
+        registry = AndroidInstrumentationRegistryImpl()
+    }
+
+    @Test
+    fun `Find and register implementations available in the classpath when querying an instrumentation`() {
+        val instrumentation = registry.get(TestAndroidInstrumentation::class.java)
+
+        instrumentation?.install(mockk(), mockk())
```
breedx-splk, line 28: Should we also assert that `installed` is false before we call install? Just to kinda assert that the registry itself doesn't do any installation.

android-agent/src/test/java/io/opentelemetry/android/instrumentation/AndroidInstrumentationRegistryImplTest.kt
```diff
@@ -0,0 +1,72 @@
...
+        val instrumentation2 = DummyInstrumentation("test2")
+
+        registry.register(instrumentation)
+
+        try {
+            registry.register(instrumentation2)
+            fail("Should not allow to register the same type of instrumentation more than once.")
+        } catch (e: IllegalStateException) {
+            assertThat(e.message).isEqualTo("Instrumentation with type '${DummyInstrumentation::class.java}' already exists.")
+        }
```
breedx-splk, line 62: Prefer to use `assertThatThrownBy()` for these kinds of exception tests.

breedx-splk (approved):
Off to a great start here! Had a few comments, but nothing to slow down this effort. Thanks.

android-agent/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentationRegistryImpl.kt
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    @Suppress("UNCHECKED_CAST")
+    override fun <T : AndroidInstrumentation> get(type: Class<out T>): T? {
+        return instrumentations[type] as? T
+    }
+
+    override fun getAll(): Collection<AndroidInstrumentation> {
+        return Collections.unmodifiableCollection(instrumentations.values)
+    }
```
marandaneto, line 24: Any reason to use the Java `Collection` types instead of the Kotlin ones?

android-agent/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentationRegistryImpl.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    override fun register(instrumentation: AndroidInstrumentation) {
```
marandaneto, line 26:
```suggestion
    @Throws(IllegalStateException::class)
    override fun register(instrumentation: AndroidInstrumentation) {
```

android-agent/src/test/java/io/opentelemetry/android/instrumentation/AndroidInstrumentationRegistryImplTest.kt
marandaneto, line 28: I'd avoid `?.` or assert that it's not null otherwise the test would be green because `instrumentation` is null.
it'd fail in L30 but because of another reason

android-agent/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentationRegistryImpl.kt
```diff
@@ -0,0 +1,32 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.android.instrumentation
+
+import java.util.Collections
+import java.util.ServiceLoader
+
+internal class AndroidInstrumentationRegistryImpl : AndroidInstrumentationRegistry {
+    private val instrumentations: MutableMap<Class<out AndroidInstrumentation>, AndroidInstrumentation> by lazy {
```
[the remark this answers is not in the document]
LikeTheSalad: Good point. My understanding is that "out Base" is the equivalent of "? extends Base", though tbh I haven't taken the time to know the details about why is that and/or if there's a better approach (at least for this particular use case).

android-agent/src/test/java/io/opentelemetry/android/instrumentation/AndroidInstrumentationRegistryImplTest.kt
marandaneto: or `org.junit.Assert.assertThrows`

android-agent/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentationRegistryImpl.kt
LikeTheSalad: > So this means that when an instance of `AndroidInstrumentationRegistryImpl` is created, the map is not yet created, right? As soon as `get()` or `getAll()` or `register()` is called, the map will be created lazily through the `ServiceLoader` mechanism?

That's right. Tbh I was a bit concerned about doing the service load implicitly as it is now, because I'm not sure if there could be a case where some users might not want to use instrumentations from the classpath and instead register manually the ones they'd like to use, and the way it is now wouldn't support that... But then I didn't want to spend too much time on a use case that I'm not sure will ever be needed, so that's why I left it as is, though I'm curious to know about other opinions on it.

android-agent/src/test/java/io/opentelemetry/android/instrumentation/AndroidInstrumentationRegistryImplTest.kt
LikeTheSalad: This is cool! I didn't know about it, I've just updated the test with it.

android-agent/src/test/java/io/opentelemetry/android/instrumentation/AndroidInstrumentationRegistryImplTest.kt
LikeTheSalad: Good point, I've just updated it.

android-agent/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentationRegistryImpl.kt
LikeTheSalad: Do you mean `kotlin.collections.Collection`?

android-agent/src/test/java/io/opentelemetry/android/instrumentation/AndroidInstrumentationRegistryImplTest.kt
LikeTheSalad: Got it, thank you. I think the `isTrue/isFalse` assertions should verify values not being null, however, I agree it's probably better to make sure. I've made the changes. Cheers!

marandaneto: yep the next assertion would fail (correctly) but the reason would be wrong (because it's null and not because it returned the non-expected result).

android-agent/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentationRegistryImpl.kt
marandaneto: yep, the `kotlin.collections` package in general.
The less we depend on `java.*` the easier it'll be to be KMP compatible in the future

LikeTheSalad: Good point. I was looking for a `Collections.unmodifiableCollection` equivalent, though it seems like `toList` should do. I've just made the changes. Cheers!

android-agent/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentationRegistryImpl.kt
breedx-splk: I don't think it's worth changing. Let's go with this and adjust as needed.

Issue #405, opened 2024-06-03
Add tests to verify the Agent works properly in Android


bidetofevil:
As discussed in a previous SIG meeting, entering an issue to create automation to verify that agent features are working when running in Android.

Feel free to assign this to me if y'all still want this - I think some Robolectric tests that run against the min and max API versions supported will be a nice, non-flakey way to verify that agent features work end-to-end on Android (as least until the exporter).

fractalwrench:
@bidetofevil what automation did you have in mind? Was this simply an instrumented test that runs on an emulator & confirms OTel logs/spans can be sent?

PR #406, opened 2024-06-03
Add basic instrumentation test to start agent on Android

bidetofevil:
This is my proposal for issue #405. It doesn't do much other than start the agent right now, but further verification can be done by adding a fake Exporter to verify that telemetry is logged as expected.

Let me know what y'all think!

### DIFF
--- a/android-agent/src/androidTest/kotlin/io/opentelemetry/android/BasicInstrumentationTest.kt
+++ b/android-agent/src/androidTest/kotlin/io/opentelemetry/android/BasicInstrumentationTest.kt
@@ -1,0 +1,30 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.android
+
+import android.app.Application
+import android.os.Build.VERSION_CODES.LOLLIPOP
+import android.os.Build.VERSION_CODES.UPSIDE_DOWN_CAKE
+import androidx.test.core.app.ApplicationProvider.getApplicationContext
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import io.opentelemetry.android.config.OtelRumConfig
+import org.junit.Assert.assertNotNull
+import org.junit.runner.RunWith
+import org.robolectric.annotation.Config
+
+@RunWith(AndroidJUnit4::class)
+internal class BasicInstrumentationTest {
+    @Config(sdk = [LOLLIPOP, UPSIDE_DOWN_CAKE])
+    fun startOpenTelemetryRumInAndroid() {
+        val application = getApplicationContext<Application>()
+        val otelRum =
+            OpenTelemetryRum.builder(
+                application,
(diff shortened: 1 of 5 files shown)


android-agent/src/androidTest/kotlin/io/opentelemetry/android/BasicInstrumentationTest.kt
```diff
@@ -0,0 +1,30 @@
...
+import androidx.test.core.app.ApplicationProvider.getApplicationContext
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import io.opentelemetry.android.config.OtelRumConfig
+import org.junit.Assert.assertNotNull
+import org.junit.runner.RunWith
+import org.robolectric.annotation.Config
+
+@RunWith(AndroidJUnit4::class)
+internal class BasicInstrumentationTest {
+    @Config(sdk = [LOLLIPOP, UPSIDE_DOWN_CAKE])
```
LikeTheSalad, line 20: (Maybe not possible) though it'd be nice if the latest version could be updated automatically as soon as a new one is available.

breedx-splk (commented):
This seems like a great start, thanks!

1. How do you run this test from gradle?
2. Where/when do we think this kind of test should run? On every PR? On merge to main? Nightly? Something else? 

Really curious about how we get this automated!

android-agent/src/androidTest/kotlin/io/opentelemetry/android/BasicInstrumentationTest.kt
bidetofevil: There is the "current development" value also but I've never used it and aren't sure of what the behaviour is. I wouldn't mind keeping it at whatever our compileSdk version is at just to reduce unexpected flakiness. That said, but I'm opposed to trying it in case y'all prefer that.

android-agent/src/androidTest/kotlin/io/opentelemetry/android/BasicInstrumentationTest.kt
```diff
@@ -0,0 +1,30 @@
...
+    fun startOpenTelemetryRumInAndroid() {
```
breedx-splk, line 21: We think `@Test` is missing.
```suggestion
    @Test
    fun startOpenTelemetryRumInAndroid() {
```

bidetofevil:
Ugh, fixed the PR. It's in the right directory now, a separate one but added to the unit test sourceset and will run as part of those

android-agent/src/androidTest/kotlin/io/opentelemetry/android/BasicInstrumentationTest.kt
marandaneto: Could also use the `minSdk` and `maxSdk` params but then it'd run for all versions I think, if it's fast it's okay otherwise probably not.

marandaneto (approved):

breedx-splk (approved):

PR #407, opened 2024-06-03
Making changelog script runnable

LikeTheSalad:
[Seems like](https://github.com/open-telemetry/opentelemetry-android/actions/runs/9348015347/job/25726243291) the script wasn't runnable.


breedx-splk (approved):
I mean, we should probably just modify the file itself rather than doing this each time, but it shouldn't hurt too much in the interim.

LikeTheSalad:
> I mean, we should probably just modify the file itself rather than doing this each time, but it shouldn't hurt too much in the interim.

Yeah that sounds better, though I'm not sure how to add the "execute permissions" in a way that is tracked by git :(

PR #408, opened 2024-06-03
Instrumentation API part 2

LikeTheSalad:
These changes involve the telemetry sent around network status changes. The amount of files is large mostly because of several files, prod code and unit tests, moved over to `android-agent` (to become part of a service).

Since the relationship around the instrumentations and the `android-agent` module is going to be inverted and a couple of tools needed across multiple modules will be turned into services, I realized that it's going to be very difficult to keep the build passing without having all the instrumentation modules already migrated over to the new API. Also, as soon as the first instrumentation is migrated (like this one), the `android-agent` will no longer provide that functionality out of the box, as now the dependency relationship is inverted, which breaks the current behavior, so in order to avoid causing troubles while all the instrumentation modules are migrated, I've created a feature branch to merge all the PRs into, one instrumentation at a time, so that they are easier to review and won't break the current functionality.


instrumentation/network/src/main/java/io/opentelemetry/android/instrumentation/network/NetworkChangeMonitor.java
```diff
@@ -20,33 +20,26 @@
...
-        this.additionalExtractors = new ArrayList<>(builder.additionalExtractors);
-    }
-
     /**
-     * Installs the network change monitoring instrumentation on the given {@link
-     * InstrumentedApplication}.
+     * Installs the network change monitoring instrumentation on the given {@link OpenTelemetry}.
      */
-    public void installOn(InstrumentedApplication instrumentedApplication) {
+    public void installOn(OpenTelemetry openTelemetry) {
```
breedx-splk, line 36: This no longer makes sense as it is written. It used to be "Hey network monitor, install yourself into this application", but it now reads "Hey install yourself into this OpenTelemetry" which I think doesn't make sense. It's also odd that the otel instance isn't just passed to the constructor, since the caller has it at creation time.

We can defer this naming choice until part 3 I suppose, but maybe we should remove the otel param, pass it in the constructor instead, and just call this method `start()`.

android-agent/src/main/AndroidManifest.xml
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
 
+    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
```
breedx-splk, line 4: It's weird to me to see this added at the same time the direct dependency on the network instrumentation goes away. Is this kinda front-loading the permission, just in case the user opts into it later? Hmmm.....if we're making the user choose which instrumentation to include, should we also require them to add the permission?

Is there a way to do this in the network instrumentation lib instead? I don't know the answer here.

instrumentation/network/build.gradle.kts
```diff
@@ -21,7 +21,7 @@ android {
 dependencies {
     api(platform(libs.opentelemetry.platform))
     api(libs.opentelemetry.api)
-    api(project(":common"))
+    api(project(":android-agent"))
```
breedx-splk, line 24: This contradicts my understanding of the original design:
image

It's odd to me that the specific instrumentation now depends on the topmost agent.  Is this something that goes away in part 3 or something?

android-agent/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
```diff
@@ -402,27 +387,17 @@ private void applyConfiguration() {
         // Network specific attributes
         if (config.shouldIncludeNetworkAttributes()) {
             // Add span processor that appends network attributes.
-            CurrentNetworkProvider currentNetworkProvider = getOrCreateCurrentNetworkProvider();
             addTracerProviderCustomizer(
                     (tracerProviderBuilder, app) -> {
                         SpanProcessor networkAttributesSpanAppender =
-                                NetworkAttributesSpanAppender.create(currentNetworkProvider);
+                                NetworkAttributesSpanAppender.create(
+                                        ServiceManager.get().getCurrentNetworkProvider());
```
breedx-splk, line 394: Does this also eventually move to the network instrumentation module? It currently strange to me because the main agent is aware that some network attributes exist, and it is including them based on config, but there's no 
clear indication about where these network attributes originate (hint: it's in a separate instrumentation module).

android-agent/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
```diff
@@ -121,20 +120,6 @@ public OtelRumConfig setDiskBufferingConfiguration(
...
-    public OtelRumConfig disableNetworkChangeMonitoring() {
-        networkChangeMonitoringEnabled = false;
-        return this;
-    }
-
-    /** Returns true if network change monitoring is enabled (default = true). */
-    public boolean isNetworkChangeMonitoringEnabled() {
-        return networkChangeMonitoringEnabled;
-    }
-
```
breedx-splk, line 137: So now the configuration choice is no longer programmatic and cannot be dynamic. It is determined statically at build-time depending on whether or not the user has included the network instrumentation dependency?

I think that's fine, I just wanted to think that through, out loud. If/when the need for additional configuration arises, I suppose we can handle that within each instrumentation module that requires it.

android-agent/src/main/java/io/opentelemetry/android/internal/services/network/CarrierFinder.java
```diff
@@ -3,22 +3,23 @@
...
+package io.opentelemetry.android.internal.services.network;
 
 import android.os.Build;
 import android.telephony.TelephonyManager;
 import androidx.annotation.RequiresApi;
+import io.opentelemetry.android.internal.services.network.data.Carrier;
 
 @RequiresApi(api = Build.VERSION_CODES.P)
-class CarrierFinder {
+public class CarrierFinder {
```
breedx-splk, line 14: These classes that are still java and have moved to `internal` packages should get the boilerplate disclaimer about not being for public use and subject to change.

android-agent/src/main/java/io/opentelemetry/android/internal/services/network/CurrentNetworkProvider.java
```diff
@@ -26,44 +30,37 @@
...
+                (ConnectivityManager) application.getSystemService(Context.CONNECTIVITY_SERVICE));
     }
 
     private final NetworkDetector networkDetector;
+    private final ConnectivityManager connectivityManager;
 
     private volatile CurrentNetwork currentNetwork = UNKNOWN_NETWORK;
-    private final List<NetworkChangeListener> listeners = new CopyOnWriteArrayList<>();
+    private final List<NetworkChangeListener> listeners =
+            new CopyOnWriteArrayList<>(); // visible for tests
```
breedx-splk, line 51: not actually visible tho....?

breedx-splk: if you're talking about the method that mutates the list, then maybe change to `// exposed thru method below for tests` or something

android-agent/src/main/java/io/opentelemetry/android/internal/services/network/CurrentNetworkProvider.java
```diff
@@ -107,11 +102,11 @@ private static NetworkRequest createNetworkMonitoringRequest() {
                 .build();
     }
 
-    CurrentNetwork getCurrentNetwork() {
+    public CurrentNetwork getCurrentNetwork() {
         return currentNetwork;
     }
 
-    void addNetworkChangeListener(NetworkChangeListener listener) {
+    public void addNetworkChangeListener(NetworkChangeListener listener) {
```
breedx-splk, line 109: Is this mutability exposed to facilitate testing? Should we make it package then?

android-agent/src/main/java/io/opentelemetry/android/internal/services/network/data/Carrier.java
```diff
@@ -12,15 +12,15 @@
 import java.util.Objects;
 
 @RequiresApi(api = Build.VERSION_CODES.P)
-final class Carrier {
+public final class Carrier {
```
breedx-splk, line 15: Add disclaimer about `internal` package.

breedx-splk: Might be a nice choice for a kotlin data class? Certainly can wait tho.

breedx-splk (commented):
Looks good to me. I do want to figure out the dependency direction issue tho.

instrumentation/network/build.gradle.kts
LikeTheSalad: The general idea is to avoid the core parts of the agent having a dependency on the instrumentations. I agree it's confusing as it's shown here. My plan after all the instrumentations are created is to rename the existing `android-agent` module to something else, not sure yet, probably `core`, and then create a new `android-agent` module which will include the core along with all the instrumentations that we'd like the agent to have "enabled" out of the box.

android-agent/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
LikeTheSalad: This feature seems to be only for adding global attrs with a SpanProcessor, so I think it should live in the core module to be available before initialization. I've added more details on it [in this other comment](https://github.com/open-telemetry/opentelemetry-android/pull/408#discussion_r1625825393).

android-agent/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
LikeTheSalad: > So now the configuration choice is no longer programmatic and cannot be dynamic. It is determined statically at build-time depending on whether or not the user has included the network instrumentation dependency?

These changes mean that the core config is not aware of the network changes instrumentation. As it is now, it can only be enabled at build-time by adding the instrumentation dependency and it cannot be disabled if it's in the classpath at runtime. However, we could expand `AndroidInstrumentation` to allow for disabling an instrumentation, even if it was already enabled (by using the [suppress context key](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/9739)) - Or we could also avoid loading instrumentations altogether from the classpath, and instead require each instrumentation to be manually "enabled" by registering it through the `AndroidInstrumentaionRegistry`. So yeah it's as you mentioned, I think we can handle different use cases if the need arises.

android-agent/src/main/java/io/opentelemetry/android/internal/services/network/CurrentNetworkProvider.java
LikeTheSalad: Ah, my bad, I missed that comment when moving code around.

android-agent/src/main/java/io/opentelemetry/android/internal/services/network/CarrierFinder.java
LikeTheSalad: Thanks, I've made the changes.

android-agent/src/main/java/io/opentelemetry/android/internal/services/network/CurrentNetworkProvider.java
LikeTheSalad: I've just fixed it, turns out the comment belonged to a constructor right below this line.

android-agent/src/main/java/io/opentelemetry/android/internal/services/network/data/Carrier.java
LikeTheSalad: Thanks, I've made the changes.

android-agent/src/main/AndroidManifest.xml
breedx-splk: Ok, I think we're on the same page anyway about it being somewhat confusing that there are "network related instrumentation responsibilities" still in both the agent and the specific instrumentation. We can circle back on that another time to figure that out....but it's good that we've recognized it.

android-agent/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
breedx-splk: Cool. I think it could be helpful to have a tracking issue for this, so I've opened #411 to track this. 👍🏻

android-agent/src/main/java/io/opentelemetry/android/internal/services/network/data/Carrier.kt
```diff
@@ -0,0 +1,74 @@
...
+) {
+    constructor(builder: Builder) : this(
+        id = builder.id,
+        name = builder.name,
+        mobileCountryCode = builder.mobileCountryCode,
+        mobileNetworkCode = builder.mobileNetworkCode,
+        isoCountryCode = builder.isoCountryCode,
+    )
+
+    class Builder {
```
breedx-splk, line 31: I think we want to not use builders for data classes. :) I think we should just give the constructor args default values and remove the builder.

android-agent/src/main/java/io/opentelemetry/android/internal/services/network/CurrentNetworkProvider.java
breedx-splk: Sorry I'm not seeing it. Can you point to where? 🙏🏻

LikeTheSalad: Ah, gotcha. Yeah it seems to be part of the code not shown by GH diff. It's in [NetworkApplicationListener](https://github.com/open-telemetry/opentelemetry-android/pull/408/files#diff-811ab73f473dc6d7c99372a0cce7f2140b7afef8e5665e4539e9ae00ba66c083) line 30 (you'd have to expand the file to see it).

breedx-splk: ok, thanks.

breedx-splk (approved):

android-agent/src/main/AndroidManifest.xml
bidetofevil: Tossing in my 2 cents re: permission

It seems like the library ought not elevate the permissions the app asks for, but instead fail gracefully if permission is not granted. If for some reason someone add this erroneously to their app that should not make network calls (and thus did not ask for the permission), that will open up a hole in the app. 

Not a huge deal, but it's ideal to not have to worry about that

instrumentation/network/build.gradle.kts
bidetofevil: In the future, it would be nice for the Agent to not have an explicit dependency on the instrumentation modules. But that will require an interface layer using which instrumentations can be enabled/disabled/configured via the Agent.

Something something Gradle plugin for build time module inclusion/exclusion based on a config file :-)

android-agent/src/main/java/io/opentelemetry/android/internal/services/network/data/Carrier.kt
bidetofevil: This is probably a matter of personal taste, but I find long-living builders that get passed around for various piece of code to contribute to the init data to be confusing and potentially error-prone - who knows what has been set when the builder is passed around before being built? At that point, the builder itself is carrying state, which partially defeats the purpose of it generating an immutable object.

That said, I'm sure there are exceptions, though I'd have to think really hard before I'd chose a classic builder over constructors+default args or factory methods.

bidetofevil (approved):
Good incremental step

instrumentation/network/build.gradle.kts
breedx-splk: > In the future, it would be nice for the Agent to not have an explicit dependency on the instrumentation modules. But that will require an interface layer using which instrumentations can be enabled/disabled/configured via the Agent.

It's my understanding that this is literally what @LikeTheSalad is working on here -- by making the instrumentations implement the API, they can be service loaded via classpath at runtime, and the install API is lightweight and dependency free (at least from the agent to the instrumentation). It's happening.

LikeTheSalad: Yes, as @breedx-splk mentioned, avoiding the core part of the agent depending on the instrumentation modules is being done in these PRs. Though it's done in a way that, in the end, end users who add the `android-agent` dependency into their projects will still get all the current default instrumentations enabled because the final `android-agent` module will have a dependency on the instrumentations, alongside the "core" module. I did it this way to keep the same behavior somewhat unchanged after the refactoring (at least when it comes to adding the agent dependency into a project). What I understood from what @bidetofevil was mentioning is to use a different approach to include the default instrumentations in the future if I understood correctly? Maybe via a grade plugin that adds the instrumentation dependencies directly into the host project or the like. I haven't thought much about alternatives to add the default instrumentations tbh (these PRs are consuming most of my energy for now) though if we need further work that could keep on improving it I'm up to discuss it once I'm done with these PRs 😅 also because I think it's better not to add too many breaking changes at once.

bidetofevil: I think what I imagined the eventual state to be is what the diagram outlines - there being no direct or indirect dependency from the eventual Agent module on any module that contains instrumentation implementations. If that's what the eventual state will be, that's great - apologies for missing the context.

The second line is a completely different thought, and like @LikeTheSalad described, is more about the *App* using the Agent not having to explicitly include the modules in its gradle files, but rather have it automatically pulled in with a gradle plugin. This will allow us to have a central location to manage versioning instead of making the app update. But this is very much a forward-looking dream and not something that should be included in this PR or even in any short term plans 😅

LikeTheSalad: > I think what I imagined the eventual state to be is what the diagram outlines - there being no direct or indirect dependency from the eventual Agent module on any module that contains instrumentation implementations.

This is what will happen with the existing `android-agent` module, which at some point will be renamed to something like `core` or similar to denote the core/init functionality of the agent. Now, the _new_ `android-agent` module will have a dependency on `core` plus all the instrumentation modules that we want to provide by default. This is because of what we discussed some time ago about providing some instrumentations by default that we thought people would expect the agent to have out of the box. The only way I think we could avoid any dependencies on the agent, while still being able to provide default instrumentations, would be via a gradle plugin that adds those dependencies straight into the host project, though I agree that it sounds like something for a not-so-near future as I think we should discuss the pros/cons/alternatives in detail first.

For now, I'm trying to keep overall functionality somewhat intact, hence the module renaming to `core` to still provide a library named `android-agent` with the same default instrumentations as the current `android-agent` provides.

bidetofevil: Understood. Maybe we can discuss this further in the future - I believe we can make it so that eventually, the whole of the Agent co de do not need to know about or directly depending on any specific instrumentation's implementation code. But that won't be for right now and may not be worth it, so we can move on 😅

PR #414, opened 2024-06-05
Progress on demo app

breedx-splk:
This attempts to flesh out a little more behavior in the demo app:

* make the session id selectable so that the id can be copied
* add exit button in "store" to return to intro screen
* add fake clients for loading product list and loading images
* display scrollable list of products names, with image and price


bidetofevil (approved):
Seems fine for a demo app. If this were real I'd have some additional comments but LGTM for what it is.

Might be nice to break out things you want to have in here into issues so folks looking to contribute can pick them up?

demo-app/src/main/java/io/opentelemetry/android/demo/clients/ProductCatalogClient.kt
```diff
@@ -0,0 +1,22 @@
...
+
+/** A fake (for now!) client for the ProductCatalog */
+// TODO: breedx-splk add instrumentation!
+// TODO: Allow this to be wired up to the go service from the actual demo
+class ProductCatalogClient(private val context: Context) {
+
+    fun get(): List<Product> {
+        val input = context.assets.open(PRODUCTS_FILE)
+        val jsonStr = input.bufferedReader().use { it.readText() }
+        val wrapper = Gson().fromJson(jsonStr, ProductDeserializationWrapper::class.java)
```
marandaneto, line 19: `fromJson` has a reader overload so you don't need to `readText()` but pass the reader directly

marandaneto (approved):

LikeTheSalad (approved):

PR #415, opened 2024-06-06
Main to instrumentation feature branch

LikeTheSalad:
To avoid potential future git conflicts

PR #416, opened 2024-06-06
Instrumentation API part 2 to main

LikeTheSalad:
Supersedes https://github.com/open-telemetry/opentelemetry-android/pull/415


breedx-splk:
There look to be a bunch of unrelated changes in this PR. @LikeTheSalad can you look into that plz?

LikeTheSalad:
> There look to be a bunch of unrelated changes in this PR. @LikeTheSalad can you look into that plz?

Do you mean changes that are not in the main branch? If you could point to one in specific @breedx-splk I'll have a deeper look tomorrow

breedx-splk:
> > There look to be a bunch of unrelated changes in this PR. @LikeTheSalad can you look into that plz?
> 
> Do you mean changes that are not in the main branch? If you could point to one in specific @breedx-splk I'll have a deeper look tomorrow

Yeah, there are a TON. Like all the demo app stuff, and some of the semconv stuff from earlier. Just look at the files list and you'll see what I'm talking about.

LikeTheSalad:
> Yeah, there are a TON. Like all the demo app stuff, and some of the semconv stuff from earlier. Just look at the files list and you'll see what I'm talking about.

I see. Yeah all the semconv changes and the demo app ones have been recently merged into `main`, so I think it makes sense that they show up here because I'm bringing all the latest changes from `main` into the `feature/instrumentation-api` branch. My goal with this PR is to keep the feature branch updated with the latest changes from `main`, otherwise I could run into a lot of git conflicts later given that the `main` branch is changing a lot these days. So for example, in [this PR](https://github.com/open-telemetry/opentelemetry-android/pull/410) is where the semconv changes were added into `main`, a lot of the changes made there are affecting the instrumentation modules so that's one of the reasons why I'd like to bring those changes into the feature branch before starting to migrate those modules, otherwise when it comes the time to merge the feature branch into main, that PR is going to be full of git conflicts if `main` changed too much while the feature branch was getting ready, and that's what I'd like to avoid by bringing the latest changes from `main` into the feature branch as often as possible.

breedx-splk:
Oh, I think I completely missed the fact that the part 1 and part 2 PRs (and now part 3) were going against this branch. Why do you want to build up all those compounded changes into a branch before merging to main? I'm not seeing the benefit. I'd much rather review and merge smaller PRs into main.

Having the changes in this other branch also means that the same changes are getting reviewed twice, as well as all these redundant changes from main (semconv, demo app, etc). How do you propose that we review this?

LikeTheSalad:
@breedx-splk Based on the above, I've just changed the base branch of this PR to `main`, to merge the "part 2" into `main` to continue there with further parts as well.

android-agent/src/main/java/io/opentelemetry/android/internal/services/network/CarrierFinder.java
```diff
@@ -0,0 +1,50 @@
...
+    public CarrierFinder(TelephonyManager telephonyManager) {
+        this.telephonyManager = telephonyManager;
+    }
+
+    public Carrier get() {
+        int id = telephonyManager.getSimCarrierId();
+        String name = null;
+        String mobileCountryCode = null;
+        String mobileNetworkCode = null;
+        String isoCountryCode = null;
```
breedx-splk, line 30: nit: should probably move these declarations closer to their usage.

breedx-splk (approved):
Déjà vu!! 😁

android-agent/src/main/java/io/opentelemetry/android/internal/services/network/CarrierFinder.java
LikeTheSalad: They are used in line 44 but they can change before reaching that line, that's why they're not close to line 44 because of the "ifs" needed to re-set them before that.

PR #421, opened 2024-06-07
Actually call listeners

breedx-splk:
As part of #397, we introduced listeners who can get notified immediately after the otel sdk is created. This is used when creating the initialization events, but users also have the ability to add. Unfortunately, we forgot to actually call the listeners!


android-agent/src/test/java/io/opentelemetry/android/OpenTelemetryRumBuilderTest.java
```diff
@@ -348,4 +344,8 @@ private static ServiceManager createServiceManager(Object... services) {
     private OpenTelemetryRumBuilder makeBuilder() {
         return OpenTelemetryRum.builder(application, buildConfig());
     }
+
+    private OtelRumConfig buildConfig() {
```
bidetofevil, line 348: Since this is called in several tests and doesn't simply create a `OtelRumConfig` instance that just uses the defaults, should the method name indicate that?

bidetofevil (approved):
LGTM - one non-blocking comment about the method name

LikeTheSalad (approved):

PR #439, opened 2024-06-17
Instrumentation API part 4

LikeTheSalad:
These changes are focused on the "slowrendering" module migration.


instrumentation/slowrendering/src/main/java/io/opentelemetry/android/instrumentation/slowrendering/SlowRenderingDetectorBuilder.java
LikeTheSalad, line 1: This was renamed to `SlowRenderingInstrumentation` and then added the `installOn` method from `SlowRenderingDetector`.

instrumentation/slowrendering/src/main/java/io/opentelemetry/android/instrumentation/slowrendering/SlowRenderingInstrumentation.java
```diff
@@ -0,0 +1,61 @@
...
+                            + interval
+                            + "; must be positive");
+            return this;
+        }
+        this.slowRenderingDetectionPollInterval = interval;
+        return this;
+    }
+
+    @Override
+    public void install(
```
bidetofevil, line 42: Perhaps we can add a `@RequiresApi` or `@ChecksSdkIntAtLeast` annotation here or to the class itself to flag attempted usage in a lower API level?1020.

instrumentation/slowrendering/src/test/java/io/opentelemetry/android/instrumentation/slowrendering/SlowRenderingInstrumentationTest.kt
```diff
@@ -0,0 +1,92 @@
...
+import io.opentelemetry.api.OpenTelemetry
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+import java.time.Duration
+
+@RunWith(RobolectricTestRunner::class)
```
bidetofevil, line 26: Can use `AndroidJUnit4` test runner here too

bidetofevil (approved):
LGTM. Nothing blocking, just a couple of suggestions

instrumentation/slowrendering/src/main/java/io/opentelemetry/android/instrumentation/slowrendering/SlowRenderingInstrumentation.java
LikeTheSalad: I didn't think of it because this method (at least for now) is meant to be called by the agent and not directly by our users, although I guess it's not impossible that someone would want to call it directly, so I guess it's probably better to make sure. I'll add the changes.

instrumentation/slowrendering/src/test/java/io/opentelemetry/android/instrumentation/slowrendering/SlowRenderingInstrumentationTest.kt
LikeTheSalad: I've made the changes, also made the missing test dependency available to all android lib modules.

instrumentation/slowrendering/src/main/java/io/opentelemetry/android/instrumentation/slowrendering/SlowRenderingInstrumentation.java
LikeTheSalad: I've made the changes.

instrumentation/slowrendering/src/main/java/io/opentelemetry/android/instrumentation/slowrendering/SlowRenderingInstrumentation.java
```diff
@@ -0,0 +1,63 @@
...
+
+    Duration slowRenderingDetectionPollInterval = Duration.ofSeconds(1);
+
+    /**
+     * Configures the rate at which frame render durations are polled.
+     *
+     * @param interval The period that should be used for polling.
+     * @return {@code this}
+     */
+    public SlowRenderingInstrumentation setSlowRenderingDetectionPollInterval(Duration interval) {
```
breedx-splk, line 29: Presumably the agent detects when this class is on the classpath, instantiates it, and installs it. How would we expect this method to ever get called? I think there are no usages outside of tests.

LikeTheSalad: @breedx-splk I'm a bit worried about not having all the parts ready before the next release, since the agent won't have any auto instrumentation working until all parts are done, so I'll go ahead and merge this PR and we can continue this discussion on the SIG if you'd like as this way of configuring instrumentations is not exclusive to the slowrendering one anyway.

PR #446, opened 2024-06-21
Instrumentation API part 6

LikeTheSalad:
I think this will be the most complicated migration. It's for the `activity` module, which is currently coupled with the `fragment`, `startup`, and the `lifecycle` ones.

Also, similarly to the `network` module, the `activity` one had 2 responsibilities:
- Creating telemetry on activity lifecycle methods.
- Providing a tool (`VisibleScreenTracker`) that was used to create a SpanProcessor to add the `screen.name` attr to all spans.

The second responsibility is needed during the OTel instance initialization, so the tool `VisibleScreenTracker` was converted into a service (`VisibleScreenService`) which is used by the init process as well as by the activity instrumentation and will also be used by the `fragment` module in a follow-up PR.

Since all these modules are related: `activity`, `fragment`, `startup`, and `lifecycle`, I was planning to migrate them all at once, but then it turned out to be too many code changes, so most of the changes I left for this part are related to the `activity` module, but might contain some smaller changes into the other modules so that they could compile and we could get a green build for this PR.


android-agent/src/main/java/io/opentelemetry/android/internal/services/visiblescreen/VisibleScreenService.kt
LikeTheSalad, line 1: This is the old [VisibleScreenTracker](https://github.com/open-telemetry/opentelemetry-android/blob/5511ba947bbf6d60555d55d16eff06c2fabab078/instrumentation/activity/src/main/java/io/opentelemetry/android/instrumentation/activity/VisibleScreenTracker.java#L27) which was turned into a service to be able to reuse it across modules and for the init process as well.

android-agent/src/main/java/io/opentelemetry/android/internal/services/visiblescreen/fragments/VisibleFragmentTracker.kt
LikeTheSalad, line 1: This provides fragment lifecycle updates that were previously sent from [this class](https://github.com/open-telemetry/opentelemetry-android/blob/5511ba947bbf6d60555d55d16eff06c2fabab078/instrumentation/fragment/src/main/java/io/opentelemetry/android/instrumentation/fragment/RumFragmentLifecycleCallbacks.java#L90).

android-agent/src/test/java/io/opentelemetry/android/internal/services/visiblescreen/VisibleScreenServiceTest.java
LikeTheSalad, line 1: This was just moved from the `activity` module.

android-agent/src/test/java/io/opentelemetry/android/internal/services/visiblescreen/activities/VisibleScreenLifecycleBindingTest.java
LikeTheSalad, line 1: This was moved from the `activity` module.

instrumentation/activity/src/main/java/io/opentelemetry/android/instrumentation/activity/ActivityLifecycleInstrumentation.kt
LikeTheSalad, line 1: This covers the activity instrumentation initialization which was previously done in the `lifecycle` module [here](https://github.com/open-telemetry/opentelemetry-android/blob/5511ba947bbf6d60555d55d16eff06c2fabab078/instrumentation/lifecycle/src/main/java/io/opentelemetry/android/instrumentation/lifecycle/AndroidLifecycleInstrumentation.java#L70).

instrumentation/activity/src/main/java/io/opentelemetry/android/instrumentation/activity/VisibleScreenTracker.java
LikeTheSalad, line 1: This was moved to the agent and renamed to `VisibleScreenService`.

instrumentation/activity/src/test/java/io/opentelemetry/android/instrumentation/activity/VisibleScreenLifecycleBindingTest.java
LikeTheSalad, line 1: This was moved to the agent.

instrumentation/activity/src/test/java/io/opentelemetry/android/instrumentation/activity/VisibleScreenTrackerTest.java
LikeTheSalad, line 1: This was moved to the agent and renamed to VisibleScreenServiceTest.

android-agent/src/main/java/io/opentelemetry/android/internal/services/visiblescreen/VisibleScreenService.kt
```diff
@@ -0,0 +1,125 @@
...
+            }
+            val lastActivity = lastResumedActivity.get()
+            if (lastActivity != null) {
+                return lastActivity
+            }
+            return "unknown"
+        }
+
+    fun activityResumed(activity: Activity) {
+        lastResumedActivity.set(activity.javaClass.simpleName)
```
bidetofevil, line 93: Do we have a R8 rule that or at least documentation that states activity class names shouldn't be obfuscated if users want the actual name of the activity to be used?

android-agent/build.gradle.kts
```diff
@@ -63,15 +63,15 @@ android {
...
     implementation(project(":instrumentation:crash"))
+    implementation(project(":instrumentation:startup")) // todo remove during the initializationevents adaptation work.
     implementation(libs.androidx.core)
     implementation(libs.androidx.navigation.fragment)
     implementation(libs.androidx.lifecycle.process)
 
     api(platform(libs.opentelemetry.platform))
     api(libs.opentelemetry.api)
-    implementation(libs.opentelemetry.sdk)
+    api(libs.opentelemetry.sdk)
```
bidetofevil, line 74: Are we saying that by including the `opentelemetry-android` project, an app is transitively importing the entire `opentelemetry-java` API and SDK, and can reference all those methods and objects even if they didn't import those projects directly?

bidetofevil (approved):
As a migration/refactor PR, this looks good to me. I have a couple of non-blocking questions though. As well, I have some questions about the implementation of the screen/view tracking, specifically if there are activities that include more than one fragment, but that's not directly about what this PR is doing, so we can talk about this in the next SIG meeting.

android-agent/src/main/java/io/opentelemetry/android/internal/services/visiblescreen/VisibleScreenService.kt
marandaneto: Activity's are never minified, its part of the default rules

bidetofevil: (Does some reading)

Right, activities are referenced in the manifest so they can't be obfuscated. But fragment class names are obfuscated, and we try to do a similar thing too. I guess it doesn't matter so much if no one uses fragments anymore?

marandaneto: I think it does matter, many apps still use fragments, or at least have a mixed setup.
IIRC we've discussed that before, ideally we'd offer a callback where the user can override the fragment's name based on the fragment instance/type, and if nothing is provided, we default to `simpleName` (which can be minified).

android-agent/build.gradle.kts
LikeTheSalad: Yes. In this case it was [needed](https://github.com/open-telemetry/opentelemetry-android/pull/446/files#diff-4ca93ba3ce26d5157044736a24cd06a5705998161d2933bf6fc1e0c64ae05cbfR8) in the `startup` module which references the [Clock](https://javadoc.io/doc/io.opentelemetry/opentelemetry-sdk-common/latest/io/opentelemetry/sdk/common/Clock.html) interface, which is available in the sdk (now that I think about it, it seems like it's a good interface for the OTel api lib). I guess we can also just duplicate that line straight into the `startup` gradle file and other modules that might need to reference an sdk specific type, is just that so far I didn't see the need for it. I'm curious to learn your thoughts about it, because based on what you mentioned it seems like exposing the OTel sdk classes to consumers of the Android agent might be a bad idea?

android-agent/src/main/java/io/opentelemetry/android/internal/services/visiblescreen/VisibleScreenService.kt
LikeTheSalad: I'm also learning more details about the existing instrumentations with these changes, so I think it's a great opportunity to bring these kinds of questions up. I've created [this issue](https://github.com/open-telemetry/opentelemetry-android/issues/459) to not loose track of this one, and added a couple of topics for tomorrow's SIG meeting.

android-agent/build.gradle.kts
marandaneto: I think it depends on what is needed, if we are exposing `api` just because we need to access some internals from OTel Java, that seems wrong.
If we are exposing because users might use features that are already implemented in the OTel Java, that seems just fine.

android-agent/src/main/java/io/opentelemetry/android/internal/services/visiblescreen/VisibleScreenService.kt
bidetofevil: Another option is to add a proguard rule to ensure they aren't obfuscated. As long as we document that, it seems reasonable.

android-agent/build.gradle.kts
bidetofevil: Yeah, basically what Manoel said. Exporting a dependency transitively isn't ideal because you're basically forcing your dependency directly to your consumer. It makes sense if it's a deliberate choice and is stated explicitly - depending on X means you are also directly importing Y, so users can accidentally take a dependency on classes in Y without knowing it, and if they decided to remove X later from their dependency change, they'll still be glued to Y.

But I can see that in our case, it makes sense - this project is just an extension of `opentelemetry-java`, so depending on us means you're explicitly depending on that as well. If that's the stance we want to take, we should do that, but also document it. And if we go down this route, we should probably export the semconv module too, so apps that use us won't need to directly depend on any of the core OTel modules explicitly if they depend on us. Doing that also ensures that there are no version conflicts, which is always good.

marandaneto: Other modules can also depend on `libs.opentelemetry.sdk` if needed as long as it's not a circular dependency :D

LikeTheSalad: Gotcha, from a first glance I guess I look at it as Hanson mentioned, i.e. "this project is an extension of the java SDK so it's fine to expose it" kind of view... Though I'm open to discuss it more in detail, given that said view that I mentioned might also mean exposing other libs as Hanson mentioned, so for now I'll just add the SDK straight into the modules that need it and will set some time in the SIG to talk this more in detail.

Issue #148, opened 2023-11-14
Add HttpUrlConnection instrumentation tests


breedx-splk:
The footprint of #133 is pretty sizable and will be experimental for a time, so we have decided to defer the tests to a future PR. This is a tracking issue to create appropriate an appropriate `InstrumentationTest` similar to the okhttp test. This should run on device and exercise the instrumentation against the testing endpoint.

surbhiia:
Do assign this to me! Thanks :)

breedx-splk:
We have this now!

PR #452, opened 2024-06-25
Add androidTests for httpUrlConnection module and create test-common module for common test utilities

surbhiia:
- This PR adds relevant androidTests for httpUrlConnection module. 
- Test-common module has also been added to park the common test utilities

Related Issue - https://github.com/open-telemetry/opentelemetry-android/issues/148

Feel free to suggest better naming for the new files. One observation I had was that, by naming module "test-common" it moves to the bottom of the list in project view in android studio and if we were to name it "common-test" it would show up right below the "common" module. Let me know if anyone has any preferences here. :)

Screenshot 2024-06-25 at 9 32 46 AM


instrumentation/httpurlconnection/testing/src/androidTest/java/io/opentelemetry/instrumentation/library/httpurlconnection/InstrumentationTest.java
```diff
@@ -0,0 +1,135 @@
...
+        assertEquals(1, inMemorySpanExporter.getFinishedSpanItems().size());
+    }
+
+    @Test
+    public void
+            testHttpUrlConnectionGetRequest_WhenNoStreamFetchedAndNoDisconnectCalledButHarvesterScheduled_ShouldBeTraced() {
+        HttpUrlConnectionTestUtil.executeCustomGet("http://httpbin.org/get", false, false);
+        ScheduledExecutorService harvester = scheduleHarvester();
+        try {
+            Thread.sleep(15000);
```
marandaneto, line 70: I'd try to write the test in a way that we don't need to `sleep` that long, usually, this slows down the test pipeline quite fast

instrumentation/httpurlconnection/testing/src/androidTest/java/io/opentelemetry/instrumentation/library/httpurlconnection/InstrumentationTest.java
```diff
@@ -0,0 +1,135 @@
...
+        executor.shutdown();
+
+        try {
+            // Wait for all tasks to finish execution or timeout
+            if (executor.awaitTermination(2, TimeUnit.SECONDS)) {
+                // if all tasks finished
+                assertEquals(4, inMemorySpanExporter.getFinishedSpanItems().size());
+            } else {
+                Assert.fail(
+                        "Test could not be completed as tasks did not complete within the 2s timeout period.");
```
marandaneto, line 97: I think this is not needed, if `awaitTermination` exceeds the 2s, it'll throw InterruptedException anyway, so the else won't ever be called IIRC.

instrumentation/httpurlconnection/testing/src/androidTest/java/io/opentelemetry/instrumentation/library/httpurlconnection/InstrumentationTest.java
```diff
@@ -0,0 +1,135 @@
...
+    @Test
+    public void
+            testHttpUrlConnectionGetRequest_WhenFourConcurrentRequestsAreMade_AllShouldBeTraced() {
+        ExecutorService executor = Executors.newFixedThreadPool(4);
+        executor.submit(() -> HttpUrlConnectionTestUtil.executeGet("http://httpbin.org/get"));
+        executor.submit(() -> HttpUrlConnectionTestUtil.executeGet("http://google.com"));
+        executor.submit(() -> HttpUrlConnectionTestUtil.executeGet("http://android.com"));
+        executor.submit(() -> HttpUrlConnectionTestUtil.executeGet("http://httpbin.org/headers"));
+
+        executor.shutdown();
```
marandaneto, line 88: I think this or `shutdownNow` should be in the finally block

instrumentation/httpurlconnection/testing/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlConnectionTestUtil.java
```diff
@@ -0,0 +1,96 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.library.httpurlconnection;
+
+import android.util.Log;
```
marandaneto, line 8: If you remove the log dependency, this can be part of the `test-common` I guess?
Its a plain Java file that can be reused across different modules

marandaneto:
Thanks for adding the tests.
It'd be cool if tests were written in Kotlin already, not a blocker at all though.

instrumentation/httpurlconnection/testing/src/androidTest/java/io/opentelemetry/instrumentation/library/httpurlconnection/InstrumentationTest.java
LikeTheSalad: ➕ @surbhiia maybe you can somehow mock the time delay instead? We did something like that in the [PeriodicRunnableTest](https://github.com/open-telemetry/opentelemetry-android/blob/475f08f78c101d0b8d9e71e28134e933cec00a29/android-agent/src/test/java/io/opentelemetry/android/internal/services/periodicwork/PeriodicRunnableTest.kt#L28) class, not sure if the same approach would work though, but probably we can come up with some alternative anyway.

instrumentation/httpurlconnection/testing/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlConnectionTestUtil.java
surbhiia: These need to be here as we're instrumenting these with our httpurlconnection bytebuddy agent so we can test the same.

instrumentation/httpurlconnection/testing/src/androidTest/java/io/opentelemetry/instrumentation/library/httpurlconnection/InstrumentationTest.java
surbhiia: Actually, awaitTermination can have three results, and we're catering to all 3 in our code:

- true -> if tasks complete before 2 sec wait time -> as tasks complete we assert if all 4 traces are created

- false -> if tasks don't complete in 2 seconds -> we fail the test with relevant message (that test could not be completed within 2 second time period instead of saying implementation under test is wrong), as we already know assertion will fail, as no traces would get created from non-completed tasks

- interrupted exception -> if thread gets interrupted -> we fail the test with relevant message (that test could not be completed due to thread interruption instead of saying implementation under test is wrong). I added assertFail for this as well so it's clear messaging what went wrong as we already know it, instead of leaving it at exception which would require us to dig in and see what exactly happened.

instrumentation/httpurlconnection/testing/src/androidTest/java/io/opentelemetry/instrumentation/library/httpurlconnection/InstrumentationTest.java
surbhiia: The shutdown here is a smooth shutdown called before executor.awaitTermination, that allows tasks to complete before shutting down the executor. And the executor.shutdownNow() in catch statement handles the scenario for shutting down executor if an exception occurs in awaitTermination call. 

That being said, the existing code does not initiate shutdown if exception were to occur in the four executor.submit calls. I'll be re-arranging the code to account for that as well.

instrumentation/httpurlconnection/testing/src/androidTest/java/io/opentelemetry/instrumentation/library/httpurlconnection/InstrumentationTest.java
```diff
@@ -0,0 +1,135 @@
...
+    private static final InMemorySpanExporter inMemorySpanExporter = InMemorySpanExporter.create();
+
+    @Before
+    public void setUp() {
+        OpenTelemetryTestUtils.setUpSpanExporter(inMemorySpanExporter);
+    }
+
+    @After
+    public void tearDown() {
+        inMemorySpanExporter.reset();
```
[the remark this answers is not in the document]
surbhiia: That's a great point! Will bring it up for discussion in our next Android Sig meeting. :)

surbhiia:
> Thanks for adding the tests. It'd be cool if tests were written in Kotlin already, not a blocker at all though.

Will convert this to Kotlin! Not many files in this PR, so should not take long. :)

instrumentation/httpurlconnection/testing/src/androidTest/java/io/opentelemetry/instrumentation/library/httpurlconnection/InstrumentationTest.java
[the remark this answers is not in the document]
marandaneto: not a blocker though, I believe more tests do this (global state).

instrumentation/httpurlconnection/testing/src/androidTest/java/io/opentelemetry/instrumentation/library/httpurlconnection/InstrumentationTest.java
marandaneto: tbh, it won't make any difference
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html#shutdown--
`shutdown` just does not allow new submits, for this test specifically unless I misunderstood something, does nothing (because you don't submit any more tasks nor any other thread does).
`shutdown` and `shutdownNow` are more for cleaning resources/avoiding memory leaks in general, in tests since GC cleans up everything after the test is finished, won't make much difference, but of course, it's still a good practice.

instrumentation/httpurlconnection/testing/src/androidTest/java/io/opentelemetry/instrumentation/library/httpurlconnection/InstrumentationTest.java
marandaneto: true, that is correct, thanks

instrumentation/httpurlconnection/testing/src/androidTest/java/io/opentelemetry/instrumentation/library/httpurlconnection/InstrumentationTest.java
surbhiia: The javadoc descriptions around these are definitely confusing! We can't call awaitTermination without calling a shutdown first, the awaitTermination java description covers that:
- https://docs.oracle.com/javase%2F8%2Fdocs%2Fapi%2F%2F/java/util/concurrent/ExecutorService.html#awaitTermination-long-java.util.concurrent.TimeUnit-

- https://howtodoinjava.com/java/multi-threading/executorservice-shutdown/#:~:text=The%20awaitTermination(long%20timeout%2C%20TimeUnit,after%20a%20shutdown()%20request.

instrumentation/httpurlconnection/testing/src/androidTest/java/io/opentelemetry/instrumentation/library/httpurlconnection/InstrumentationTest.java
```diff
@@ -0,0 +1,141 @@
...
+                        "Test could not be completed as tasks did not complete within the 2s timeout period.");
+            }
+        } catch (InterruptedException e) {
+            // print stack trace to decipher lines that threw InterruptedException as it can be
+            // possibly thrown by multiple calls above.
+            e.printStackTrace();
+            Assert.fail("Test could not be completed due to an interrupted exception.");
+        } finally {
+            if (!executor.isShutdown()) {
+                executor.shutdownNow();
```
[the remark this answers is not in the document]
surbhiia: I did come across this. After some lookup, I decided that it is not required here - as we don't want any other things to happen post a interruption - we just catch it and fail the test, other parts of the code don't need to do anything based on the interruption status so we don't need to reset it. Does that make sense?

instrumentation/httpurlconnection/testing/src/androidTest/java/io/opentelemetry/instrumentation/library/httpurlconnection/InstrumentationTest.java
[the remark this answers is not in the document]
surbhiia: @marandaneto @LikeTheSalad Do you see any concerns with having an inMemorySpanExporter per test case? We can do that right away right now to avoid issues later if we were to run tests in parallel? I can make the change in okhttp3.InstrumentationTest.java too for same in this PR.

instrumentation/httpurlconnection/testing/src/androidTest/java/io/opentelemetry/instrumentation/library/httpurlconnection/InstrumentationTest.java
surbhiia: One simpler approach that comes to mind is - when we make connection's inactivity timeout configurable (in next PR), we can set it 0s in this test case and run the harvester just once, right after, to end the trace, we can use ExecutorService and do a 2s awaitTimeout (mostly will finish before that) for this task to complete and do our assertion. Example:

`ExecutorService executorService = Executors.newSingleThreadExecutor();
  try {
        executorService.submit(HttpUrlInstrumentationConfig.getReportIdleConnectionRunnable());

        executorService.shutdown();

       if (executorService.awaitTermination(2, TimeUnit.SECONDS)) {
               Assert.assertEquals(1, inMemorySpanExporter.finishedSpanItems.size)
       } else {
               Assert.fail(
                    "Test could not be completed as tasks did not complete within the 2s timeout period.",
                )
        }
        } catch (InterruptedException e) {
         
        } finally {}
`

We can remove this test for now and add it in next PR (when we make timeout configurable), with this new approach. Does this look good?

instrumentation/httpurlconnection/testing/src/androidTest/java/io/opentelemetry/instrumentation/library/httpurlconnection/InstrumentationTest.java
marandaneto: true, it's been a while since I last used such methods.

instrumentation/httpurlconnection/testing/src/androidTest/java/io/opentelemetry/instrumentation/library/httpurlconnection/InstrumentationTest.java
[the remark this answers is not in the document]
marandaneto: if tests are not running in parallel, all good, if they do, it is a must to avoid concurrency issues and flaky tests

surbhiia:
I'll be on PTO next week. Will address any new comments starting next to next week! :)

instrumentation/httpurlconnection/testing/src/androidTest/java/io/opentelemetry/instrumentation/library/httpurlconnection/InstrumentationTest.java
bidetofevil: Another way to test this is to fake the clock this is using, but it requires a lot of plumbing will be beyond the scope of this change. 

I agree with @marandaneto - sleeps not only slow the test down, it can also introduce flakiness down the line because execution in CI is often unpredictable. So if you have to do it, make the value configurable so it won't take that long, and rely on `CountDownLatch` to verify conditions before proceeding rather than expecting certain things to be done after a certain amount of time.

instrumentation/httpurlconnection/testing/src/androidTest/java/io/opentelemetry/instrumentation/library/httpurlconnection/InstrumentationTest.java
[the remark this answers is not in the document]
bidetofevil: I would not recommend putting any test infrastructure in static scope. At minimum, have one per test instance, but ideally you create a new one per test run if it involves additional threads. This project isn't so big that reinits are goign to slow the test suite runs down, so it's probably better to be as clean as possible to avoid weird flakiness.

bidetofevil:
We talked about wanting all of the code to eventually be Kotlin (I will begin converting soon), so new code, especially tests, should be in Kotlin. 

We should add something to the contribution guide about that so anything new is Kotlin unless it really can't be. I can find a place for that.... CC @LikeTheSalad @breedx-splk

instrumentation/httpurlconnection/testing/src/androidTest/java/io/opentelemetry/instrumentation/library/httpurlconnection/InstrumentationTest.java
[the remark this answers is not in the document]
LikeTheSalad: I agree, I think we should avoid statics in general unless it's necessary for a particular test case.

LikeTheSalad:
> We talked about wanting all of the code to eventually be Kotlin (I will begin converting soon), so new code, especially tests, should be in Kotlin.
> 
> We should add something to the contribution guide about that so anything new is Kotlin unless it really can't be. I can find a place for that.... CC @LikeTheSalad @breedx-splk

I agree, we should add it to the contributing guide.

instrumentation/httpurlconnection/testing/src/androidTest/java/io/opentelemetry/instrumentation/library/httpurlconnection/InstrumentationTest.java
surbhiia: I've removed the harvester test from this PR. Will add it back in next PR when we make the inactivity timeout configurable, so we can set it to 0 and assert after 1 run of the harvester. I can wrap the runnable with a new runnable with CountDownLatch.

instrumentation/httpurlconnection/testing/src/androidTest/java/io/opentelemetry/instrumentation/library/httpurlconnection/InstrumentationTest.java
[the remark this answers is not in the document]
surbhiia: I’ve added a new commit which addresses the following:

- Makes a per test instance of inMemorySpanExporter - thereby making it possible to run these tests in parallel
- For test purposes only - providing a function for re-setting the instrumenter instance for each test - thereby allowing us to re-set opentelemetry instance (with new exporter) for each test and not depending on GlobalOpentelemetry instance (If tests are run in parallel but in same JVM process, GlobalOpentelemetry instance would be shared, so we can’t use that). 

Some other considerations: 
- For now, for general use case, still depending on GlobalOpentelemetry instance until we have a way around that. (We cannot use a installer like AndroidInstrumentation.kt as we do not want to introduce the requirement of client having to make a manual call to install(opentelemetryInstance) after initializing their OpentelemetryInstance. That adds manual parts to this auto-instrumentation. And if we were to automate this install call, it would be difficult/tedious for auto-instrumentations to figure out where exactly to place this call in client code as it needs to be after their opentelemetryInstance initialization which could vary for each customer.)
- For now, I have not used CachedSupplier for getting the instrumenter instance lazily as it synchronizes every read access to instrumenter which I believe has more performance downside. Instead I’ve only synchronized the read access until instrumenter is null and made the variable volatile. We can discuss this more and establish the best approach and unify them. 

I can also make the change for okhttp3 androidTests, if this approach seems okay.

instrumentation/httpurlconnection/testing/src/androidTest/kotlin/io/opentelemetry/instrumentation/library/httpurlconnection/InstrumentationTest.kt
```diff
@@ -0,0 +1,112 @@
...
+import java.util.concurrent.Executors
+import java.util.concurrent.TimeUnit
+
+class InstrumentationTest {
+    @Test
+    fun testHttpUrlConnectionGetRequest_ShouldBeTraced() {
+        val inMemorySpanExporter = InMemorySpanExporter.create()
+        HttpUrlConnectionSingletons.setInstrumenterForTesting(OpenTelemetryTestUtils.setUpSpanExporter(inMemorySpanExporter))
+        executeGet("http://httpbin.org/get")
+        Assert.assertEquals(1, inMemorySpanExporter.finishedSpanItems.size)
```
breedx-splk, line 24: nit: we should use static imports for these, or use the assertj fluent style of `assertThat(...).isEqualTo(1)`. This applies to all these asserts in this PR.

instrumentation/httpurlconnection/testing/src/androidTest/kotlin/io/opentelemetry/instrumentation/library/httpurlconnection/InstrumentationTest.kt
```diff
@@ -0,0 +1,112 @@
...
+import io.opentelemetry.sdk.testing.exporter.InMemorySpanExporter
+import org.junit.Assert
+import org.junit.Test
+import java.util.concurrent.Executors
+import java.util.concurrent.TimeUnit
+
+class InstrumentationTest {
+    @Test
+    fun testHttpUrlConnectionGetRequest_ShouldBeTraced() {
+        val inMemorySpanExporter = InMemorySpanExporter.create()
```
breedx-splk, line 21: This `inMemorySpanExporter` is created at the beginning of every test method and shutdown at the bottom of each. I think this can be better handled by making the exporter a class field and creating it in a `@BeforeEach` method and shutting it down in the `@AfterEach` method.

breedx-splk (approved):
I appreciate that this increases the tests. Some small room remaining for improvements (more kotlin, etc), but I think it's worth moving forward.

test-common/src/main/kotlin/io/opentelemetry/android/test/common/OpenTelemetryTestUtils.kt
```diff
@@ -0,0 +1,43 @@
...
+import io.opentelemetry.api.GlobalOpenTelemetry
+import io.opentelemetry.api.OpenTelemetry
+import io.opentelemetry.api.trace.Span
+import io.opentelemetry.sdk.OpenTelemetrySdk
+import io.opentelemetry.sdk.trace.SdkTracerProvider
+import io.opentelemetry.sdk.trace.export.SimpleSpanProcessor
+import io.opentelemetry.sdk.trace.export.SpanExporter
+
+object OpenTelemetryTestUtils {
+    lateinit var openTelemetry: OpenTelemetry
```
LikeTheSalad, line 17: Maybe after the okhttp inst changes, I think we should update this to avoid relying on storing the otel instance statically for every test. Seems like a test extension/rule could help.

LikeTheSalad (approved):

PR #465, opened 2024-07-03
Instrumentation API part 7

LikeTheSalad:
Continuation of the `activity`-`fragment`-`lifecycle`-`startup` migration. This part covers the `fragment` instrumentation implementation and removes the `lifecycle` module that used to be an entry point for both, `activity` and `fragment` instrumentations, but now each of them has its own entry point (AndroidInstrumentation impl).


instrumentation/fragment/src/main/java/io/opentelemetry/android/instrumentation/fragment/FragmentLifecycleInstrumentation.kt
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    override fun install(
+        application: Application,
+        openTelemetryRum: OpenTelemetryRum,
+    ) {
+        application.registerActivityLifecycleCallbacks(buildFragmentRegisterer(openTelemetryRum))
+    }
+
+    private fun buildFragmentRegisterer(openTelemetryRum: OpenTelemetryRum): ActivityLifecycleCallbacks {
```
LikeTheSalad, line 38: This code was moved from [here](https://github.com/open-telemetry/opentelemetry-android/pull/465/files#diff-3699b73481a1f4e4e9f99c7361324d8395de5d012b7c5b8c0da0afac1e969b4bL91).

breedx-splk (approved):
Looking good to me. Thanks!

PR #480, opened 2024-07-16
Instrumentation API part 8

LikeTheSalad:
These changes cover the crash module migration.


android-agent/src/main/java/io/opentelemetry/android/OpenTelemetryRumImpl.java
```diff
@@ -19,7 +18,7 @@ final class OpenTelemetryRumImpl implements OpenTelemetryRum {
     }
 
     @Override
-    public OpenTelemetry getOpenTelemetry() {
+    public OpenTelemetrySdk getOpenTelemetry() {
```
LikeTheSalad, line 21: I don't like having to use `OpenTelemetrySdk` instead of `OpenTelemetry`, though it was needed because of the way that the existing crash instrumentation works, as it needs to call `SdkLoggerProvider.forceFlush()` inside `CrashReportingExceptionHandler.uncaughtException` to try and get the crash data exported before the app is closed. I hope it's something we can improve later.

LikeTheSalad: Never mind, it seems like just casting it to the SDK impl for the crash instrumentation does the trick without any side effects I could spot.

instrumentation/crash/src/test/java/io/opentelemetry/android/instrumentation/crash/CrashReporterTest.java
```diff
@@ -11,49 +11,59 @@
...
+    private Thread.UncaughtExceptionHandler existingHandler;
 
-    @RegisterExtension
-    static final OpenTelemetryExtension testing = OpenTelemetryExtension.create();
-
-    static Thread.UncaughtExceptionHandler existingHandler;
-
-    @BeforeAll
-    static void setUp() {
+    @Before
```
breedx-splk, line 38: Shouldn't we be using `@BeforeEach` and `@AfterEach`?

breedx-splk (approved):
Had a question about the test, but looks good to me.

instrumentation/crash/src/test/java/io/opentelemetry/android/instrumentation/crash/CrashReporterTest.java
LikeTheSalad: I'd wish, but the robolectric runner doesn't support junit5 :(

PR #486, opened 2024-07-23
Instrumentation API part 9

LikeTheSalad:
Migrates the `startup` module to the new instrumentation API.

This is the last instrumentation module that the `android-agent` module has a dependency on.


android-agent/src/main/java/io/opentelemetry/android/initialization/InitializationEvents.kt
```diff
@@ -25,12 +27,39 @@ interface InitializationEvents {
...
+                } else {
+                    set(NO_OP)
+                }
+            }
+
+            return instance!!
+        }
+
+        @JvmStatic
+        fun set(initializationEvents: InitializationEvents) {
```
breedx-splk, line 47: Is this just visible for testing? Could use a comment clarifying such.

breedx-splk: Or can it be made internal? It would be an error for a user to call this multiple times etc.

android-agent/src/test/java/io/opentelemetry/android/initialization/TestInitializationEvents.kt
```diff
@@ -0,0 +1,35 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.android.initialization
+
+import io.opentelemetry.android.config.OtelRumConfig
+import io.opentelemetry.sdk.trace.export.SpanExporter
+
+class TestInitializationEvents : InitializationEvents {
```
breedx-splk, line 11: Is this the same as the noop version? Can we reuse that instead?

buildSrc/src/main/kotlin/otel.java-library-conventions.gradle.kts
```diff
@@ -5,6 +5,7 @@ import ru.vyarus.gradle.plugin.animalsniffer.AnimalSniffer
 plugins {
     id("java-library")
     id("org.jetbrains.kotlin.jvm")
+    id("kotlin-kapt")
```
breedx-splk, line 8: Sure, the android libs need it but do the java libs really also need this?

instrumentation/startup/src/main/java/io/opentelemetry/android/instrumentation/startup/SdkInitializationEvents.kt
```diff
@@ -5,33 +5,30 @@
...
-        val map = mutableMapOf<String, AnyValue<*>>()
-        config.entries.forEach(
-            Consumer { e: Map.Entry<String, String> ->
-                map[e.key] = AnyValue.of(e.value)
-            },
-        )
-        val body = AnyValue.of(map)
-        addEvent(RumConstants.Events.INIT_EVENT_CONFIG, body = body)
+    override fun recordConfiguration(config: OtelRumConfig) {
+        // TODO convert config to AnyValue and add an event for it named RumConstants.Events.INIT_EVENT_CONFIG
```
breedx-splk, line 31: Can we get an issue for this please?

android-agent/src/main/java/io/opentelemetry/android/initialization/InitializationEvents.kt
LikeTheSalad: Good point, I don't think there's any reason to expose it for now, so I've moved it to the internal package.

android-agent/src/test/java/io/opentelemetry/android/initialization/TestInitializationEvents.kt
LikeTheSalad: It's kinda the same, though the reason I created it was for [this test](https://github.com/open-telemetry/opentelemetry-android/pull/486/files#diff-e8aa70e2db0e749932aacc44941988c9754c369a96aba4d3afb02ea8e8505756R21). So that the test can verify that the `InitializationEvents.get()` function does search for implementations in the classpath. If no implementation is found, then the function returns the noop impl, so that's why I couldn't use the same one for the test.

buildSrc/src/main/kotlin/otel.java-library-conventions.gradle.kts
LikeTheSalad: Good point, I've removed it from the java lib conventions.

instrumentation/startup/src/main/java/io/opentelemetry/android/instrumentation/startup/SdkInitializationEvents.kt
LikeTheSalad: Sure, I've just created [this one](https://github.com/open-telemetry/opentelemetry-android/issues/489).

android-agent/src/test/java/io/opentelemetry/android/initialization/TestInitializationEvents.kt
breedx-splk: ok, I think that makes sense then.

breedx-splk (approved):

PR #498, opened 2024-07-29
Instrumentation API part 10

LikeTheSalad:
This is the last part to migrate the instrumentations that were included by default in the agent and to replace the old API with the new one when initializing OpenTelemetryRum.

The changes made here are the following:
- Renaming the `android-agent` module to `core`.
- Creating a new `android-agent` module as a convenience lib for providing a set of default instrumentations alongside the `core` one. This is done in a way that all the previously default instrumentations are still provided by default for projects that include the `android-agent` library as a dependency.
- Removing the no longer needed `common` module.

Note: Even though it looks like this PR has many file changes, **most of them** are just related to the renaming of the old `android-agent` module to `core`.


android-agent/src/main/kotlin/io/opentelemetry/android/agent/OtelRumConfigExtensions.kt
```diff
@@ -0,0 +1,80 @@
...
+import io.opentelemetry.android.instrumentation.crash.CrashReporterInstrumentation
+import io.opentelemetry.android.instrumentation.fragment.FragmentLifecycleInstrumentation
+import io.opentelemetry.android.instrumentation.network.NetworkChangeInstrumentation
+import io.opentelemetry.android.instrumentation.slowrendering.SlowRenderingInstrumentation
+import io.opentelemetry.android.internal.services.network.data.CurrentNetwork
+import io.opentelemetry.api.trace.Tracer
+import io.opentelemetry.instrumentation.api.instrumenter.AttributesExtractor
+import java.time.Duration
+
+/**
```
LikeTheSalad, line 23: This change isn't really needed, I added it because it seemed handy to have those options in `OtelRumConfig`.

core/build.gradle.kts
LikeTheSalad, line 1: This is not a new file, it's just the old build.gradle.kts file from before renaming the old `android-agent` module to `core`.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/OtelRumConfigExtensions.kt
breedx-splk: Oh yeah interesting, I hadn't thought of doing it with extension methods.... 🤔

core/src/main/java/io/opentelemetry/android/SdkPreconfiguredRumBuilder.kt
```diff
@@ -56,13 +56,21 @@ class SdkPreconfiguredRumBuilder
...
-            //        }
-            return OpenTelemetryRumImpl(sdk, sessionId)
+            val openTelemetryRum = OpenTelemetryRumImpl(sdk, sessionId)
+
+            // Install instrumentations
+            for (instrumentation in getInstrumentations()) {
+                instrumentation.install(application, openTelemetryRum)
+            }
+
+            return openTelemetryRum
```
breedx-splk, line 66: Hooray, we made it! 🥳

core/src/test/java/io/opentelemetry/android/OpenTelemetryRumBuilderTest.java
```diff
@@ -187,27 +192,43 @@ public void shouldBuildLogRecordProvider() {
...
+                        application,
+                        buildConfig().disableInstrumentationDiscovery(),
+                        timeoutHandler)
+                .addInstrumentation(localInstrumentation)
+                .build(serviceManager);
+
+        verify(serviceManager.getAppLifecycleService()).registerListener(timeoutHandler);
+
+        verify(localInstrumentation).install(eq(application), notNull());
+        verifyNoInteractions(classpathInstrumentation);
```
breedx-splk, line 231: So I like this test because I think it is highlighting a current shortcoming with this approach....which I think is fine for now in this PR, but I'd like to see this addressed in the very near future. 

If I'm a user looking at this code for the first time, and I have a bit of manual instrumentation that I've created or initialized or whatever, it's unclear if I'm supposed to register it directly with the registry or if I should `addInstrumentation()` on the `OpenTelemetryRumBuilder`. A causal reader might think that these are two ways of solving the same problem...and they kind of are, until `discoverInstrumentations` gets into the mix. 

The configuration via `discovrerInstrumentations` just prevents the contents of the registry from being added to the final list of instrumentations. It doesn't actually prevent the discovery! I think that even though the `instrumentations` field on the `AndroidInstrumentationRegistryImpl` is lazy, any interactions with it will basically cause it to use the `ServiceLoader` to load all instrumentations on the classpath....but just won't ever install those. For the most part that should be fine, simply creating an instrumentation shouldn't have much overhead and definitely shouldn't have side effects without `install()`...but we have no guarantees.

One way I can think of handling this is to decide if the registry really needs to have the `register` method at all. Maybe we can remove that method and then rename `AndroidInstrumentationRegistry` (and impl) to `AutodetectedClasspathInstrumentations` or something similar.

breedx-splk: Opened #502 to track this and not block this work.

breedx-splk (approved):
This for sticking this out, Cesar @LikeTheSalad !! 🥳

core/src/test/java/io/opentelemetry/android/OpenTelemetryRumBuilderTest.java
LikeTheSalad: This is a great point, we should definitely avoid this ambiguity for adding custom instrumentations.

> One way I can think of handling this is to decide if the registry really needs to have the register method at all. Maybe we can remove that method and then rename AndroidInstrumentationRegistry (and impl) to AutodetectedClasspathInstrumentations or something similar.

I think this sounds good.

PR #500, opened 2024-07-29
Update HttpUrlConnection androidTests to make them compatible for parallel running

surbhiia:
Improving androidTests to make them compatible for running in parallel: 

- Tested by running them parallel in android studio:
Screenshot parallel run

- Tested synchronous run:
Screenshot sync run

- Also tested the actual instrumentation (non test scenario) works as expected. 

- No change for the Okhttp3 androidTests. They run correctly synchronously:
Screenshot okhttp3 sync

### DIFF
--- a/instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/internal/HttpUrlConnectionSingletons.java
+++ b/instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/internal/HttpUrlConnectionSingletons.java
@@ -23,11 +23,17 @@ import java.net.URLConnection;
 
 public final class HttpUrlConnectionSingletons {
 
-    private static volatile Instrumenter<URLConnection, Integer> instrumenter;
     private static final String INSTRUMENTATION_NAME =
             "io.opentelemetry.android.http-url-connection";
-    private static final Object lock = new Object();
+
     private static OpenTelemetry openTelemetryInstance;
+    private static volatile Instrumenter<URLConnection, Integer> instrumenter;
+    private static final Object lock = new Object();
+
+    // ThreadLocal instances for parallel test execution.
+    private static ThreadLocal<OpenTelemetry> testOpentelemetryInstance = new ThreadLocal<>();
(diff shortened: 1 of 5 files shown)


breedx-splk (commented):
I'm struggling with this. I think it's a pretty strong antipattern to mix test code explicitly into the production code, which includes `HttpUrlConnectionSingletons`. I know we sometimes expand visibility to facilitate testing, but it's not that common and this seems like a considerably deeper intrusion. 

I'd like to give some additional consideration to how me might solve this concurrency challenge without compromising the instrumentation code.

instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/internal/HttpUrlConnectionSingletons.java
```diff
@@ -76,23 +80,41 @@ public static Instrumenter<URLConnection, Integer> createInstrumenter(
...
     }
 
     public static OpenTelemetry openTelemetryInstance() {
-        return openTelemetryInstance;
+        OpenTelemetry testOpenTelemetry = testOpentelemetryInstance.get();
+        if (testOpenTelemetry != null) {
+            return testOpenTelemetry;
+        } else {
+            return openTelemetryInstance;
+        }
```
breedx-splk, line 106: nit: always prefer uninverting the `if` and omit the redundant `else`.

surbhiia:
> I'm struggling with this. I think it's a pretty strong antipattern to mix test code explicitly into the production code, which includes `HttpUrlConnectionSingletons`. I know we sometimes expand visibility to facilitate testing, but it's not that common and this seems like a considerably deeper intrusion.
> 
> I'd like to give some additional consideration to how me might solve this concurrency challenge without compromising the instrumentation code.

I will also look into other ways of solving it in a cleaner way! :)

surbhiia:
Closing this PR as per our discussion in Android Sig today morning. 

One possible approach we discussed we could look at for future is - Using roboelectric to test most of the functionality without BCI (as BCI might not be possible with it) and having 1 androidTest to test that BCI works (end-to-end integration).

For now, we will have 6 androidTests for httpurlconnection that can be run synchronously.

Issue #433, opened 2024-06-13
Issue with Android Instrumentation for OkHttp


esdrasdl:
Hello. I'm having some issues to use Open telemetry in my android app. Because I want to track every request as a span, I followed the instructions [here](https://github.com/open-telemetry/opentelemetry-android/tree/main/instrumentation/okhttp/okhttp-3.0) 

I took a look at the code and I figured out I need to setup my OpenTelemetry instance at GlobalOpenTelemetry to make it works. Let me know if I miss understood it.

The first time I load the app, usually I have no problem. But If I kill the app and open it again, the SDK throws an exception.

Here is the exception.
```
java.lang.IllegalStateException: GlobalOpenTelemetry.set has already been called. GlobalOpenTelemetry.set must be called only once before any calls to GlobalOpenTelemetry.get. If you are 
using the OpenTelemetrySdk, use OpenTelemetrySdkBuilder.buildAndRegisterGlobal instead. Previous invocation set to cause of this exception.
14:12:35.926  W  	at io.opentelemetry.api.GlobalOpenTelemetry.set(GlobalOpenTelemetry.java:107)
14:12:35.927  W  	at io.opentelemetry.api.GlobalOpenTelemetry.set(GlobalOpenTelemetry.java:115)
14:12:35.927  W  	at io.opentelemetry.api.GlobalOpenTelemetry.get(GlobalOpenTelemetry.java:85)
14:12:35.927  W  	at io.opentelemetry.instrumentation.library.okhttp.v3_0.internal.OkHttp3Singletons.lambda$static$2(OkHttp3Singletons.java:43)
14:12:35.927  W  	at io.opentelemetry.instrumentation.library.okhttp.v3_0.internal.OkHttp3Singletons$$ExternalSyntheticLambda2.get(D8$$SyntheticClass:0)
14:12:35.928  W  	at io.opentelemetry.instrumentation.library.okhttp.v3_0.internal.CachedSupplier.get(CachedSupplier.java:31)
14:12:35.928  W  	at io.opentelemetry.instrumentation.library.okhttp.v3_0.internal.OkHttp3Singletons.lambda$static$5(OkHttp3Singletons.java:86)
14:12:35.928  W  	at io.opentelemetry.instrumentation.library.okhttp.v3_0.internal.OkHttp3Singletons$$ExternalSyntheticLambda5.get(D8$$SyntheticClass:0)
[...]

PR #505, opened 2024-08-01
OkHttp migration to instrumentaion API

LikeTheSalad:
Migrating the okhttp instrumentation.

This should also address this issue: #433 since now `GlobalOpenTelemetry` isn't needed.


instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/library/okhttp/v3_0/OkHttpInstrumentation.java
```diff
@@ -109,13 +114,21 @@ public static PeerServiceResolver newPeerServiceResolver() {
...
+
+    public OpenTelemetryRum getOpenTelemetryRum() {
+        return openTelemetryRum;
+    }
+
+    @Override
+    public void install(
+            @NotNull Application application, @NotNull OpenTelemetryRum openTelemetryRum) {
+        this.openTelemetryRum = openTelemetryRum;
+    }
```
breedx-splk, line 133: Hmmmm, this feels weird to me. Installing does nothing but hold onto the instance, so that we can add a get() method to this instrumentation so that the lazy singletons can get at the instance. Shouldn't the `install()` method actually help do the creational work? 

The lifecycle is further complicated / obfuscated by the lazy caching stuff in the singletons class.

Here's an idea: What if the `install()` method just passed the needed otel instance to a static factory method in the singletons class, and that is responsible for setting up the instrumenter and the dependent interceptors?

I made a demonstration PR to this branch to show you more what I'm talking about: https://github.com/LikeTheSalad/opentelemetry-android/pull/1

Note that in addition to the lifecycle being somewhat clearer, it also allows us to remove the otelRum instance, remove the lazy interceptor, and the cached supplier.

breedx-splk (commented):
This seems largely ok to me, but I want us to discuss the lifecycle a bit more before we get too far down the road and establish a pattern for other library instrumentations.

instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/library/okhttp/v3_0/OkHttpInstrumentation.java
LikeTheSalad: Got it, thanks! I took a quick look and your approach looks a lot cleaner, although I'm concerned about removing the _lazyness_ of the singletons due to the timing between the instantiation of OkHttpClient objects and the OTel rum initialization. I remember it was a key issue when using GlobalOpenTelemetry, though maybe it's no longer the case. I'd like to discuss it further in the SIG just to make sure.

LikeTheSalad: Thanks. I've applied the changes suggested [here](https://github.com/LikeTheSalad/opentelemetry-android/pull/1) taking into account what was discussed in the SIG meeting as well 👍

marandaneto (approved):

instrumentation/okhttp/okhttp-3.0/README.md
```diff
@@ -40,4 +40,17 @@ automatically.
...
+```java
+OkHttpInstrumentation instrumentation = AndroidInstrumentationLoader.getInstrumentation(OkHttpInstrumentation.class);
+
+// Call `instrumentation` setters.
+```
+
+> [!NOTE]
+> You must make sure to apply any configurations **before** initializing your OpenTelemetryRum
+> instance (i.e. calling OpenTelemetryRum.builder()...build()). Otherwise your configs won't be
+> taken into account during the RUM initialization process.
```
breedx-splk, line 56: I like these docs and am wondering how we might document this more broadly in the near future.

breedx-splk (approved):
👍🏻

instrumentation/okhttp/okhttp-3.0/README.md
LikeTheSalad: Good point, I've created this issue for it: https://github.com/open-telemetry/opentelemetry-android/issues/532

PR #507, opened 2024-08-05
Added a integer spin box that allows too choose quantity of a product…

magda-woj:
… added to cart - range from 1 to 10, like on the otel demo astronomy shop website.

### DIFF
--- a/demo-app/src/main/java/io/opentelemetry/android/demo/ui/shop/products/ProductDetails.kt
+++ b/demo-app/src/main/java/io/opentelemetry/android/demo/ui/shop/products/ProductDetails.kt
@@ -85,72 +85,93 @@ fun QuantityChooser() {
 
     Row(
         verticalAlignment = Alignment.CenterVertically,
-        modifier = Modifier.padding(vertical = 8.dp)
+        modifier = Modifier
+            .padding(vertical = 8.dp)
+            .fillMaxWidth()
+            .padding(horizontal = 16.dp)
     ) {
         Text(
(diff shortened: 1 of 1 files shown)


breedx-splk:
The number itself is oddly wide, but I think it's fine for now. Thanks!
image

breedx-splk (approved):

breedx-splk:
This was also an amusing read: https://rmirabelle.medium.com/there-is-no-material-design-spinner-for-android-3261b7c77da8 🙃

breedx-splk:
@magda-woj do you know if the demo web client version allows choosing more than 10 items?

demo-app/src/main/java/io/opentelemetry/android/demo/ui/shop/products/ProductDetails.kt
```diff
@@ -71,3 +77,80 @@ fun ProductDetails(product:Product){
...
+                }
+            }
+            DropdownMenu(
+                expanded = expanded,
+                onDismissRequest = { expanded = false }
+            ) {
+                for (i in 1..10) {
+                    DropdownMenuItem(
+                        text = {
+                            Row(
```
LikeTheSalad, line 130: nit: I think we should (whenever possible) try to split big widgets into multiple composable functions to avoid "indentation hell".

magda-woj:
@breedx-splk web client version also allows to choose from 1 to 10, that's why i did it like that here.

Issue #501, opened 2024-07-30
Run androidTests as part of CI


breedx-splk:
Right now, we have some androidTests that do not yet automatically run. We should build automation that runs these as part of our  CI builds. 

Open question: At what step(s) should we run them? PR? Main merge? Nightly? 

Open question: What devices with what android versions?

Are there pros/cons to not just having it be part of every PR build?

breedx-splk:
One options for this: https://github.com/marketplace/actions/android-emulator-runner

LikeTheSalad:
Done in #509

PR #509, opened 2024-08-06
Running android tests as part of daily checks

LikeTheSalad:
Addresses #501

### DIFF
--- a/.github/workflows/codeql-daily.yml
+++ b/.github/workflows/codeql-daily.yml
@@ -37,6 +37,16 @@ jobs:
 
       - name: Perform CodeQL analysis
         uses: github/codeql-action/analyze@v3
+      - name: Enable KVM for Android tests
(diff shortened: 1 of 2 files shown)


breedx-splk:
Oof, this adds 23 minutes to the build! Are we ok with that?

LikeTheSalad:
> Oof, this adds 23 minutes to the build! Are we ok with that?

That seems quite a long time, especially considering that we don't have many android tests right now.

I've just moved it to the daily build checks as I don't think it's necessary for now to run these on every PR, it would have been nice though, but based on the numbers we saw here I think it's not worth waiting that long unless it becomes really necessary in the future.

breedx-splk (approved):

breedx-splk:
Yeah, might be nice down the road to understand where the time is being spent. It does seem excessive.

Issue #502, opened 2024-07-30
AndroidInstrumentationRegistry should have a clearer purpose/api


breedx-splk:
This is a [follow-up from this comment](https://github.com/open-telemetry/opentelemetry-android/pull/498/files#r1697661012).

The `AndroidInstrumentationRegistryImpl` seems like it's intended to ONLY encapsulate the auto-discovered instrumentations available on the classpath. Even though `internal`, having a `register()` method is slightly confusing (it's unclear when that should/could be used). Furthermore, it appears that any interactions with the registry (`get()` or `getAll()`) will cause the lazy field to be initialized, which _will_ load instrumentation classes. Depending on the configuration, this may not be what users expect, and at worst could have side effects if instrumentations are behaving poorly during construction.

Propose removing `register()` from the registry and renaming the registry to something like `AutodetectedClasspathInstrumentations`, and only calling `get()` or `getAll()` when the configuration has `discoverInstrumenations = TRUE`.

LikeTheSalad:
Done in #516

PR #516, opened 2024-08-08
Instrumentation registry changes

LikeTheSalad:
Addresses #502


core/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentationServices.kt
```diff
@@ -5,17 +5,19 @@
...
 package io.opentelemetry.android.instrumentation
 
+import io.opentelemetry.android.internal.instrumentation.AndroidInstrumentationServicesImpl
+
 /**
- * Stores and provides all the available instrumentations.
+ * Loads and provides Java SPI services of type [AndroidInstrumentation] from the runtime classpath.
  */
-interface AndroidInstrumentationRegistry {
+interface AndroidInstrumentationServices {
```
breedx-splk, line 13: Sorry to say so, but I don't love this name. It's leaking out the fact that instrumentations are discovered via `AutoService` (hence "Services") and suggests that the things you can `get()` from it are services (but they're not, they're instrumentation). We also have the `ServiceManager`, so putting `Service` in the name implies a relationship there, where we might not want one.

Some alternative names:
* `AndroidInstrumentionDiscovery`
* `DetectedInstrumentationLoader`
* `InstrumentationFinder`
* `AndroidInstrumentationLoader`
* `AutoDiscoveredAndroidInstrumentationLoader`

My preference is for the last one. It's more verbose, but I hope it's very clear in what it does.

core/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentationServices.kt
```diff
@@ -24,26 +26,25 @@ interface AndroidInstrumentationRegistry {
...
+                instance = AndroidInstrumentationServicesImpl()
             }
             return instance!!
         }
 
+        /**
+         * Convenience method for [AndroidInstrumentationServices.getByType].
+         */
+        @JvmStatic
+        fun <T : AndroidInstrumentation> getService(type: Class<out T>): T? {
```
breedx-splk, line 44: Same concerns here about `Service`. Would prefer to call this `getInstrumentation()`.

breedx-splk (commented):
Thanks @LikeTheSalad , I appreciate where this is going. I am hung up on a couple of names, but otherwise seems good.

core/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentationServices.kt
LikeTheSalad: I'm always looking for name ideas, as that's the part I struggle the most with, so thanks for your suggestions!

I went with `AndroidInstrumentationLoader` because I think it's the most concise. Since users might need to access this type to configure some instrumentations I think it's better to keep the name as short as possible.

core/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentationServices.kt
LikeTheSalad: Thanks, it's updated now.

breedx-splk (approved):
Thanks!

Issue #506, opened 2024-08-01
Unable to get traces exported out of the demo app


MrHadiSatrio:
Hello! First of all, thank you for publishing this project. It will undoubtedly help many Android developers improve their application's health and performance.

Unfortunately, I'm unable to export traces from the demo app, even though I have Jaeger and the collector running on my machine. Logcat consistently reports the following error:

```
Failed to export logs. The request could not be executed. Full error message: unexpected end of stream on http://10.0.2.2:4318/... 
java.io.IOException: unexpected end of stream on http://10.0.2.2:4318/...
	at okhttp3.internal.http1.Http1ExchangeCodec.readResponseHeaders(Http1ExchangeCodec.kt:210)
	(...)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:637)
	at java.lang.Thread.run(Thread.java:1012)
Caused by: java.io.EOFException: \n not found: limit=0 content=…
	at okio.RealBufferedSource.readUtf8LineStrict(RealBufferedSource.kt:335)
	(...)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:637) 
	at java.lang.Thread.run(Thread.java:1012) 
```

I've verified using `nc` that port `4318` and `4317` is accessible. Did I miss any configuration steps for Jaeger or the collector that might be causing this issue?

LikeTheSalad:
Hi! It seems to be an issue with Jaeger, because I was also having this issue until I locally rolled back Jaeger's version from `1.59` to `1.58` [here](https://github.com/open-telemetry/opentelemetry-android/blob/059a829df4de2286b62709f72da5d48b9834c3bf/demo-app/compose.yaml#L12). Said version was [recently set](https://github.com/open-telemetry/opentelemetry-android/pull/473) to `1.59`.

Maybe something about Jaeger's config has changed recently, do you know anything about this @breedx-splk ?

breedx-splk:
Yeah, this is called out in the release notes: https://github.com/jaegertracing/jaeger/releases/tag/v1.59.0

I opened #508 to fix this....just remember to `docker compose build` to get the update.

julian-sanz-code-seat:
Hi @LikeTheSalad and @breedx-splk! why do you refer to jaeger when @MrHadiSatrio are talking about demo-app and sending traces to collector??
[...]

PR #517, opened 2024-08-08
Configure OTLP receiver to listen on 0.0.0.0

MrHadiSatrio:
This change configures the OTLP receiver in the demo application to listen on 0.0.0.0. Previously, this was part of the default config but changed since v0.104.0.

Related to #506.


breedx-splk (approved):
Thanks!

PR #518, opened 2024-08-08
Demo app cart 

magda-woj:
Adding product of specified quantity to cart and emptying cart works. QuantityChooser component moved to a separate file.

### DIFF
--- a/demo-app/src/main/java/io/opentelemetry/android/demo/ui/shop/components/QuantityChooser.kt
+++ b/demo-app/src/main/java/io/opentelemetry/android/demo/ui/shop/components/QuantityChooser.kt
@@ -88,27 +88,37 @@ fun QuantityDropdown(
         onDismissRequest = onDismissRequest
     ) {
         for (i in 1..10) {
-            DropdownMenuItem(
-                text = {
-                    Row(
-                        verticalAlignment = Alignment.CenterVertically
-                    ) {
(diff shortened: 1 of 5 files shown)


demo-app/src/main/java/io/opentelemetry/android/demo/ui/shop/components/QuantityChooser.kt
```diff
@@ -0,0 +1,114 @@
...
+                            )
+                            Spacer(modifier = Modifier.width(8.dp))
+                        }
+                        Text(
+                            text = i.toString(),
+                        )
+                    }
+                },
+                onClick = { onQuantitySelected(i) }
+            )
```
breedx-splk, line 111: I like that this was broken out into its own file. I think @LikeTheSalad had a comment previously about indent levels, and there's still some room to improve this. If you make a new topmost item called something like `QuantityDropDownItem` (whose constructor takes `i` and a boolean `isSelected`) then you can basically move this whole block into that new composable and save 1 or 2 indent levels. The for loop then looks something like
```
for (i in 1..10) {
    QuantityDropDownItem(i, i == selectedQuantity)
}
```

breedx-splk (approved):
One small recommendation about reducing indent, but otherwise looking good to me. Thanks for the help!

LikeTheSalad (approved):
Thanks!

PR #529, opened 2024-08-12
remove proguard config for InstrumenationUtil

SylvainJuge:
With SDK 1.41.0 `io.opentelemetry.exporter.internal.InstrumentationUtil` is now deprecated and replaced with `io.opentelemetry.api.internal.InstrumentationUtil` directly in the API.

Update to the 1.41.0 SDK has been done in https://github.com/open-telemetry/opentelemetry-android/pull/521 and is now merged.

As a consequence, the proguard configuration is no more needed for this class.

@LikeTheSalad can you tell me if there is anything related to `InstrumentationUtil` usage that needs to be updated here ?

### DIFF
--- a/instrumentation/okhttp/okhttp-3.0/library/consumer-rules.pro
+++ b/instrumentation/okhttp/okhttp-3.0/library/consumer-rules.pro
@@ -1,1 +1,0 @@
--keep class io.opentelemetry.exporter.internal.InstrumentationUtil { *; }


LikeTheSalad:
Hi! Thank you for bringing this up, the proguard rule removed in this PR has 2 dependencies, one is the otel java API, which is now fine, and the other one is the otel java instrumentation API. So we should technically be able to remove this rule after [this other PR](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/11985) is merged and a new version of the instrumentation API is released and used here. This is because the rule `keep class io.opentelemetry.exporter.internal.InstrumentationUtil` ensures that said class name isn't changed during obfuscation, so that the reflection [used here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/2af49b4c044db526dd0741eccfe0bdc07baeda71/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SpanSuppressors.java#L100) doesn't get affected.

**However**, it seems like we're [still planning to use reflection](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/11985/files#diff-3b7ccbad06d354ba3f49a3d9076c73105af0a6e141c4ec69bb231adac15db9ccR100) in the java instrumentation lib even after the otel java changes. If that PR gets merged as is, then instead of removing the rule `keep class io.opentelemetry.exporter.internal.InstrumentationUtil`, we would instead have to change the rule to target the new reflection, i.e. `io.opentelemetry.api.internal.InstrumentationUtil`.

LikeTheSalad:
I've added a comment [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/11985) because I think the new reflection might affect this functionality for Android unless we modify our rule.

LikeTheSalad:
> Usage of reflection was removed from [open-telemetry/opentelemetry-java-instrumentation#11985](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/11985) and it's now merged, so I think we should be good to merge it as-is.

Yep, we can merge this change as-is since they've removed reflection. I just think we should wait for a new version of the instrumentation API to be created with those changes and we should be good to go then.

LikeTheSalad (approved):

PR #531, opened 2024-08-13
Updating changelog for the next release

LikeTheSalad:


breedx-splk (approved):
I think there should also be a note at the top that points out the considerable modularization that took place since the last release, and also maybe we should enumerate the separate instrumentation modules. Just a thought.

LikeTheSalad:
> I think there should also be a note at the top that points out the considerable modularization that took place since the last release, and also maybe we should enumerate the separate instrumentation modules. Just a thought.

Good point, I've added some more details about those changes 👍

PR #545, opened 2024-08-19
Checkout info

magda-woj:
Demo-app: implemented a buyer's details screen after clicking checkout button in cart.

### DIFF
--- a/demo-app/src/main/java/io/opentelemetry/android/demo/ui/shop/cart/CheckoutInfo.kt
+++ b/demo-app/src/main/java/io/opentelemetry/android/demo/ui/shop/cart/CheckoutInfo.kt
@@ -12,7 +12,11 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.platform.LocalFocusManager
 import androidx.compose.ui.text.input.KeyboardType
+import androidx.compose.ui.text.input.PasswordVisualTransformation
+import androidx.compose.ui.text.input.VisualTransformation
 import androidx.compose.ui.unit.dp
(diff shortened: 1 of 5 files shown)


breedx-splk:
Nice! So after clicking on "Checkout" in the cart, the user is taken to this screen:

image

demo-app/src/main/java/io/opentelemetry/android/demo/ui/shop/cart/CheckoutInfo.kt
```diff
@@ -0,0 +1,150 @@
...
+    }
+}
+
+fun <T : Any> T.allFieldsNotBlank(): Boolean {
+    return this::class.memberProperties.all { property ->
+        property.isAccessible = true
+        val value = property.call(this)
+        value is String && value.isNotBlank()
+    }
+}
```
breedx-splk, line 148: I think we should avoid this reflection and instead build explicit methods on each of the data classes. Something like this:

```
fun isComplete(): Boolean {
    return arrayOf(email, streetAddress, zipCode, city, state, country)
        .all { it.isNotBlank() }
}
```

The reflection feels a bit forced, and it modifies the permission/accessibility in an unwanted way. It also limits the ability add an optional field (like address line 2 or whatever) down the road.

marandaneto: also if minified is enabled, the methods might be removed unless its added in the proguard file

LikeTheSalad (approved):

breedx-splk (approved):
Solid, thanks!

PR #554, opened 2024-08-23
Demo app: Recommendations

magda-woj:
A simple recommendation service based on the web demo was implemented: recommended products appear on the product details and cart screen. Additionally, a "go back" arrow button was added to the product details and the checkout-info screen that allows to successfully navigate up.
Zrzut ekranu 2024-08-23 o 16 33 43

### DIFF
--- a/demo-app/src/main/java/io/opentelemetry/android/demo/clients/RecommendationService.kt
+++ b/demo-app/src/main/java/io/opentelemetry/android/demo/clients/RecommendationService.kt
@@ -8,13 +8,6 @@ class RecommendationService(
     private val cartViewModel: CartViewModel
 ) {
 
-    private fun getAllNonCartProducts(): List<Product>{
-        val allProducts = productCatalogClient.get()
-        val cartItems = cartViewModel.cartItems.value
-
-        return allProducts.filter { product -> cartItems.none { it.product.id == product.id } }
-    }
-
     fun getRecommendedProducts(currentProduct: Product, numberOfProducts: Int = 4): List<Product> {
         return getAllNonCartProducts().filter { it.id != currentProduct.id }
         .shuffled().take(numberOfProducts)
@@ -22,6 +15,13 @@ class RecommendationService(
 
     fun getRecommendedProducts(numberOfProducts: Int = 4): List<Product> {
         return getAllNonCartProducts().shuffled().take(numberOfProducts)
+    }
+
+    private fun getAllNonCartProducts(): List<Product>{
+        val allProducts = productCatalogClient.get()
+        val cartItems = cartViewModel.cartItems.value
+
+        return allProducts.filter { product -> cartItems.none { it.product.id == product.id } }
     }
 }
 
--- a/demo-app/src/main/java/io/opentelemetry/android/demo/clients/RecommendationService.kt
+++ b/demo-app/src/main/java/io/opentelemetry/android/demo/clients/RecommendationService.kt
@@ -8,22 +8,20 @@ class RecommendationService(
     private val cartViewModel: CartViewModel
 ) {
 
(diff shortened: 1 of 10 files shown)


demo-app/src/main/java/io/opentelemetry/android/demo/clients/RecommendationService.kt
```diff
@@ -0,0 +1,29 @@
...
+    private val cartViewModel: CartViewModel
+) {
+
+    fun getRecommendedProducts(currentProduct: Product): List<Product> {
+        val allProducts = productCatalogClient.get()
+        val cartItems = cartViewModel.cartItems.value
+
+        return allProducts.filter { product ->
+            product.id != currentProduct.id && cartItems.none { it.product.id == product.id }
+        }.shuffled().take(4)
```
breedx-splk, line 17: This 4 is kinda hidden and makes the reader wonder how many might be returned. What do you think about making it a parameter with a default value?

demo-app/src/main/java/io/opentelemetry/android/demo/clients/RecommendationService.kt
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    fun getRecommendedProducts(): List<Product> {
+        val allProducts = productCatalogClient.get()
+        val cartItems = cartViewModel.cartItems.value
+
+        return allProducts.filter { product ->
+            cartItems.none { it.product.id == product.id }
+        }.shuffled().take(4)
```
breedx-splk, line 26: Same comment about the 4 being buried here.

demo-app/src/main/java/io/opentelemetry/android/demo/clients/RecommendationService.kt
```diff
@@ -0,0 +1,29 @@
...
+    }
+}
```
breedx-splk, line 28: I think there's a small opportunity to improve this by factoring out a method called something like `getAllNonCartProducts()`. Just like it says in the name, that method would return all products from the catalog that are not currently in the cart. Both existing methods here can then delegate to that new method.

demo-app/src/main/java/io/opentelemetry/android/demo/ui/shop/products/ProductList.kt
```diff
@@ -15,7 +15,7 @@ fun ProductList(products: List<Product>, onProductClick: (String) -> Unit) {
     LazyColumn(modifier = Modifier.fillMaxSize()) {
         items(products.size) { index ->
             Row {
-                ProductCard(products[index], onClick = onProductClick)
+                ProductCard(products[index], onProductClick= onProductClick)
```
breedx-splk, line 18: formatting nitpick:
```suggestion
                ProductCard(products[index], onProductClick = onProductClick)
```

breedx-splk (commented):
Looking good. Just had a few small comments.

demo-app/src/main/java/io/opentelemetry/android/demo/clients/RecommendationService.kt
```diff
@@ -0,0 +1,27 @@
...
+    private val productCatalogClient: ProductCatalogClient,
+    private val cartViewModel: CartViewModel
+) {
+
+    private fun getAllNonCartProducts(): List<Product>{
+        val allProducts = productCatalogClient.get()
+        val cartItems = cartViewModel.cartItems.value
+
+        return allProducts.filter { product -> cartItems.none { it.product.id == product.id } }
+    }
```
breedx-splk, line 16: Another small nit -- call flow should generally go from top to bottom, so this method should be moved below the two methods where it is called. It's kinda silly, but it helps the reader to follow a vertical flow, rather than having to scroll or move their eyes up and down to follow.

PR #569, opened 2024-08-30
Make HttpURLConnection connection inactivity timeout configurable and add test for harvester code

surbhiia:
Made connection inactivity timeout ms configurable and added a test case for harvester code. 

Next items for future PRs:
- Discuss the requirement for custom spanName/status function configuration - is it needed per instrumentation or at the sdk level works. 
- Migrate HttpURL instrumentation to utilize AndroidInstrumentation API for install and configuration.


instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlInstrumentationConfig.java
```diff
@@ -119,20 +119,32 @@ public static boolean emitExperimentalHttpClientMetrics() {
...
+     * connection was last active. It is used by the idle connection harvester thread to find idle
+     * connections that should be reported. To schedule the idle connection harvester follow
+     * instructions defined in the <a
+     * href="https://github.com/open-telemetry/opentelemetry-android/blob/96ea4aa9fe709838a91811deeb85b0f1baceb8cd/instrumentation/httpurlconnection/README.md#scheduling-harvester-thread">
+     * README</a>.
+     *
+     * <p>Default value: "10000"
+     */
+    public static void setConnectionInactivityTimeoutMs(long timeoutMs) {
+        HttpUrlInstrumentationConfig.connectionInactivityTimeoutMs = timeoutMs;
```
marandaneto, line 133: should we check if the value here is at least positive and non zero?

surbhiia: It does make sense for it to be positive for production workflows. I can throw an IllegalArgumentException if it's negative or zero. Does that make sense? 

We will need another test method than similar to this where we allow for negative values:
 public static void setConnectionInactivityTimeoutMsForTesting(long timeoutMs) {
        HttpUrlInstrumentationConfig.connectionInactivityTimeoutMs = timeoutMs;
}

If the customer wants to test the harvester with this they can as well so no issue in exposing this to the customer?

breedx-splk: Conventionally, zero has been used for a long time to mean "no timeout, wait forever". Not sure how much we need or care about that here. Just throwing an exception for negative seems reasonable to me.

instrumentation/httpurlconnection/README.md
```diff
@@ -62,12 +62,15 @@ byteBuddy("io.opentelemetry.android:httpurlconnection-agent:AUTO_HTTP_URL_INSTRU
...
 Executors.newSingleThreadScheduledExecutor().scheduleWithFixedDelay(HttpUrlInstrumentationConfig.getReportIdleConnectionRunnable(), 0, HttpUrlInstrumentationConfig.getReportIdleConnectionInterval(), TimeUnit.MILLISECONDS);
 ```
 
-`HttpUrlInstrumentationConfig.getReportIdleConnectionRunnable()` is the API to get the runnable. `HttpUrlInstrumentationConfig.getReportIdleConnectionInterval()` is the API to get the fixed interval (10s) in milli seconds.
+- `HttpUrlInstrumentationConfig.getReportIdleConnectionRunnable()` is the API to get the runnable.
+- By default, connections inactive for 10000ms are reported. To optionally change the connection inactivity timeout call `HttpUrlInstrumentationConfig.setConnectionInactivityTimeoutMs(customTimeoutValue)` API as shown in above code snippet.
+- It is efficient to run the harvester thread at the same time interval as the connection inactivity timeout used to identify the connections to be reported. `HttpUrlInstrumentationConfig.getReportIdleConnectionInterval()` is the API to get the same connection inactivity timeout interval (milli seconds) you have configured or the default value of 10000ms if not configured.
```
breedx-splk, line 73:
```suggestion
- It is efficient to run the harvester thread at the same time interval as the connection inactivity timeout used to identify the connections to be reported. `HttpUrlInstrumentationConfig.getReportIdleConnectionInterval()` is the API to get the same connection inactivity timeout interval (milliseconds) you have configured or the default value of 10000ms if not configured.
```

breedx-splk (approved):
Just a small nit, but looks good to me otherwise.

LikeTheSalad (approved):

instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlInstrumentationConfig.java
surbhiia: Added an exception for negative values! Also added a test only function to allow negative values for test, so wait times are not needed in test. 

Added some good documentation to prevent misuse. We can also add @VisibleForTesting annotation on the test function. Right now held off on that as I think we're not using the annotation beyond documentation purposes and we need to introduce additional dependency in this module for just one function. Let me know if it's better to have that annotation instead.

PR #570, opened 2024-09-03
Demo app: activity with fragments

magda-woj:
Developed a new activity (AboutActivity) to showcase Fragment Lifecycle instrumentation. The activity contains 2 fragments: AppFeaturesFragment (a little bit more complicated with a Recycler View) and AboutOpenTelemetryFragment (for now very simple, just with some text written into the layout file) and a bottom navigation to navigate between them and return to the main activity.

### DIFF
--- a/demo-app/src/main/AndroidManifest.xml
+++ b/demo-app/src/main/AndroidManifest.xml
@@ -23,6 +23,11 @@
             android:label="@string/title_activity_astronomy_shop"
             />
         <activity
+            android:name=".about.AboutActivity"
+            android:exported="true"
+            android:label="AboutActivity"
+            />
+        <activity
             android:name=".MainActivity"
             android:exported="true"
             android:label="@string/app_name"
--- a/demo-app/src/main/AndroidManifest.xml
+++ b/demo-app/src/main/AndroidManifest.xml
@@ -20,12 +20,12 @@
         <activity
             android:name=".shop.ui.AstronomyShopActivity"
             android:exported="true"
-            android:label="@string/title_activity_astronomy_shop"
+            android:label="Astronomy Shop"
             />
         <activity
             android:name=".about.AboutActivity"
             android:exported="true"
-            android:label="AboutActivity"
+            android:label="About OpenTelemetry Android"
             />
         <activity
             android:name=".MainActivity"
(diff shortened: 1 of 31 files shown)


demo-app/src/main/java/io/opentelemetry/android/demo/about/AppFeaturesFragment.kt
```diff
@@ -0,0 +1,32 @@
...
+import androidx.recyclerview.widget.LinearLayoutManager
+import androidx.recyclerview.widget.RecyclerView
+import io.opentelemetry.android.demo.R
+
+class AppFeaturesFragment : Fragment() {
+
+    private lateinit var recyclerView: RecyclerView
+    private lateinit var featureAdapter: FeatureAdapter
+
+    private val features = getFeatureList()
```
marandaneto, line 17: if this is not reused anywhere, they don't need to be global vars

marandaneto (approved):

demo-app/src/main/AndroidManifest.xml
```diff
@@ -18,10 +18,15 @@
...
         <activity
-            android:name=".ui.shop.AstronomyShopActivity"
+            android:name=".shop.ui.AstronomyShopActivity"
             android:exported="true"
             android:label="@string/title_activity_astronomy_shop"
             />
+        <activity
+            android:name=".about.AboutActivity"
+            android:exported="true"
+            android:label="AboutActivity"
```
breedx-splk, line 28:
```suggestion
            android:label="About OpenTelemetry Android"
```

demo-app/src/main/java/io/opentelemetry/android/demo/MainActivity.kt
```diff
@@ -73,9 +74,13 @@ class MainActivity : ComponentActivity() {
                             painterResource(id = R.drawable.otel_icon),
                         )
                         val context = LocalContext.current
-                        OpenStoreButton(text = "Click to begin", onClick = {
+                        OpenStoreButton(text = "Go shopping", onClick = {
                             context.startActivity(Intent(this@MainActivity, AstronomyShopActivity::class.java))
                         })
+                        OpenStoreButton(text = "Learn more", onClick = {
```
breedx-splk, line 80: It's clear now that this button isn't just for opening the store....maybe it can be renamed to something more generic, like `LauncherButton` or something.

LikeTheSalad (approved):

Issue #560, opened 2024-08-27
"AppStart" instrumentation span is never started.


beekhc:
I'm using `ActivityLifecycleInstrumentation`, but never getting `"AppStart"` events. It's possible that I'm using the instrumentation wrong, but looking through the code, it looks like nothing ever calls `AppStartupTimer#start(Tracer)`. It's a public method, but `startupTimer` is a private member in `ActivityLifecycleInstrumentation` and there's no setter, so I don't see any way to call it manually, either.

To reproduce the issue, use the `ActivityLifecycleInstrumentation` in an example app and observe that there are no `"AppStart"` events. Alternatively, set a breakpoint in `AppStartupTimer#start(Tracer)` and run in debug mode and observe that the function is never called.

I'd be happy to contribute the PR implementing the call, but it's not obvious to me where it should be called. My best guess is that it should be called in `ActivityLifecycleInstrumentation#install`, since that's the earliest point we can get a tracer. Admittedly, that would miss out on the time spent before OTel was initialized.

LikeTheSalad:
Hi @beekhc 

Thank you for taking the time! I just confirmed that we're not calling `AppStartupTimer#start(Tracer)` which is wrong, so I agree that we should call it in `ActivityLifecycleInstrumentation#install` since that's the earliest we can get a tracer for it. If you're interested in creating a PR for it it would be very helpful for us, so I'd say go ahead! :)

What you mentioned about the time `ActivityLifecycleInstrumentation#install` is currently called not being the earliest we should start the timer is a very valid point, and I also believe this event shouldn't be a span as it is now, so I think we need to work on a refactor in the future to do it properly, i.e. starting the timer as soon as possible and sending an event instead of a span. For now, our priority is mostly focusing on the API foundations so that we can mark it stable as soon as possible, and I believe that once that's done we can start focusing on adding more features and enhancing the existing ones. So in the meantime, I think it's fine to keep the span and start it in `ActivityLifecycleInstrumentation#install`.

I'll loop @breedx-splk in just in case I missed something.

breedx-splk:
I agree with what @LikeTheSalad said, and I think this is probably a bug/shortcoming....however @beekhc you didn't specify if you were attempting to use the instrumentation alone (without the agent) or with the agent? While we sure may want the instrumentation to be usable alone, our focus right now is definitely to get it working first within the context of the agent. 👍🏻

beekhc:
[...]

PR #571, opened 2024-09-03
add SessionIdEventSender and wire up into lifecycle

breedx-splk:
This is a follow-up from #564. It reintroduces the `SessionIdEventSender` class and wires it up at `build()` time.

I know there's some hesitation to put "opinionated" session behavior into the core module. I'm open to discussion and ideas here, but I do think that:

* It's important that the initial Session be created as early as `OpenTelemetryRum` can allow. Right now that happens with the builder. This is important for several reasons, not the least of which is that we want telemetry to be associated with the current session.
* It's important to generate the `session.start` event as close to the session actually being started/created.
* The `session.start` event can't be created until after the otel sdk is created.


core/src/main/java/io/opentelemetry/android/SdkPreconfiguredRumBuilder.kt
```diff
@@ -50,6 +51,14 @@ class SdkPreconfiguredRumBuilder
             // might turn off/on additional telemetry depending on whether the app is active or not
             appLifecycleService.registerListener(timeoutHandler)
 
+            val eventLogger =
+                SdkEventLoggerProvider.create(sdk.logsBridge)
+                    .get(OpenTelemetryRum::class.java.simpleName)
+
+            sessionManager.addObserver(SessionIdEventSender(eventLogger))
+            // After addObserver(), we call getSessionId() to trigger a session.start event
```
LikeTheSalad, line 59: This seems a bit forced. I'm guessing it will happen naturally when the first telemetry event is recorded, is there a reason why we must start the session earlier?

marandaneto: probably needed if we want a session to be active as early as possible.
eg SDK is initiated during the Application#onCreate class.
If the session starts (automatically) when the OS gets a notification that the app is in the foreground (using androidx.lifecycle), that might be too late because the whole Application#onCreate code has been run, maybe some telemetry data was collected, but there was not an active session yet, in this case, makes sense that the SDK starts as session ASAP.

core/src/main/java/io/opentelemetry/android/SdkPreconfiguredRumBuilder.kt
```diff
@@ -50,6 +51,14 @@ class SdkPreconfiguredRumBuilder
...
+            sessionManager.getSessionId()
```
marandaneto, line 60: should the `sessionManager` have a `start` method instead for this use case so it's semantically correct?

core/src/main/java/io/opentelemetry/android/SdkPreconfiguredRumBuilder.kt
LikeTheSalad: Also, if we must do it earlier than the first telemetry event, which I guess would depend on the definition of a session, for example, if the definition ends up telling something around the lines of "the session starts when the app is launched", then I think the initialization of the SDK is too late to do so as well. It reminds me of the app launch timer that was recently mentioned [here](https://github.com/open-telemetry/opentelemetry-android/issues/560), where we might have to come up with some other mechanisms to track the start time temporarily until the Otel instance is ready.

LikeTheSalad: > maybe some telemetry data was collected, but there was not an active session yet

The session is queried and applied in the processors, so it will start as soon as the first telemetry data is collected. We do have an exception though where we collect events before the SDK is finished initializing which is [here](https://github.com/open-telemetry/opentelemetry-android/blob/5a230fc59e85b244ff253c8faabe023c353c9bd4/instrumentation/startup/src/main/java/io/opentelemetry/android/instrumentation/startup/SdkInitializationEvents.kt#L22) in the startup instrumentation where we store events in memory until those can be sent after the SDK is ready. It's probably worth considering a similar solution for this case in case we really need to trigger the event before the first telemetry data is collected.

bidetofevil: I too think the first session should start as early as possible, but if the OTel SDK isn't started, can we even do anything with it? If we simply initialize it lazily when an instance of the processor is created, it should ensure that every piece of OTel telemetry recorded will have a session ID?

bidetofevil:
Seems like we should have a list of features/instrumentations defined as "must haves" that cannot be opt out of, and another set that are default on, and another that is default off. Do we have a list like that so we can determine where each feature belongs to, so we can better figure out where initialization and configuration should be happening?

core/src/main/java/io/opentelemetry/android/SdkPreconfiguredRumBuilder.kt
LikeTheSalad: > If we simply initialize it lazily when an instance of the processor is created, it should ensure that every piece of OTel telemetry recorded will have a session ID?

The session is started when `getSessionId()` is called for the first time. It happens as soon as the [first span is started](https://github.com/open-telemetry/opentelemetry-android/blob/56dbb88b5e07235eb14c30ea521788070b81fe19/core/src/main/java/io/opentelemetry/android/SessionIdSpanAppender.java#L26) or the [first log is emitted](https://github.com/open-telemetry/opentelemetry-android/blob/56dbb88b5e07235eb14c30ea521788070b81fe19/core/src/main/java/io/opentelemetry/android/SessionIdLogRecordAppender.kt#L20), so there's no way that a signal is created without a session id. So it seems to me that we don't need to force it unless we must create a session before any telemetry is sent, like when the app is created for example, in which case this wouldn't be a good place to do so either.

I'd prefer to avoid adding behaviors to `core` unless it's a must for everyone, so it seems to me that the agent would be a nice place for this kind of behavior. I'm just back from PTO so I was planning to continue the work on [this prototype](https://github.com/open-telemetry/opentelemetry-android/pull/624) as it seems like the overall idea has gotten thumbs-up, probably during the same work or right after I can add this session event sender behavior to the agent, wdyt @breedx-splk ?

breedx-splk:
To address the concerns around "Not every vendor may want these events" critique, I have moved the session id event generation into an instrumentation module. This module is NOT included (yet) by the `android-agent` module, so by default the events will not be generated. 

In order to make this work, the instrumentation needs to access the `SessionManager` so that it can register its observer. The `SessionManager` was previously an internal concrete class, so I renamed it to `SessionManagerImpl` and extracted the tag interface `SessionManager` which now just extends `SessionProvider` and `SessionPublisher`...and this was added to the `InstallationContext` for any future instrumentations that might need to muck with sessions. It becomes apparent now too that the session package could even likely be its own module (not sure how necessary tho).

I did include the session event instrumentation explicitly in the demo app, so that we could verify that events can still be generated. I think this also raises an interesting point about being able to disable instrumentations, either programmatically or via config, just like we do with java auto-instrumentation. 

@open-telemetry/android-approvers  @open-telemetry/android-maintainers Please have another look.

breedx-splk:
Ok sorry wait, this has gotten too spread out. I'm going to move this back to draft and try and backfill with a series of smaller commits.

PR #582, opened 2024-09-10
Group java core dependency updates

breedx-splk:
We currently have a state where renovate can push version updates for some of the core repos separately from the same `-alpha` artifacts. This can (and has) introduced breakage, so I think we should look at grouping the core versions with their alpha counterparts.


bidetofevil (approved):
LGTM

LikeTheSalad (approved):

PR #584, opened 2024-09-11
Demo App crash

### LINKED ISSUE #347: [demo] Add a way to crash the application
Not sure where to put it, but could be nice to have a menu item or a button on a fragment in the shop activity that causes the app to crash. This allows us to manually test the crash detection instrumentation.

For starters, we can just pick whatever crash mechanism we like and if we want to try other ways in the future we can do that.

magda-woj:
May resolve #347.
Added a way to crash the demo app through unhandled exceptions and documented it. To invoke it, try adding to cart exactly 10 of the first product on the list. The way in which `multiThreadCrashing` function works is copied from the [splunk-otel-android sample app](https://github.com/signalfx/splunk-otel-android/tree/main/sample-app) function of the same name.

### DIFF
--- a/demo-app/README.md
+++ b/demo-app/README.md
@@ -34,6 +34,11 @@ The OpenTelemetry Android Demo App currently supports the following features:
   - Provides detailed insights into each lifecycle phase.
   - Can be observed in the "About OpenTelemetry Android" activity, entered via "Learn more" on the main screen.
 
+* Crash Reporting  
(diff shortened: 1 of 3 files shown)


breedx-splk (approved):
Nice! I had one small wording suggestion, otherwise looks great. Thanks.

marandaneto (approved):

LikeTheSalad (approved):

PR #586, opened 2024-09-11
Explicitly enumerate core packages

breedx-splk:
The attempt in #582 did not work as expected, resulting in horrible renovate ideas like #585. 

So now we just explicitly list out all of the core packages, group them together, and tell renovate to not ignore unstable, which I think means to keep the alphas (and don't favor non-alphas).


breedx-splk:
I also noticed that our contrib/disk-buffering dependency was 4 months behind as well.

marandaneto (approved):

Issue #590, opened 2024-09-12
"Paused" span for a fragment incorrectly becomes parent of future spans after exiting activity


magda-woj:
Observed this using the demo app, where we recently implemented an activity (`AboutActivity.kt`) with fragments. Upon exiting the activity, the `Paused` span for the last fragment (with `fragmentPaused` and `fragmentStopped` events), which is a child of the activity's `Paused` span, incorrectly becomes the parent for unrelated future spans not connected to the fragment or activity.

Below is a screenshot from the Jaeger UI with the relevant spans expanded:
<img width="1585" alt="Zrzut ekranu 2024-09-12 o 09 46 09" src="https://github.com/user-attachments/assets/7056ad1b-ea08-4aa4-8ca9-6bc774090c2a">
In the screenshot, the `Created`, `Paused` etc spans under the expanded ones are for different activities, and `logo.clicked` is a custom span emitted after clicking a button on the main screen.
To reproduce this behaviour:
1. In the demo app, click "Learn more" on the main screen.
2. Exit the activity with fragments by selecting "Exit" from the bottom navigation bar.
3. Click around the app to generate new spans.

It seems to me that simply ending the `Paused` span and creating a new `Stopped` span for `fragmentStopped` resolves this issue and i'll submit a pr for this.

PR #591, opened 2024-09-12
Ending "Paused" span for a fragment.

magda-woj:
This may be a simple fix for the #590 issue I just submitted. However I don't know the reasoning for the current way, so I don't know if it creates new problems.

Currently `onFragmentPaused` and `onFragmentStopped` add events to the same span, which is ended in `onFragmentStopped`. I propose to end an active span in `onFragmentPaused` and create a new one in `onFragmentStopped`. With those changes `Paused` for About activity is a parent of `Paused` for the last fragment and analogously with `Stopped` (so 2 traces with 2 spans each) . Unrelated spans get their own traces in between and after.

In addition to fixing #590 this is consistent with how in Activity Lifecycle Instrumentation `Paused` and `Stopped` are separate spans.


magda-woj:
(I've been meaning to add this before but forgot) I haven't changed the tests for this because I wanted to wait for a review of the idea not to to do it needlessly, as I haven't dug into our unit tests. That's why it doesn't pass checks.

LikeTheSalad (approved):
Thanks!

marandaneto (approved):

PR #596, opened 2024-09-13
Expose additional disk buffering configuration

breedx-splk:
Because one size can't necessarily fit all, we need to expose additional settings for the disk buffering feature.

* maxFileAgeForWriteMillis
* minFileAgeForReadMillis
* maxFileAgeForReadMillis

These map over directly to the same configurations in the contrib module, and use the same sane defaults as well.


marandaneto (approved):

LikeTheSalad (approved):
🚀

PR #598, opened 2024-09-16
Demo app: Order confirmation

magda-woj:
Created `CheckoutConfirmation.kt` with an order confirmation screen, basing it off the web demo. User sees this screen after clicking "Place Order" on the CheckoutInfo screen. The cart is also cleared in the process.
image
Additionally, I made the form for shipping and payment information pre-filled, to be able to go to that screen without filling them in every time.

### DIFF
--- a/demo-app/src/main/java/io/opentelemetry/android/demo/shop/ui/cart/CheckoutConfirmation.kt
+++ b/demo-app/src/main/java/io/opentelemetry/android/demo/shop/ui/cart/CheckoutConfirmation.kt
@@ -35,6 +35,7 @@ fun CheckoutConfirmationScreen(
             lifecycleOwner.lifecycle.removeObserver(observer)
         }
     }
+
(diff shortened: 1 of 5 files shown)


demo-app/src/main/java/io/opentelemetry/android/demo/shop/ui/cart/CheckoutConfirmation.kt
```diff
@@ -0,0 +1,141 @@
...
+    val shippingInfo = checkoutInfoViewModel.shippingInfo
+
+    Column(
+        modifier = Modifier
+            .fillMaxSize()
+            .padding(16.dp)
+            .verticalScroll(rememberScrollState())
+    ) {
+        Text(
+            text = "Your order is complete!",
```
breedx-splk, line 48: Ideally these would be translatable strings, but for now it's perfectly fine to just have them coded.

breedx-splk (approved):
Solid. Thanks!

PR #604, opened 2024-09-18
Demo app: don't start the main activity again, just finish the current one

magda-woj:
Basically the title. I don't think we should be starting the main activity again, when clicking the `Exit` button on another activity. Just finishing the current activity seems to work fine. The motivation to look into this were multiple `App start` spans emitted by the app. After those changes it doesn't happen, there are `Restarted` spans for the main activity instead. The behaviour after the change is identical to the one using the phone's build-in "go back" button while on the first screen of a non-main activity.

### DIFF
--- a/demo-app/src/main/java/io/opentelemetry/android/demo/about/AboutActivity.kt
+++ b/demo-app/src/main/java/io/opentelemetry/android/demo/about/AboutActivity.kt
@@ -1,10 +1,8 @@
 package io.opentelemetry.android.demo.about
 
-import android.content.Intent
 import android.os.Bundle
 import androidx.appcompat.app.AppCompatActivity
 import com.google.android.material.bottomnavigation.BottomNavigationView
-import io.opentelemetry.android.demo.MainActivity
 import io.opentelemetry.android.demo.R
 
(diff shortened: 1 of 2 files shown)


LikeTheSalad:
> multiple `App start` spans emitted by the app

I agree we shouldn't relaunch the main activity, so thanks for these changes! However, I was also curious why doing so would trigger a new app launch event. I found [this part](https://github.com/open-telemetry/opentelemetry-android/blob/d774f72761cc81d936e332e84934d97bd7d3054c/instrumentation/activity/src/main/java/io/opentelemetry/android/instrumentation/activity/ActivityTracer.java#L85) of the code where I think it's happening and it looks like the logic is to create the event every time the "initial activity" starts, which doesn't seem like a good idea to trigger it (there are also some comments there acknowledging so) so maybe we should do some work in the future to properly cover the "hot" app start case.

LikeTheSalad (approved):

breedx-splk (approved):

breedx-splk:
I opened this issue to track the AppStart thing: https://github.com/open-telemetry/opentelemetry-android/issues/606

Issue #556, opened 2024-08-26
1.0.0 RUM initialization API proposals


LikeTheSalad:
In the last Android SIG meeting we discussed potential API surface options before going GA/stable. Ideally, the API should be easy to use for users who aren't familiar with OTel in general and/or don't have special requirements for their apps and are happy with the defaults provided by the Android agent, yet flexible enough to accommodate users with more expertise around OTel and/or with specific requirements that aren't available by default.

It was also mentioned that, when it comes to API surface, it's easier to add features later as the need arises rather than having to remove/modify prematurely added ones in the future due to the breaking changes/deprecations/refactorings, etc, that could arise as a consequence.

Based on the above, we decided to try and come up with some API surface proposals, keeping in mind the 2 types of users mentioned earlier, and show how would each of them make use of the API for different use cases.

This issue is a placeholder for said proposals so please feel free to add yours below.

LikeTheSalad:
# Proposal

<img width="1356" alt="Screenshot 2024-08-26 at 15 52 50" src="https://github.com/user-attachments/assets/71e128ff-997d-46e5-924c-c7344990a016">

## Description

- `OpenTelemetryRumBuilder` gets the bare minimum, low-level setters/adders for it to work. Also, the non-serializable config options from `OtelRumConfig` are moved over to `OpenTelemetruRumBuilder`.
- A new type, `EndpointConfig` is created to allow users to provide their exporter endpoint parameters. Being an interface it could potentially be queried multiple times to allow for "dynamic endpoint params" in the future, while it can also provide final values, as its default impl `SimpleEndpointConfig` should do.
- A new type, `AndroidAgentBuilder`, is created in the higher-level module `android-agent` and will take care of configuring the lower-level builder `OpenTelemetryRumBuilder` with all the defaults that we consider should be present in most Android apps. This new type will take care of:
  - Adding the app name, app version, and environment name to the OTel resources.
  - Enabling disk buffering and configuring it.
  - Configuring a `SessionManager` (not currently present, though it could be similar to the interface created for JS RUM) implementation that would apply the same session logic we have right now (time-based).
  - Creating OtlpHttp exporter implementations for spans and logs using the parameters provided by the provided `EndpointConfig` object.
  - Allowing for customization of the created exporters.
  - (Potentially) installing the default instrumentations without having to rely on SPIs.
  - (Potentially) providing config options for the default instrumentations.

## Use cases

### I just want the defaults

```kotlin
[...]

PR #624, opened 2024-10-08
[Do not merge] Agent api prototype

LikeTheSalad:
Prototype that should cover what's being discussed [here](https://github.com/open-telemetry/opentelemetry-android/issues/556).

- A Kotlin [factory method](https://github.com/open-telemetry/opentelemetry-android/pull/624/files#diff-836d93c05e9bdac10749d4a9358ef9eed93a95f975dc29257e482bc403e4cc7dR62) that returns OpenTelemetryRumBuilder, which to me it seems to be the simplest way of using the agent as there's only one method that users would need to be aware of. It also contains config params for the default instrumentations.
- Following the diagram, it removes some API surface from OpenTelemetryRumBuilder.
- Following the diagram, It moves the "unserializable" config params from OtelRumConfig into OpenTelemetryRumBuilder.
- Following the diagram (kind of, as the interface initially named SessionManager is now changed to SessionProvider https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2358) it sets the SessionProvider as the only session-related type in core, and moves all of our session-related types to the agent so that the existing session handling tools (the ones that create sessions based on a timeout) are provided by the agent, and "power-users" can provide their own session handling logic directly to core, if needed.
- Following the diagram, it creates [EndpointConfig](https://github.com/open-telemetry/opentelemetry-android/pull/624/files#diff-9c2d3bac896bbf1cfba28553fec9c4b8ff7c8016e668013381080ec04f28c287) and a default impl that provides HTTP endpoints.
- Adds the [OTLP exporters](https://github.com/open-telemetry/opentelemetry-android/pull/624/files#diff-5c25489af737ef1fb30c51ae820b29c9cf99496b69c57a323193d3a98beff2f0R13) dependency to the agent module and makes the agent to automatically [set otlp-http span and log exporters](https://github.com/open-telemetry/opentelemetry-android/pull/624/files#diff-836d93c05e9bdac10749d4a9358ef9eed93a95f975dc29257e482bc403e4cc7dR108) that use the provided endpoint config params.
[Here](https://github.com/open-telemetry/opentelemetry-android/pull/624/files#diff-cd2d20c56219b8123f1890c60908d015dfe529acb9b2320ad08cca590950aeedR25) we can see how this agent API is used in our demo app.


android-agent/src/main/kotlin/io/opentelemetry/android/agent/AndroidAgent.kt
```diff
@@ -0,0 +1,167 @@
...
+        AndroidInstrumentationLoader.getInstrumentation(NetworkChangeInstrumentation::class.java)!!
+    }
+    private val slowRenderingInstrumentation by lazy {
+        AndroidInstrumentationLoader.getInstrumentation(SlowRenderingInstrumentation::class.java)!!
+    }
+
+    fun createRumBuilder(
+        application: Application,
+        otelRumConfig: OtelRumConfig = OtelRumConfig(),
+        endpointConfig: EndpointConfig = EndpointConfig.getDefault("http://localhost"),
```
breedx-splk, line 65: I like the idea of having a simple and direct configuration approach, but we should still make it easy for users to configure separate endpoints for logs and spans. We can't assume that they're the same.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/AndroidAgent.kt
```diff
@@ -0,0 +1,167 @@
...
+    private val networkChangeInstrumentation by lazy {
+        AndroidInstrumentationLoader.getInstrumentation(NetworkChangeInstrumentation::class.java)!!
+    }
+    private val slowRenderingInstrumentation by lazy {
+        AndroidInstrumentationLoader.getInstrumentation(SlowRenderingInstrumentation::class.java)!!
+    }
+
+    fun createRumBuilder(
+        application: Application,
+        otelRumConfig: OtelRumConfig = OtelRumConfig(),
```
breedx-splk, line 64: as a user, when I see that I can pass in a "rum config" object, I instantly wonder what's in there. ...but then I see this other stuff alongside it and wonder even more why some of this configurable stuff is in the config but some of it is passed directly to the builder. Seems inconsistent.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/AndroidAgent.kt
LikeTheSalad: Got it, yeah it makes sense. Right now the [EndpointConfig](https://github.com/open-telemetry/opentelemetry-android/pull/624/files#diff-9c2d3bac896bbf1cfba28553fec9c4b8ff7c8016e668013381080ec04f28c287) interface provides a separate getter for each, although it's not the most user-friendly approach having to implement it just to provide different URLs. Do you have any preferences of what could be a nice way to address this? The one thing I was thinking right now was to remove the `EndpointConfig` param and instead add string params for each url, and maybe a "defaultUrl" one that gets overridden if there's a value provided for a specific signal url.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/AndroidAgent.kt
LikeTheSalad: That's true, although I'd argue that we're already in a similar situation by having an `OtelRumConfig` object passed to the `OpenTelemetryRumBuilder` which itself has some config options too! 😅 

At some point, I was thinking about `OtelRumConfig` as a place for flags so that we could switch things on/off and also provide a serializable part of the config which might be useful for debug logs, and then the rest of config options in the builder as a place for more complex objects... Although maybe we would also like to show something else in the debug logs apart from just flags, in which case I'm not sure how can I justify having both tbh. What's your view on this?

android-agent/src/main/kotlin/io/opentelemetry/android/agent/AndroidAgent.kt
bidetofevil: Separate for each and fallback to the default sounds good. There aren't so many signals that it would be onerous to specify the URL twice on init if they have a custom one

bidetofevil (approved):
Added some suggestions but I think this is an incremental improvement that we can build on.

Between this PR and [this](https://github.com/open-telemetry/opentelemetry-android/pull/671), a larger discussion can probably be had about separating the project into layers so that they can be individually configured and used. There's still a bit too much mixing of concerns between the Agent, the platform services it provides, the SDK instance it wraps, and the instrumentation that depend directly on one or more of these layers.  Until the separation is more formal, it'll be hard to get away from some of these challenges of partial configuration and reuse.

LikeTheSalad:
I think we got a rough idea from this draft, so I'll close it and start working on small changes towards some form of initialization API on other PRs.

Issue #608, opened 2024-09-20
[demo-app] Add ability to cause an ANR (application not responding)


breedx-splk:
Much like the ability to crash the application, we want to be able to manually/intentionally trigger an ANR. We should build this capability into the demo application.

Relates to the questions in #496.

Victorsesan:
@breedx-splk Can you please assign this to me? i can give it a try and add the ANR simulation code to the main activity and Implement the ANR Logic to it and you review. Also the repo Readme.md file shows some of the features included in the demo app already includes the ANR feature , does that mean we just have to modify an already implemented ANR feature? if so where can i find the file to work on. Thank You

breedx-splk:
@Victorsesan thank you so much to offering to work on this, but another contributor @magda-woj explained on the weekly SIG meeting today that the code is written and a PR will be submitted right after #627 is merged. 

The help is definitely welcome, so please feel free to look into other issues or just directly submit a PR for something else without being assigned (it's not required). Thanks again.

PR #627, opened 2024-10-09
Demo app: slow renders animation

magda-woj:
Added a way to trigger slow rendering in the demo app. `SlowCometAnimation.kt` contains a jetpack compose animation of a "meteor shower" that's intentionally slowed down through delays, expensive drawing, and unnecessary random number generation. To see the animation, add any amount of The Comet Book (the last product on the product list) to the cart.


marandaneto (approved):

marandaneto:
I think this PR closes https://github.com/open-telemetry/opentelemetry-android/issues/608

breedx-splk (approved):
The animation is pretty cool and I love that you were able to keep it on-theme (astronomy!). Thanks!

PR #634, opened 2024-10-16
Demo app ANR

magda-woj:
Added a way to intentionally trigger application not responding in the demo app, forcing Thread.sleep() on the main thread. To cause the ANR, try try to add to cart exactly 9 National Park Foundation Explorascopes (first product on the product list) and click "Yes, I'm sure." on the alert pop-up. Addresses #608.


breedx-splk (approved):
LGTM!

PR #653, opened 2024-10-22
A few low-hanging Kotlin migrations

breedx-splk:
While looking at some code, decided to pick up a few small kotlin migrations.


MaciejCzekanski (approved):
Looking good

marandaneto (approved):

instrumentation/activity/src/main/java/io/opentelemetry/android/instrumentation/activity/ActivityCallbacks.kt
```diff
@@ -0,0 +1,95 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.android.instrumentation.activity
+
+import android.app.Activity
+import android.os.Bundle
+import io.opentelemetry.android.internal.services.visiblescreen.activities.DefaultingActivityLifecycleCallbacks
+
+class ActivityCallbacks(private val tracers: ActivityTracerCache) : DefaultingActivityLifecycleCallbacks {
```
bidetofevil, line 12: You may be able to just extend the base interface rather than `DefaultingActivityLifecycleCallbacks` given you're overriding everything explicitly already, but this is fine.

bidetofevil (approved):
LGTM

PR #656, opened 2024-10-23
Make the spotless usage block level

breedx-splk:
Most of the other instructions are quoted as a block, which I think made the spotless bit a little too easy for new contributors to overlook.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,8 +44,13 @@ The output artifacts will be in `instrumentation/build/outputs/`.
 We use [spotless](https://github.com/diffplug/spotless) to enforce a consistent code style
 throughout the project. This includes reformatting (linting) of both source code and markdown.
 
-Before submitting a PR, you should ensure that your code is linted by
-running `./gradlew spotlessApply` and committing your changes.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -44,8 +44,13 @@ The output artifacts will be in `instrumentation/build/outputs/`.
...
 
-Before submitting a PR, you should ensure that your code is linted by
-running `./gradlew spotlessApply` and committing your changes.
+Before submitting a PR, you should ensure that your code is linted. We use the
+[spotless gradle plugin](https://github.com/diffplug/spotless/tree/main/plugin-gradle) to
+make this easy. You should run it like this:
+
+```
+./gradlew spotlessApply
+```
```
marandaneto, line 53: What I like doing is having a `Makefile` for such things and git pre-commit hooks, so it's "forced" to fix such things before committing.
Even better if CI runs auto-fix andcommitst changes ^^
Not a blocker, just ideas.

breedx-splk: Yeah, or something like this: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/.githooks/pre-commit

Issue #658, opened 2024-10-23
Extract AndroidInstrumentation Interface and related classes into their own module + update install method to take generic OpenTelemetry instance 


surbhiia:
Right now all the instrumentations implement [the AndroidInstrumentation Interface](https://github.com/open-telemetry/opentelemetry-android/blob/c6007a9309f2733462f258f793fa750443efa913/core/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentation.kt) which resides in [the bigger core module](https://github.com/open-telemetry/opentelemetry-android/tree/c6007a9309f2733462f258f793fa750443efa913/core/src/main/java/io/opentelemetry/android). If suppose, some vendors just want to utilize the individual instrumentation modules like [OkHttp3](https://github.com/open-telemetry/opentelemetry-android/tree/c6007a9309f2733462f258f793fa750443efa913/instrumentation/okhttp/okhttp-3.0) / [HttpURLConnection ](https://github.com/open-telemetry/opentelemetry-android/tree/c6007a9309f2733462f258f793fa750443efa913/instrumentation/httpurlconnection)without all of the core SDK, they'd be dependent on the bigger core module without needing it. So, it would be helpful to have the AndroidInstrumentation Interface and related interface (AndroidInstrumentationLoader) and classes ([AndroidInstrumentationLoaderImpl](https://github.com/open-telemetry/opentelemetry-android/blob/01a801d4eff21dee5e143d8116daf58b85003e83/core/src/main/java/io/opentelemetry/android/internal/instrumentation/AndroidInstrumentationLoaderImpl.kt#L8)) reside in their own separate module. 

Secondly, If a vendor does not want to use OpenTelemetryRum Interface and just the individual instrumentations, they'd not be able to do that because the [AndroidInstrumentation install method requires an OpenTelemetryRum instance](https://github.com/open-telemetry/opentelemetry-android/blob/c6007a9309f2733462f258f793fa750443efa913/core/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentation.kt#L37). We should change it to generic OpenTelemetry instance instead. For example: if I want to utilize the HttpURLConnection instrumentation, I'd have to do the following:
`OpenTelemetryRum oTelRumInstance = OpenTelemetryRum.builder(application).build();
[...]

PR #661, opened 2024-10-24
Relax install arg from OpenTelemetryRum to OpenTelemetry instance.

breedx-splk:
As @surbhiia described in #658, there is an unnecessary tight coupling from the `AndroidInstrumentation` implementations back to the core module.  Literally every one of the instrumentations was taking the rum instance only so that it could get access to the otel instance. 

So let's simplify this and have fewer law-of-demeter violations! This changes the `install()` API to take the OpenTelemetry instance directly.

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/SdkPreconfiguredRumBuilder.kt
+++ b/core/src/main/java/io/opentelemetry/android/SdkPreconfiguredRumBuilder.kt
@@ -54,7 +54,7 @@ class SdkPreconfiguredRumBuilder
 
             // Install instrumentations
             for (instrumentation in getInstrumentations()) {
-                instrumentation.install(application, openTelemetryRum)
+                instrumentation.install(application, openTelemetryRum.openTelemetry)
             }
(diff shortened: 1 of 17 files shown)


surbhiia (approved):
Thanks for this! It looks great :) 

(I noticed the volley http README mentions usage of OpenTelemetryRum as well, we can mention a line there saying users can bypass it and use their own OpenTelemetry instance as well)

bidetofevil (approved):
LGTM. 

Some of the instrumentation may also not need to entire OpenTelemetry object, so in the future, it may be good to narrow the scope of what they need to further narrow this (e.g. just pass in the TracerProvider or something)

breedx-splk:
> LGTM.
> 
> Some of the instrumentation may also not need to entire OpenTelemetry object, so in the future, it may be good to narrow the scope of what they need to further narrow this (e.g. just pass in the TracerProvider or something)

Yup, totally agree that keeping dependencies minimal is the way to go.

Issue #638, opened 2024-10-17
Add ability to throttle exports when reading from disk.


breedx-splk:
If the user is offline or in airplane mode for a long time and a lot of data has buffered on disk, it is possible for the agent to use a lot of bandwidth. There is currently nothing that limits how much or how fast the data should be sent. Instead, we just try and export all the data on disk as fast as we can...and this can be a problem because it hogs a limited resource.

This feature could be opt-in or opt-out, but there should be a way to configure the max "bandwidth" used by the SignalFromDisk exporters. I used the quotes around "bandwidth" because I'm not sure that it needs to be strict bytes/time style, and we might just be able to do batch/time or spans/time or something similar.

This would ideally allow the data to flow more smoothly/consistently instead of "all at once".

breedx-splk:
There is some prior art [here in the Splunk ThrottlingExporter](https://github.com/signalfx/splunk-otel-android/blob/main/splunk-otel-android/src/main/java/com/splunk/rum/ThrottlingExporter.java), but this is not exactly what we want because it favors dropping data instead of allowing it to stay buffered (on disk).

Victorsesan:
@breedx-splk Can i be allowed to help with this one? thanks

breedx-splk:
Sure, thanks!

Victorsesan:
@breedx-splk Please help me out, can we create a new class/file in the android/export directory in this repo with a name like RateLimitedExporter.java to implement a SpanExporter interface which can help with this? or in which file can we implement the bandwidth-limiting feature in?. Thanks for the help

breedx-splk:
@Victorsesan Yes, you would likely want to create a new package and new class(es) to implement this new functionality. Please review [CONTRIBUTING.md](https://github.com/open-telemetry/opentelemetry-android/blob/main/CONTRIBUTING.md) and make sure include tests and to format with `./gradlew spotlessApply`.

PR #663, opened 2024-10-27
Add ability to throttle exports when reading from disk. 

Victorsesan:
Added an implementation which provides a flexible way to manage bandwidth usage when exporting spans, allowing for smoother data flow and preventing resource hogging. It can further refine the size estimation logic based on a specific use case.
Relate to #638

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/export/RateLimitedExporter.java
+++ b/core/src/main/java/io/opentelemetry/android/export/RateLimitedExporter.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.opentelemetry.android;
+
+import android.util.Log;
+import io.opentelemetry.sdk.common.CompletableResultCode;
+import io.opentelemetry.sdk.trace.data.SpanData;
+import io.opentelemetry.sdk.trace.export.SpanExporter;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.function.Function;
+
+class BandwidthThrottlingExporter implements SpanExporter {
+    private final SpanExporter delegate;
+    private final Function<SpanData, String> categoryFunction;
+    private final long maxBytesPerSecond;
+    private final long timeWindowInMillis;
+    private long lastExportTime;
+    private long bytesExportedInWindow;
+
+    private BandwidthThrottlingExporter(Builder builder) {
+        this.delegate = builder.delegate;
+        this.categoryFunction = builder.categoryFunction;
+        this.maxBytesPerSecond = builder.maxBytesPerSecond;
+        this.timeWindowInMillis = builder.timeWindow.toMillis();
+        this.lastExportTime = System.currentTimeMillis();
+        this.bytesExportedInWindow = 0;
+    }
+
+    static Builder newBuilder(SpanExporter delegate) {
+        return new Builder(delegate);
+    }
+
+    @Override
+    public CompletableResultCode export(Collection<SpanData> spans) {
+        List<SpanData> spansToExport = new ArrayList<>();
+        long totalBytes = 0;
+
+        for (SpanData span : spans) {
+            // Estimate the size of the span (this can be adjusted based on actual size)
+            long spanSize = estimateSpanSize(span);
+            totalBytes += spanSize;
+
+            // Check if we can export this span based on the current bandwidth limit
+            if (canExport(spanSize)) {
+                spansToExport.add(span);
+                bytesExportedInWindow += spanSize;
+            } else {
+                Log.d("BandwidthThrottlingExporter", "Throttled span: " + span.getName());
+            }
+        }
+
+        return delegate.export(spansToExport);
+    }
+
+    private boolean canExport(long spanSize) {
+        long currentTime = System.currentTimeMillis();
+        if (currentTime - lastExportTime > timeWindowInMillis) {
+            // Reset the window
+            bytesExportedInWindow = 0;
+            lastExportTime = currentTime;
+        }
+
+        return (bytesExportedInWindow + spanSize)
+                <= maxBytesPerSecond * (timeWindowInMillis / 1000);
+    }
+
+    private long estimateSpanSize(SpanData span) {
+        // This is a placeholder for actual size estimation logic
+        return span.getAttributes().size() * 8; // Example: 8 bytes per attribute
+    }
+
+    @Override
+    public CompletableResultCode flush() {
+        return delegate.flush();
+    }
+
+    @Override
+    public CompletableResultCode shutdown() {
+        return delegate.shutdown();
+    }
+
+    static class Builder {
+        final SpanExporter delegate;
+        Function<SpanData, String> categoryFunction = span -> "default";
+        long maxBytesPerSecond = 1024; // Default to 1 KB/s
+        Duration timeWindow = Duration.ofSeconds(1); // Default to 1 second
+
+        private Builder(SpanExporter delegate) {
+            this.delegate = delegate;
+        }
+
+        Builder maxBytesPerSecond(long maxBytesPerSecond) {
+            this.maxBytesPerSecond = maxBytesPerSecond;
+            return this;
+        }
+
+        Builder timeWindow(Duration timeWindow) {
+            this.timeWindow = timeWindow;
+            return this;
+        }
+
+        BandwidthThrottlingExporter build() {
+            return new BandwidthThrottlingExporter(this);
+        }
+    }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 demo-app/local.properties
 .DS_Store
 **/build/
+


core/src/main/java/io/opentelemetry/android/export/RateLimitedExporter.java
```diff
@@ -0,0 +1,112 @@
...
+    @Override
+    public CompletableResultCode shutdown() {
+        return delegate.shutdown();
+    }
+
+    static class Builder {
+        final SpanExporter delegate;
+        Function<SpanData, String> categoryFunction = span -> "default";
+        long maxBytesPerSecond = 1024; // Default to 1 KB/s
+        Duration timeWindow = Duration.ofSeconds(1); // Default to 1 second
```
marandaneto, line 92: Function and Duration require newer APIs (24 and 26), most likely not a problem tho since we require desugaring right?

core/src/main/java/io/opentelemetry/android/export/RateLimitedExporter.java
```diff
@@ -0,0 +1,112 @@
...
+    private final long timeWindowInMillis;
+    private long lastExportTime;
+    private long bytesExportedInWindow;
+
+    private BandwidthThrottlingExporter(Builder builder) {
+        this.delegate = builder.delegate;
+        this.categoryFunction = builder.categoryFunction;
+        this.maxBytesPerSecond = builder.maxBytesPerSecond;
+        this.timeWindowInMillis = builder.timeWindow.toMillis();
+        this.lastExportTime = System.currentTimeMillis();
```
marandaneto, line 31: Use https://github.com/open-telemetry/opentelemetry-android/blob/e785fc627e688fa664f21ba5e68bebae98d07ca2/core/src/main/java/io/opentelemetry/android/internal/tools/time/SystemTime.kt#L31 for `System.currentTimeMillis()` so you can unit test the changes

core/src/main/java/io/opentelemetry/android/export/RateLimitedExporter.java
Victorsesan: You're right! Thank you for pointing it out, I had figured we already had desugaring configured which is why i added them. However i think they're very essential to use and we should have desugaring configured regardless to make them work as expected. I have opened another PR with the needed config and changes, let me know what you think. Thanks

core/src/main/java/io/opentelemetry/android/export/RateLimitedExporter.java
Victorsesan: Also it seems like i cannot use return in an assignment statement, it doesn't pass spotless test. Or should i add and ignore the build test error?

core/src/main/java/io/opentelemetry/android/export/RateLimitedExporter.java
marandaneto: desugaring has to be configured in the hosting app, see [note in the readme](https://github.com/open-telemetry/opentelemetry-android#getting-started)
I think if its fine if we are using the Duration and Function elsewhere in the SDK, are we?

core/build.gradle.kts
```diff
@@ -12,6 +12,12 @@ android {
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
         consumerProguardFiles("consumer-rules.pro")
         buildConfigField("String", "OTEL_ANDROID_VERSION", "\"$version\"")
+
+// Enable desugaring for Java 8 features
+        compileOptions {
+            sourceCompatibility = JavaVersion.VERSION_1_8 // Set source compatibility to Java 8
+            targetCompatibility = JavaVersion.VERSION_1_8 // Set target compatibility to Java 8
+        }
```
marandaneto, line 20: Probably not needed because depending on the AGP version, 1.8 is already the default, unless its already Java 11 and you want to force Java 1.8

core/src/main/java/io/opentelemetry/android/export/RateLimitedExporter.java
marandaneto: You should be able to just do `SystemTime.get().getCurrentTimeMillis()`, what s the error?

core/build.gradle.kts
Victorsesan: @marandaneto How does the new changes look? I'm still making some improvements, just want to make sure I'm on the right track and following everything right👍

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "java.configuration.updateBuildConfiguration": "automatic"
```
breedx-splk, line 2: Let's keep the IDE specific configurations out of the repo please.

core/src/main/java/io/opentelemetry/android/export/RateLimitedExporter.java
```diff
@@ -0,0 +1,112 @@
...
+import io.opentelemetry.sdk.common.CompletableResultCode;
+import io.opentelemetry.sdk.trace.data.SpanData;
+import io.opentelemetry.sdk.trace.export.SpanExporter;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.function.Function;
+
+class BandwidthThrottlingExporter implements SpanExporter {
```
breedx-splk, line 18: I think we would also need to see a unit test for this class.

core/src/main/java/io/opentelemetry/android/export/RateLimitedExporter.java
marandaneto: somehow relates to https://github.com/open-telemetry/opentelemetry-android/issues/682

core/src/main/java/io/opentelemetry/android/export/RateLimitedExporter.java
```diff
@@ -0,0 +1,112 @@
...
+            } else {
+                Log.d("BandwidthThrottlingExporter", "Throttled span: " + span.getName());
+            }
+        }
+
+        return delegate.export(spansToExport);
+    }
+
+    private boolean canExport(long spanSize) {
+        long currentTime = System.currentTimeMillis();
```
marandaneto, line 62: should use `SystemTime.get().getCurrentTimeMillis();` as well

core/src/main/java/io/opentelemetry/android/export/RateLimitedExporter.java
```diff
@@ -0,0 +1,112 @@
...
+    @Override
+    public CompletableResultCode shutdown() {
+        return delegate.shutdown();
+    }
+
+    static class Builder {
+        final SpanExporter delegate;
+        CategoryFunction categoryFunction = span -> "default";
+        long maxBytesPerSecond = 1024; // Default to 1 KB/s
+        long timeWindowInMillis = 1000; // Default to 1 second
```
marandaneto, line 92: should they be private since they have a setter anyway (using the builder pattern)?

Victorsesan: Since the fields are intended to be set only through the builder methods i have tried to add and make them private so they can only be modified through it's provided builder methods. I'm sure it will help enhance encapsulation and help maintain the integrity of it's object's state.

marandaneto: `delegate` can also be private I guess

breedx-splk:
@Victorsesan are you able to come back to this any time soon? Thanks!

Victorsesan:
Hey @breedx-splk Yes i will, i think i lastly made a change that needed a mod review.  Still waiting

core/src/main/java/io/opentelemetry/android/export/RateLimitedExporter.java
```diff
@@ -0,0 +1,112 @@
...
+        return delegate.flush();
+    }
+
+    @Override
+    public CompletableResultCode shutdown() {
+        return delegate.shutdown();
+    }
+
+    static class Builder {
+        final SpanExporter delegate;
```
marandaneto, line 89:
```suggestion
        private final SpanExporter delegate;
```

core/src/main/java/io/opentelemetry/android/export/RateLimitedExporter.java
marandaneto: https://github.com/open-telemetry/opentelemetry-android/pull/663/files#r1941464425

marandaneto:
> * https://docs.github.com/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches

https://github.com/open-telemetry/opentelemetry-android/pull/663#discussion_r1941464425 left and ready to go, I will approve meanwhile.

marandaneto (approved):
https://github.com/open-telemetry/opentelemetry-android/pull/663#discussion_r1941464425 pending otherwise LGTM

breedx-splk:
@Victorsesan seems like we're close, but the build is broken again.

LikeTheSalad:
> So I had a look at the PR, and I think it needs a few additional changes to be production ready: namely, the algorithm to determine the size of a span in bytes is just a placeholder, and when the threshold is reached, the exported spans are not cached, but simply dropped and not passed onto the delegate.
> 
> If it was an in-progress change, it may be reasonable to merge it, but unless there is commit to get this production-ready, I don't think we should be in the repo.

I agree. Also, to add to those points:

- The algorithm to determine the size of a span might not be straightforward to create, and even if we come up with a nice one, it might not be as processing-friendly as other options, such as the one about using a batch/time approach that's mentioned in the issue.
- Dropping data should not be part of this solution. The closest I think we can get to an implementation that addresses this issue without dropping data, would be by somehow [breaking this loop](https://github.com/open-telemetry/opentelemetry-android/blob/32fd8e1681dde7a0fc91381c047320bed0946c8e/core/src/main/java/io/opentelemetry/android/features/diskbuffering/scheduler/DefaultExportScheduler.kt#L28C42-L28C61) before all the available data in disk is exported.

Victorsesan:
> Thank you for creating this PR, @Victorsesan. The approach proposed here to solve the issue brings some important concerns, mentioned in the latest comments, that don't make it feasible to get merged, unless we change the overall approach.
> 
> Going with a different approach would most likely require discarding all the existing changes in this PR, which is totally understandable if that’s more work than you planned for. So please let us know if you’re up for spending more time on it — if not, no worries, we can close this one and revisit it in a future PR.

Hi @LikeTheSalad i can give another go, since the PR has been opened for too long i will be happy to have it completed regardless

bidetofevil:
A couple of suggestions that I think might simplify the solution:

1. We can approach this from the read-from-disk side of the house 
- Basically, replace the timed job mechanism to export batches, but rather have the read-side be triggered on-demand and read read from disk when it's ready. Basically, when a batch is written to disk, it'll inform the reader that there is a batch ready to go. The reader can decide if it's ready to process it, and do when when it's ready. Once it it exports a batch, it can schedule itself to determine when they should check next, and so on, until there are no more batches to read. The reader will be triggered again when a new batch is written to disk.
- The advantage of this is that you won't read from disk until you're ready to send, thereby limiting losing data when there's a crash during export, say, if you were using a buffering exporter to send data out.
2. Instead of counting by bytes, just count by spans. And instead of cutting it off right at the limit, just let the last batch go through. 
- We are just approximating things here to limit data flow, so there's no need to eat the complexity to try to be that fine grain. I think doing so by counting logs and spans is sufficient for most cases, even if not entirely accurate.

Victorsesan:
Thanks for the suggestions @bidetofevil will keep that in mind while working on it

PR #664, opened 2024-10-27
add-disk-buffering-debug-mode

### LINKED ISSUE #597: Add setting to enable disk buffering debug mode
After https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1455 lands and is published, we will want to be able to expose this in the `DiskBufferingConfiguration` so that users can turn it on for troubleshooting purposes.

Annosha:
**Description**

Fixes #597  Add setting to enable disk buffering debug mode 

This PR introduces a configurable debug mode in the `DiskBufferingConfiguration` class within OpenTelemetry Android, allowing users to enable verbose logging for troubleshooting disk buffering behavior.
**Changes Made**

1. Added `setDebugMode` Method:

A new method, `setDebugMode`, has been added to the `DiskBufferingConfiguration.Builder `class. This method accepts a boolean parameter to enable or disable debug mode.
When enabled, a log message indicates that debug mode for disk buffering is active, providing enhanced logging details for easier troubleshooting.

2. Enhanced `setEnabled `Method:

The `setEnabled` method now includes a check for `enableDebugMode`. If debug mode is active, an additional log message provides further information on the state of disk buffering.
This conditional logging gives users more visibility into the disk buffering configuration when debug mode is enabled.

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfiguration.java
+++ b/core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfiguration.java
@@ -22,6 +22,7 @@ public final class DiskBufferingConfiguration {
     private final long maxFileAgeForWriteMillis;
     private final long minFileAgeForReadMillis;
     private final long maxFileAgeForReadMillis;
+    private final boolean enableDebugMode;
 
     private DiskBufferingConfiguration(Builder builder) {
         enabled = builder.enabled;
@@ -30,7 +31,7 @@ public final class DiskBufferingConfiguration {
         maxFileAgeForWriteMillis = builder.maxFileAgeForWriteMillis;
         minFileAgeForReadMillis = builder.minFileAgeForReadMillis;
         maxFileAgeForReadMillis = builder.maxFileAgeForReadMillis;
-        this.enableDebugMode = builder.enableDebugMode;
+        enableDebugMode = builder.enableDebugMode;
     }
 
     public static Builder builder() {
--- a/core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfiguration.java
+++ b/core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfiguration.java
@@ -30,7 +30,7 @@ public final class DiskBufferingConfiguration {
         maxFileAgeForWriteMillis = builder.maxFileAgeForWriteMillis;
         minFileAgeForReadMillis = builder.minFileAgeForReadMillis;
         maxFileAgeForReadMillis = builder.maxFileAgeForReadMillis;
-        this.enableDebugMode = builder.enableDebugMode; 
+        this.enableDebugMode = builder.enableDebugMode;
     }
 
     public static Builder builder() {
@@ -78,9 +78,12 @@ public final class DiskBufferingConfiguration {
         // Sets the debug mode for disk buffering, enabling additional logging.
         public Builder setDebugMode(boolean enableDebugMode) {
             this.enableDebugMode = enableDebugMode;
-            Logger.getLogger(DiskBufferingConfiguration.class.getName()).log(Level.INFO, 
-                "Disk buffering has been " + (enabled ? "enabled." : "disabled.") + 
-                    (enableDebugMode ? " Debug mode is active." : ""));
+            Logger.getLogger(DiskBufferingConfiguration.class.getName())
+                    .log(
+                            Level.INFO,
+                            "Disk buffering has been "
+                                    + (enabled ? "enabled." : "disabled.")
+                                    + (enableDebugMode ? " Debug mode is active." : ""));
             return this;
         }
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfiguration.java
```diff
@@ -70,12 +72,30 @@ public static final class Builder {
         private long maxFileAgeForWriteMillis = TimeUnit.SECONDS.toMillis(30);
         private long minFileAgeForReadMillis = TimeUnit.SECONDS.toMillis(33);
         private long maxFileAgeForReadMillis = TimeUnit.HOURS.toMillis(18);
+        private boolean enableDebugMode = false;
 
         private ExportScheduleHandler exportScheduleHandler = DefaultExportScheduleHandler.create();
 
+        // Sets the debug mode for disk buffering, enabling additional logging.
```
marandaneto, line 79:
```suggestion
        /** Sets the debug mode for disk buffering, enabling additional logging. */
```

core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfiguration.java
```diff
@@ -22,6 +22,7 @@ public final class DiskBufferingConfiguration {
     private final long maxFileAgeForWriteMillis;
     private final long minFileAgeForReadMillis;
     private final long maxFileAgeForReadMillis;
+    private final boolean enableDebugMode;
```
marandaneto, line 25: I think a debug mode should be part of the `OtelRumConfig` or the lowest level of configuration (if there's one) so the whole SDK respects the debug mode flag, and not only for the disk buffering, this would be useful across the SDK and its configurations

breedx-splk: I think there's a case to be made for both. Top-level debug mode could imply disk buffering debug mode, but not the other way around. In any case, that should be a separate PR IMO. Feel free to log an issue to track.

marandaneto: All good, having an SDK debug mode for all options feels too much IMO, a top-level debug mode that logs everything everywhere and respects eg a severity log level for filtering would be cleaner and easier to enable/disable.
If we introduce a debug flag per feature eg diskbuffering, that just adds more complexity everywhere, I'd rather achieve this with TAG filtering if the global debug mode is enabled.

marandaneto: I unresolved the comment so that I can copy-paste the comment URL otherwise the anchor link does not work, but it's not that you have to address this in this PR.

marandaneto: https://github.com/open-telemetry/opentelemetry-android/issues/668

breedx-splk: Thanks. I'm not entirely sure I made my point, so let's continue chatting about it at SIG sometime.

core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfiguration.java
```diff
@@ -70,12 +72,30 @@ public static final class Builder {
...
 
+        /** Sets the debug mode for disk buffering, enabling additional logging. */
+        public Builder setDebugMode(boolean enableDebugMode) {
+            this.enableDebugMode = enableDebugMode;
+            Logger.getLogger(DiskBufferingConfiguration.class.getName())
+                    .log(
+                            Level.INFO,
+                            "Disk buffering has been "
+                                    + (enabled ? "enabled." : "disabled.")
+                                    + (enableDebugMode ? " Debug mode is active." : ""));
```
breedx-splk, line 87: I don't think we need to log the `enabled` setting here -- that's a separate concern below. I recommend changing the log message to be simpler and only reflect the new value of `enableDebugMode`.

core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfiguration.java
```diff
@@ -70,12 +72,30 @@ public static final class Builder {
...
+            return this;
+        }
+
         /** Enables or disables disk buffering. */
         public Builder setEnabled(boolean enabled) {
             this.enabled = enabled;
+            if (enableDebugMode) {
+                Logger.getLogger(DiskBufferingConfiguration.class.getName())
+                        .log(Level.INFO, "Debug log message here");
```
breedx-splk, line 96: I think this message is probably not what you intended?

breedx-splk (changes requested):
Thanks for the PR. This is a good first step, but we're currently missing the glue that uses this setting to actually turn on this feature in the disk buffering module. Can you please add that part, where the disk buffering code is initialized?

breedx-splk:
Hey @Annosha just checking in to see if you're still able to work on some revisions/improvements to this PR. Thanks again!

breedx-splk:
@Annosha checking back again after the holidays. How's it going?

breedx-splk:
Superseded by #753. Sorry, but seemed abandoned after 3 months. 🤷🏻

PR #667, opened 2024-10-28
release.yml - Remove commented grep

breedx-splk:
Looks like #665 did not do the trick, [still broken](https://github.com/open-telemetry/opentelemetry-android/actions/runs/11559453062)....so I used a yaml validator and it caught an issue related to a comment:

image

So let's just remove that commented line and hope that does the trick.


breedx-splk:
If that does turn out to be the case, then GitHub is just flagging the wrong line maybe as invalid. 🤷🏻

marandaneto (approved):

Issue #669, opened 2024-10-29
Extract Services out in their own modules so they can be plugged-in as needed


surbhiia:
Currently, [services](https://github.com/open-telemetry/opentelemetry-android/tree/main/core/src/main/java/io/opentelemetry/android/internal/services) are part of the bigger core module. It would be beneficial to extract each of them in their own separate modules so they can be plugged-in as needed. 

**Example use case** - Some services like **PeriodicWorkService** are needed if Disk Buffering is used (to export spans stored in disk to backend). If users/ vendors implement their own disk buffering mechanism, they do not need this service. 

There are mainly 4 services (AppLifecycle, VisibleScreenService, CurrentNetworkProvider, PeriodicWorkService) and 2 classes (CacheStorage and Preferences). Objects are created from the 2 classes and the objects can be obtained from the service interface but they're not really services. 

- CacheStorage is also only used for disk buffering. I think this can perhaps to moved to the module hosting disk buffering related stuff (under [persistence](https://github.com/open-telemetry/opentelemetry-android/tree/main/core/src/main/java/io/opentelemetry/android/internal/features/persistence) which hosts DiskManager which is the only place it is used right now in).  
-  Preferences can be provided as a generic utility class maybe via the [common](https://github.com/open-telemetry/opentelemetry-android/tree/main/core/src/main/java/io/opentelemetry/android/common) module. Right now  it is used only in [DiskManager for storage and retrieval of  MAX_FOLDER_SIZE_KEY](https://github.com/open-telemetry/opentelemetry-android/blob/main/core/src/main/java/io/opentelemetry/android/internal/features/persistence/DiskManager.kt#L75)), 

Following bullets cover what instrumentations rely on what services for more clarity, so we know not all instrumentations require all services
-  ANR -> AppLifecycle
-  Activity / Fragment Lifecycle -> VisibleScreenService
-  NetworkChangeMonitor -> AppLifecycle and CurrentNetworkProvider

surbhiia:
**Currently**, services are housed under core -> internal -> services

**Changes proposal:**

[...]

PR #671, opened 2024-10-30
Relax the ServiceManager singleton nature

breedx-splk:
As noted elsewhere and discussed in the Android SIG today, the static singleton nature of the `ServiceManager` both causes modularity challenges and makes testing in vendor distros that wrap the `OpenTelmetryRumBuilder` difficult if not impossible. So this makes a first attempt at the following:

1. Eliminate the static singleton nature of the `ServiceManager`. This primarily means that any class that depends on this must have it injected (via construction). There is still only one instance used at runtime, but no longer can classes reach out statically to randomly call `ServiceManager.get()` in an ad-hoc manner.
2. Allows a custom `ServiceManager` instance to be injected into the `OpenTelemetryRumBuilder` in case the (power) user needs to override some defaults or provide their own implementation of a service.
2. Several instrumentations need to get access to services during installation time, so the instrumentation API needed to change to accommodate this. As a result, we now have an `InstallationContext` data class in the instrumentation package that bundles up the app, the otel instance, and service manager instance.
3. Unfortunately, some of the disk buffering configuration ended up hitting the static `ServiceManager.get()` singleton to get the `PeriodicWorkService` when creating the default implementations of the (from disk) export scheduler. I decided that having functional components as "configuration" was probably stupid, so I just pulled them out all the way up to the builder.
4. The list of services that are implied/default is no longer coupled inside the `ServiceManager` interface, but instead fixed in the sole implementation: `ServiceManagerImpl`. 

Two things that this PR does not yet do that will need to be addressed in follow-up efforts:

1. `ServiceManager` interface is still marked as internal and lives in an internal package. This is problematic because the OTRB now has a public interface method that can take a `ServiceManager` instance. Maybe we just make the interface no longer internal?
2. As @surbhiia has pointed out elsewhere, the list of created and started services still doesn't map well or appropriately to the requirements of the actual instrumentation. In other words, there could still be services created and started that are not used by any instrumentation or exporters or anything.

In any case, I think the testability is greatly improved and the lifecycle is less complex.

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentation.kt
+++ b/core/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentation.kt
@@ -5,9 +5,7 @@
 
 package io.opentelemetry.android.instrumentation
 
-import android.app.Application
 import io.opentelemetry.android.OpenTelemetryRum
-import io.opentelemetry.api.OpenTelemetry
 
 /**
  * This interface defines a tool that automatically generates telemetry for a specific use-case,
@@ -30,11 +28,7 @@ interface AndroidInstrumentation {
      * only be called from [OpenTelemetryRum]'s builder once the [OpenTelemetryRum] instance is initialized and ready
      * to use for generating telemetry.
      *
-     * @param application The Android application being instrumented.
-     * @param openTelemetry The [OpenTelemetry] instance to use for creating signals.
+     * @param ctx The InstallationContext under which the instrumentation is being installed
      */
-    fun install(
-        application: Application,
-        openTelemetry: OpenTelemetry,
-    )
+    fun install(ctx: InstallationContext)
 }
--- a/core/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentation.kt
+++ b/core/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentation.kt
@@ -28,7 +28,7 @@ interface AndroidInstrumentation {
      * only be called from [OpenTelemetryRum]'s builder once the [OpenTelemetryRum] instance is initialized and ready
      * to use for generating telemetry.
      *
-     * @param ctx The InstallationContext under which the instrumentation is being installed
+     * @param ctx The InstallationContext under which the instrumentation is being installed.
      */
     fun install(ctx: InstallationContext)
 }
--- a/core/src/main/java/io/opentelemetry/android/instrumentation/InstallationContext.kt
+++ b/core/src/main/java/io/opentelemetry/android/instrumentation/InstallationContext.kt
@@ -1,0 +1,11 @@
+package io.opentelemetry.android.instrumentation
+
+import android.app.Application
+import io.opentelemetry.android.internal.services.ServiceManager
+import io.opentelemetry.api.OpenTelemetry
+
+data class InstallationContext(
+    val application: Application,
+    val openTelemetry: OpenTelemetry,
+    val serviceManager: ServiceManager
+)
--- a/core/src/main/java/io/opentelemetry/android/instrumentation/InstallationContext.kt
+++ b/core/src/main/java/io/opentelemetry/android/instrumentation/InstallationContext.kt
@@ -1,3 +1,8 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
 package io.opentelemetry.android.instrumentation
 
 import android.app.Application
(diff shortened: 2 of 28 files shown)


core/src/main/java/io/opentelemetry/android/instrumentation/InstallationContext.kt
```diff
@@ -0,0 +1,16 @@
...
+package io.opentelemetry.android.instrumentation
+
+import android.app.Application
+import io.opentelemetry.android.internal.services.ServiceManager
+import io.opentelemetry.api.OpenTelemetry
+
+data class InstallationContext(
+    val application: Application,
+    val openTelemetry: OpenTelemetry,
+    val serviceManager: ServiceManager,
```
surbhiia, line 15: Can we make the application and serviceManager nullable as they're not needed in all instrumentations?

breedx-splk: Possibly, but I think we mostly want to use the same context when installing all instrumentations at startup -- in which case _some_ of the instrumentations will need some of these, so they won't be null. 

I think it's simpler if instrumentations just ignore (don't use) the parts they don't need.

surbhiia: That makes sense! We can perhaps provide a getter for a no-op ServiceManager instance to support the use cases when a user is utilizing one/some of the many instrumentations like the okhttp3 instrumentation that does not require a ServiceManager?

core/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentation.kt
```diff
@@ -30,11 +28,7 @@ interface AndroidInstrumentation {
...
      *
-     * @param application The Android application being instrumented.
-     * @param openTelemetry The [OpenTelemetry] instance to use for creating signals.
+     * @param ctx The InstallationContext under which the instrumentation is being installed.
      */
-    fun install(
-        application: Application,
-        openTelemetry: OpenTelemetry,
-    )
+    fun install(ctx: InstallationContext)
```
LikeTheSalad, line 33: I like this idea. I still need to take a deeper look at the whole PR so I can provide a better review overall, though this part reminded me of [this other PR](https://github.com/open-telemetry/opentelemetry-android/pull/661) which makes it a bit confusing to me because there it seemed like we wanted to avoid passing core-specific types to instrumentations? I'm ok with doing so btw, I just want to make sure what's the overall consensus of the kind of use cases we want to support for instrumentations (whether we need them to be fully independent or not).

breedx-splk: And when you say "core-specific types" in this case you mean both the `InstallationContext` itself but also the `ServiceManager` contained within?

In any case, the api needs to have _something_ to pass to `install()`, and instrumentations might want to leverage the consistent way we've started setting up to get at services. I'm open to other ways that don't involve global/static/singletons. :)

LikeTheSalad: > And when you say "core-specific types" in this case you mean both the `InstallationContext` itself but also the `ServiceManager` contained within?

Yes, that's what I was thinking.

> In any case, the api needs to have something to pass to install(), and instrumentations might want to leverage the consistent way we've started setting up to get at services. I'm open to other ways that don't involve global/static/singletons. :)

Makes sense, we still need to pass something to instrumentations, and I like this idea of passing a single object of ours because it means that we could add more tools in the future if needed without having to change the signature of `install()`.

I think the solution to this will become clearer once we fully understand the problem, if it's a matter of having "zero dependency" on anything that's outside of the upstream java repo, then things might get quite complicated, however, if it's just a matter of package size concerns and having a lot of code from a dependency that might not be fully used, such as the Android core, then it's probably a matter of extracting the instrumentation-related files out of core, as @surbhiia mentioned [here](https://github.com/open-telemetry/opentelemetry-android/issues/658), where we can also add a new type like `InstallationContext`. If I understood correctly, I think the problem is the latter, so maybe we can get away with just moving this type into the separate "instrumentation API" module. Now, what to do about the service manager is another issue though, I hope we can discuss the details properly in a SIG meeting.

bidetofevil (approved):
LGTM. The change moves the project ahead, so I'm good with it overall (see specific comments for non-blocking suggestions)

core/src/main/java/io/opentelemetry/android/instrumentation/InstallationContext.kt
LikeTheSalad: Based on yesterday's SIG meeting, we should try to find a way to allow instrumentations to reuse services/utilities without having to expose them as public apis. There's more details about it all on [this issue](https://github.com/open-telemetry/opentelemetry-android/issues/669).

breedx-splk:
There's still a lot of open discussion in this PR, but it's been approved for 2 weeks without any blocking reviews, so I'm going to merge to try and keep momentum. Like always, happy to continue revising incrementally as we see fit.

Issue #672, opened 2024-10-30
Wired transport type is not supported


MaciejCzekanski:
`SimpleNetworkDetector` class is missing mapping from `ConnectivityManager.TYPE_ETHERNET` to wired connection type.

`NetworkState.TRANSPORT_WIRED` also doesn't exist, even though there is already corresponding value in `NetworkConnectionTypeValues`

PR #673, opened 2024-10-30
Handle wired network transport type

MaciejCzekanski:
This fixes https://github.com/open-telemetry/opentelemetry-android/issues/672


breedx-splk (approved):
Thanks for adding the completeness. I imagine that ethernet is somewhat rare for Android devices, but it's nice to have it covered.

Issue #83, opened 2023-09-18
Add consumer proguard rules


LikeTheSalad:
To prevent obfuscation to break OTel Java SDK functionality in production apps.

LikeTheSalad:
This is quite strange, though I just tried to use the OTel Java SDK (version `1.28.0`) in an app with code obfuscation and I didn't get any errors (which I remember getting some time ago). So I'm not sure if it's due to some code refactoring in the OTel Java SDK that removed the usage of reflection or something like that. But it seems like we no longer need to add proguard rules, at least not for now.

GerardPaligot:
Maybe something change from 2023 but today, if I didn't add these proguard rules in my project with OpenTelemetry Android, I got a crash on the obfuscation:

```
-keep class io.opentelemetry.exporter.** { *; }
-keep class io.grpc.** { *; }
-keep class com.google.protobuf.** { *; }
-keep class com.fasterxml.jackson.core.** { *; }
-dontwarn com.google.auto.value.AutoValue$CopyAnnotations
-dontwarn com.google.auto.value.extension.memoized.Memoized
-dontwarn io.grpc.**
-dontwarn com.fasterxml.jackson.core.**
```

Can we consider to add them in opentelemetry-android to avoid this configuration in consumers?

LikeTheSalad:
Thank you for the heads-up, I'll take a look at it.

LikeTheSalad:
I think the reason might be that they've [enabled R8 full mode by default](https://android.googlesource.com/platform/tools/base/+/379d2f864025130c01bb6ded1e840229dff25d6a), I'm not fully sure what's the AGP version where the switch was made from disabled by default, though it seems like the tests we've run so far were counting on it being disabled only. I'll make some changes and add some automatic checks for R8 full mode.

marandaneto:
> I think the reason might be that they've [enabled R8 full mode by default](https://android.googlesource.com/platform/tools/base/+/379d2f864025130c01bb6ded1e840229dff25d6a), I'm not fully sure what's the AGP version where the switch was made from disabled by default, though it seems like the tests we've run so far were counting on it being disabled only. I'll make some changes and add some automatic checks for R8 full mode.

https://developer.android.com/build/releases/past-releases/agp-8-0-0-release-notes#default-changes >= v8 is enabled by default

LikeTheSalad:
[...]

PR #685, opened 2024-11-07
R8 consumer rules

LikeTheSalad:
Related to #83

### DIFF
--- a/core/consumer-rules.pro
+++ b/core/consumer-rules.pro
@@ -2,3 +2,11 @@
 -dontwarn com.google.auto.value.AutoValue$CopyAnnotations
 -dontwarn com.google.auto.value.AutoValue
 -dontwarn com.google.auto.value.extension.memoized.Memoized
+-dontwarn io.grpc.Channel
+-dontwarn io.grpc.MethodDescriptor$Builder
+-dontwarn io.grpc.MethodDescriptor$Marshaller
+-dontwarn io.grpc.MethodDescriptor$MethodType
+-dontwarn io.grpc.MethodDescriptor
+-dontwarn io.grpc.stub.AbstractFutureStub
+-dontwarn io.grpc.stub.AbstractStub$StubFactory
+-dontwarn io.grpc.stub.AbstractStub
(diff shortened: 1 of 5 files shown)


core/consumer-rules.pro
```diff
@@ -0,0 +1,12 @@
+-dontwarn com.google.auto.value.AutoValue$Builder
```
LikeTheSalad, line 1: [Open question] Is there a way we can easily automate checking whether it's ok to discard these warnings, or whether we should instead keep these classes from getting modified/removed?

I ran some local tests by running the app and I didn't get any crashes due to missing classes, so I think it's ok to just use `dontwarn` for now, but I'm concerned that in the future we might actually have to keep classes instead. The only automated solution I can see now would be to create Android functional tests that would run in an emulator, though that's very slow to run in GH actions. However, if it's the only way, we should add them.

marandaneto: I think reflection is the only problem, if auto value does reflection, better to have proper testing in place (I am not sure it does tho)

LikeTheSalad: Makes sense, thanks! AutoValue doesn't work with reflection so it seems like it should be fine.

demo-app/src/main/java/io/opentelemetry/android/demo/OtelDemoApplication.kt
```diff
@@ -62,6 +61,26 @@ class OtelDemoApplication : Application() {
...
+            }
+            .addLogRecordExporterCustomizer {
+                OtlpGrpcLogRecordExporter.builder().build()
+            }
+
+        // This is an overly-cautious measure to prevent R8 from discarding away the whole method
+        // in case it identifies that it's actually not doing anything meaningful.
+        if (System.currentTimeMillis() < 0) {
+            print(builder)
+        }
```
marandaneto, line 83: I guess we can remove this now?

marandaneto (approved):

demo-app/src/main/java/io/opentelemetry/android/demo/OtelDemoApplication.kt
breedx-splk: @LikeTheSalad I'm going to merge this because it looks good and this change isn't in the critical path, but hopefully you can revisit/respond to it. Thanks!

Issue #682, opened 2024-11-05
Make coreLibraryDesugaring optional (when minSdk >= 26)


hannesbolmstedt:
coreLibraryDesugaring should only be required for Android API levels lower than 26 (this is stated both in the README.md and VERSIONING.md).

However, `isCoreLibraryDesugaringEnabled = true` is currently set in both [otel.android-app-conventions.gradle.kts](https://github.com/open-telemetry/opentelemetry-android/blob/main/buildSrc/src/main/kotlin/otel.android-app-conventions.gradle.kts) and [otel.android-library-conventions.gradle.kts](https://github.com/open-telemetry/opentelemetry-android/blob/main/buildSrc/src/main/kotlin/otel.android-library-conventions.gradle.kts) making it impossible to import the project without enabling coreLibraryDesugaring, even when using a minSdk >= 26 (correct me if I am wrong).

My suggestion is to remove these two lines in order to make coreLibraryDesugaring optional.

marandaneto:
I use `ru.vyarus.animalsniffer` as well so it is def. a good thing, another option is to have a sample project (just a template) with low minSdk that compiles on CI, so if compilation breaks, you know you have used something you should not, it's not guaranteed that it won't break at runtime though.

bidetofevil:
Yeah I don't think it's necessary - the app is responsible for deciding whether desugaring is necessary, so the SDK doing it is a bit of an over reach. I'd remove even the library dependency and let the app take care of that. A warning or failure during build time or in CI would be nice, but not strictly necessary IMO if we've documented this.

breedx-splk:
> I use `ru.vyarus.animalsniffer` as well so it is def. a good thing, another option is to have a sample project (just a template) with low minSdk that compiles on CI, so if compilation breaks, you know you have used something you should not, it's not guaranteed that it won't break at runtime though.

Yeah I also see the value in this. Maybe if it's slow we just do it nightly, but it seems like catching the obvious/easy stuff automatically is helpful.

hannesbolmstedt:
Closing as completed, thank you

breedx-splk:
Lemme click that. :) Reopen if wanted/needed.

PR #701, opened 2024-11-19
Remove core library desugaring from android lib conventions

surbhiia:
As discussed in today's sig meeting about issue #682, I removed the added by default core library desugaring from android lib conventions. 

Project assembles fine but `./gradlew check `errors out on `lintDebug` task with 56 instances of requiring API level 24/26 or desugaring, as our minSdk = 21 :

Lint errors

We [earlier](https://github.com/open-telemetry/opentelemetry-android/blob/0e1d3fd9e4a3b83855facfecfba7c62532895250/instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlReplacements.java#L109) used a `@SuppressLint('NewApi')` annotation to suppress such warning instead of creating a lint baseline file (as maintaining that is tedious) : 

Screenshot 2024-11-19 at 11 41 30 AM

@RequiresApi might not be correct here as it does not necessarily need that API level but can do with desugaring which is what we're going with here. 

So, I think we need to add @SuppressLint('NewApi')` annotation at all 56 places? Wanted to cross check before going forward with that :D

### DIFF
--- a/instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlReplacements.java
+++ b/instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlReplacements.java
@@ -76,7 +76,6 @@ public class HttpUrlReplacements {
         return replace(connection, () -> connection.getContentLength());
     }
 
-    @SuppressLint("NewApi")
     public static long replacementForContentLengthLong(URLConnection connection) {
         return replace(connection, () -> connection.getContentLengthLong());
     }
--- a/buildSrc/src/main/kotlin/otel.android-library-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.android-library-conventions.gradle.kts
@@ -27,7 +27,6 @@ android {
     compileOptions {
         sourceCompatibility(javaVersion)
         targetCompatibility(javaVersion)
-        isCoreLibraryDesugaringEnabled = true
     }
 
     kotlinOptions {
@@ -53,5 +52,4 @@ dependencies {
     testRuntimeOnly(libs.findLibrary("junit-platform-launcher").get())
     testImplementation(libs.findLibrary("opentelemetry-sdk-testing").get())
     testImplementation(libs.findLibrary("androidx-junit").get())
-    coreLibraryDesugaring(libs.findLibrary("desugarJdkLibs").get())
 }
--- a/buildSrc/src/main/kotlin/otel.android-library-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.android-library-conventions.gradle.kts
@@ -22,6 +22,7 @@ android {
         // we rely on dependabot for dependency updates
         disable.add("GradleDependency")
         disable.add("AndroidGradlePluginVersion")
+        disable.add("NewApi")
     }
 
     compileOptions {


marandaneto:
If you want to make desugaring optional (and keeping the same features), you'd need to do runtime checks and use alternative classes/methods eg:

```
if API >= X
  do this
else
  do that
```

if that's the case, we'd need to check the otel-java as well, see https://developer.android.com/studio/write/lint and `checkDependencies = true`, maybe the dependency uses stuff that requires desugaring and we are not aware of.

if desugaring is mandatory for this SDK, then it's just safe to suppress all linters.

surbhiia:
Thanks @marandaneto for your feedback! 

oTel-java dependencies do cause us to require desugaring for API levels below 26, like this issue that often comes up: 
` java.lang.NoSuchMethodError: No static method stringKey(Ljava/lang/String;)Lio/opentelemetry/api/common/AttributeKey; `

Adding checkDependencies = true will increase the time it takes for the ./gradlew check command. 

I think we can go with [clearly documenting the requirement](https://github.com/open-telemetry/opentelemetry-android/blob/main/README.md#getting-started), [keeping the responsibility of enabling desugaring to the apps ](https://github.com/open-telemetry/opentelemetry-android/issues/682#issuecomment-2463845011) if they're using lower API levels and suppressing these lint errors in our libraries. What do you think?

LikeTheSalad:
Thank you for putting up this PR, @surbhiia!

> if desugaring is mandatory for this SDK, then it's just safe to suppress all linters.

It's mandatory for projects with minSdk lower than 26, this is something we can't easily get away from since the [Java upstream](https://github.com/open-telemetry/opentelemetry-java) project makes use of a lot of Java 8 tools that are only available on Android 26+ (or lower but with corelibs).

> Adding checkDependencies = true will increase the time it takes for the ./gradlew check command.

This shouldn't be necessary as the Java upstream project uses [animal sniffer](https://github.com/open-telemetry/opentelemetry-java/blob/main/animal-sniffer-signature/build.gradle.kts#L31) to ensure they don't use apis that aren't available in the corelibs.

> I think we can go with [clearly documenting the requirement](https://github.com/open-telemetry/opentelemetry-android/blob/main/README.md#getting-started), https://github.com/open-telemetry/opentelemetry-android/issues/682#issuecomment-2463845011if they're using lower API levels and suppressing these lint errors in our libraries. What do you think?

I agree, although since we're going to suppress the "NewApi" lint error in this project, we should replace it with a similar approach, ideally the same as the Java upstream repo by implementing animal sniffer checks to ensure we don't add code that's not supported by the corelibs. The animal sniffer plugin doesn't currently support android lib projects though, which is why I [created this PR](https://github.com/xvik/gradle-animalsniffer-plugin/pull/99) to enable that support, so as soon as that code is released I'm planning to implement it here to ensure we stay compliant with android apis.

In the meantime, I think it's fine to suppress the "NewApi" lint checks, however, I'd wait until we have a replacement before releasing a new version of OTel Android to be fully sure that we're not going to break any consumer projects.

> So, I think we need to add @SuppressLint('NewApi')` annotation at all 56 places?

While that will work, I would prefer to not use the annotation because it might be tedious having to remember it every time we are going to use apis that are available via the corelib. Instead I would try to add the suppression in the `otel.android-library-conventions.gradle.kts` file using the [lint block](https://developer.android.com/reference/tools/gradle-api/7.3/com/android/build/api/dsl/Lint). I haven't tested this, but probably something like this should work:

```kotlin
android {
   lint {
      disable += ['NewApi']
   }
}
```

Maybe that's not the exact way to set it up because I'm not sure if the examples are using kotlin DSL or groovy, but it gives the overall idea. If this works, it'd be nicer because we wouldn't have to add a lot of annotations and we also wouldn't need to create an XML file for lint rules.

marandaneto:
> Adding checkDependencies = true will increase the time it takes for the ./gradlew check command.

Well yes, but this is a linter that would prevent runtime checks, if otel-java adds a method that is available only on eg Android API 30 and you don't do runtime checks either on the Java or Android side, the app will crash at runtime.
I'm not sure animal sniffer can detect that (because the dependency is a compiled jar already), if it is, then it's redundant I agree.

marandaneto:
IMO an app that compiles successfully but requires a desugaring set otherwise it crashes at runtime and is always a bad call.
I understand changing that now would require some refactoring, just sharing my 2 cents :) - crashes will come because of that for sure.

marandaneto:
@LikeTheSalad yeah I understand that.
I say that because of https://github.com/open-telemetry/opentelemetry-android/pull/701#issuecomment-2489887167

This is a runtime error, animal sniffer will catch it during the SDK build, but not when people's apps are built, so the issue persists.

Right now if they don't have desugaring enabled, they will either:

1. read the docs and know they need it (this is not a common behavior) - leading to 2.
2. face a crash - leading to 3.
3. find out they need desugaring enabled.

As I've seen a few threads and issues related to not having desugaring enabled already, that proves that a subset of users will have a bad experience after installing the SDK or even after rolling out to production which is even worse.

LikeTheSalad:
> I say that because of https://github.com/open-telemetry/opentelemetry-android/pull/701#issuecomment-2489887167

Got it, I didn't see that in detail.

@surbhiia - regarding this part:

> like this issue that often comes up:
> java.lang.NoSuchMethodError: No static method stringKey(Ljava/lang/String;)Lio/opentelemetry/api/common/AttributeKey;

I'm a bit confused by it because it doesn't seem to be related to desugaring. Usually, desugaring-related issues show missing JDK classes or methods, though the method you mentioned should come from the OTel Java lib instead, so I'm wondering if that  "NoSuchMethodError: No static method stringKey" message might come from another type of issue, such as the use of reflection in an obfuscated build?

surbhiia:
> @surbhiia - regarding this part:
> 
 > like this issue that often comes up:
 > java.lang.NoSuchMethodError: No static method stringKey(Ljava/lang/String;)Lio/opentelemetry/api/common/AttributeKey;
> 
> I'm a bit confused by it because it doesn't seem to be related to desugaring. Usually, desugaring-related issues show missing JDK classes or methods, though the method you mentioned should come from the OTel Java lib instead, so I'm wondering if that "NoSuchMethodError: No static method stringKey" message might come from another type of issue, such as the use of reflection in an obfuscated build?

[This](https://issuetracker.google.com/issues/334281968) thread described the issue nicely. I took another look at it - looks like it is an issue in dexing when desugaring is enabled. So, you're right, it is not directly related to missed desugaring.

Given that, I hope the following holds true:

> otel-java does check that they're not using unsupported android corelib apis with animal sniffer checks, so (at least from the side of OTel Java) we shouldn't need to worry about it.

surbhiia:
> I would prefer to not use the annotation because it might be tedious having to remember it every time we are going to use apis that are available via the corelib. Instead I would try to add the suppression in the otel.android-library-conventions.gradle.kts file using the [lint block](https://developer.android.com/reference/tools/gradle-api/7.3/com/android/build/api/dsl/Lint).

Disabled NewApi lint check in android lib conventions. Thanks @LikeTheSalad , this was a quicker change :) 

> In the meantime, I think it's fine to suppress the "NewApi" lint checks, however, I'd wait until we have a replacement before releasing a new version of OTel Android to be fully sure that we're not going to break any consumer projects.

We can wait on merging this PR if we think we might have a release before we have chance to incorporate animal sniffer otherwise we can merge it now itself. :)

surbhiia:
> As I've seen a few threads and issues related to not having desugaring enabled already, that proves that a subset of users will have a bad experience after installing the SDK or even after rolling out to production which is even worse.

Can we perhaps add a non-missable callout regarding the desugaring requirement in the README. Hopefully that'll help. It's in the [Getting Started ](https://github.com/open-telemetry/opentelemetry-android/blob/main/README.md#getting-started) section already but we can add a better heading to the desugaring note to call attention to it like "IMPORTANT REQUIREMENT".  

Also thinking that we might not always have ways to support lower API levels everywhere, where we have newer features / APIs used. Also, graceful handling of just logging or preventing runtime crash might be problematic as app users would not know why certain functionality doesn't work as expected until they dig into the logs.

instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlReplacements.java
```diff
@@ -76,7 +76,6 @@ public static int replacementForContentLength(URLConnection connection) {
         return replace(connection, () -> connection.getContentLength());
     }
 
-    @SuppressLint("NewApi")
```
marandaneto, line 79: This is a simple example that would require:

```
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
	return replace(connection, () -> connection.getContentLengthLong());
} else {
	return -1
}
```

And the linter wouldn't need to be suppressed but its out of the scope of this PR, just sharing.

marandaneto:
I'm still not a big fan of silencing the linter in the hopes that animalsnifer (a 3rd party) would prevent all the cases that the linter would catch, afaik animal sniffer is always behind at least for new releases (its well maintained at least).
animalsnifer IMO is a complementary tool.
Approaches like [this](https://github.com/open-telemetry/opentelemetry-android/pull/701/files#r1853431253) is way more efficient IMO, but it has a lil performance impact at runtime, its just an `if` though.
I'll let @LikeTheSalad do the green mark :D

instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlReplacements.java
LikeTheSalad: That's right, I wasn't hoping for animal sniffer to be a replacement for these kinds of approaches which we should still use IMO. We do so already in some services like [this one](https://github.com/open-telemetry/opentelemetry-android/blob/c10adbaf11dbc415190d5fdf8dcf5183485d3ef6/core/src/main/java/io/opentelemetry/android/internal/services/CacheStorage.java#L47) for example.

The example highlighted here is a special one because the class `HttpUrlReplacements` is meant to be used during bytecode instrumentation to replace existing calls to the original methods, so for this case, the method `HttpUrlReplacements.replacementForContentLengthLong` will only be used in apps that are already calling the original method [getContentLengthLong()](https://developer.android.com/reference/java/net/URLConnection#getContentLengthLong()), so it implicitly means that their minSdk is at least 24, otherwise they'd get an AGP compilation error.

That being said, I'm not fully sure what would happen in this case when animal sniffer is involved, I'm assuming it will complain because the signature we're going to use is of API 21 and also because I don't think this method has nothing to do with the desugaring corelibs either, so what I can propose is to follow what @surbhiia mentioned earlier about waiting for the animal sniffer integration before removing the "NewApi" checks just to make sure it behaves the same way and wouldn't just ignore these types of scenarios which are a potential runtime crash in most cases.

marandaneto: makes sense.

LikeTheSalad:
> I'm still not a big fan of silencing the linter in the hopes that animalsnifer (a 3rd party) would prevent all the cases that the linter would catch, afaik animal sniffer is always behind at least for new releases (its well maintained at least). animalsnifer IMO is a complementary tool. Approaches like [this](https://github.com/open-telemetry/opentelemetry-android/pull/701/files#r1853431253) is way more efficient IMO, but it has a lil performance impact at runtime, its just an `if` though. I'll let @LikeTheSalad do the green mark :D

I think we should be pretty confident about animal sniffer capabilities before making the switch, so as I mentioned [here](https://github.com/open-telemetry/opentelemetry-android/pull/701#discussion_r1853562754), we should probably wait for it to be implemented to make sure it covers the same cases as the "NewApi" lint checks. 

> afaik animal sniffer is always behind at least for new releases (its well maintained at least).

Animal sniffer itself doesn't do the checks, it relies on "signatures" that contain the types and methods that are "allowed". You're right that not all the signatures are up-to-date as soon as a new release is made, this is actually an [issue we had in the past](https://github.com/open-toast/gummy-bears/issues/25) where google released a new corelib version with some changes we needed but the animal sniffer signature we were using for corelibs wasn't up-to-date. However, we later found out that we could [create our own signatures](https://github.com/open-telemetry/opentelemetry-java/pull/5973) by using google's corelib artifact directly, which allows us to stay updated right away. So there's that workaround if it helps.

Ultimately, I would prefer to use native android checks as we do right now, but unfortunately it throws [these false-positives](https://github.com/open-telemetry/opentelemetry-android/issues/682) that are very confusing to users with minSdk >= 26 (and rightfully so) so that's why I'm willing to give animal sniffer a try, which I hope it can do the same but without enforcing desugaring when it's not needed. Actually, to be honest I'd prefer we wouldn't have to add support to apis < 26 😅 but I know that's not feasible for now, so I hope people understand that they need to do these extra steps of configuring corelib desugaring when targeting apis < 26, though not when >= 26, that doesn't make sense.

bencehornak:
As the owner of an app, which doesn't need desugaring I couldn't agree more with the objective of this PR, and having it is a must have for trying out the Otel Android SDK.

@LikeTheSalad / @surbhiia what's the next step to move this PR forward? What is your preferred way to resolve the linter rules vs Animal sniffer dilemma?

LikeTheSalad:
Hi @bencehornak 

It seems like [this month](https://github.com/xvik/gradle-animalsniffer-plugin/issues/8#issuecomment-2563395945) we might get an animal sniffer release with Android support. As soon as that's available, we can safely merge this PR with that alternative verification method. We could also just remove the Android native checks and leave no other validations in place in the meantime until animal sniffer is ready, though given that it seems like the alternative check will be ready soon I don't think that should be needed.

juanchosaravia:
I completely agree with @bencehornak, and I'm extremely excited about the news, @LikeTheSalad! I'm currently conducting an evaluation for a client, and this will be incredibly useful, as they're very concerned about the necessity to desugar the app despite targeting a greater minsdk.

bencehornak:
And gradle-animalsniffer-plugin@2.0.0 just released :tada:
https://github.com/xvik/gradle-animalsniffer-plugin/releases/tag/2.0.0

juanchosaravia:
Does it mean we will have this PR merged soon?

LikeTheSalad (approved):
Cheers! I'll follow-up with animal sniffer checks [here](https://github.com/open-telemetry/opentelemetry-android/issues/770).

juanchosaravia:
I'm using the latest 0.9.1-alpha version and it still force me to use `isCoreLibraryDesugaringEnabled = true`. I thought this was fixed

bencehornak:
@juanchosaravia it will be included in the next release (0.10.0-alpha) coming very soon.

@breedx-splk I noticed that this PR is not referenced in the changelog (#870).

PR #709, opened 2024-11-22
Prevent ANR during SDK initialization

ansman:
## Description
When initializing the OpenTelemetry Android SDK with disk buffering enabled, we discovered that synchronous disk space checks were causing ANRs in production. These checks occur during the creation of disk buffering exporters, specifically in `DiskManager.getMaxFolderSize()`, which makes blocking IPC calls through `StorageManager.getAllocatableBytes()` on the main thread. The issue manifests in the following ANR stacktrace:
```
android.os.BinderProxy.transact (BinderProxy.java:662)
android.os.storage.IStorageManager$Stub$Proxy.getAllocatableBytes (IStorageManager.java:2837)
android.os.storage.StorageManager.getAllocatableBytes (StorageManager.java:2414)
android.os.storage.StorageManager.getAllocatableBytes (StorageManager.java:2404)
io.opentelemetry.android.internal.services.CacheStorage.getAvailableSpace (CacheStorage.java:66)
io.opentelemetry.android.internal.services.CacheStorage.ensureCacheSpaceAvailable (CacheStorage.java:50)
io.opentelemetry.android.internal.features.persistence.DiskManager.getMaxFolderSize (DiskManager.kt:58)
io.opentelemetry.android.OpenTelemetryRumBuilder.createStorageConfiguration (OpenTelemetryRumBuilder.java:338)
io.opentelemetry.android.OpenTelemetryRumBuilder.build (OpenTelemetryRumBuilder.java:286)
```

## Solution
To fix this we moved initialization to run on a background executor and buffer the data in memory until it completes.

The process works like this:
1. Initialize the SDK with `BufferDelegatingExporter` instances that can immediately accept telemetry data.
2. Move exporter initialization off the main thread.
3. Once async initialization completes, flush buffered signals to initialized exporters and delegate all future signals.

The primary goal of this solution is to be unobtrusive and prevent ANRs caused by initialization of disk exporters, while preventing signals from being dropped.

## Testing
We have added unit tests to cover the buffering, delevation, and RUM building. We've also verified this with both disk enabled and disk disabled.


ansman:
FYI we are working on getting the CLA signed.

ansman:
@breedx-splk CLA signed

core/src/main/java/io/opentelemetry/android/export/BufferedDelegatingExporter.kt
```diff
@@ -0,0 +1,100 @@
...
+ * An in-memory buffer delegating signal exporter that buffers signal in memory until a delegate is set.
+ * Once a delegate is set, the buffered signals are exported to the delegate.
+ *
+ * The buffer size is set to 5,000 by default. If the buffer is full, the exporter will drop new signals.
+ */
+internal abstract class BufferedDelegatingExporter<T, D>(private val bufferedSignals: Int = 5_000) {
+    @Volatile
+    private var delegate: D? = null
+    private val buffer = arrayListOf<T>()
+    private val lock = Any()
```
breedx-splk, line 21: I'm a little concerned/confused about the concurrency handling here. I appreciate that effort was put into this, but I don't completely follow it. 

1. The delegate is both volatile and also sometimes (but not always) guarded by the `lock`. Do we really need both?  Can we perhaps simplify this by removing volatile and always use the `lock` when accessing the `delegate`?
2. The buffer that stores the telemetry doesn't seem to have any concurrency provisions...and I think maybe it should. Might be as simple as using a `ConcurrentLinkedQueue` or something?

core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
```diff
@@ -353,8 +396,11 @@ public OpenTelemetryRumBuilder setServiceManager(ServiceManager serviceManager)
      * If not specified, the default schedule exporter will be used.
      */
     public OpenTelemetryRumBuilder setExportScheduleHandler(
-            ExportScheduleHandler exportScheduleHandler) {
-        this.exportScheduleHandler = exportScheduleHandler;
+            @NonNull ExportScheduleHandler exportScheduleHandler) {
+        requireNonNull(exportScheduleHandler, "exportScheduleHandler cannot be null");
```
breedx-splk, line 400: I'm not sure we need both the annotation and the runtime check, do we?

breedx-splk (commented):
Thanks @ansman for taking this on, I think this is going to be really helpful...and there are bound to be a few other areas that are problematic at startup, so I appreciate you both shining some light on this and for going the distance in submitting this PR.

core/src/main/java/io/opentelemetry/android/export/BufferedDelegatingExporter.kt
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /**
+     * Buffers the given data if the delegate has not been set, otherwise exports the data to the delegate.
+     *
+     * @param data the data to buffer or export
+     */
+    protected fun bufferOrDelegate(data: Collection<T>): CompletableResultCode =
+        withDelegateOrNull {
+            if (it != null) {
```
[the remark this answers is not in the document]
ansman: No problem, will do

core/src/main/java/io/opentelemetry/android/export/BufferDelegatingLogExporter.kt
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * An in-memory buffer delegating log exporter that buffers log records in memory until a delegate is set.
+ * Once a delegate is set, the buffered log records are exported to the delegate.
+ *
+ * The buffer size is set to 5,000 log entries by default. If the buffer is full, the exporter will drop new log records.
+ */
+internal class BufferDelegatingLogExporter(
+    maxBufferedLogs: Int = 5_000,
+) : BufferedDelegatingExporter<LogRecordData, LogRecordExporter>(bufferedSignals = maxBufferedLogs),
```
breedx-splk, line 20: I've experience so much pain from unnecessarily complicated class inheritance problems over my career, so it will often make me sad to see inheritance used primarily as a means to code reuse. Delegation is hard here without a common interface for log/span export, I get it, but maybe we can collectively circle back on this after this PR work is complete.

core/src/main/java/io/opentelemetry/android/export/BufferedDelegatingExporter.kt
ansman: Sure, happy to elaborate.

1. Volatile is needed to correctly implement the double check pattern. You can read more about it [here](https://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html).
2. It's not entirely obvious but the only time the block is called with a `null` delegate is when it's inside the `synchronized` block so no additional synchronization is required. This code was written with high performance in mind but this could perhaps be documented if nothing else.

core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
ansman: I find it best to be clear with your assertions, otherwise it might crash down the line. FWIW Kotlin implements non nullable parameters like this. I can remove if you prefer.

core/src/main/java/io/opentelemetry/android/export/BufferDelegatingLogExporter.kt
ansman: Yeah agreed. In this case the main benefits is that you don't have to implement some methods but the delegation pattern would probably work with little overhead if you prefer we use that instead.

core/src/main/java/io/opentelemetry/android/export/BufferedDelegatingExporter.kt
```diff
@@ -0,0 +1,100 @@
...
+                exportToDelegate(it, data)
+            } else {
+                val amountToTake = bufferedSignals - buffer.size
+                buffer.addAll(data.take(amountToTake))
+                CompletableResultCode.ofSuccess()
```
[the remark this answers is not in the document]
ansman: Yeah I suppose, but there might be _a lot_ of warnings. But I you probably want to know. How do you prefer we log this? I would avoid logging using `Log` since logging in production is frowned upon.

core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
```diff
@@ -279,6 +286,56 @@ public OpenTelemetryRum build() {
...
+                        application,
+                        sdk,
+                        timeoutHandler,
+                        sessionManager,
+                        config.shouldDiscoverInstrumentations(),
+                        getServiceManager());
+
+        // AsyncTask is deprecated but the thread pool is still used all over the Android SDK
+        // and it provides a way to get a background thread without having to create a new one.
+        android.os.AsyncTask.THREAD_POOL_EXECUTOR.execute(
```
LikeTheSalad, line 322: Selling the use of deprecated tools is tough, though I understand the concern about creating many threads in Android. If this PR gets merged I think I can help address this by creating a work for `PeriodicWorkService` that executes once. It's an internal tool that allows us to run stuff in the background by reusing a single thread. Alternatively, using the default Kotlin coroutine dispatcher as an executor via `ExecutorsKt.asExecutor(Dispatchers.getDefault()).execute(...);` could be another way to do the same but without having to use deprecated tools.

core/src/main/java/io/opentelemetry/android/export/BufferedDelegatingExporter.kt
LikeTheSalad: My understanding of the DCL pattern is that it's an anti-pattern. I agree with @breedx-splk that it's probably best to use synchronization only.

LikeTheSalad (commented):
Thank you for taking the time to contribute 🙏 I like the overall idea. I left some questions and suggestions, so please have a look when you get a chance.

core/src/main/java/io/opentelemetry/android/export/BufferedDelegatingExporter.kt
```diff
@@ -0,0 +1,100 @@
...
+package io.opentelemetry.android.export
+
+import io.opentelemetry.sdk.common.CompletableResultCode
+import java.util.concurrent.atomic.AtomicBoolean
+
+/**
+ * An in-memory buffer delegating signal exporter that buffers signal in memory until a delegate is set.
+ * Once a delegate is set, the buffered signals are exported to the delegate.
+ *
+ * The buffer size is set to 5,000 by default. If the buffer is full, the exporter will drop new signals.
```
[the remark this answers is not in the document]
ansman: It was arbitrarily picked. It was chosen to be high enough to handle most logging needs while not taking up too much space. It can absolutely be changed

core/src/main/java/io/opentelemetry/android/export/BufferedDelegatingExporter.kt
ansman: The main reason this was chosen is that, as a library, it's important to think about performance, especially in low level functionality like this. Users could potentially be logging a lot and synchronizing would incur a cost for every log entry and albeit tiny, could add up.

I also wouldn't say it's an anti pattern. It's well established and used frequently (for example by [Dagger](https://github.com/google/dagger/blob/master/java/dagger/internal/DoubleCheck.java)) It's complicated and should be avoided unless needed, but I think it's warranted here.

Unless you want to also lock while dispatching to the delegate (which would be a risky idea) you'd end up with fairly similar code, just lacking the outer check and the volatile statement:
```kotlin
val delegate = synchronized(lock) {
  delegate ?: run {
    writeToBuffer(entry)
    return
  }
}
delegate.export(entry)
```

core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
```diff
@@ -94,8 +96,13 @@ public final class OpenTelemetryRumBuilder {
...
 
-    @Nullable private ServiceManager serviceManager;
-    @Nullable private ExportScheduleHandler exportScheduleHandler;
+    private final Object lock = new Object();
+
+    // Writes guarded by "lock"
+    @Nullable private volatile ServiceManager serviceManager;
+
+    // Writes guarded by "lock"
+    @Nullable private volatile ExportScheduleHandler exportScheduleHandler;
```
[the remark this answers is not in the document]
ansman: It was mostly due to an uncertainty if they were allowed to be changed after calling `build`. Happy to remove this and perhaps add a check to ensure you're not calling these after calling `build`.

core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
ansman: Yes it's unfortunate. But like the comment mentions, the thread pool is actually still used in a lot of places, including by AndroidX and Android internally. It felt worthwhile to not spin up new threads.

I was under the impression that coroutines is not used in this library, otherwise I would have used the coroutine dispatchers.

core/src/main/java/io/opentelemetry/android/export/BufferedDelegatingExporter.kt
ansman: That being said, this is your library and I'm happy to change this to whatever you think is best. I wanted to offer an explanation of why we chose this pattern but ultimately you're the ones who have to main this so the call is yours to make.

LikeTheSalad: Got it, thank you for the explanation. I agree that we should make an effort to ensure that this library doesn't have a poor performance and I appreciate the thoughtfulness put into this contribution in that regard! My concern is that, generally speaking, attempting to do premature optimizations might make the code a bit more difficult to understand and in the worst case might cause some other issues that are difficult to spot, especially for concurrency-related code as liveness issues might arise under specific conditions only, so that's why I think we should start simple and then gradually make things more complex as the need arises. In this case, if during your tests you've spotted some considerable performance-related issues that are only addressable by using this approach, then I think we can keep it, otherwise, I'd prefer to start with the option of using synchronization only.

core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
breedx-splk: I don't think it's a big deal, I just thought it was surprising to see both. Let's continue with it. 👍🏻

core/src/main/java/io/opentelemetry/android/export/BufferedDelegatingExporter.kt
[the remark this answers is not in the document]
breedx-splk: Seems like a reasonable default to me. 🍰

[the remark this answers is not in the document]
LikeTheSalad: I don't have a clear reference of how much memory each signal can take, mostly because it also depends on its attributes/general content, though iirc a simple proto-serialized span size in bytes can get to around 700 (I might be very wrong though because last time I checked was quite long ago), let's say it's 500 bytes per span for the sake of the example, then we might be talking about ~2.5 mb for 5000 of those, plus other ~2.5 mb for the logs, we might be talking about ~5 mb total in the worst case, though again it depends on the size of each signal and also on how accurately I remember these rough proto sizes and how they translate into runtime memory, so we should take this with a grain of salt. 

However, before getting into the weeds of how much memory is available for apps/OS version/hardware/etc, I'd like to take a step back and try to figure out for how long do we think the agent would need to use this mechanism of storing signals in memory? If we base it on the ANR issue that's the reason for these changes, it means that it would roughly take as long as it takes `android.os.storage.StorageManager.getAllocatableBytes` to execute, which I haven't measured but I'm guessing it should take less than a second, and, based on that assumption, I'm finding it difficult to see how 10000 signals can be created within that timeframe. I tend to err on the side of using as few resources as possible, so I'd prefer to choose a much lower number if possible. From a couple of local examples I've run (not real-world examples though) I think I've gotten at most ~20 signals created at app launch, considering it's not a real-world use case I'm conscious that the number might be higher, but 5000 seems a bit excessive. I might be missing something though, so I was wondering if you could share a bit more info on the kind of use-cases that we should look out for where this number might be needed?

core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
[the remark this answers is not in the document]
LikeTheSalad: Got it, yeah none of the setters should be called after calling `build()`, though I see your point that it's not impossible to do so right now as we don't have validations in place for it, so yeah I think we should add checks to prevent it from happening. We could add those validations here or as part of a follow-up PR too, so no worries if wouldn't like to address it now 👍 though I think we should add those validations for sure at some point.

core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
LikeTheSalad: Got it, that's fair enough. I think it's fine to leave it as is for now.

core/src/main/java/io/opentelemetry/android/export/BufferedDelegatingExporter.kt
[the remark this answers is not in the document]
ansman: Your analysis matches our own. Our goal was to pick a value high enough that nobody is likely to hit it but low enough to have little impact on memory usage.

We've seen that the binder call can take up to several seconds to complete on a low hardware phone that is currently busy with background services so I think it's fair to say that 5s is around a common worst case (something like P99).

We don't mind setting it lower but the risk of dropping events without any warning seems potentially very very bad. It's worth noting that we do not allocate more space than needed (or rather the array list doesn't). It doubles in size when more space is needed so you'd be logging more than 2500 spans to get the full 5000 capacity. And if you're logging that much you would likely trade 2.5MB (per exporter) of memory over dropping them I'd assume.

That being said, this value could definitely be lowered if you feel like it's absurdly high but I wanted to outline our thinking here. Let us know how you want to proceed.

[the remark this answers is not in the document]
LikeTheSalad: Fair points. I wasn't aware that this call could take up to 5 seconds to execute in some cases, and I was mostly curious to know about some use cases that would require that amount of signals within that timeframe. However, as you mentioned, the number is high enough for it to be difficult to get hit, and also, the memory isn't allocated upfront anyway, so it should be fine to leave it as is. Thanks for the explanation.

core/src/main/java/io/opentelemetry/android/export/BufferedDelegatingExporter.kt
bidetofevil: A bit late to the game, but I think it's reasonable to take extra care in handling thorny race conditions during transition. If the transition is unclean, there could be buffered data that gets lost, so locking this up, even with some more complicated code, is a fine thing to do. We might want to add more robust testing of edge cases where we simulate race conditions (good times), but I think it's beneficial overall to have something like this to increase the guarantee.

That said, are your concerns, @breedx-splk and @LikeTheSalad , more with readability or with the approach, specifically with the locking at every export call?

Another approach could be to lock only during the transition when the delegate is being set and not lock during export. When the transition is done, we can flush the in-memory data into the delegate, taking extra care on the in-memory queue to ensure that data coming in during the transition is picked picked. We just have to ensure that there isn't anything left in the in-memory queue after the transition, even if it comes in late.

What do you think, @ansman?

ansman: I have a rewrite that doesn't use double checking and I think it's fine. I don't think there is any reduction in edge cases with the double locking, it would be a pure performance improvement.

One thing that I do want to ask is if it's ok to call the delegate while holding the lock. This is risky because if the delegate ends up waiting for something that in turn tries to log something you'd get a deadlock. This is true with or without the double locking I believe. 

But on the other hand, calling the delegate without holding the lock can lead to out of order exports during initialization. I don't think the latter is a big deal perhaps and preferable over deadlocking. But It's also possible that this isn't a scenario that can actually happen so let me know how you want to proceed.

bidetofevil: There's no guarantee of ordering in terms of export, nor the order in which it's received in the Collector, so as long as everything that's buffered gets exported without an unreasonable delay, I think that's all the guarantee we'd need. And it eliminates any possibility of a deadlock by not requiring the lock to be held :-)

core/src/main/java/io/opentelemetry/android/export/BufferedDelegatingExporter.kt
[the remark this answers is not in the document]
breedx-splk: Let's also not forget that we can change this default in a future PR....there's nothing holding us to this number. 🙃

core/src/main/java/io/opentelemetry/android/export/BufferedDelegatingExporter.kt
LikeTheSalad: I agree with @bidetofevil, the order shouldn't matter in this case, the signal's timestamp should be used later when querying the data to ensure they're properly ordered. So I think the option that doesn't have the deadlock risk is the one we should go with.

core/src/main/java/io/opentelemetry/android/export/BufferedDelegatingExporter.kt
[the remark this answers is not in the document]
ansman: Fixed

core/src/main/java/io/opentelemetry/android/export/BufferedDelegatingExporter.kt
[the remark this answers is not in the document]
ansman: @breedx-splk What would you like to do here? I think logging is a good idea but not in production. Is there an existing solution for this already?

core/src/main/java/io/opentelemetry/android/export/BufferedDelegatingExporter.kt
```diff
@@ -0,0 +1,100 @@
...
+        exportToDelegate(delegate, buffer)
+        buffer.clear()
+        buffer.trimToSize()
+    }
+
+    private fun bufferedShutDown(): CompletableResultCode {
+        isShutDown.set(true)
+
+        return withDelegateOrNull {
+            if (it != null) {
```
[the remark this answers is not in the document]
ansman: Fixed

core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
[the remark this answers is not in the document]
ansman: Updated and checks added

ansman:
I've updated the PR with the feedback and everything except the logging has been addressed.

In addition the returned result will not be success any longer but rather resolve when the delegate is set.

core/src/main/java/io/opentelemetry/android/export/DelegatingExporter.kt
```diff
@@ -0,0 +1,161 @@
...
+            } else {
+                val throwable = other.failureThrowable
+                if (throwable == null) {
+                    fail()
+                } else {
+                    failExceptionally(throwable)
+                }
+            }
+        }
+    }
```
ansman, line 160: @LikeTheSalad, @breedx-splk Please verify that this does what you'd expect. It's meant to resolve the receiver with the same result as `other`.

core/src/main/java/io/opentelemetry/android/export/DelegatingExporter.kt
```diff
@@ -0,0 +1,161 @@
...
+        other.whenComplete {
+            if (other.isSuccess) {
+                succeed()
+            } else {
+                val throwable = other.failureThrowable
+                if (throwable == null) {
+                    fail()
+                } else {
+                    failExceptionally(throwable)
+                }
```
ansman, line 157: I'm pretty sure this can be replaced with
```suggestion
                failExceptionally(throwable)
```
But the docs for `failExceptionally` does not mention nullability so I opted to be safe

core/src/main/java/io/opentelemetry/android/export/BufferedDelegatingExporter.kt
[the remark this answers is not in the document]
breedx-splk: I still think that if/when this happens it's going to be challenging to see without ANY signs...so maybe doing a one-shot log is better than nothing. Kind-of a stupid amount of extra work, but something like this:

```
private val haveLoggedErrorOnce = AtomicBoolean(false)
...
if(haveLoggedErrorOnce.compareAndSet(false, true)) {
    Log.w(OTEL_RUM_LOG_TAG,"Telemetry buffer was filled before export delegate set...")
    Log.w(OTEL_RUM_LOG_TAG,"This has resulted in a loss of telemetry data!")
}
```

core/src/main/java/io/opentelemetry/android/export/DelegatingExporter.kt
breedx-splk: I think this works, yeah. Makes sense to me. @LikeTheSalad ?

core/src/main/java/io/opentelemetry/android/export/DelegatingExporter.kt
breedx-splk: Yeah that's fair. [The implementation](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/common/src/main/java/io/opentelemetry/sdk/common/CompletableResultCode.java#L129) looks like it would be fine as well, but I think without a `@Nullable` on the param or any javadoc about it, better to play it safe.

breedx-splk: I also opened https://github.com/open-telemetry/opentelemetry-java/pull/6963 for our future selves. :)

breedx-splk (approved):
Thanks @ansman for sticking this out, I like how it's come together! Would still like to see the log warning emitted, and I'll wait to let @LikeTheSalad do the merge with his approval, but I think things are looking solid now. Thanks for the contribution.

core/src/main/java/io/opentelemetry/android/export/BufferedDelegatingExporter.kt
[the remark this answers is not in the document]
ansman: I suppose these days it doesn't matter too much since logcat filtering is much better than it was. I wouldn't really mind logging every time honestly.

Also, you could imagine a small logging framework that otel ships with that you can hook into. I'd love to use open telemetry isefl to log this but if logcat is used there is no way to hook in. But this sounds like an issue for a later date.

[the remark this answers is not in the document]
breedx-splk: Thanks yeah I'm fine with doing the simpler thing and just logging it every time, but I thought you were reluctant. 🙃 
Anyway, yeah, probably worth mentioning that we've had a log related issue here https://github.com/open-telemetry/opentelemetry-android/issues/142 for a year, looking for contributors who want to take that on.

core/src/main/java/io/opentelemetry/android/export/DelegatingExporter.kt
```diff
@@ -0,0 +1,171 @@
...
+ */
+
+package io.opentelemetry.android.export
+
+import android.util.Log
+import io.opentelemetry.android.common.RumConstants.OTEL_RUM_LOG_TAG
+import io.opentelemetry.api.internal.GuardedBy
+import io.opentelemetry.sdk.common.CompletableResultCode
+import java.nio.BufferOverflowException
+import kotlin.math.log
```
LikeTheSalad, line 13: nit: it doesn't seem like we need this import.

core/src/main/java/io/opentelemetry/android/export/DelegatingExporter.kt
LikeTheSalad: Looks good to me as well 👍

LikeTheSalad (approved):
Thank you for your contribution and patience, @ansman! 🙏  I'm looking forward to getting your PR merged, though it looks like there are some formatting issues that are making the build fail, if you could please update this PR after running `./gradlew spotlessApply` locally, I think that should do. I'm happy to proceed with the merge once the GH build passes :)

core/src/main/java/io/opentelemetry/android/export/BufferedDelegatingExporter.kt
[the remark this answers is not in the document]
ansman: If we're gonna log we might as well log haha

core/src/main/java/io/opentelemetry/android/export/DelegatingExporter.kt
ansman: Fixed

Issue #702, opened 2024-11-20
Modularize project so Android instrumentations can used by SDKs that provide their own OpenTelemetry instance


bidetofevil:
# Background

Conceptually, the `core` module provides serves three purposes:

1. Initializes an instance of `OpenTelemetry` with reasonable defaults 
2. Defines an API for Android instrumentations and orchestrates the starting of the configured instrumentations 
3. Provides a set of opinionated abstractions over the Android platform that instrumentations can depend on (i.e. Services)

As an app developer, using this project is straightforward: include the `android-agent` module, configure and start the SDK, profit.

However, as @surbhiia points out in #669 , vendors that have their own version of `android-agent` that are used with their own custom instrumentation can't use the instrumentations provided by this directly.

Part of that is intentional:- the platform abstraction services are a dependency, so they must be provided by the project to the instrumentations for them to work. But is there a good reason why the `OpenTelemetry` instance has to be created in `core`? To put in another way, does the use case outlined in #669 seems reasonable? I think the answer is yes.

This issue is about doing all the work to enable the use of instrumentations defined in this project independent from the initialization of the OTel SDK.

# Goals

- Restructure `core` so that purpose 1 listed about (i.e. initialize `OpenTelemetry` instance) is outside of its direct purview
- Refactor the initialization of the services so they are done lazily, in case they are not all needed for the specific set of instrumentation that are in use.

# Non-Goals

- Make the interfaces of the platform abstractions public, i.e. they should remain internal and not be part of the supported public API. In other words, those interfaces and implementations should not be explicitly depended on by app and they can change form version to version with no warning.
- Make the interface for the Android instrumentation public, i.e. apps can't create their own instances of instrumentation that are then managed by `core`. In other words, `AndroidInstrumentation` will NOT be a publicly supported API.

LikeTheSalad:
Thank you, @bidetofevil!

It'd be awesome if we could come up with a way to use instrumentations independently 👍 

You touch on different topics, some of which I fully understand and agree with (such as the part that reads: "Refactor the initialization of the services so they are done lazily, in case they are not all needed for the specific set of instrumentation that are in use") though I'm not fully sure I follow some others, which are the ones I'd like to expand on below:

> Restructure core so that purpose 1 listed about (i.e. initialize OpenTelemetry instance) is outside of its direct purview

[...]

PR #711, opened 2024-11-27
Extract modules from core

surbhiia:
As a first step towards #669 , I've extracted following 3 modules from core:

**:common**
- Holds classes which are required in :core, :services and some instrumentations too like RumConstants.
- Not published so keeping it internal. Added as implementation dependency everywhere. 
- Had to extract `networkattrs->data` and `CurrentNetworkAttributesExtarctor` here as it is utilized by both `:services` in `CurrentNetworkProvider `and `:core `in `NetworkAttributesSpanAppender. `
- Extracted `RuntimeDetailsExtractor` used by crash instrumentation here. 

**:services**
- The `services` folder in `core` is extracted as a module
- Moved `SystemTime` class here as well as it is a good contender for a shared utility. It is currently only used by the `PeriodicWorkService`. Alternatively, We can move it to :common module (Update - moved to `:common` now)
- Not published so keeping it internal. Also used internal keyword in package name. Added as implementation dependency everywhere. 

**:instrumentation:android-instrumentation** 
- AndroidInstrumentation AutoService interface and related classes
- This is published as currently we're allowing users to bring in their own custom instrumentations that implement this interface and add via the addInstrumentation() API we have exposed. Added as api dependency everywhere right now. 

**Other callouts:**
- `Preferences.java` in service module depended on the generated `BuildConfig.java` for LIBRARY_PACKAGE_NAME for getting the shared preference object. It does not make sense to generate `BuildConfig.java` in any other module other than :core as it is project specific. So, I've replaced the dependence by direct replacement with "io.opentelemetry.android" string. Also, it did not make sense to add this string to RumConstants and use from there.
- ~~`test/resources/META-INF/services` and `TestAndroidInstrumentation` is replicated right now in :core and :instrumentation:android-instrumentation modules as it is just providing an AndroidInstrumentation AutoService on the classpath for tests in both modules as those tests are specific to the individual modules (`AgentInitTest`,  `OpentelemetryRumBuilderTest`, `AndroidInstrumentationLoaderImplTest`).  (While writing this, it comes to my mind that we can perhaps move `TestAndroidInstrumentation` as a shared test class in test-common module.)~~

This is just a first step, we've a long way to go in determining the future possibilities of how we can support instrumentations to be used independently of this SDKs core (initializing and configuring) module. Those discussions are covered in following issues - #658, #669 , #702. 

Screenshots of the new modules structure from android studio project explorer:
Screenshot 2024-11-27 at 3 37 16 PM
Screenshot 2024-11-27 at 3 36 43 PM
Screenshot 2024-11-27 at 3 35 53 PM


common/src/main/java/io/opentelemetry/android/common/internal/features/networkattrs/CurrentNetworkAttributesExtractor.java
```diff
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package io.opentelemetry.android.internal.features.networkattrs;
+package io.opentelemetry.android.common.internal.features.networkattrs;
```
keciman, line 6: I suppose networkattrs is shortening of networkattributes. 
Personally I suggest to use networkattributes so it is more obvious, attrs has already a meaning in tech. 
I undestand these were original changes so it is up to you :)

keciman (approved):

LikeTheSalad (approved):
Thanks! I think it's a good start. I do believe we should move SystemTime to common though, as you mentioned:

> Moved SystemTime class here as well as it is a good contender for a shared utility. It is currently only used by the PeriodicWorkService. Alternatively, We can move it to :common module

It's pretty common having to use the current time and it's tricky to test those use cases without something like SystemTime. Here's [an example](https://github.com/open-telemetry/opentelemetry-android/pull/663/files#diff-ef0c17b34e141dfdff4b7093ce2d697347aa1ed85a313b5e21a3812195d35459R62) of a usage outside of the services scope where having it in common should help.

Other than that I think we can move forward with these changes, though I'll leave this PR open until next week to give a chance for people in the US to have a look too, as I've noticed that this week has been pretty quiet in OTel in general due to the holidays.

breedx-splk (approved):
This feels like progress. Thanks for contributing this!

PR #717, opened 2024-12-11
Refactor session manager

breedx-splk:
This relates to #563 and #571. Hopefully this is a more contained, smaller, reviewable unit.

It does the following:
* renames the `SessionManager` implementation to `SessionManagerImpl` and makes it internal
* extracts a `SessionManager` interface from the impl
* moves the session api to its own top-level module
* moves `SessionidTimeoutHandler` from java to kotlin and makes it internal

The intended next step is to add `SessionManager` to the `InstallationContext` and create a new instrumentation module that can generate session start/end events, as in #571.


marandaneto (approved):
since its mostly extracting and moving stuff around, looks all good, LGTM

PR #719, opened 2024-12-11
Add new session instrumentation

breedx-splk:
This will rebase after #717 and is basically a continuation of #571.

Creates a new instrumentation module that is responsible for generation `session.end` and `session.start` events, per the semantic conventions. In order to achieve this, the `SessionManager` is added to the `InstallationContext`, which (to me) seems reasonable that instrumentations might want to have access to the session.

It is currently NOT enabled by default through the agent (opt-in)...but I think we should reconsider this and should make it a default.


instrumentation/sessions/build.gradle.kts
```diff
@@ -0,0 +1,35 @@
...
+    api(project(":instrumentation:android-instrumentation"))
+//    api(platform(libs.opentelemetry.platform.alpha))
+    implementation(libs.opentelemetry.api.incubator)
+    implementation(libs.opentelemetry.sdk)
+    implementation(libs.opentelemetry.semconv.incubating)
+    api(project(":core"))
+    api(project(":common"))
+    api(project(":session"))
+
+//    implementation(project(":services"))
```
bidetofevil, line 26: Any particular reason you want these commented out lines here? For easy debugging?

bidetofevil (approved):
LGTM. Comment added, but not a blocker

instrumentation/sessions/build.gradle.kts
breedx-splk: Nope, I think it's leftover copypasta....I'll remove it. 👍🏻

instrumentation/sessions/build.gradle.kts
```diff
@@ -0,0 +1,35 @@
...
+//    implementation(project(":common"))
+//    api(platform(libs.opentelemetry.platform.alpha))
+//    api(libs.opentelemetry.api)
+//    implementation(libs.androidx.core)
+//    implementation(libs.opentelemetry.semconv)
+//    implementation(libs.opentelemetry.sdk)
+//    implementation(libs.opentelemetry.instrumentation.api)
+//    testImplementation(libs.robolectric)
```
breedx-splk, line 34:
```suggestion
```

instrumentation/sessions/build.gradle.kts
```diff
@@ -0,0 +1,26 @@
...
+    namespace = "io.opentelemetry.android.instrumentation.sessions"
+
+    defaultConfig {
+        consumerProguardFiles("consumer-rules.pro")
+    }
+}
+
+dependencies {
+    api(project(":instrumentation:android-instrumentation"))
+//    api(platform(libs.opentelemetry.platform.alpha))
```
breedx-splk, line 18:
```suggestion
```

LikeTheSalad (approved):

Issue #735, opened 2025-01-09
Modernize SimpleNetworkDetector with updated Android network APIs


orioonyx:
### Description
The `SimpleNetworkDetector` class currently uses deprecated Android APIs:
1. `getActiveNetworkInfo()` (deprecated in API 29).
2. `ConnectivityManager.TYPE_*` constants (deprecated in API 28).

These deprecated APIs may cause compatibility issues with Android 10 (API 29) and above, and could lead to maintainability challenges as Android evolves.

### Why This Is Important
- Modernizing the implementation ensures compatibility with Android 10 and newer versions.
- Reduces reliance on deprecated methods, making the codebase more maintainable and future-proof.
- Aligns the project with modern Android development best practices, benefiting both maintainers and users.

Thank you for considering this issue, and I look forward to contributing to its resolution.

breedx-splk:
Duplicate of #49

PR #736, opened 2025-01-09
fix(network): modernize simplenetworkdetector with updated android apis

orioonyx:
### Related issue
This pull request resolves [issue #735](https://github.com/open-telemetry/opentelemetry-android/issues/735).

### Summary of changes
This PR updates the `SimpleNetworkDetector` class to replace deprecated Android APIs with their modern counterparts while ensuring backward compatibility:
1. For API 29 and above:
   - Replaced `getActiveNetworkInfo()` with `getActiveNetwork()` and `getNetworkCapabilities()`.
   - Used `NetworkCapabilities.TRANSPORT_*` constants to determine network type.
2. For API 28 and below:
   - Retained the existing logic using `getActiveNetworkInfo()` and `TYPE_*` constants to support older devices.

Additional improvements include:
- Added `null` checks to handle cases where modern APIs might return `null`.
- Simplified and aligned the code for better readability.

### Why these changes are important
- Aligning with modern Android APIs ensures compatibility with Android 10 and newer versions.
- Maintaining backward compatibility supports older Android versions, benefiting a broader range of users.
- Improving code readability and structure enhances maintainability for future contributors.

### Acknowledgment
Thank you for providing this opportunity to contribute to the project. I have done my best to follow the contribution guidelines and ensure the changes adhere to project standards. 

If there are any further improvements or adjustments required, please let me know. I will be more than happy to address any feedback promptly.

Thank you for your time and consideration!


services/src/main/java/io/opentelemetry/android/internal/services/network/detector/SimpleNetworkDetector.java
```diff
@@ -25,17 +27,49 @@ class SimpleNetworkDetector implements NetworkDetector {
...
+
+            NetworkCapabilities capabilities = connectivityManager.getNetworkCapabilities(network);
+            if (capabilities == null) {
+                return UNKNOWN_NETWORK;
+            }
+
+            // Determine network type based on transport capabilities
+            if (capabilities.hasTransport(NetworkCapabilities.TRANSPORT_CELLULAR)) {
+                return CurrentNetwork.builder(NetworkState.TRANSPORT_CELLULAR)
+                        .subType("") // Additional details can be added
```
marandaneto, line 45: Not possible to do `activeNetwork.getSubtypeName()` like the older APIs?

marandaneto (approved):
left a question otherwise LGTM

services/src/main/java/io/opentelemetry/android/internal/services/network/detector/SimpleNetworkDetector.java
orioonyx: Thank you for the question!

Unfortunately, getSubtypeName() is not available in the newer APIs like NetworkCapabilities. The newer APIs focus on transport-level details (e.g., cellular, WiFi) rather than providing specific subtype information like "LTE" or "HSPA."

Thank you for your review and feedback! 😊

services/src/main/java/io/opentelemetry/android/internal/services/network/detector/SimpleNetworkDetector.java
```diff
@@ -25,17 +27,49 @@ class SimpleNetworkDetector implements NetworkDetector {
...
+                return CurrentNetwork.builder(NetworkState.TRANSPORT_WIRED).subType("").build();
+            }
+
+            // Default to UNKNOWN_NETWORK for other types
+            return UNKNOWN_NETWORK;
+        }
+
+        // For API 28 and below, use deprecated methods
         NetworkInfo activeNetwork =
                 connectivityManager.getActiveNetworkInfo(); // Deprecated in API 29
```
breedx-splk, line 61: The fact that this exists means that we will still get a build-time warning, right? 

I think that if you factor out the body of the above `if(api_29)` block into a method, you could then also factor out the lower half of this method into a new method that we could suppress the warnings for. Just an idea.

breedx-splk (approved):
Thanks for the contribution. I think this is a step forward, but I made one suggestion about refactoring.

I think once this lands, we can convert it to kotlin and leverage pattern matching to make it even nicer. 👍🏻

services/src/main/java/io/opentelemetry/android/internal/services/network/detector/SimpleNetworkDetector.java
orioonyx: Thank you so much for your valuable feedback and suggestions! 😊

I've addressed your comments by:
- Adding `@RequiresApi(api = Build.VERSION_CODES.Q)` to the `detectUsingModernApi` method to align with the API requirements for modern implementations.
- Improving code readability by refactoring the `buildCurrentNetwork` calls to adhere to the project's formatting standards.
- Running `spotlessApply` to ensure the code meets the required formatting rules.

Please let me know if there's anything else you'd like me to adjust or further refine. I'm more than happy to make additional changes based on your feedback.

Thank you again for taking the time to review my PR and for providing such helpful guidance!

LikeTheSalad (approved):
Thank you!

PR #753, opened 2025-01-16
Add ability to enable debug logging for disk buffering

### LINKED ISSUE #597: Add setting to enable disk buffering debug mode

breedx-splk:
Supersedes #664 (which seems to be stalled out).

Resolves #597.

This PR adds a debugEnabled flag to the disk buffering config and converts the config to kotlin.


core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfig.kt
```diff
@@ -0,0 +1,60 @@
...
+    ) {
+        companion object {
+            /**
+             * Convenience factory method that validates the min/max and fixes
+             * those up if needed. Users should prefer this method over the
+             * vanilla non-validating constructor.
+             */
+            @JvmOverloads
+            @JvmStatic
+            fun create(
```
breedx-splk, line 34: This is kinda stupid, but the helpful logic was there in the verbose java builder and I didn't know of a better way of doing this, short of building a kotlin builder. 🤷🏻

core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfig.kt
```diff
@@ -0,0 +1,60 @@
...
+const val MAX_CACHE_FILE_SIZE: Int = 1024 * 1024
+
+data class DiskBufferingConfig
+    @JvmOverloads
+    constructor(
+        val enabled: Boolean = false,
+        val maxCacheSize: Int = DEFAULT_MAX_CACHE_SIZE,
+        val maxFileAgeForWriteMillis: Long = TimeUnit.SECONDS.toMillis(30),
+        val minFileAgeForReadMillis: Long = TimeUnit.SECONDS.toMillis(33),
+        val maxFileAgeForReadMillis: Long = TimeUnit.HOURS.toMillis(18),
```
marandaneto, line 22: should they be const as well? (30, 33, 18)?

core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfig.kt
```diff
@@ -0,0 +1,60 @@
...
+                maxCacheSize: Int = DEFAULT_MAX_CACHE_SIZE,
+                maxFileAgeForWriteMillis: Long = TimeUnit.SECONDS.toMillis(30),
+                minFileAgeForReadMillis: Long = TimeUnit.SECONDS.toMillis(33),
+                maxFileAgeForReadMillis: Long = TimeUnit.HOURS.toMillis(18),
+                maxCacheFileSize: Int = MAX_CACHE_FILE_SIZE,
+                debugEnabled: Boolean = false,
+            ): DiskBufferingConfig {
+                var minRead = minFileAgeForReadMillis
+                if (minFileAgeForReadMillis <= maxFileAgeForWriteMillis) {
+                    minRead = maxFileAgeForWriteMillis + 5
```
marandaneto, line 45: nasty :D but yeah its there in the Java file already

marandaneto (approved):

core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfig.kt
marandaneto: Kotlin way is just data classes with default values.
there are a few ways in kotlin:

validate/change values on the `init {}` block.

```kotlin
data class User(
    val age: Int = 18
) {
    init {
        require(age >= 0) { "Age must be non-negative" } // this will throw but you get me
    }
}
```

factory function:

```kotlin
data class User private constructor(
    val age: Int
) {
    companion object {
        fun create(age: Int): User {
            require(age >= 0) { "Age must be non-negative" }
            return User(age)
        }
    }
}
```

validate/change on custom setters, prob not ideal in this case.

builder in kotlin is more verbose but not a blocker here.

breedx-splk: Yeah, and as you've seen, the purpose of the create/init thing is to _change_ one of the values if it's "wrong"...which can't be done in an init (with immutable fields), which is why I resorted to the static factory `create` method...which I still don't love. :) 

Appreciate the input suggestions tho!

PR #763, opened 2025-01-22
Instrumentation context change

LikeTheSalad:
- Decoupling of Services from the instrumentation context. 
- Renaming types to avoid having the word "Service" on them.
- Making services "lazy" to initialize.

This is based on a discussion from a previous Android SIG meeting where we decided to decouple services from the Instrumentation API to avoid passing them to each instrumentation, making instrumentations easier to reuse independently, and also avoiding expanding the public API surface.


core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
```diff
@@ -544,7 +532,7 @@ private SdkLoggerProvider buildLoggerProvider(
                         .addLogRecordProcessor(new SessionIdLogRecordAppender(sessionProvider))
                         .addLogRecordProcessor(
                                 new ScreenAttributesLogRecordProcessor(
-                                        getServiceManager().getVisibleScreenService()))
+                                        services.getVisibleScreenTracker()))
```
surbhiia, line 535: It is not related to this PR but something that came to my mind looking at this. We use config "config.shouldIncludeScreenAttributes()" to determine if screen attributes should be added to the spans but for logs we're adding it always irrespective of the config.Wanted to confirm if this was by choice or a miss? I also checked currently only crash instrumentation is utilizing the logger provider.

services/src/main/java/io/opentelemetry/android/internal/services/Services.kt
```diff
@@ -0,0 +1,97 @@
...
+
+    val visibleScreenTracker: VisibleScreenTracker by lazy {
+        factory.createVisibleScreenTracker()
+    }
+
+    companion object {
+        private var instance: Services? = null
+
+        @JvmStatic
+        fun get(application: Application): Services =
```
surbhiia, line 53: The synchronization here has a slight performance downside as it synchronizes on every read as well. I came across [this](https://www.baeldung.com/java-singleton-double-checked-locking) nice article which describes an [alternate way](https://www.baeldung.com/java-singleton-double-checked-locking#2-initialization-on-demand) which might not need synchronization at all.

LikeTheSalad: Thanks Surbhi. I understand what you mean, although the type of initialization you shared is ideal when the instance can be created without dependencies, or when all its dependencies are static, which is not the case here with `Application`. The only alternative approach I think we could use here is the "double-checked locking" pattern, which I tend to avoid because it's tricky to implement and difficult to read, and most of the time the performance gain it provides is negligible depending on the use case. So I'd prefer to use the simple synchronization-of-it-all approach unless we really need that bigger hammer here, which I don't think is the case unless I'm missing something. If you'd like to get more details of what I mean, [this page](https://codeql.github.com/codeql-query-help/java/java-unsafe-double-checked-locking/) manages to explain it concisely.

core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
LikeTheSalad: I didn't notice it before, nice catch! I think it's a miss 😅

breedx-splk (approved):
Thanks @LikeTheSalad. I like the `ServiceManager` -> `Services` rename quite a bit (terser, simpler!), and the lazy initialization of the individual services is nice....if not a bit less flexible (no more interface). 

This does appear to keep the internal bits more internal -- that is, without `ServiceManager` (now `Services`) exposed on the instrumentation api (via `InstallationContext`), there's no externally exposed services api...which is a net-positive from a maintenance perspective.

One downside here is that instrumentations must now decide how they are going to resolve the things we're hiding behind `Services`. Our instrumentations are free to just call the static methods to get at the `Services` instance because the instrumentations are considered "internal" to the otel android project I guess? However, I wonder what a 3rd-party user who is writing their own instrumentation will think when they look inside one of our implementations as an example. 

I think we've made it harder to do the "wrong" thing (can't use our internal classes), but it might still be rather frustrating for instrumentation authors to do things with services in a way that consistently interops cleanly with our service implementations.

Overall I think this is worth a shot and we should try out this approach.

instrumentation/activity/src/main/java/io/opentelemetry/android/instrumentation/activity/ActivityLifecycleInstrumentation.kt
```diff
@@ -36,7 +37,7 @@ class ActivityLifecycleInstrumentation : AndroidInstrumentation {
     }
 
     private fun buildActivityLifecycleTracer(ctx: InstallationContext): DefaultingActivityLifecycleCallbacks {
-        val visibleScreenService = ctx.serviceManager.getVisibleScreenService()
+        val visibleScreenService = Services.get(ctx.application).visibleScreenTracker
```
breedx-splk, line 40: The static call here is the thing that I think I'm most concerned about in this PR. I think it's manageable in these internal instrumentations (I hope 🤞🏻)....and we can provide overloads or whatever to make it more testable as needed.

services/src/main/java/io/opentelemetry/android/internal/services/Services.kt
surbhiia: Ohh.. I missed the argument part! Double checked locking seems to be tedious and not beneficial here. Thanks for looking into this :)

LikeTheSalad:
> Our instrumentations are free to just call the static methods to get at the Services instance because the instrumentations are considered "internal" to the otel android project I guess?

Yes, for our internal instrumentations, it's fine to use the Services singleton, which is not recommended for third-party ones.

> However, I wonder what a 3rd-party user who is writing their own instrumentation will think when they look inside one of our implementations as an example.

Services mostly keep calls to the Android SDK tools nicely grouped for us to reuse and for maintenance to get easier in the long run. By this, I mean that there's nothing done there that users couldn't do by themselves using the same `Application` dependency that is provided to all instrumentations. If however somebody decides to use Services for their own instrumentation, they are free to do so but we're not forced to ensure that it will work properly or that it will not be broken in the future, so I'd say that users should take a look at something they need from Services, if that's the case, and make those Android API calls directly for their own instrumentations instead.

instrumentation/activity/src/main/java/io/opentelemetry/android/instrumentation/activity/ActivityLifecycleInstrumentation.kt
LikeTheSalad: Definitely, also if there are any issues we can also make breaking changes if needed as it's no longer part of the public api.

core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
LikeTheSalad: I've created [this issue](https://github.com/open-telemetry/opentelemetry-android/issues/765) to address this in a future PR.

PR #767, opened 2025-01-25
Configure metrics export

### LINKED ISSUE #527: Initializing metrics exporter
We're currently initializing tracer and logger rexporters but there's no metrics exporter provided by default, nor there is an "exporter customizer" for metrics either.

breedx-splk:
Resolves #527.

A few notes about this:

* I'm still not convinced that most users want/need metrics, but we've had enough interest lately that I think it's warranted.
* The logging exporter is based on `java.util.logging` and doesn't seem to actually template the `MetricData` values correctly, so the actual output in logcat is less than helpful.
* Maybe eventually we can make it easy for users to specify which resource attributes they want on metrics, because the exporter takes a Resource, we could customize this at creation time. It's a rabbit hole that we should discuss. 
* With the metrics exporter customizer functionality now, users can override and customize every aspect (including resource), but it's honestly pretty challenging to get correct, especially if you want disk buffering too...which most users want and which we want to be default.

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
+++ b/core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
@@ -61,6 +61,8 @@ import io.opentelemetry.sdk.logs.export.LogRecordExporter;
 import io.opentelemetry.sdk.metrics.SdkMeterProvider;
 import io.opentelemetry.sdk.metrics.SdkMeterProviderBuilder;
 import io.opentelemetry.sdk.metrics.export.MetricExporter;
+import io.opentelemetry.sdk.metrics.export.MetricReader;
+import io.opentelemetry.sdk.metrics.export.PeriodicMetricReader;
 import io.opentelemetry.sdk.resources.Resource;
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
```diff
@@ -552,14 +594,23 @@ private SpanExporter buildSpanExporter() {
         return spanExporterCustomizer.apply(defaultExporter);
     }
 
+    private MetricExporter buildMetricExporter() {
+        // TODO: Default to otlp...but how can we make endpoint and auth mandatory?
```
marandaneto, line 598: do we need to answer that before merging? if not, LGTM

marandaneto (approved):

LikeTheSalad (approved):

core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
breedx-splk: Nah, that's been in there for a while and is applicable across all the signal types. 👍🏻

Issue #770, opened 2025-01-27
Add animal sniffer check


LikeTheSalad:
As a follow-up for https://github.com/open-telemetry/opentelemetry-android/pull/701

PR #771, opened 2025-01-27
Animalsniffer check

LikeTheSalad:
Closes https://github.com/open-telemetry/opentelemetry-android/issues/770

Follow-up for https://github.com/open-telemetry/opentelemetry-android/pull/701


breedx-splk:
Also supersedes #768

buildSrc/src/main/kotlin/otel.java-library-conventions.gradle.kts
```diff
@@ -1,58 +0,0 @@
-import org.jetbrains.kotlin.gradle.dsl.JvmTarget
```
breedx-splk, line 1: Oof, I guess this was cruft just left over from a previous rename?

animal-sniffer-signature/build.gradle.kts
```diff
@@ -0,0 +1,70 @@
...
+    configurations.create("generatedSignature") {
+        isCanBeConsumed = true
+        isCanBeResolved = false
+    }
+configurations.add(signatureJar)
+configurations.add(signatureJarClasspath)
+configurations.add(generatedSignature)
+
+dependencies {
+    signature("com.toasttab.android:gummy-bears-api-${project.property("android.minSdk")}:0.9.0@signature")
```
breedx-splk, line 37: What's all this about?

instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlReplacements.java
```diff
@@ -77,7 +78,15 @@ public static int replacementForContentLength(URLConnection connection) {
     }
 
     public static long replacementForContentLengthLong(URLConnection connection) {
-        return replace(connection, () -> connection.getContentLengthLong());
+        return replace(
```
breedx-splk, line 81: Oh, did this catch one already then!?

animal-sniffer-signature/build.gradle.kts
marandaneto: this is the plugin that will be compiled against (pinned to the min version SDK) so the sniffer knows if there's code that's not compatible with

buildSrc/src/main/kotlin/otel.java-library-conventions.gradle.kts
LikeTheSalad: The convention configs for animalsniffer checks on plain Java libs are slightly different than for Android libs, for a moment I thought about adding all that config in a separate convention file and so on, but then I realized that the plain Java lib conventions aren't used anywhere right now, so it seems like it's better to just remove them instead of making this PR bigger by adding all that other config that's not needed right now.

instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlReplacements.java
LikeTheSalad: Yes, the method [getContentLengthLong()](https://developer.android.com/reference/java/net/URLConnection#getContentLengthLong()) is available starting from API 24 so animalsniffer complains about it because our min is 21. However, for this particular case, this method is guaranteed to be called on apps with at least min API 24 so there's no need to address it, so to disable animalsniffer for this case only we have to annotate the class/method/field with `@RequiresApi` where this call is made, and I tried first to annotate the parent method `replacementForContentLengthLong`, though that didn't work, because the call to `getContentLengthLong()` was made in a lambda and it seems like animalsniffer doesn't check those scenarios, so for it to work I had to turn the lambda into an anonymous class and then annotate the actual method inside where the call to `getContentLengthLong()` is made.

marandaneto (approved):

Issue #803, opened 2025-02-09
Improve OSSF scorecard and CLOMonitor scores


trask:
Similar to work going on in other Java repos, e.g. https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/13157

See

* https://securityscorecards.dev/viewer/?uri=github.com/open-telemetry/opentelemetry-android (doesn't exist yet)
* https://clomonitor.io/projects/cncf/open-telemetry#opentelemetry-android

breedx-splk:
@trask do we want to leave this open for a bit while we figure out the remaining CLO Monitor items or ?

trask:
nah, let's close, thanks!

PR #807, opened 2025-02-09
Gradle wrapper validation all the time

trask:
Related to #803, this ensures that the OSSF "binary artifacts" verification for gradle wrappers passes


.github/workflows/gradle-wrapper-validation.yml
```diff
@@ -0,0 +1,16 @@
+name: Gradle wrapper validation
+
+on:
+  push:
+  pull_request:
```
marandaneto, line 5: Should we keep the `paths` tho? Is there You don't need to run if the wrapper hasn't changed.

trask: yeah, the OSSF "binary artifacts" check only goes green if it is run on every PR that is merged to main (it looks at CI runs for PRs that are merged to main and verifies that gradle wrapper validation was performed)

breedx-splk: There might also be some trickery (that I'm not specifically aware of) that could convince a gradle wrapper to run from a different location, so this is probably still nice to have.

breedx-splk (approved):
Tempted to reject outright with the name change from `yaml` -> `yml` 😬 but we're already inconsistent in our use of file suffixes for workflows...... 🤷🏻

breedx-splk:
Whoops, I think merging #801 broke this. Sorry.

trask:
> Tempted to reject outright with the name change from `yaml` -> `yml` 😬 but we're already inconsistent in our use of file suffixes for workflows...... 🤷🏻

yeah, the other "Java" repos are using .yml, sorry

trask:
> 8x3 lifers

😆

PR #809, opened 2025-02-10
Bom setup

### LINKED ISSUE #348: Publish a bom?
The other java-based projects (core, instrumentation, maybe even semconv?) publish a bom (bill-of-materials) that gradle can use to keep dependency versions aligned. As we continue to modularize the repository and publish more independent, smaller artifacts, this becomes increasingly important.

Is a bom a normal/standard thing to do for Android projects as well? I suspect so, since gradle is being used.

The goal is to prevent end-users and distros from having to specify individual versions. Specifying the version number for the bom should suffice.

LikeTheSalad:
Closes #348


marandaneto (approved):

breedx-splk (commented):
Wow, the right evil incantation from the gradle depths does wonders.

The java core bom is called `openetelemetry-bom` and the java agent instrumentation bom is called `opentelemetry-instrumentation-bom`. Can we name this one `opentelemetry-android-bom` to match?

LikeTheSalad:
> Wow, the right evil incantation from the gradle depths does wonders.

Agree 😄 🪄 

> The java core bom is called openetelemetry-bom and the java agent instrumentation bom is called opentelemetry-instrumentation-bom. Can we name this one opentelemetry-android-bom to match?

Sure, I've just changed it 👍

breedx-splk (approved):
Thanks!

PR #813, opened 2025-02-10
Restrict token permissions

trask:


.github/workflows/release.yml
```diff
@@ -104,16 +109,3 @@ jobs:
                             --title "Version $VERSION" \
                             --notes-file /tmp/release-notes.txt \
                             v$VERSION
-
-          echo "version=$VERSION" >> $GITHUB_OUTPUT
-
-      - name: Create release tag for release builds
-        env:
-          GH_TOKEN: ${{ secrets.OPENTELEMETRYBOT_GITHUB_TOKEN }}
-        if: "${{ inputs.release_type == 'final' }}"
```
trask, line 113: I didn't find `release_type` anywhere so assume this is all unused (and deleted it so I didn't have to figure out what permissions to give it 😄)

marandaneto: will let @breedx-splk or @LikeTheSalad double check, 0 context here.

LikeTheSalad: iirc it's probably there from when we used to have a single action to create both snapshots and release builds, although I might be mixing things up, but anyway we need to pass the gradle param "[final=true](https://github.com/open-telemetry/opentelemetry-android/blob/ca67529b1c295f0f22346cd624a6835b45984525/.github/workflows/release.yml#L85)" to make gradle set the [proper version](https://github.com/open-telemetry/opentelemetry-android/blob/ca67529b1c295f0f22346cd624a6835b45984525/build.gradle.kts#L28), so that's probably where this comes from.

However, after a quick search, it seems like this is the only place where we're creating the release tag? If that's the case I'm wondering how it has worked so far 😅 maybe it's getting created somewhere else that I missed, I think @breedx-splk might have a better idea.

LikeTheSalad: Got it, thanks! Then I think it's safe to delete this part.

LikeTheSalad (approved):

PR #823, opened 2025-02-13
Compare cached cache size with current request

deividasstr:
Currently, if developer changes cache size, the users who have cache, will not have their cache size reduced (until some cache clear).
This might be harmful for when the cache change is made due to memory issue (too little cache was set initially, or too much)

The change simply compares caches cache size vs requested.

There could be one case when recalculation of cache size will still happen - when cached cache size was not different from requested, but the requested cache size was more than available space at that time (so available space size was cached instead). But it does not sound like a very important case.

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/internal/features/persistence/DiskManager.kt
+++ b/core/src/main/java/io/opentelemetry/android/internal/features/persistence/DiskManager.kt
@@ -45,15 +45,15 @@ internal class DiskManager(
          * size is stored in the preferences and returned otherwise.
          */
         get() {
+            val requestedSize = diskBufferingConfig.maxCacheSize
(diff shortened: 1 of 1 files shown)


LikeTheSalad (approved):
Thank you!

breedx-splk (approved):

breedx-splk:
Thanks @deividasstr . Can you fix up that test?

deividasstr:
> Thanks @deividasstr . Can you fix up that test?

Sure, but after [this PR](https://github.com/open-telemetry/opentelemetry-android/pull/828) is merged - it simplifies the storate calculation logic.

deividasstr:
This [PR](https://github.com/open-telemetry/opentelemetry-android/pull/828) removes max file size caching, thus rendering this PR obsolete.

PR #836, opened 2025-02-18
Emit log for every instrumentation init event

bidetofevil:
Store init events in a thread safe queue and keep a reference to the created `EventLogger` when `finish()` is called after all the instrumentation has initialized. 

When an event is added - if it's `EventLogger` reference is not null, emit all the stored event, as it means an event has arrived after finished as been called.

If init events never appear after `finish()` is called, this will simply be a no-op. If they do, this will ensure they are properly emitted.

### DIFF
--- a/instrumentation/startup/src/main/java/io/opentelemetry/android/instrumentation/startup/SdkInitializationEvents.kt
+++ b/instrumentation/startup/src/main/java/io/opentelemetry/android/instrumentation/startup/SdkInitializationEvents.kt
@@ -14,15 +14,19 @@ import io.opentelemetry.api.common.AttributeKey
 import io.opentelemetry.api.common.Attributes
 import io.opentelemetry.api.common.Value
 import io.opentelemetry.api.incubator.logs.ExtendedLogRecordBuilder
+import io.opentelemetry.api.logs.Logger
 import io.opentelemetry.sdk.trace.export.SpanExporter
 import java.time.Instant
+import java.util.concurrent.ConcurrentLinkedQueue
+import java.util.concurrent.atomic.AtomicReference
 import java.util.function.Supplier
 
 @AutoService(InitializationEvents::class)
 class SdkInitializationEvents(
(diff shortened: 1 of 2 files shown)


breedx-splk:
The [latest upstream update](https://github.com/open-telemetry/opentelemetry-android/pull/826/files) was disruptive to events -- you'll have some work to do in this PR to fix it up. Sorry. Most notably, the `EventLogger` is gone in favor of the ExtendedLogger that has a `setEventName()` method. 🥚 🍳

bidetofevil:
Updated branch after rebasing

instrumentation/startup/src/main/java/io/opentelemetry/android/instrumentation/startup/SdkInitializationEvents.kt
```diff
@@ -59,23 +63,35 @@ class SdkInitializationEvents(
...
+                }
+            }.emit()
     }
 
     private fun addEvent(
         name: String,
         attr: Attributes? = null,
     ) {
         events.add(Event(clock.get(), name, attr))
+        eventLogger.get()?.emitInitEvents()
```
breedx-splk, line 94: Maybe a comment to let the reader know that the `eventLogger` will only be set if `finish()` has been called?

breedx-splk (approved):
Looks good to me. It would be great to get confirmation that this resolves the issue, but it seems like it should. I'll ask in the issue as well.

LikeTheSalad (approved):
Looks good! Cheers.

instrumentation/startup/src/main/java/io/opentelemetry/android/instrumentation/startup/SdkInitializationEvents.kt
bidetofevil: Sounds good

bidetofevil: Done

Issue #861, opened 2025-02-28
Generate events for OkHttp Websocket events


cleverchuk:
I'd like to able to collect websocket events to track activity.

cleverchuk:
please assign to me. i have something already cooked

LikeTheSalad:
Done. Thank you for volunteering, @cleverchuk.

PR #863, opened 2025-03-03
Generate events for OkHttp Websocket events

cleverchuk:
This PR adds a build time instrumentation that wraps `okhttp3.WebSocketListener` to provide event generations for each of the WebSocket events. I haven't added unit test because the change is mostly delegated write operations and I'd gladly add them when there's a recommendation. Also, the WebSocket event's attributes are open for discussion. I leaned towards making them similar to HTTP.

Addresses #861


instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/library/okhttp/v3_0/internal/WebsocketListenerWrapper.java
```diff
@@ -0,0 +1,75 @@
...
+    private final WebSocketListener delegate;
+
+    public WebsocketListenerWrapper(WebSocketListener delegate) {
+        this.delegate = delegate;
+    }
+
+    @Override
+    public void onClosed(@NonNull WebSocket webSocket, int code, @NonNull String reason) {
+        Attributes attributes = WebsocketAttributeExtractor.extractAttributes(webSocket);
+        WebsocketEventGenerator.generateEvent("onClosed", attributes);
```
breedx-splk, line 26: So, we don't yet have semantic conventions for websockets, but I think we should put our best foot forward here because this can be used as a reference when creating the semconv/spec, and other implementations will look to us first for examples.

The ws events are named `close`, `error`, `message`, and `open` ([reference](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/open_event)). As such, I think we should map those over, inside of a `websocket` prefix. So, you'd end up with these dotted event names instead of the camel-case ones proposed right now:

* `websocket.open`
* `websocket.close`
* `websocket.message`
* `websocket.error`

The `onClosing()` callback is interesting, because it's not part of the spec, but something nice that okhttp provides. We have little guarantee that ios/web can capture this. So I think we should remove the closing event for now, with the intent to add it later if/when users ask for it.

breedx-splk (commented):
This is cool, thanks for the PR!

Please change the event names to be more aligned with standard otel dot notations, and please add an integration test. Thanks!

breedx-splk:
Oh yeah, also as discussed in SIG earlier today, let's split websocket out into its own instrumentation.

instrumentation/okhttp/okhttp-3.0/agent/build.gradle.kts
```diff
@@ -11,6 +11,6 @@ android {
 
 dependencies {
     implementation(project(":instrumentation:okhttp:okhttp-3.0:library"))
-    implementation(libs.okhttp)
+    implementation(libs.okhttp) // TODO - should this be compileOnly?
```
cleverchuk, line 14: @breedx-splk what do you think about this?

breedx-splk: I think I agree that it should be compileOnly. :)

marandaneto:
LGTM to me.
I wonder if the code could be Kotlin, but I understand it's all done already.

marandaneto (approved):
@breedx-splk left a few comments I agree with otherwise LGTM

LikeTheSalad (approved):
Thank you!

PR #867, opened 2025-03-04
Add attribute extractors to auto-http instrumentations and a null check

surbhiia:
- Added attribute extractors for both the auto-http instrumentations - `Okhttp3` and `HttpURLConnection`
- Added null check to prevent null pointer exception if `instrumenter = null `in `HttpURL` code

Tested both the extractors in my own sample app.

### DIFF
--- a/instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlReplacements.java
+++ b/instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlReplacements.java
@@ -290,7 +290,9 @@ public class HttpUrlReplacements {
 
     private static void startTracingAtFirstConnection(URLConnection connection) {
         Context parentContext = Context.current();
-        if (!HttpUrlConnectionSingletons.instrumenter().shouldStart(parentContext, connection)) {
+        if (HttpUrlConnectionSingletons.instrumenter() == null
+                || !HttpUrlConnectionSingletons.instrumenter()
+                        .shouldStart(parentContext, connection)) {
             return;
         }
 
--- a/instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/library/okhttp/v3_0/OkHttpInstrumentation.java
+++ b/instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/library/okhttp/v3_0/OkHttpInstrumentation.java
@@ -9,6 +9,7 @@ import com.google.auto.service.AutoService;
 import io.opentelemetry.android.instrumentation.AndroidInstrumentation;
 import io.opentelemetry.android.instrumentation.InstallationContext;
 import io.opentelemetry.instrumentation.api.incubator.semconv.net.PeerServiceResolver;
+import io.opentelemetry.instrumentation.api.instrumenter.AttributesExtractor;
 import io.opentelemetry.instrumentation.api.internal.HttpConstants;
 import io.opentelemetry.instrumentation.library.okhttp.v3_0.internal.OkHttp3Singletons;
 import java.util.ArrayList;
@@ -17,16 +18,25 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import okhttp3.Interceptor;
+import okhttp3.Response;
 import org.jetbrains.annotations.NotNull;
 
 /** Instrumentation for okhttp requests. */
 @AutoService(AndroidInstrumentation.class)
 public class OkHttpInstrumentation implements AndroidInstrumentation {
+    private final List<AttributesExtractor<Interceptor.Chain, Response>> additionalExtractors =
+            new ArrayList<>();
     private List<String> capturedRequestHeaders = new ArrayList<>();
     private List<String> capturedResponseHeaders = new ArrayList<>();
     private Set<String> knownMethods = HttpConstants.KNOWN_METHODS;
     private Map<String, String> peerServiceMapping = new HashMap<>();
     private boolean emitExperimentalHttpClientMetrics;
+
+    /** Adds an {@link AttributesExtractor} that will extract additional attributes. */
+    public void addAttributesExtractor(AttributesExtractor<Interceptor.Chain, Response> extractor) {
+        additionalExtractors.add(extractor);
+    }
 
     /**
      * Configures the HTTP request headers that will be captured as span attributes as described in
@@ -123,6 +133,6 @@ public class OkHttpInstrumentation implements AndroidInstrumentation {
 
     @Override
     public void install(@NotNull InstallationContext ctx) {
-        OkHttp3Singletons.configure(this, ctx.getOpenTelemetry());
+        OkHttp3Singletons.configure(this, ctx.getOpenTelemetry(), additionalExtractors);
     }
 }
--- a/instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlInstrumentation.java
+++ b/instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlInstrumentation.java
@@ -9,8 +9,10 @@ import com.google.auto.service.AutoService;
 import io.opentelemetry.android.instrumentation.AndroidInstrumentation;
 import io.opentelemetry.android.instrumentation.InstallationContext;
 import io.opentelemetry.instrumentation.api.incubator.semconv.net.PeerServiceResolver;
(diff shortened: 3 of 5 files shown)


instrumentation/httpurlconnection/testing/src/androidTest/kotlin/io/opentelemetry/instrumentation/library/httpurlconnection/InstrumentationTest.kt
```diff
@@ -55,12 +56,11 @@ class InstrumentationTest {
...
-            // Wait for all tasks to finish execution or timeout
-            assertThat(executor.awaitTermination(2, TimeUnit.SECONDS))
-                .withFailMessage("Test could not be completed as tasks did not complete within the 2s timeout period.")
-                .isTrue()
-
-            assertThat(openTelemetryRumRule.inMemorySpanExporter.finishedSpanItems.size).isEqualTo(4)
+            if (executor.awaitTermination(2, TimeUnit.SECONDS)) {
+                assertThat(openTelemetryRumRule.inMemorySpanExporter.finishedSpanItems.size).isEqualTo(4)
+            } else {
+                Log.e(TAG, "Test could not be completed as tasks did not complete within the 2s timeout period.")
```
marandaneto, line 62: in this case, the test should fail, or the test should be rewritten in a way that you can be notified once the tasks are completed

instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlReplacements.java
```diff
@@ -290,7 +290,9 @@ private static void endTracing(URLConnection connection, int responseCode, Throw
 
     private static void startTracingAtFirstConnection(URLConnection connection) {
         Context parentContext = Context.current();
-        if (!HttpUrlConnectionSingletons.instrumenter().shouldStart(parentContext, connection)) {
+        if (HttpUrlConnectionSingletons.instrumenter() == null
```
marandaneto, line 293: I'd assign the `HttpUrlConnectionSingletons.instrumenter()` to a var and reuse it

instrumentation/httpurlconnection/testing/src/androidTest/kotlin/io/opentelemetry/instrumentation/library/httpurlconnection/InstrumentationTest.kt
marandaneto: I'd use eg https://github.com/awaitility/awaitility/wiki/Usage

marandaneto: > The one I've changed it to for now - Prevent false negative by logging that execution did not complete within given max timeout, so you can go and increase timeout and run manually and test. Would adding a comment saying - test this manually and look for any logged error messages help?

its more about that this is run in CI and tests should break, no one is looking at logs on CI, so if its broken, CI is green and we will merge broken things.

> Increase the timeout (it is working in 4s now) and let it fail - this would mean the test is flaky (depends on the time the parallel execution of 4 requests take) and a false negative is possible.

increasing timeout is a workaround, and no solution, not telling we should not do this, but usually just incase testing time and the flakiness is still there, fixing the test (or even removing it) is usually better IMO

https://github.com/awaitility/awaitility/blob/master/awaitility-kotlin/src/test/kotlin/org/awaitility/kotlin/KotlinTest.kt
there are a few options, for example, `await().untilAsserted`, `await().until`, `await.forever until`, etc... so you don't specify a timer but rather a condition.

marandaneto: an example:

```kotlin
val executor = Executors.newFixedThreadPool(4)
        val counter = AtomicInteger(0)

        executor.submit {
            executeGet("http://httpbin.org/get")
            counter.incrementAndGet()
        }
        executor.submit {
            executeGet("http://google.com")
            counter.incrementAndGet()
        }
        executor.submit {
            executeGet("http://android.com")
            counter.incrementAndGet()
        }
        executor.submit {
            executeGet("http://httpbin.org/headers")
            counter.incrementAndGet()
        }

        executor.shutdown()

        await.untilAsserted {
            assertEquals(4, counter.get())
        }
```

marandaneto: another option is a `CountDownLatch(4)` calling `countDown()` for each of the submits.
and `cdl.await()` after

surbhiia: Infinite timeout in my opinion would be dangerous, it can be achieved with all 3 APIs - executor.await, latch.await, or await.forever, await.untilAsserted. But I did not want to go with it as I don't think we want to stall the test pipeline if in case anything were to go wrong with executing the test. 

By suggesting to log, I wanted to at least keep the code there for manual testing during our development (Mentioning that in function comments properly). I still think we can do this. 

Alternatively, we can remove the test if that's what everyone thinks is better? @LikeTheSalad @breedx-splk

marandaneto: I dont think its a problem since the executor is created in the test, so we have control over its lifecycle.
as far as I can tell, `executeGet` swallows all exceptions, so its guaranteed that eg `counter.incrementAndGet()` will be called.
the only difference is that this way you allow `executeGet` to take as much time as needed so its not flaky, if `executeGet` takes forever or does not finish for some reason, its an even more serious bug.

breedx-splk: I'm with @marandaneto on this one, and I like the suggestion to use a `Latch` to accomplish fast success with a longer timeout to prevent flapping. There's more than one way to do this, but I put in a PR here to demonstrate: https://github.com/surbhiia/opentelemetry-android/pull/2

surbhiia: Thanks @marandaneto and @breedx-splk for your feedback and for taking the time to demonstrate the code changes. 

I still think it would be the same as `executor.awaitTermination(1, TimeUnit.MINUTES)` as `awaitTermination `is primarily concerned with waiting for task completion and doesn’t wait for the shutdown of the executor itself. 

As everyone prefers the `Latch`, I’ve updated the code as per the suggested changes now. :)

marandaneto (approved):

instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/library/okhttp/v3_0/OkHttpInstrumentation.java
```diff
@@ -123,6 +133,6 @@ public boolean emitExperimentalHttpClientMetrics() {
 
     @Override
     public void install(@NotNull InstallationContext ctx) {
-        OkHttp3Singletons.configure(this, ctx.getOpenTelemetry());
+        OkHttp3Singletons.configure(this, ctx.getOpenTelemetry(), additionalExtractors);
```
LikeTheSalad, line 136: The existing configs have getters that are called from `OkHttp3Singletons`. Is there any reason to choose a different approach (adding a new OkHttp3Singletons.configure argument) for the extractors?

surbhiia: Thanks for pointing this out! I earlier followed the pattern of passing in attributeExtractors as argument like some other instrumentations. Updated the code now to follow the pattern we follow here in http instrumentations i.e to fetch the configs via getters :)

LikeTheSalad (approved):
Thank you! 🙏

instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/library/okhttp/v3_0/OkHttpInstrumentation.java
LikeTheSalad: Thank you!

surbhiia:
@breedx-splk Does this PR seem okay to be merged? Was just wondering if there was anything for me to take care on this one?

PR #871, opened 2025-03-05
Add `DiskBufferingConfig.signalsBufferDir`

### LINKED ISSUE #859: Move the `signalsBufferDir` under `context.getFilesDir()`
At the moment the `signalsBufferDir`, which is used to store un-exported telemetry batches, is stored within [`context.getCacheDir()`](https://developer.android.com/reference/android/content/Context#getCacheDir()):

https://github.com/open-telemetry/opentelemetry-android/blob/a317b11a5a8787c6e60eafdc92c6711d3296bf90/core/src/main/java/io/opentelemetry/android/internal/features/persistence/DiskManager.kt#L20-L25

https://github.com/open-telemetry/opentelemetry-android/blob/a317b11a5a8787c6e60eafdc92c6711d3296bf90/services/src/main/java/io/opentelemetry/android/internal/services/CacheStorage.java#L24-L26

According to the docs there are no persistency guarantees for [`context.getCacheDir()`](https://developer.android.com/reference/android/content/Context#getCacheDir()):

> The system will automatically delete files in this directory as disk space is needed elsewhere on the device. 

For those applications, where telemetry is a nice-to-have this sounds like an acceptable trade-off. In my case I cannot afford to lose telemetry because of the internal quirks of the OS, as there might be important business events standing in the device's queue.

So my recommendation is to move the `signalsBufferDir` under the more persistent [`getFilesDir()`](https://developer.android.com/reference/android/content/Context#getFilesDir()) to prevent telemetry data loss caused by OS clean-ups.

bencehornak:
Fixes #859 by giving the flexibility to the SDK's users to choose a `signalsBufferDir` other than one used by default.

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfig.kt
+++ b/core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfig.kt
@@ -7,6 +7,7 @@ package io.opentelemetry.android.features.diskbuffering
 
 import android.util.Log
 import io.opentelemetry.android.common.RumConstants.OTEL_RUM_LOG_TAG
+import java.io.File
 import java.util.concurrent.TimeUnit
 
 const val DEFAULT_MAX_CACHE_SIZE: Int = 10 * 1024 * 1024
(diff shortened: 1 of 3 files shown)


breedx-splk (approved):
LGTM! Seems like a nice thing to be able to customize.

LikeTheSalad (approved):
Thank you 🙏

PR #880, opened 2025-03-08
Remove old/outdated explicit build tools version

breedx-splk:
The presence of these lines is causing a build warning:

image

Seems like removing them with AGP 8.9.0 should be safe.


marandaneto:
the reason is that AGP now chooses the build tools version, you cannot force it anymore

marandaneto (approved):

Issue #860, opened 2025-02-28
Use of SessionIdTimeoutHandler class


ddas09:
Hi Team, I am trying to understand the use of the `SessionIdTimeoutHandler` class. As per my understanding there are two classes - 

- [SessionManagerImpl](https://github.com/open-telemetry/opentelemetry-android/blob/main/core/src/main/java/io/opentelemetry/android/internal/session/SessionManagerImpl.kt) - this allows a session to be active for a maximum of 4 hours (if not overridden using the `OtelRumConfig` [here](https://github.com/open-telemetry/opentelemetry-android/blob/main/core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java#L322)
- [SessionIdTimeoutHandler](https://github.com/open-telemetry/opentelemetry-android/blob/main/core/src/main/java/io/opentelemetry/android/internal/session/SessionIdTimeoutHandler.kt) - this class handles expiring session if the app remains if background for more than 15 minutes (if not overridden using the `OtelRumConfig` [here](https://github.com/open-telemetry/opentelemetry-android/blob/main/core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java#L119) without any activity (span, logs).

The only scenario when I think the `SessionIdTimeoutHandler` class [hasTimedOut check](https://github.com/open-telemetry/opentelemetry-android/blob/main/core/src/main/java/io/opentelemetry/android/internal/session/SessionIdTimeoutHandler.kt#L51C9-L51C20) to be useful is **if the timeout is not overridden from the `OtelRumConfig` API**. In this case the handler class would timeout the session **If the app is in the background and no activity (spans) happens for >15 minutes**

But if **the timeout is set in the config API** it would override the value in both the `SessionIdTimeoutHandler` and `SessionManagerImpl` class. In this case If the app is in the background and no activity (spans) happens for >15 minutes, the next time the app comes foreground and any span is created, the check in the `SessionManagerImpl` class only would fail. Since the `sessionLifetimeNanos` is 15 minutes now.

Please help me to understand if I am missing something!

breedx-splk:
I think you got it.

Sessions are intended to timeout after some period of inactivity, default is 15 minutes as you noted. Inactivity is defined as no user behavior, which means no telemetry being created. A foregrounded app is not considered inactive.

Hopefully that helps.

breedx-splk:
[...]

PR #887, opened 2025-03-14
Factor out and make session timeouts more explicit.

breedx-splk:
Relates to #860.  See the thread over there for additional details.

Basically, we have always intended to have two session "timeouts" -- one that is a maximum absolute lifespan (4 hour default) and a background inactivity timeout (15 minute default). Somewhere along the way, these two values were blended, and the same timeout value was ultimately used for both.

The purpose of this PR is to once again separate these two times, and to make their names and usages more explicit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### ⚠️⚠️ Breaking changes
+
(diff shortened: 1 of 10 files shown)


ddas09 (approved):

ddas09:
@breedx-splk, LGTM! However, I see similar changes also done in this [PR](https://github.com/open-telemetry/opentelemetry-android/pull/889).

marandaneto (approved):

Issue #795, opened 2025-02-04
[Discuss] SessionManager argument in InstallationContext API


surbhiia:
I wanted to see what everyone thinks about having [the SessionManager argument in InstallationContext API.](https://github.com/open-telemetry/opentelemetry-android/blob/ae83d01b197fee5b385f80b722578fe330ea92d3/instrumentation/android-instrumentation/src/main/java/io/opentelemetry/android/instrumentation/InstallationContext.kt#L15). 

- It is only used in session-instrumentation where it's used to fetch [the SessionPublisher.addObserver function](https://github.com/open-telemetry/opentelemetry-android/blob/ae83d01b197fee5b385f80b722578fe330ea92d3/instrumentation/sessions/src/main/kotlin/io/opentelemetry/android/instrumentation/sessions/SessionInstrumentation.kt#L20) for attaching a listener that listens to session.start and session.end events.   
- We cannot apply the same logic as we applied for decoupling [Services](https://github.com/open-telemetry/opentelemetry-android/blob/ae83d01b197fee5b385f80b722578fe330ea92d3/services/src/main/java/io/opentelemetry/android/internal/services/Services.kt) from InstallationContext as SessionManager needs to be initialized at a specific time (Example - on application start) and not lazily when the first time it is referred and secondly, [it also needs non-static parameters - like SessionIdTimeoutHandler and timeout config.](https://github.com/open-telemetry/opentelemetry-android/blob/ae83d01b197fee5b385f80b722578fe330ea92d3/core/src/main/java/io/opentelemetry/android/internal/session/SessionManagerImpl.kt#L74) 

Use cases for possibly decoupling this:

- If someone is not utilizing [session management module](https://github.com/open-telemetry/opentelemetry-android/tree/ae83d01b197fee5b385f80b722578fe330ea92d3/session) or[ session-instrumentation ](https://github.com/open-telemetry/opentelemetry-android/tree/ae83d01b197fee5b385f80b722578fe330ea92d3/instrumentation/sessions)from here and just some other instrumentations they need to anyway depend on session module and define a dummy SessionManager object to pass in the InstallationContext which will use it for nothing. 

Questions:
- Can we make [the SessionManager argument](https://github.com/open-telemetry/opentelemetry-android/blob/ae83d01b197fee5b385f80b722578fe330ea92d3/instrumentation/android-instrumentation/src/main/java/io/opentelemetry/android/instrumentation/InstallationContext.kt#L15) nullable? 
[...]

Issue #794, opened 2025-02-04
Same session timeout config value used for foreground & background


surbhiia:
**Introduce a separate config for foreground session timeout**

- SDK provides only [one config for session timeout duration](https://github.com/open-telemetry/opentelemetry-android/blob/main/core/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java#L28C22-L28C36), instead there should be two, one for each foreground & background session timeout duration. 

- Also, as a result the SDK is setting the same value (default 15 mins - set on [background here](SessionTimeout), set on [foreground here](https://github.com/open-telemetry/opentelemetry-android/blob/main/core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java#L338)) in both instead of setting default 4 hours for foreground sessions.

Issue #793, opened 2025-02-04
Refactoring proposals for session related code


surbhiia:
Proposals

- Move [SessionManagerImpl](https://github.com/open-telemetry/opentelemetry-android/blob/main/core/src/main/java/io/opentelemetry/android/internal/session/SessionManagerImpl.kt) & [SessionIdTimeoutHandler](https://github.com/open-telemetry/opentelemetry-android/blob/main/core/src/main/java/io/opentelemetry/android/internal/session/SessionIdTimeoutHandler.kt) inside a package called "internal" under [:session](https://github.com/open-telemetry/opentelemetry-android/tree/main/session/src/main/kotlin/io/opentelemetry/android/session) module. It allows others to use the :session module as is, if they wanted to use our opinionated implementation of the SessionManager (i.e SessionManagerImpl) too.

- Similarly, [SessionIdLogRecordAppender](https://github.com/open-telemetry/opentelemetry-android/blob/main/core/src/main/java/io/opentelemetry/android/internal/processors/SessionIdLogRecordAppender.kt), [SessionIdSpanAppender](https://github.com/open-telemetry/opentelemetry-android/blob/main/core/src/main/java/io/opentelemetry/android/SessionIdSpanAppender.java) & [SessionIdRatioBasedSampler](https://github.com/open-telemetry/opentelemetry-android/blob/8641af428ece7fa381d8faf299bf09ec9e1753db/core/src/main/java/io/opentelemetry/android/SessionIdRatioBasedSampler.java#L24) can be moved in the same module under a "processors" package.

- Rename the SessionIdTimeoutHandler class to include the word background as it essentially handles background session expiry.

Issue #781, opened 2025-01-30
First session is not created on application start


surbhiia:
-  First session will start when the first time anyone calls [getSessionId()](https://github.com/open-telemetry/opentelemetry-android/blob/main/core/src/main/java/io/opentelemetry/android/internal/session/SessionManagerImpl.kt#L36), until then we just have a [default (id = "") session](https://github.com/open-telemetry/opentelemetry-android/blob/main/core/src/main/java/io/opentelemetry/android/internal/session/SessionManagerImpl.kt#L25) in place. 

-  And `getSessionId()` is only getting called when you call `SessionIdSpanAppender `the first time, so new session is created only when first trace is calling the span processor, **not when application starts**

I was able to vet the same with debugging the` demo-app`:

- Looks like when `ActivityLifecycleInstrumentation` is installed, it calls `AppStartupTimer` to emit the **app start event,** that's when `SessionIdSpanAppender` is called, which calls `session.getSessionId()` and leads to 1st session creation (attaching screenshot with debug stack trace for same):

<img width="1337" alt="Image" src="https://github.com/user-attachments/assets/823b36de-d07a-4d08-a3ed-760bcea0450c" />

- **Solution**: We can initialize the first session in the [init](https://github.com/open-telemetry/opentelemetry-android/blob/8641af428ece7fa381d8faf299bf09ec9e1753db/core/src/main/java/io/opentelemetry/android/internal/session/SessionManagerImpl.kt#L28) block of `SessionManagerImpl`, that way session will start as soon as the `SessionManagerImpl` object is created in `OpenTelemetryRumBuilder` (which should be called in customer's application `onCreate`). Below is a stack trace showing when `SessionManagerImpl` is first called while debugging demo app:

<img width="1285" alt="Image" src="https://github.com/user-attachments/assets/e4f7aad3-c695-412a-9bd3-83def53b98bc" />

breedx-splk:
Thanks @surbhiia. There's a lot going on with this issue. Even though most concerns here are session related, I think it could help to foster discussion to have it broken down into a few separate single-issue issues.

surbhiia:
Divided it into separate issues!

breedx-splk:
I think this was resolved in #1413. Please reopen if that is not the case and if you are still having problems.

PR #889, opened 2025-03-14
Session management updates

surbhiia:
This PR covers the following:
- Starting the first session on application start (as soon as `ServiceManagerImpl` is loaded), resolving #781 
- Adding two separate `configs` for session `foreground` and `background` `timeout`, resolving #794 
- Adding a no-op session manager as default in `InstallationContext` API, interim resolution of #795 
- Renaming `SessionIdTimeoutHandler` to `SessionBackgroundTimeoutHandler`, part of #793

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
+++ b/core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
@@ -118,7 +118,7 @@ public final class OpenTelemetryRumBuilder {
         return new OpenTelemetryRumBuilder(
                 application,
                 config,
-                new SessionBackgroundTimeoutHandler(config.getSessionTimeout()));
+                new SessionBackgroundTimeoutHandler(config.getSessionBackgroundTimeout()));
     }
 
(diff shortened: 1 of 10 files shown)


LikeTheSalad:
Hi @surbhiia 

Thank you for these changes. There's one that I think we need to review if it's needed:

> Adding two separate configs for session foreground and background timeout, resolving https://github.com/open-telemetry/opentelemetry-android/issues/794

We discussed in the last SIG meeting that it's probably not a good idea to have different behaviors based on whether the app is in the "foreground" or "background", because those terms might mean different things depending on the app. For example, if an audio streaming app is "backgrounded", but the audio is still playing, probably its developers would want the same "foreground-behavior" of OTel to be running. So, in general, it's best not to depend on "app states" to define different behaviors, because we cannot guarantee that those would mean the same for all apps. So regarding this particular feature, it seems like it's better if we just define one timeout that's always present, regardless of the app state.

Please let me know if I missed something, @breedx-splk @bidetofevil

breedx-splk:
Reopen if you want this to continue. Thanks!

PR #894, opened 2025-03-18
Use event name for crash event (instead of attr)

breedx-splk:
During SIG review today, we noticed that the log record for crash still had the event name attribute set. This changes it to use the top-level event field on the LogRecord.


LikeTheSalad (approved):
Thank you!

PR #895, opened 2025-03-18
Migrate network change event from zero-duration span to event

breedx-splk:
The network change event is currently a zero-duration span. We want to migrate these to actual OpenTelemetry log-based events. This does that work, and also introduces a (global) `LogRecordProcessor` named `NetworkAttributesLogRecordAppender` that (when configured) always appends the global log attributes to all logs/events (similar to the NetworkAttributesSpanAppender).

### DIFF
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/OtelRumConfigExtensions.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/OtelRumConfigExtensions.kt
@@ -16,9 +16,11 @@ import io.opentelemetry.android.instrumentation.crash.CrashReporterInstrumentati
 import io.opentelemetry.android.instrumentation.fragment.FragmentLifecycleInstrumentation
 import io.opentelemetry.android.instrumentation.network.NetworkChangeInstrumentation
 import io.opentelemetry.android.instrumentation.slowrendering.SlowRenderingInstrumentation
+import io.opentelemetry.api.common.AttributesBuilder
 import io.opentelemetry.api.trace.Tracer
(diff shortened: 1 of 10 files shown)


android-agent/src/main/kotlin/io/opentelemetry/android/agent/OtelRumConfigExtensions.kt
```diff
@@ -73,7 +75,7 @@ fun OtelRumConfig.addCrashAttributesExtractor(extractor: AttributesExtractor<Cra
     return this
 }
 
-fun OtelRumConfig.addNetworkChangeAttributesExtractor(extractor: AttributesExtractor<CurrentNetwork, Void>): OtelRumConfig {
+fun OtelRumConfig.addNetworkChangeAttributesExtractor(extractor: BiConsumer<AttributesBuilder, CurrentNetwork>): OtelRumConfig {
```
LikeTheSalad, line 78: Not to be addressed here, though I think it'd be nice to convert the underlying types to Kotlin (I'm referring to the NetworkChangeInstrumentation one in this case) to avoid having to propagate upward the usage of types that have better alternatives in Kotlin, such as `BiConsumer`, which can be a Kotlin function with 2 params.

LikeTheSalad (approved):

android-agent/src/main/kotlin/io/opentelemetry/android/agent/OtelRumConfigExtensions.kt
breedx-splk: Yeah good point. I'll get that in a follow-up PR.

PR #900, opened 2025-03-20
Add AssertionExtensions to work around upstream stability

breedx-splk:
Given [this unfortunate situation with "stability" in the core assertions](https://github.com/open-telemetry/opentelemetry-java/pull/7207), let's come up with our own extension(s) for doing assertions. This starts with `hasEventName()` on the `LogRecordDataAssert`. 

This PR also migrates the `NetworkChangeMonitorTest` to kotlin and mockk in order to demonstrate usage of the extension assert.


test-common/src/main/kotlin/io/opentelemetry/android/test/common/AssertionExtensions.kt
```diff
@@ -0,0 +1,18 @@
...
+package io.opentelemetry.android.test.common
+
+import io.opentelemetry.sdk.logs.data.internal.ExtendedLogRecordData
+import io.opentelemetry.sdk.testing.assertj.LogRecordDataAssert
+import io.opentelemetry.sdk.testing.assertj.OpenTelemetryAssertions.assertThat
+
+fun LogRecordDataAssert.hasEventName(eventName: String): LogRecordDataAssert {
+    isNotNull()
+    // TODO: This will be removed after the event bits of logs gets out of incubating.
+    assertThat(this.actual()).isInstanceOf(ExtendedLogRecordData::class.java)
```
marandaneto, line 15: this is likely redundant since the method below will fail if `this.actual() as ExtendedLogRecordData` fails anyway

test-common/src/main/kotlin/io/opentelemetry/android/test/common/AssertionExtensions.kt
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+package io.opentelemetry.android.test.common
+
+import io.opentelemetry.sdk.logs.data.internal.ExtendedLogRecordData
+import io.opentelemetry.sdk.testing.assertj.LogRecordDataAssert
+import io.opentelemetry.sdk.testing.assertj.OpenTelemetryAssertions.assertThat
+
+fun LogRecordDataAssert.hasEventName(eventName: String): LogRecordDataAssert {
+    isNotNull()
```
marandaneto, line 13: i think this is redundant since `LogRecordDataAssert.hasEventName` so `LogRecordDataAssert` is never null

marandaneto (approved):

test-common/src/main/kotlin/io/opentelemetry/android/test/common/AssertionExtensions.kt
LikeTheSalad: +1

LikeTheSalad (approved):

test-common/src/main/kotlin/io/opentelemetry/android/test/common/AssertionExtensions.kt
breedx-splk: Are we sure about this? The extension is on the `LogRecordDataAssert` (the assertion class, NOT the log record itself), and the method `isNotNull()` is on the assetj parent class `AbstractAssert` and it verifies that the actual result isn't null. I expect that the error message would be easier to read by checking explicitly here, so like `assertThat(null).hasEventName("foo")` would give a clearer message 

Are we talking about the same thing here?

The `isInstanceOf()` also contains an internal `isNotNull()` check, but when the eventName comes out of incubating, then the instance checks go away and we will lose the specificity.

breedx-splk:
Given that there are two approvals, I'm going to merge in spite of the unresolved conversation. It's not a big deal to make changes after, but I would like to get the rebase of the following PR going. Thanks for the reviews!

PR #901, opened 2025-03-20
Migrate network instrumentation to kotlin

breedx-splk:
This will rebase after #900 is merged.

Based on [this comment](https://github.com/open-telemetry/opentelemetry-android/pull/895#discussion_r2005716560), we want to simplify the network instrumentation by migrating it to kotlin, which eliminates the explicit BiConsumer ugliness. In the process of doing this, it also became apparent that the `NetworkChangeMonitor` class wasn't really pulling its weight, so its only small method was just brought into the `install()` for the instrumentation....further simplifying things.


test-common/src/main/kotlin/io/opentelemetry/android/test/common/AssertionExtensions.kt
```diff
@@ -0,0 +1,18 @@
...
+
+package io.opentelemetry.android.test.common
+
+import io.opentelemetry.sdk.logs.data.internal.ExtendedLogRecordData
+import io.opentelemetry.sdk.testing.assertj.LogRecordDataAssert
+import io.opentelemetry.sdk.testing.assertj.OpenTelemetryAssertions.assertThat
+
+fun LogRecordDataAssert.hasEventName(eventName: String): LogRecordDataAssert {
+    isNotNull()
+    // TODO: This will be removed after the event bits of logs gets out of incubating.
```
marandaneto, line 14: just a note I guess? or should we create an issue?

marandaneto (approved):

test-common/src/main/kotlin/io/opentelemetry/android/test/common/AssertionExtensions.kt
breedx-splk: I think the note is fine....once incubator changes it should force us to notice.

PR #941, opened 2025-04-11
Relax fragment dependency

breedx-splk:
This changes the dependency on navigation-fragment to compileOnly for all modules except fragment intsrumentation.

This should prevent a transitive dependency for users who do not use fragments, and doesn't force a newer version on users who might use an older version.

I tried the demo app, and it came up just fine. I believe that this is because the demo-app uses the agent, and the agent includes fragment instrumentation, so it picks up the transitive dependency as normal. It's a little worrying to have references to Fragment classes with a mere `compileOnly()`, so we should figure out what our end state should look like.

### DIFF
--- a/instrumentation/activity/build.gradle.kts
+++ b/instrumentation/activity/build.gradle.kts
@@ -23,7 +23,8 @@ dependencies {
     implementation(project(":common"))
     implementation(libs.opentelemetry.sdk)
     implementation(libs.androidx.core)
-    implementation(libs.androidx.navigation.fragment)
+    compileOnly(libs.androidx.navigation.fragment)
     implementation(libs.opentelemetry.instrumentation.api)
+    testImplementation(libs.androidx.navigation.fragment)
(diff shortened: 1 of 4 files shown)


marandaneto:
I think the only problem is if any classes that depend on the fragment lib try to be loaded/instantiated.
is that the case?
if so, we'd need to do reflection and check if the fragment lib is available at runtime before doing so:
eg

```
val clazz = "androidx.fragment.app.FragmentManager$FragmentLifecycleCallbacks"
val isFragmentAvailable = try Class.forName(clazz) catch ...
if (!isFragmentAvailable) return
// import and use android fragment classes
```

otherwise the app will crash at runtime

bidetofevil:
Feels like some instrumentation, especially ones with special deps, should be pulled into its own module and can thus be not included for folks who don't want it.

bidetofevil:
What remains of this should be handled in #961

PR #943, opened 2025-04-14
 Adding item to cart fix

### LINKED ISSUE #940: [demo-app] Sometimes can't add items to cart
Something weird happens when you first "Go Shoping" and then choose "National Park Foundation Explorascope" and then "Add To Cart" without touching the quantity (use the default 1). Then view the cart. There is nothing added to the cart.

I think it doesn't happen with other items, though, which is weird.

magda-woj:
Fixes #940 - took a line that adds an item to cart out of `else` for item being National Park Foundation Explorascope.


breedx-splk (approved):
Perfect, thanks!

PR #953, opened 2025-04-17
Capture click events for non-compose views

### LINKED ISSUE #930: Capture click events for non-compose views
Capturing touch events can be helpful during troubleshooting to better understand which UI element triggered a sequence of events leading up to the failure. It may also be useful for session replays.

cleverchuk:
Fixes #930 

Adds `android.view.View` click capture instrumentation. This instrumentation uses the `Window.Callback` API to intercept window motion events. During the interception, it walks the view tree to figure out which view was clicked by doing a boundary overlap check of the each view position against the touch position. `ComposeView` aren't eligible for click capturing in this approach.


instrumentation/view-click/src/main/java/io/opentelemetry/android/instrumentation/view/click/ViewClickInstrumentation.java
```diff
@@ -0,0 +1,28 @@
...
+import com.google.auto.service.AutoService;
+import io.opentelemetry.android.instrumentation.AndroidInstrumentation;
+import io.opentelemetry.android.instrumentation.InstallationContext;
+
+@AutoService(AndroidInstrumentation.class)
+public class ViewClickInstrumentation implements AndroidInstrumentation {
+    public static final String INSTRUMENTATION_NAME = "android.view.click";
+
+    @Override
+    public void install(@NonNull InstallationContext ctx) {
```
bidetofevil, line 18: It seems like instead of creating all those standalone `object` singletons, you can just make them classes and instantiate and pass the here, i.e. 

A instance of `EventBuilderCreator` can be created here that takes a `Logger` as a parameter, which can be obtained from the `InstallationContext`. That can then be passed into a new instance of `ViewClickEventGenerator` which can in turn be passed into `ViewClickActivityCallback`. The callback itself can then access the `ViewClickEventGenerator` by having it passed in via its constructor.

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/EventBuilderCreator.kt
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @JvmStatic
+    fun createEvent(name: String): ExtendedLogRecordBuilder =
+        eventLogger
+            .logRecordBuilder()
+            .setEventName(name)
+
+    @JvmStatic
+    fun createViewAttributes(view: View): Attributes {
```
bidetofevil, line 38: this and `viewToName` seems to contain implementation details - any reason why these are public rather than have their setting be encapsulated in a method that will emit the event?

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewUtils.kt
```diff
@@ -0,0 +1,22 @@
...
+
+const val APP_SCREEN_CLICK_EVENT_NAME = "app.screen.click"
+const val VIEW_CLICK_EVENT_NAME = "event.app.widget.click"
+val viewNameAttr: AttributeKey<String> = stringKey("app.widget.name")
+
+val xCoordinateAttr: AttributeKey<Double> = doubleKey("app.screen.coordinate.x")
+val yCoordinateAttr: AttributeKey<Double> = doubleKey("app.screen.coordinate.y")
+val viewIdAttr: AttributeKey<Long> = longKey("app.widget.id")
+
+internal val View.isVisible: Boolean
```
bidetofevil, line 21: Why not just stick extension function into the one class that needs it and make it private?

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewUtils.kt
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+package io.opentelemetry.android.instrumentation.view.click
+import android.view.View
+import io.opentelemetry.api.common.AttributeKey
+import io.opentelemetry.api.common.AttributeKey.doubleKey
+import io.opentelemetry.api.common.AttributeKey.longKey
+import io.opentelemetry.api.common.AttributeKey.stringKey
+
+const val APP_SCREEN_CLICK_EVENT_NAME = "app.screen.click"
```
bidetofevil, line 13: Why not just declare these constants in the main instrumentation class?

instrumentation/view-click/src/test/java/io/opentelemetry/android/instrumentation/view/click/ViewClickInstrumentationTest.kt
```diff
@@ -0,0 +1,320 @@
...
+import io.opentelemetry.sdk.testing.junit4.OpenTelemetryRule
+import org.junit.Before
+import org.junit.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
+@ExtendWith(MockKExtension::class)
+class ViewClickInstrumentationTest {
```
bidetofevil, line 39: You can probably test this without mocks and just observe the events/logs that are generated after you hook up the callback to an Activity and simulate the click. At worse, you can manually invoke the methods the on the callback. Either way, you should be able to see log with the expected values generated rather than having to wire up these mocks.

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
```diff
@@ -0,0 +1,122 @@
...
+import android.os.Build
+import android.os.Build.VERSION_CODES
+import android.view.MotionEvent
+import android.view.View
+import android.view.ViewGroup
+import android.view.Window
+import java.lang.ref.WeakReference
+import java.util.LinkedList
+
+internal object ViewClickEventGenerator {
```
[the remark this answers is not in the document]
LikeTheSalad: +1 - That could also help make this PR smaller. Also, we're not sure if making this instrumentation work with API 21 is needed right now, so I think it's worth keeping it simple for now and waiting to see if someone asks for API level 21 support later.

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/EventBuilderCreator.kt
```diff
@@ -0,0 +1,55 @@
...
+        val builder = Attributes.builder()
+        builder.put(viewNameAttr, viewToName(view))
+        builder.put(viewIdAttr, view.id.toLong())
+
+        builder.put(xCoordinateAttr, view.x.toDouble())
+        builder.put(yCoordinateAttr, view.y.toDouble())
```
MustafaHaddara, line 44: In addition to what we have here, how do we feel about doing something like this:

``
if (view instanceof TextView) {
    setAttribute("view.text", (TextView)view.text)
}
```

Since most clickable things (ie. `ButtonView` etc.) will be subclasses of `TextView`, and this may be more descriptive than the resourceEntry/id strategy in `viewToName`

**Review by MustafaHaddara — APPROVED** [review]

File: instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/EventBuilderCreator.kt [hunk 8]
```diff
@@ -0,0 +1,55 @@
...
+package io.opentelemetry.android.instrumentation.view.click
+import android.view.View
+import io.opentelemetry.android.instrumentation.InstallationContext
+import io.opentelemetry.api.OpenTelemetry
+import io.opentelemetry.api.common.Attributes
+import io.opentelemetry.api.incubator.logs.ExtendedLogRecordBuilder
+import io.opentelemetry.api.incubator.logs.ExtendedLogger
+
+object EventBuilderCreator {
+    private var eventLogger =
```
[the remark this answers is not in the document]
cleverchuk: i'm trying to avoid branching.

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/EventBuilderCreator.kt
```diff
@@ -0,0 +1,55 @@
...
+
+package io.opentelemetry.android.instrumentation.view.click
+import android.view.View
+import io.opentelemetry.android.instrumentation.InstallationContext
+import io.opentelemetry.api.OpenTelemetry
+import io.opentelemetry.api.common.Attributes
+import io.opentelemetry.api.incubator.logs.ExtendedLogRecordBuilder
+import io.opentelemetry.api.incubator.logs.ExtendedLogger
+
+object EventBuilderCreator {
```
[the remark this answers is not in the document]
cleverchuk: It looks like you're saying this is best turned in a class? If that's the case, it looks like making it a class requires more boilerplate code than how it is now? I think it's okay because it's already a singleton anyway.

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewUtils.kt
cleverchuk: field so it could be accessed as property.

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewUtils.kt
cleverchuk: i like them separate.

instrumentation/view-click/src/test/java/io/opentelemetry/android/instrumentation/view/click/ViewClickInstrumentationTest.kt
cleverchuk: How about an example?

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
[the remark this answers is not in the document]
cleverchuk: please suggest another way to deal with animalsniffer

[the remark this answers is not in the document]
LikeTheSalad: > please suggest another way to deal with animalsniffer

What we do about those is to annotate the class or method that makes use of higher-level apis with the RequiresApi annotation, [as shown here](https://github.com/open-telemetry/opentelemetry-android/blob/728f004784bfd38511e5718deb4c3ebb8069c014/instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/HttpUrlReplacements.java#L86) for example.

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
```diff
@@ -0,0 +1,147 @@
...
+        view.getLocationInWindow(viewCoordinates)
+        val vx = viewCoordinates[0]
+        val vy = viewCoordinates[1]
+
+        val w = view.width
+        val h = view.height
+        return !(x < vx || x > vx + w || y < vy || y > vy + h)
+    }
+
+    private fun isJetpackComposeView(view: View): Boolean = view::class.java.name.startsWith("androidx.compose.ui.platform.ComposeView")
```
marandaneto, line 143: if minify is enabled, this might fail? is `androidx.compose.ui.platform.ComposeView` minified or an exception (Added to proguard etc?)

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
```diff
@@ -0,0 +1,147 @@
...
+            }
+        }
+    }
+
+    private fun hitTest(
+        view: View,
+        x: Float,
+        y: Float,
+    ): Boolean {
+        view.getLocationInWindow(viewCoordinates)
```
marandaneto, line 134: wondering if it should be this one or `getLocationOnScreen`
Another thing to consider is the density of the screen since this slightly changes the x/y pos

marandaneto, line 134: wondering if it should be this one or `getLocationOnScreen`
Another thing to consider is the density of the screen since this slightly changes the x/y pos

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
```diff
@@ -0,0 +1,147 @@
...
+            .logRecordBuilder()
+            .setEventName(name)
+
+    private fun createViewAttributes(view: View): Attributes {
+        val builder = Attributes.builder()
+        builder.put(viewNameAttr, viewToName(view))
+        builder.put(viewIdAttr, view.id.toLong())
+
+        builder.put(xCoordinateAttr, view.x.toDouble())
+        builder.put(yCoordinateAttr, view.y.toDouble())
```
marandaneto, line 74: should we also call `getLocationInWindow` or `getLocationOnScreen` here?

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
```diff
@@ -0,0 +1,147 @@
...
+            }
+
+            if (view is ViewGroup) {
+                handleViewGroup(view, x, y, queue)
+            }
+        }
+        return target
+    }
+
+    private fun isValidClickTarget(view: View): Boolean = view.isClickable && view.isVisible
```
marandaneto, line 111: this can be a bit more clever. not sure if needed though.
eg https://github.com/PostHog/posthog-android/blob/d232586ce6b5043d716f108d06316d8de4714c31/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt#L479-L520

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
marandaneto: eg https://github.com/PostHog/posthog-android/blob/d232586ce6b5043d716f108d06316d8de4714c31/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt#L707-L710

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
marandaneto: eg https://github.com/PostHog/posthog-android/blob/d232586ce6b5043d716f108d06316d8de4714c31/posthog-android/consumer-rules.pro#L93-L94

cleverchuk: yea, the test will fail and result in traversing compose view tree. however, i don't see how an exception will result here.

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
cleverchuk: The current works fine since we're interested in the window the app is running in and it's taken from [here](https://github.com/DataDog/dd-sdk-android/blob/0e0e229941686159808dc7d127b7b893231132fa/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/internal/instrumentation/gestures/AndroidActionTrackingStrategy.kt#L59-L73)

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
cleverchuk: i don't think it's necessary here because we're not doing bound checks here.

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/EventBuilderCreator.kt
breedx-splk: We agreed in SIG that we think it's a good idea but would prefer to defer this to a later enhancement.

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
marandaneto: ok, i'm not sure if `view.x` and `view.y` are relative to the parent or absolute to the window, if its relative to the parent, you'd need `getLocationInWindow` right otherwise the position will be wrong

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/internal/ViewUtils.kt
```diff
@@ -0,0 +1,19 @@
...
+import io.opentelemetry.api.common.AttributeKey.doubleKey
+import io.opentelemetry.api.common.AttributeKey.longKey
+import io.opentelemetry.api.common.AttributeKey.stringKey
+
+const val APP_SCREEN_CLICK_EVENT_NAME = "app.screen.click"
+const val VIEW_CLICK_EVENT_NAME = "event.app.widget.click"
+val viewNameAttr: AttributeKey<String> = stringKey("app.widget.name")
+
+val xCoordinateAttr: AttributeKey<Double> = doubleKey("app.screen.coordinate.x")
+val yCoordinateAttr: AttributeKey<Double> = doubleKey("app.screen.coordinate.y")
```
breedx-splk, line 18: These should not be `Double` they should be `Long`. It is [specified](https://github.com/open-telemetry/semantic-conventions/tree/2073d8ab5e6274e6f883b8ff89d358fdee510c58/docs/app#click-events) in pixels, which is never fractional/decimal and is always a whole number.

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
cleverchuk: Not sure, I understand what you're asking. However, it sounds like you're saying I should create a consumer rules with [this](https://github.com/PostHog/posthog-android/blob/d232586ce6b5043d716f108d06316d8de4714c31/posthog-android/consumer-rules.pro#L93-L94) as the content?

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
cleverchuk: From what i understand from the [doc](https://developer.android.com/reference/android/view/View#getX()), it's absolute.

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
cleverchuk: no need for that.

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/internal/ViewUtils.kt
cleverchuk: They're floats in the platform between.

instrumentation/view-click/src/test/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickInstrumentationTest.kt
```diff
@@ -0,0 +1,326 @@
...
+
+        val events = openTelemetryRule.logRecords
+        assertThat(events).hasSize(2)
+
+        var event = events[0] as ExtendedLogRecordData
+        assertThat(event)
+            .hasEventName(APP_SCREEN_CLICK_EVENT_NAME)
+            .hasAttributesSatisfyingExactly(
+                equalTo(xCoordinateAttr, motionEvent.x.toDouble()),
+                equalTo(yCoordinateAttr, motionEvent.y.toDouble()),
```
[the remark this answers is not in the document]
cleverchuk: same as above

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/internal/ViewUtils.kt
breedx-splk: Ok, I read these docs https://developer.android.com/reference/android/view/MotionEvent#getX(int) and it sounds like there is the possibility of some devices providing subpixel precision, in which case it sounds like we should probably [change the current semconv](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/app.md) to match. 

What do you think?

breedx-splk: Looks like you changed it to long..thanks!

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
```diff
@@ -0,0 +1,147 @@
...
+import io.opentelemetry.api.incubator.logs.ExtendedLogger
+import java.lang.ref.WeakReference
+import java.util.LinkedList
+
+class ViewClickEventGenerator(
+    private val eventLogger: ExtendedLogger,
+) {
+    private var windowRef: WeakReference<Window>? = null
+
+    private val viewCoordinates = IntArray(2)
```
bidetofevil, line 29: The underlying coordinates are Longs - why not go with a `Pair<Long, Long>`?

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
bidetofevil: What if there's another class that implements `AbstractComposeView` directly?

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/WindowCallbackWrapper.kt
```diff
@@ -0,0 +1,34 @@
...
+package io.opentelemetry.android.instrumentation.view.click
+
+import android.os.Build.VERSION_CODES
+import android.view.ActionMode
+import android.view.MotionEvent
+import android.view.SearchEvent
+import android.view.Window.Callback
+import androidx.annotation.RequiresApi
+
+class WindowCallbackWrapper(
```
bidetofevil, line 15: You may just be able to annotate the class rather than the methods so it'll detect the creation

bidetofevil (approved):
LGTM. Some non-blocking comments, but the one about using a data structure that stores Longs instead of Ints for the coordinates is probably a good idea to do before merging.

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
cleverchuk: @bidetofevil it looks like that doesn't exist in the Android compose framework yet and we can address them when that changes?

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
cleverchuk: `view.getLocationInWindow` API demands it.

breedx-splk (approved):
Thanks for the contribution!

PR #958, opened 2025-04-18
Updated Export Frequency to 60 Seconds and Added Disk Export Parameter

### LINKED ISSUE #947: Create configuration parameter for disk-buffering exporting frequency
### Background
Currently the disk buffering exporter attempts to export every 10 seconds - the `WorkDelegator` has a loop of 10 second to check, if the `PeriodicRunnable` `isReadyToRun()`:

https://github.com/open-telemetry/opentelemetry-android/blob/7238ac498aa2b74760f0a9958800fd8138fea23c/services/src/main/java/io/opentelemetry/android/internal/services/periodicwork/PeriodicWork.kt#L35

... and the `DefaultExportScheduler` uses the same 10 seconds waiting period to show its readiness:

https://github.com/open-telemetry/opentelemetry-android/blob/7238ac498aa2b74760f0a9958800fd8138fea23c/core/src/main/java/io/opentelemetry/android/features/diskbuffering/scheduler/DefaultExportScheduler.kt#L20

### Problem
Since my internal enterprise apps are used by colleagues 8 hours a day it is critical to not put too much strain on their phone's batteries. The 10 second exporting frequency means that the phone has to do IO 2880 times during an 8 hour workday, if my Maths is correct, which feels like a lot both for the devices and for the backend to handle as well, if the number of devices is high.

### Proposed solution

I would increase the default exporting frequency to 1 minute. Additionally, I would create a new config parameter in `DiskBufferingConfig`, which allows setting the exporting frequency catering for various preferences on the tradoff between real-time-ness and battery consumption.

Alternatively or additionally, the `WorkManager` API could be used (#882), which raises the minimum frequency to 1x in 15 minutes and optimizes the battery life even better.

Yadavanurag13:
This PR closes #947 

I have updated the export frequency to 60 sec and added the parameter in the config.

I'm raising this pr. I'm new to this repo so, In case there any review comments pls let me know.

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfig.kt
+++ b/core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfig.kt
@@ -27,7 +27,7 @@ data class DiskBufferingConfig
         val maxFileAgeForReadMillis: Long = TimeUnit.HOURS.toMillis(DEFAULT_MAX_FILE_AGE_FOR_READ_MS),
         val maxCacheFileSize: Int = MAX_CACHE_FILE_SIZE,
         val debugEnabled: Boolean = false,
-        val exportIntervalInSeconds: Long = DEFAULT_EXPORT_INTERVAL_IN_SECONDS,
+        val exportIntervalInSeconds: Duration = DEFAULT_EXPORT_INTERVAL_IN_SECONDS,
         /**
          * The directory where the SDK stores the buffered signals before they are exported. If
          * `null`, a default directory inside the application's cache directory will be used.
--- a/core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfig.kt
+++ b/core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfig.kt
@@ -15,6 +15,7 @@ const val MAX_CACHE_FILE_SIZE: Int = 1024 * 1024
 const val DEFAULT_MAX_FILE_AGE_FOR_WRITE_MS = 30L
 const val DEFAULT_MIN_FILE_AGE_FOR_READ_MS = 33L
 const val DEFAULT_MAX_FILE_AGE_FOR_READ_MS = 18L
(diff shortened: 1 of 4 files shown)


breedx-splk:
Hi @Yadavanurag13 you will need to sign the CLA and get the build passing. Thanks!

Yadavanurag13:
@breedx-splk I was trying to sign in as an individual contributor, but I got an error that this project is only for employer contributors is it so ?

breedx-splk:
> @breedx-splk I was trying to sign in as an individual contributor, but I got an error that this project is only for employer contributors is it so ?

It's been a while since I signed it, but you should totally be able to sign as an individual contributor.

https://opentelemetry.io/docs/contributing/prerequisites/#cla

Please make sure that the git commits were made with the same email address that you are signing with (that sometimes trips people up).

core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfig.kt
```diff
@@ -26,6 +27,7 @@ data class DiskBufferingConfig
         val maxFileAgeForReadMillis: Long = TimeUnit.HOURS.toMillis(DEFAULT_MAX_FILE_AGE_FOR_READ_MS),
         val maxCacheFileSize: Int = MAX_CACHE_FILE_SIZE,
         val debugEnabled: Boolean = false,
+        val exportIntervalInSeconds: Long = DEFAULT_EXPORT_INTERVAL_IN_SECONDS,
```
marandaneto, line 30: should we use `Duration` instead of `Long`?
eg https://github.com/open-telemetry/opentelemetry-android/blob/c2ddbe09784e1a5e24d16ac29b876e245153e138/session/src/main/kotlin/io/opentelemetry/android/session/SessionConfig.kt#L13-L14

breedx-splk:
@Yadavanurag13 can you address the few comments above please? Thanks again!

Yadavanurag13:
@breedx-splk Sure !!

PR #959, opened 2025-04-21
Move Session Config up

MdNaushad:
* When we are using opentelemetry-android in a app and want to configure the SessionConfig, it gives error. 
`Unresolved reference: SessionConfig`. 
* It looks like SessionConfig is declared inside `session/src/main/kotlin/io/opentelemetry/android/session/SessionConfig.kt`
* When `opentelemetry-android` is built, it doesn't expose the `sessionConfig` out for the child projects to use.

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
+++ b/core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
@@ -37,7 +37,7 @@ import io.opentelemetry.android.internal.services.network.CurrentNetworkProvider
 import io.opentelemetry.android.internal.services.periodicwork.PeriodicWork;
 import io.opentelemetry.android.internal.session.SessionIdTimeoutHandler;
 import io.opentelemetry.android.internal.session.SessionManagerImpl;
-import io.opentelemetry.android.session.SessionConfig;
+import io.opentelemetry.android.config.SessionConfig;
 import io.opentelemetry.android.session.SessionManager;
 import io.opentelemetry.android.session.SessionProvider;
 import io.opentelemetry.api.baggage.propagation.W3CBaggagePropagator;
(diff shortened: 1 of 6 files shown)


breedx-splk:
@MdNaushad it looks like tis PR simply moves the `SessionConfig` class out of the session module and into the `config` package of the `core` module. Is that right? 

I'm not convinced that we should do this -- the session related stuff is currently all grouped together in the `session` module, and this starts to change that. I do understand that it's a little bit confusing...but I think the current way that you can solve this is to include a dependency on `io.opentelemetry.android:session`. Let us know if that works for you.

MdNaushad:
> @MdNaushad it looks like tis PR simply moves the `SessionConfig` class out of the session module and into the `config` package of the `core` module. Is that right?
> 
> I'm not convinced that we should do this -- the session related stuff is currently all grouped together in the `session` module, and this starts to change that. I do understand that it's a little bit confusing...but I think the current way that you can solve this is to include a dependency on `io.opentelemetry.android:session`. Let us know if that works for you.

Yes, it's simply moving the class. 
I am thinking from user's perspective. If I want to just configure session timeout or background activity time out, I would need to explicitly add a dependency on `io.opentelemetry.android:session` even though I am not using anything out of it. 

If we want to keep SessionConfig inside the session module, I believe may be we can have methods on OtelRumConfig like
OtelRumConfig.setBackgroundActivityTimeout
OtelRumConfig.setMaxLifeTimeout

This way user won't have to add dependency on session module. Let me know your thoughts

breedx-splk:
@MdNaushad Ok that's cool...we may just want to move the other config(s) into a common place as well.
I've added an item for the SIG meeting tomorrow to discuss unless people chime in here first. :) 

Regardless, you'll need to sign the CLA before we can accept the PR. Thanks so much!

LikeTheSalad (approved):
Thank you 🙏

breedx-splk (approved):

PR #961, opened 2025-04-22
Remove runtime dep on androidx fragment navigiation

bidetofevil:
Remove runtime dependency by looking at whether the fragment is a `NavHost`, which is not part of the `navigation-fragment` package. We are still depending on AndroidX's Navigation API because that's what the logic checks. Further decoupling will require the implementation to not know about the concept of `NavHost`, which will be trickier to do and out of scope for what I'm trying to achieve, i.e. remove the dep to the fragments navigation package only.

### DIFF
--- a/services/build.gradle.kts
+++ b/services/build.gradle.kts
@@ -17,10 +17,11 @@ dependencies {
     implementation(project(":common"))
 
     implementation(libs.androidx.core)
-    implementation(libs.androidx.navigation.fragment)
+    implementation(libs.androidx.navigation.runtime.ktx)
     implementation(libs.androidx.lifecycle.process)
     implementation(libs.androidx.preference.ktx)
 
     testImplementation(libs.androidx.test.core)
     testImplementation(libs.robolectric)
+    testImplementation(libs.androidx.navigation.fragment)
 }
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -13,11 +13,13 @@ kotlin = "2.1.20"
 androidPlugin = "8.9.1"
 junitKtx = "1.2.1"
 autoService = "1.1.1"
+androidx-navigation = "2.7.7"
 
 [libraries]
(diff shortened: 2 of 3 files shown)


bidetofevil:
Added additional commit that made `ScreenNameExtractor` agnostic to the input class. This enabled me to remove the dependency from all the modules that don't actually need it.

LikeTheSalad (approved):
Thank you 🙏

instrumentation/common-api/src/test/java/io/opentelemetry/android/instrumentation/common/ScreenNameExtractorTest.kt
```diff
@@ -0,0 +1,58 @@
...
+import io.opentelemetry.android.instrumentation.annotations.RumScreenName
+import org.junit.jupiter.api.Assertions
+import org.junit.jupiter.api.Test
+
+internal class ScreenNameExtractorTest {
+    @Test
+    fun testActivity() {
+        val activity = Activity()
+        val name = ScreenNameExtractor.DEFAULT.extract(activity)
+        Assertions.assertEquals("Activity", name)
```
breedx-splk, line 19: Please use static imports for these, or ever better, migrate to the fluent assertj assertions: `assertThat(name).isEqualTo("Activity")`. 

I know you didn't write these and migrating to kotlin was just part of the cleanup, but it would help here. Thanks!

services/build.gradle.kts
```diff
@@ -17,10 +17,11 @@ dependencies {
     implementation(project(":common"))
 
     implementation(libs.androidx.core)
-    implementation(libs.androidx.navigation.fragment)
+    implementation(libs.androidx.navigation.runtime.ktx)
```
breedx-splk, line 20: Is this a better dependency? Do we think that users might have the same concern?

breedx-splk (approved):

services/build.gradle.kts
marandaneto: > Do we think that users might have the same concern?

most likely but this one only has a few transitive dependencies and its mainly the common/runtime bits, compared to the `fragment`, its much more niche/less likely to be an issue

marandaneto (approved):

Issue #841, opened 2025-02-19
Impossible to apply `SessionIdRatioBasedSampler` after session refactoring


GerardPaligot:
Now that we should use `SessionManager` to generate the session id and because `ServiceManager getServiceManager` from `OpenTelemetryRumBuilder` or `SessionManagerImpl` aren't public, it is impossible to call `setSampler` from `SdkTracerProviderBuilder` to initialize `SessionIdRatioBasedSampler` class.

```kotlin
otelRumBuilder.addTracerProviderCustomizer { tracerProvider, _ ->
  return@addTracerProviderCustomizer tracerProvider
    .setSampler(SessionIdRatioBasedSampler(ratio, /* session provider here but we can't get it */))
}
```

LikeTheSalad:
cc @breedx-splk @surbhiia @bidetofevil 

This sounds related to what we discussed in the last SIG meeting, in which we should allow the session manager to be retrieved from OTel RUM. Then we would be able to enable a lot of use cases by providing the OTel RUM object, including the automatic instrumentation that needs access to the session manager to work.

GerardPaligot:
Getting an access from OTeL RUM is too late because we need it during its initialization (builder phase). Another solution is to provide our own `SessionManager` (and by extension, the `SessionProvider`) but `SessionManagerImpl` isn't public.

Or another idea is to create `setSampler` function (from my first message) directly in `OpenTelemetryRumBuilder` to avoid exposing the session manager during the builder phase.

What do you think about this? This issue is a blocker to upgrade to latest versions.

LikeTheSalad:
> Or another idea is to create `setSampler` function (from my first message) directly in `OpenTelemetryRumBuilder` to avoid exposing the session manager during the builder phase.

Can you elaborate on this, please? If I understood correctly, you need to be able to set a sampler that is aware of the current session. If we create a setSampler in the builder, how could it help to address the missing session issue? I think this option sounds good, I just want to understand the details of what's needed and the overall approach that you mentioned.

GerardPaligot:
I want to apply a sampling based on the session that is handled by opentelemetry-android. To do that, I need to customize my trace provider with the sampler `SessionIdRatioBasedSampler` that needs a ratio and the session provider. But the session provider isn't yet accessible during the builder phase (before we've initialized the OpenTelemetry instance).

[...]

PR #971, opened 2025-05-07
feat: add session-based sampling ratio support

GerardPaligot:
* Integrated SessionIdRatioBasedSampler in OpenTelemetryRumBuilder.
* Enhanced SessionConfig to support sampling ratio configuration.

Issue #841

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
+++ b/core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
@@ -545,6 +545,13 @@ public final class OpenTelemetryRumBuilder {
                 customizer : tracerProviderCustomizers) {
             tracerProviderBuilder = customizer.apply(tracerProviderBuilder, application);
         }
+
+        if (config.getSessionConfig().getRatio() != null) {
+            tracerProviderBuilder.setSampler(
+                    new SessionIdRatioBasedSampler(
+                            config.getSessionConfig().getRatio(), sessionProvider));
+        }
+
         return tracerProviderBuilder.build();
     }
 
--- a/core/src/main/java/io/opentelemetry/android/config/SessionConfig.kt
+++ b/core/src/main/java/io/opentelemetry/android/config/SessionConfig.kt
@@ -12,9 +12,20 @@ import kotlin.time.Duration.Companion.minutes
 data class SessionConfig(
     val backgroundInactivityTimeout: Duration = 15.minutes,
     val maxLifetime: Duration = 4.hours,
+    val ratio: Double? = null,
 ) {
+    constructor(
+        backgroundInactivityTimeout: Int = 15,
+        maxLifetime: Int = 4,
+        ratio: Double? = null,
+    ) : this(
+        backgroundInactivityTimeout = backgroundInactivityTimeout.minutes,
+        maxLifetime = maxLifetime.hours,
+        ratio = ratio,
+    )
+
     companion object {
         @JvmStatic
-        fun withDefaults(): SessionConfig = SessionConfig()
+        fun withDefaults(): SessionConfig = SessionConfig(15.minutes, 4.hours)
     }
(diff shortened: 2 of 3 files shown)


breedx-splk:
Yes @GerardPaligot I will second that this PR is appreciated and especially thanks for bringing up the problems with the sampler in #841. 

In addition to the things that @LikeTheSalad mentioned that we discussed in the SIG meeting, I was also enlightened in the later logs SIG about the idea of storing the session into the otel Context. This allows us to grab the Session without the need for an observer, which I think is pretty interesting and underutilized. Please have a look at #974 and see if that approach will work for you and/or if you'd have any suggestions.

Again, we really do appreciate the work and I'm not trying to take this effort away or diminish it, just trying an alternative that is hopefully agreeable to everyone.

GerardPaligot:
Don't worry guys, it is totally okay to not accept an external contribution. To be honest, I wasn't convince by myself about my own contribution. It seems that it breaks SOLID principles and I don't like the indirect link between the session config and the sampling.

@breedx-splk About your lest contribution, yes probably but let me add some comments there.

breedx-splk:
We do require the Contributor License Agreement (CLA) to be signed though.

GerardPaligot:
Thanks!

I'm currently building an internal SDK based on OpenTelemetry Android and OpenTelemetry Java to support features that aren’t yet available in this SDK, or that work differently. In particular, I avoid auto-instrumentation in my projects because I prefer to explicitly define what gets instrumented, maintaining a clear contract within the application.

That said, I think it could be valuable to open discussions on certain features that might be generic and reusable enough to contribute upstream.

Here are a few examples of what I’ve been working on:

* KSP compiler plugin: Observes a `@Monitor` annotation on Kotlin interfaces and generates a proxy class that instruments all function calls. This proxy can be injected between the interface and its implementation in a DI setup, enabling instrumentation across architectural layers.
* Lifecycle-aware observer: Similar to your activity/fragment observer, but built to be as generic as possible. In my case, it integrates with Compose’s DisposableEffect and the compose navigation library to observe screen-level navigation events.
* Coroutine entry point extensions: I created `launchMonitored`, `withContextMonitored`, and `asyncMonitored` as wrappers around coroutine builders to trace logic flowing from the presentation to domain and data layers.

These tools let me build a trace structure like this:

Capture d’écran 2025-04-29 à 16 17 50
(No critical data from this capture, so it is ok to share)

> We do require the Contributor License Agreement (CLA) to be signed though.

I'm not sure yet if I should sign it for a personal contribution or professional. Probably the second one so the process seems longer.

LikeTheSalad:
> That said, I think it could be valuable to open discussions on certain features that might be generic and reusable enough to contribute upstream.

Definitely! That'd be great 🙏 

Also, I understand the benefits of letting people choose the automatic instrumentations that they need, without having them enabled by default. This project should allow for that use case as it's split into 3 components:

- The [core](https://github.com/open-telemetry/opentelemetry-android/tree/main/core) component, where all the rum initialization is done and all the configs are as generic as possible.
- The [automatic instrumentations](https://github.com/open-telemetry/opentelemetry-android/tree/main/instrumentation), which are all independent from core, though they are installed as part of the core rum initialization.
- The [agent](https://github.com/open-telemetry/opentelemetry-android/tree/main/android-agent), which is the enduser-friendly route, which also adds high-level opinionated config options. It combines core + some instrumentations, for people who want to go with our recommended configurations out of the box and avoid having to write much code to get everything up and running as quick as possible.

It seems like, for your use case, you might want to use [core](https://github.com/open-telemetry/opentelemetry-android/tree/main/core) as base for your SDK. If you happen to spot something that's impossible to configure via core, similarly to what happened with the session sample rate, it would be very helpful to get feedback on that so that we can ensure that core is as flexible as possible to work as a generic base for any SDK.

marandaneto:
Should this be closed because other PRs have superseded it?

Issue #842, opened 2025-02-19
Expose API for user provided `SessionProvider`


cleverchuk:
There isn't a public API for users to provide their own session manager or provider implementation. The existing [API](https://github.com/open-telemetry/opentelemetry-android/blob/9c1061f39f7fb95321ffc1faf697b32c0dc49c50/core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java#L413) is package private. How about we expose an API for users to provide their own `SessionProvider` and `SessionManager`?

bidetofevil:
Feels like this is something we should allow given the interface is public for a reason. I'll take a look

breedx-splk:
I still feel like we're being too generous with the session abstraction and should be more opinionated about it. We can/should expose the id generator for advanced users, but the behavior I think we should be better about managing. 🤷🏻

Ejdangerfield:
Just my take: Many companies implementing Otel will already have their own definition of a "session" and their or format for session id. If you want to enable them to add telemetry to their existing system they ought to be able to implement their own provider.

LikeTheSalad:
I've created [this PR](https://github.com/open-telemetry/opentelemetry-android/pull/979) to address this issue.

PR #974, opened 2025-05-07
Enable session-id based probabilistic sampling via OpenTelemetryRumBuilder.

breedx-splk:
Here's an alternative to #971. ~~This updates the `SessionManagerImpl` to ensure that the session is always stored in the otel context~~, This uses a new `SessionObserver` to track the Session, and it creates a new opt-in method on the `OpenTelemetryRumBuilder` called `enableSessionBasedSampling(ratio)`.

This new method wires up an instance of the `SessionIdRatioBasedSampler` via the existing `addTracerProviderCustomizer()` mechanism. 

There's currently no real way to sample logs, so there's a placeholder/TODO to add that later.


core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
```diff
@@ -301,6 +304,36 @@ public OpenTelemetryRumBuilder addLogRecordExporterCustomizer(
...
     }
 
+    /**
+     * Enables a ratio-based probabilistic sampling of sessions. When a session is sampled
+     * (included), then the telemetry for that session will be created and exported. If a session is
+     * not sampled (excluded) then the telemetry for that session will not be exported.
+     *
+     * @param ratio - A number between 0 and 1, representing a percentage of sessions to sample.
+     */
+    public OpenTelemetryRumBuilder enableSessionBasedSampling(double ratio) {
```
GerardPaligot, line 314: The class `SessionFromContextProvider` is pretty much explicit about where we are fetching the current session but not this method. If we want to continue to use a custom `SessionManager` (with `setSessionManager`) and use this method to apply a sampling, we need to know that we should save our session id in the context with a specific context key. Hard to know for someone that has limited knowledge about the internal implementation of this project.

GerardPaligot: Another proposition without the need of a context and allow us to provide a custom `SessionManager` if it is require by the consumer.

```java
    public OpenTelemetryRumBuilder enableSessionBasedSampling(double ratio) {
        return enableSessionBasedSampling(ratio, SessionManagerImpl.create());
    }

    public OpenTelemetryRumBuilder enableSessionBasedSampling(double ratio, SessionManager sessionManager) {
        Sampler sampler = buildSampler(ratio, sessionManager);
        addTracerProviderCustomizer((builder, application) -> builder.setSampler(sampler));

        // TODO: When there is a mechanism for sampling logs, we need to wire
        // it up here...

        this.sessionManager = sessionManager;

        return this;
    }

    private static Sampler buildSampler(double ratio, SessionProvider sessionProvider) {
        if (ratio < 0.0 || ratio > 1.0) {
            throw new IllegalArgumentException("ratio must be in range [0.0, 1.0]");
        }
        if (ratio == 0.0) {
            return Sampler.alwaysOff();
        }
        if (ratio == 1.0) {
            return Sampler.alwaysOn();
        }
        return new SessionIdRatioBasedSampler(ratio, sessionProvider);
    }
```

core/src/main/java/io/opentelemetry/android/internal/session/SessionFromContextProvider.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import io.opentelemetry.android.session.Session
+import io.opentelemetry.android.session.SessionProvider
+import io.opentelemetry.context.Context
+
+/**
+ * A SessionProvider instance that requires the session to be stored
+ * in the current otel Context.
+ */
+class SessionFromContextProvider : SessionProvider {
```
GerardPaligot, line 16: If we keep context, this class should be probably `internal`.

core/src/main/java/io/opentelemetry/android/internal/session/SessionManagerImpl.kt
```diff
@@ -11,10 +11,14 @@ import io.opentelemetry.android.session.SessionIdGenerator
 import io.opentelemetry.android.session.SessionManager
 import io.opentelemetry.android.session.SessionObserver
 import io.opentelemetry.android.session.SessionStorage
+import io.opentelemetry.context.Context
+import io.opentelemetry.context.ContextKey
 import io.opentelemetry.sdk.common.Clock
 import java.util.Collections.synchronizedList
 import kotlin.time.Duration
 
+val SESSION_CONTEXT_KEY: ContextKey<Session> = ContextKey.named("rum.session")
```
GerardPaligot, line 20: Why not just `session` like for the trace attribute?

core/src/main/java/io/opentelemetry/android/internal/session/SessionManagerImpl.kt
```diff
@@ -57,6 +61,10 @@ internal class SessionManagerImpl(
         if (newSession != session) {
             val previousSession = session
             session = newSession
+            Context
+                .current()
```
LikeTheSalad, line 65: Considering that a context can change when making a span current, I'm wondering what would happen when this code is called within a nested span, for example:

```
Span A
├─ Span B <-- This is the current context
Span C
```

In that example, the new session will get stored within the Span B context, which will end before the Span C is created. Will the "Span C" have access to the new session? Given that it's not the same context?

core/src/main/java/io/opentelemetry/android/internal/session/SessionManagerImpl.kt
```diff
@@ -57,6 +61,10 @@ internal class SessionManagerImpl(
...
+                .with(SESSION_CONTEXT_KEY, session)
+                .makeCurrent()
```
LikeTheSalad, line 67: Could this cause a context leak?

For example:

```
Span A
├─ Span B
```

If "Span B" is made current, it has a Scope that must be closed by the time it ends so that the context is restored to what it was before Span B (i.e. Span A in this case). However, if we change the current context before Span B ends, it seems like by the time its Scope is closed, it won't be able to return to Span A's context, based on [this logic](https://github.com/open-telemetry/opentelemetry-java/blob/eec21201b1b94a8838fd935b2eb5839211c5695e/context/src/main/java/io/opentelemetry/context/ThreadLocalContextStorage.java#L48).

core/src/main/java/io/opentelemetry/android/internal/session/SessionManagerImpl.kt
breedx-splk: I'm open to it, but I wanted it to be explicit that it was the rum session and not any other session that might be in the context in the future....like a user session or a shopping session or I dunno.

core/src/main/java/io/opentelemetry/android/internal/session/SessionManagerImpl.kt
breedx-splk: Yeah this is a good point that I didn't consider. I don't think that spans are relevant, but the fact that this calls `makeCurrent()` to create new scopes that never close is definitely a leak situation.

breedx-splk:
I really liked the idea of "just store the session in the context" when I first heard it. I thought it was a super interesting thing. Now after having implemented it and gathering some feedback, I think isn't viable....and I'll now attempt to summarize why before closing this out:

1. There isn't one context...at least in Java. The idea of "store the session in the otel context" should be met with the question of "which otel context"? The "current" context changes all the time, depending on several factors, not the least of which is thread context and span context. And the root context (and all children of it) are all immutable...which makes storing something "global" like the session kind of moot, as it's impossible.
2. Having two places to get the session id from is an anti-pattern IMO. If code sometimes consults the `SessionManager` and sometimes it reads it from the context, then this is inconsistent. In fact, being able to read the session from the context bypasses the current mechanism that only changes the session id on read...effectively breaking the timeout mechanism that is initiated on read (get).
3. Having a user-providable `SessionManager` interface means that none of the SessionManager behaviors can be assumed...and I hate this. There should just be one session manager, and it should opinionated...this would make all of our lives easier.

Thanks for playing out this exercise with me!

For now, I'm going to close this with the intention of replacing the Context bit with an observer and see how that's received.

core/src/main/java/io/opentelemetry/android/internal/session/SessionManagerImpl.kt
breedx-splk: I was pretty sure that by setting this early initially we would be on the root context, but I think there's nothing to ensure this, and so it's risky. Good catch.

breedx-splk:
Another attempt that uses a `SessionObserver` that follows the session instead of relying on the Context.

core/src/main/java/io/opentelemetry/android/internal/session/SessionObservingSessionProvider.kt
```diff
@@ -0,0 +1,28 @@
...
+import io.opentelemetry.android.session.Session
+import io.opentelemetry.android.session.SessionObserver
+import io.opentelemetry.android.session.SessionProvider
+import java.util.concurrent.atomic.AtomicReference
+
+/**
+ * A SessionObserver that listens for session start events, stores the session,
+ * and is also a SessionProvider, that can return the session id.
+ */
+internal class SessionObservingSessionProvider :
```
GerardPaligot, line 17: From what I understand, this class was introduced due to the belief that:

> Having a user-providable SessionManager interface means that none of the SessionManager behaviors can be assumed...and I hate this. There should just be one session manager, and it should opinionated...this would make all of our lives easier.

However, I'm not fully convinced this is actually a problem. The current `SessionManager` already extends both `SessionProvider` and `SessionPublisher`, which together define a clear contract. Any implementation (whether user-provided or internal) must satisfy this contract. While the observer pattern might make sense for external libraries that shouldn't depend directly on the OpenTelemetry Android sdk, I’m not sure it’s the best fit within the sdk itself.

The more I think about it, the more I believe that the SDK may not need to own the concept of a session at all. On Android, sessions are often used beyond telemetry (for example, for analytics or app-specific behaviors). In my own experience, I work on a kiosk application that resets itself after a configurable timeout, triggering a new session. In such scenarios, having a centralized, user-defined session manager becomes critical, as it allows multiple tools (including OpenTelemetry) to share a consistent session lifecycle.

If that's the case, then perhaps the SDK should act only as a consumer of session data rather than its owner. In this context, adopting an observer pattern makes more sense. But to fully embrace that model, maybe all session-related logic should be removed from the core sdk, with the option for users to plug in either a custom or provided basic implementation.

Just my two cents, curious to hear yours.

GerardPaligot (commented):
While I still have some questions about adopting the observer pattern in this contribution, I prefer this approach over the previous one that relied on the OpenTelemetry context.

core/src/main/java/io/opentelemetry/android/internal/session/SessionObservingSessionProvider.kt
LikeTheSalad: I think I see what you mean. In this case, core is being opinionated about what a session must be. Based on the project structure that I [mentioned here](https://github.com/open-telemetry/opentelemetry-android/pull/971#issuecomment-2865971815), I think one approach we could take to avoid it, is to move our opinionated implementation of session manager from core into the agent, and then allow the interface to be set in core. That way, endusers will still have our opinionated version of a session, while other SDKs and/or expert users can define their own with core.

LikeTheSalad: I've created [this PR](https://github.com/open-telemetry/opentelemetry-android/pull/979) with the approach that I was mentioning in yesterday's SIG meeting, @breedx-splk, please have a look to see what you think. I think it should address both @GerardPaligot's issue and also [this other one](https://github.com/open-telemetry/opentelemetry-android/issues/842#issuecomment-2877907916) and it will also simplify the core API surface.

core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
```diff
@@ -301,6 +305,21 @@ public OpenTelemetryRumBuilder addLogRecordExporterCustomizer(
...
+        if (ratio < 0.0 || ratio > 1.0) {
+            throw new IllegalArgumentException("ratio must be in range [0.0, 1.0]");
```
marandaneto, line 317: should we log out and default to 0.0 instead? I usually don't like to throw exceptions because maybe a simple mistake and a wrong release might crashpeople'ss apps, and a new update alwaytakeske days

bidetofevil:
Given that #979 solves this use case and others, I think I prefer that to this. If y'all decide go narrow the scope, this is OK for me too.

core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
breedx-splk: Yeah that's reasonable. I think I agree. 👍🏻

core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java
```diff
@@ -360,6 +381,30 @@ public OpenTelemetryRum build() {
         return delegate.build();
     }
 
+    private void configureSessionSampling() {
+        if ((sessionIdSamplingRatio < 0) || (sessionManager == null)) {
```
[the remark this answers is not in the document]
breedx-splk: we can't skip it, because the default is negative to indicate "no sampling".

bidetofevil (approved):
LGTM. It adds a bit of surface area to the API but it's kind of necessary if we are offering sessionId based sampling as an option.

breedx-splk:
Closing for now, in favor of #979 -- which solves the sampler problem by allowing the user to inject their own session provider....which means they can create their own instance of the existing `SessionIdRatioBasedSampler`.

PR #976, opened 2025-05-12
Agent default instrumentation config

LikeTheSalad:
Moving existing configs to the agent api


android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
```diff
@@ -70,4 +145,8 @@ object OpenTelemetryRumInitializer {
                     .setHeaders(metricEndpointConnectivity::getHeaders)
                     .build()
             }.build()
+    }
+
+    private inline fun <reified T : AndroidInstrumentation> getInstrumentation(): T =
+        AndroidInstrumentationLoader.getInstrumentation(T::class.java)!!
```
marandaneto, line 151:
```suggestion
    private inline fun <reified T : AndroidInstrumentation> getInstrumentation(): T? =
        AndroidInstrumentationLoader.getInstrumentation(T::class.java)
```

this would avoid possible crashes, apparently we already check if the return is not nullable for a few of them, lets do for all of them

breedx-splk (approved):
LGTM

PR #1002, opened 2025-06-09
Capture click events for compose

### LINKED ISSUE #986: Capture click events for compose views
This is a follow-on to #930

cleverchuk:
fixes #986 

This PR takes inspiration from [here](https://github.com/DataDog/dd-sdk-android/blob/develop/integrations/dd-sdk-android-compose/src/main/kotlin/com/datadog/android/compose/internal/ComposeActionTrackingStrategy.kt) and [here](https://github.com/embrace-io/embrace-android-sdk/blob/main/embrace-android-compose/src/main/java/io/embrace/android/embracesdk/compose/internal/EmbraceNodeIterator.kt) to realize the feature.


breedx-splk:
Thanks, looking forward to reviewing!

One initial thought -- the new module is `instrumentation/compose` and only contains the click instrumentation. Compose is asked about a lot...do you foresee adding more instrumentation to the same module, or do we think it might be better broken out, such as `instrumentation/compose/click` and `instrumentation/compose/navigation` etc.?  This isn't a blocking question, just curious what people think.

bidetofevil:
I like @breedx-splk's suggestion. Compose is an entire framework, and there could be instrumentation for many aspects of it. Having them live in separate modules seem like the right thing to do, as you might not want all of them.

gradle/libs.versions.toml
```diff
@@ -14,6 +14,8 @@ androidPlugin = "8.10.1"
 junitKtx = "1.2.1"
 autoService = "1.1.1"
 androidx-navigation = "2.7.7"
+#noinspection GradleDependency
+compose = "1.3.0"
```
bidetofevil: What's the reason you chose `1.3.0`? Is it that you can only support versions at runtime that is greater than or equal to this version?

instrumentation/compose/src/main/kotlin/io/opentelemetry/instrumentation/compose/ComposeClickInstrumentation.kt
```diff
@@ -0,0 +1,29 @@
...
+    override val name: String = "compose"
+
+    override fun install(ctx: InstallationContext) {
+        ctx.application.registerActivityLifecycleCallbacks(
+            ComposeClickActivityCallback(
+                ComposeClickEventGenerator(
+                    ctx.openTelemetry
+                        .logsBridge
+                        .loggerBuilder("io.opentelemetry.android.instrumentation.compose")
+                        .build() as ExtendedLogger,
```
bidetofevil: Will this work with disk buffering, @LikeTheSalad ?

instrumentation/compose/src/main/kotlin/io/opentelemetry/instrumentation/compose/WindowCallbackWrapper.kt
```diff
@@ -0,0 +1,34 @@
...
+package io.opentelemetry.instrumentation.compose
+
+import android.os.Build.VERSION_CODES
+import android.view.ActionMode
+import android.view.MotionEvent
+import android.view.SearchEvent
+import android.view.Window.Callback
+import androidx.annotation.RequiresApi
+
+internal class WindowCallbackWrapper(
```
bidetofevil: Will this work with versions less than M if two of these methods are not callable? If the behavior in L is unpredictable as a result, I would just make the whole class require M and simply not add the instrumentation if it's not at least M

instrumentation/compose/src/main/kotlin/io/opentelemetry/instrumentation/compose/internal/ComposeLayoutNodeUtil.kt
```diff
@@ -0,0 +1,43 @@
...
+
+const val APP_SCREEN_CLICK_EVENT_NAME = "app.screen.click"
+const val VIEW_CLICK_EVENT_NAME = "event.app.widget.click"
+val viewNameAttr: AttributeKey<String> = stringKey("app.widget.name")
+
+val xCoordinateAttr: AttributeKey<Long> = longKey("app.screen.coordinate.x")
+val yCoordinateAttr: AttributeKey<Long> = longKey("app.screen.coordinate.y")
+val viewIdAttr: AttributeKey<Long> = longKey("app.widget.id")
+
+class ComposeLayoutNodeUtil {
```
bidetofevil: Why not make the whole class internal so you don't have a public object with inaccessible methods outside the module?

gradle/libs.versions.toml
cleverchuk: > What's the reason you chose 1.3.0?

`1.3.0` because that's when the `layoutDelegate` field was added to `LayoutNode` class.

> Is it that you can only support versions at runtime that is greater than or equal to this version?

yes

instrumentation/compose/src/main/kotlin/io/opentelemetry/instrumentation/compose/WindowCallbackWrapper.kt
cleverchuk: this is similar to view-click one. IIRC animal-sniffer gave me a lot of grief when i made it class only and i think the tests didn't run correctly either with what you're suggesting.

instrumentation/compose/src/main/kotlin/io/opentelemetry/instrumentation/compose/internal/ComposeLayoutNodeUtil.kt
cleverchuk: it'll cause a compile-time error when put on the class because it's being accessed [here](https://github.com/open-telemetry/opentelemetry-android/pull/1002/files#diff-16cad923fd92359921ca6f2f9a8009d193505f8b7a8d56a2faafd2262479e686R35) and [here](https://github.com/open-telemetry/opentelemetry-android/pull/1002/files#diff-f50854a1fce92f2a9ac6c480dd191e4d5d612188f9f87210cf8afd364b4720dbR60).

instrumentation/compose/src/main/kotlin/io/opentelemetry/instrumentation/compose/ComposeClickInstrumentation.kt
LikeTheSalad: This is a good point. I'm not sure it would cause issues, but in general I don't think we should rely on extended types unless we have no choice. Can we change this to use plain [Logger](https://github.com/open-telemetry/opentelemetry-java/blob/4a33b6699603ca667425820f92e2f206737ed0c1/api/all/src/main/java/io/opentelemetry/api/logs/Logger.java) instead, @cleverchuk ?

breedx-splk: Unfortunately, the event name stuff is all incubating, which means that any time we want to create events we have to do the cast until it's done incubating. See the network change instrumentation for an existing precedent.

instrumentation/compose/src/main/kotlin/io/opentelemetry/instrumentation/compose/ComposeClickActivityCallback.kt
```diff
@@ -0,0 +1,23 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.compose
+
+import android.app.Activity
+import io.opentelemetry.android.internal.services.visiblescreen.activities.DefaultingActivityLifecycleCallbacks
+
+internal class ComposeClickActivityCallback(
```
breedx-splk: Would be good to have a unit test for this new class.

instrumentation/compose/src/main/kotlin/io/opentelemetry/instrumentation/compose/ComposeClickEventGenerator.kt
```diff
@@ -0,0 +1,220 @@
...
+        x: Float,
+        y: Float,
+    ): Boolean {
+        val bounded =
+            composeLayoutNodeUtil.getLayoutNodeBoundsInWindow(node)?.let { bounds ->
+                x >= bounds.left && x <= bounds.right && y >= bounds.top && y <= bounds.bottom
+            } == true
+
+        return bounded && isValidClickTarget(node)
+    }
```
breedx-splk: Wow, kinda surprised at how much work it was to get the target! Thanks for taking that on. I think this class has gotten a little long and has multiple responsibilities. Would it be possible to factor out the target finding capabilities into a separate collaborator class?

instrumentation/compose/src/main/kotlin/io/opentelemetry/instrumentation/compose/ComposeClickEventGenerator.kt
```diff
@@ -0,0 +1,220 @@
...
+                val className = modifier::class.qualifiedName
+                isClickable = (
+                    className == CLASS_NAME_CLICKABLE_ELEMENT ||
+                        className == CLASS_NAME_COMBINED_CLICKABLE_ELEMENT ||
+                        className == CLASS_NAME_TOGGLEABLE_ELEMENT
+                )
+            }
+        }
+
+        return isClickable
```
breedx-splk: We're reassigning `isClickable` in the loop, so it looks like it's possible for one pass thru the `for` loop to set `isClickable = true` and then the next pass the modifier might set it back to `false`. Is that right? Should we return `true` from the method as soon as we find any modifier that makes it true?

instrumentation/compose/src/main/kotlin/io/opentelemetry/instrumentation/compose/internal/ComposeLayoutNodeUtil.kt
```diff
@@ -0,0 +1,43 @@
...
+import io.opentelemetry.api.common.AttributeKey.longKey
+import io.opentelemetry.api.common.AttributeKey.stringKey
+
+const val APP_SCREEN_CLICK_EVENT_NAME = "app.screen.click"
+const val VIEW_CLICK_EVENT_NAME = "event.app.widget.click"
+val viewNameAttr: AttributeKey<String> = stringKey("app.widget.name")
+
+val xCoordinateAttr: AttributeKey<Long> = longKey("app.screen.coordinate.x")
+val yCoordinateAttr: AttributeKey<Long> = longKey("app.screen.coordinate.y")
+val viewIdAttr: AttributeKey<Long> = longKey("app.widget.id")
```
breedx-splk: These attributes are now available as constants from the semantic conventions, and so we should use those instead. See #1008 as an example of how to do that.

instrumentation/compose/src/test/kotlin/io/opentelemetry/instrumentation/compose/ComposeClickEventGeneratorTest.kt
```diff
@@ -0,0 +1,303 @@
...
+            targetX = motionEvent.x,
+            targetY = motionEvent.y,
+            hitIndexes = listOf(3, 4),
+            clickableIndexes = listOf(2, 3, 4),
+        )
+
+        composeClickEventGenerator.generateClick(motionEvent)
+
+        val events = openTelemetryRule.logRecords
+        Assertions.assertThat(events).hasSize(2)
```
breedx-splk: static import please

instrumentation/compose/src/test/kotlin/io/opentelemetry/instrumentation/compose/ComposeClickEventGeneratorTest.kt
```diff
@@ -0,0 +1,303 @@
...
+            targetY = motionEvent.y,
+            hitIndexes = listOf(3, 4),
+            clickableIndexes = listOf(2, 3, 4),
+            describableIndexes = listOf(3, 4),
+        )
+
+        composeClickEventGenerator.generateClick(motionEvent)
+
+        val events = openTelemetryRule.logRecords
+        Assertions.assertThat(events).hasSize(2)
```
breedx-splk:
```suggestion
       assertThat(events).hasSize(2)
```

instrumentation/compose/src/main/kotlin/io/opentelemetry/instrumentation/compose/ComposeClickActivityCallback.kt
cleverchuk: it's mainly delegation so i didn't think it was necessary. however, i can add one if you feel strongly about it.

instrumentation/compose/src/main/kotlin/io/opentelemetry/instrumentation/compose/ComposeClickEventGenerator.kt
cleverchuk: okidoki

instrumentation/compose/src/main/kotlin/io/opentelemetry/instrumentation/compose/WindowCallbackWrapper.kt
bidetofevil: Tests that use the AndroidX or Roboletric runner will by default use the minApiVersion. You can use the `@Config` annotation to specify a working Android version and then test that a non-working one wouldn't have this added. Have a look at `AgentInitTest` among others.

instrumentation/compose/src/main/kotlin/io/opentelemetry/instrumentation/compose/internal/ComposeLayoutNodeUtil.kt
bidetofevil: Right. It's because `ComposeClickEventGenerator` is public and this is part of the interface. Can that be internal as well? What you have here is a concept that is partially leaking publicly, but not really accessible which isn't great. 

I think you have a few options:

- Make everything public
- Make everything internal
- Make the Util class completely private/internal and don't expose it in the interface of `ComposeClickEventGenerator`

Any of these would work

instrumentation/compose/src/main/kotlin/io/opentelemetry/instrumentation/compose/ComposeClickActivityCallback.kt
bidetofevil: A test would be good in case it is modified in the future and the delegation gets screwed up or becomes incomplete

instrumentation/compose/src/main/kotlin/io/opentelemetry/instrumentation/compose/WindowCallbackWrapper.kt
cleverchuk: what are you suggesting, change the min api to M?

instrumentation/compose/src/main/kotlin/io/opentelemetry/instrumentation/compose/ComposeClickActivityCallback.kt
cleverchuk: roger

breedx-splk (approved):
Thanks again! 🚀

gradle/libs.versions.toml
LikeTheSalad: I think there's no need to pin it to this version in the project, since we add it as `compileOnly` and the README explains that the minimum version required is `1.3.0`. So I'd remove the `#noinspection GradleDependency` and set it to the latest version available.

cleverchuk:
I'm gonna close the PR and reopen it later because it looks like the compiler suppression doesn't work from 1.6.0+.

cleverchuk:
okay, we can only claim to support 1.3.0 - 1.5.4. i'll work on figuring out how to support the rest. also moved to new package because it seemed like everyone liked the idea.

instrumentation/compose/src/main/kotlin/io/opentelemetry/instrumentation/compose/WindowCallbackWrapper.kt
bidetofevil: I'm suggesting changing the tests to run as if it were M, and only including this instrumentation if it were M+. 

No worries about this. I'll see if I can refactor and throw a test up quickly today since this was merged already

PR #1008, opened 2025-06-12
Use semantic conventions in click instrumentation

breedx-splk:
We don't want to duplicate these constants now that they are available in the incubating semconv.


bidetofevil (approved):
LGTM

PR #1011, opened 2025-06-12
Return the less specific interface

breedx-splk:
Small internal refactoring to use `SessionProvider` (smaller, broader interface) instead of `SessionManager`.


LikeTheSalad (approved):
Cheers!

PR #1037, opened 2025-07-01
Aligns the link checking with the upstream instrumentation repo

### LINKED ISSUE #1033: Make markdown link check consistent between projects
There's an inconsistency problem with how markdown links are checked between the java-related projects. Here's a description of the current state of affairs:

* android - [executes tool with custom script and json config](https://github.com/open-telemetry/opentelemetry-android/blob/main/.github/scripts/markdown-link-check-with-retry.sh#L9)
* java core - [uses lychee-action](https://github.com/open-telemetry/opentelemetry-java/blob/main/.github/workflows/reusable-markdown-link-check.yml#L15) with args for config 
* java contrib - [uses lychee-action](https://github.com/breedx-splk/opentelemetry-java-contrib/blob/main/.github/workflows/reusable-markdown-link-check.yml#L15) with args for config
* java instrumentation - [uses lychee-action with a `.toml` config file](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/.github/workflows/reusable-markdown-link-check.yml#L18)

We should figure out a single, more consistent approach and use it across these projects. I like the `.toml` approach and suggest we standardize on that.

breedx-splk:
Resolves #1033.

This also makes it considerably simpler to run the link checker locally, on your laptop (just run the `.github/scripts/link-check.sh` script, assuming you have docker).


.github/workflows/build.yaml
```diff
@@ -30,6 +30,7 @@ jobs:
         run: ./gradlew check assemble
       - name: publish snapshot
         run: ./gradlew publishToSonatype
+        run: ./gradlew publishMavenPublicationToSonatypeRepository closeSonatypeStagingRepository releaseSonatypeStagingRepository
```
LikeTheSalad: I think the line above does the same as `publishMavenPublicationToSonatypeRepository`, so this might cause that we upload the artifact twice. Also, for the other two tasks (closeSonatypeStagingRepository releaseSonatypeStagingRepository), [the plugin](https://github.com/gradle-nexus/publish-plugin) has the `closeAndReleaseSonatypeStagingRepository` one for when it's needed to run them both. I'm not sure if closing and releasing a staging repo is necessary when publishing a SNAPSHOT version though, as for the "legacy maven central" it wasn't the case.

breedx-splk: Oof, I mean, this line wasn't even supposed to be part of this PR, so good catch. Let me remove it!

breedx-splk:
```suggestion
```

Issue #76, opened 2023-09-13
Alter FilteringSpanExporter to leverage common code from contrib


breedx-splk:
[stub] We think that InterceptableSpanExporter from contrib is well suited to be used in the FilteringSpanExporter and builder. Same for event filtering, which doesn't exist yet.

RiyazurRazak:
@breedx-splk, I'm interested in working on this task if it's still available.

breedx-splk:
Thanks @RiyazurRazak , I'll assign you. I'm not sure if it'll impact this, but we expect a contrib release in the next day or two. Cheers!

PR #1043, opened 2025-07-04
Alter FilteringSpanExporter to leverage common code from contrib

RiyazurRazak:
Resolves https://github.com/open-telemetry/opentelemetry-android/issues/76

- Integrates processors from the Java Contrib module
- Refactors FilteringSpanExporterBuilder to utilize InterceptableSpanExporter provided by the contrib library

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/export/FilteringSpanExporterBuilder.java
+++ b/core/src/main/java/io/opentelemetry/android/export/FilteringSpanExporterBuilder.java
@@ -5,8 +5,11 @@
 
 package io.opentelemetry.android.export;
 
+import androidx.annotation.Nullable;
 import io.opentelemetry.api.common.AttributeKey;
 import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.contrib.interceptor.InterceptableSpanExporter;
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/opentelemetry/android/export/FilteringSpanExporterBuilder.java
```diff
@@ -16,6 +19,19 @@ public final class FilteringSpanExporterBuilder {
...
+            new Interceptor<SpanData>() {
+
+                @Nullable
+                @Override
+                public SpanData intercept(SpanData item) {
+                    if (predicate.test(item)) {
+                        return null;
+                    } else {
+                        return item;
+                    }
```
breedx-splk: Prefer to remove the redundant `else`.
```suggestion
                    if (predicate.test(item)) {
                        return null;
                    } 
                    return item;                   
```

breedx-splk (approved):
Thanks for the contribution! I think the `FilteringSpanExporter` has so little content in it now, that I would probably just move the `FilteringSpanExporterBuilder` content inside of it, but that could always happen in a followup PR.

LikeTheSalad (approved):
Thank you 🙏

PR #1046, opened 2025-07-07
Fix several build warnings around deprecations

breedx-splk:


build.gradle.kts
```diff
@@ -18,6 +19,7 @@ plugins {
 }
 
 extra["java_version"] = JavaVersion.VERSION_1_8
+extra["jvm_target"] = JvmTarget.JVM_1_8
```
LikeTheSalad, line 22: Can we reuse the `java_version` value?

breedx-splk: Sadly no, the types don't match.

PR #1048, opened 2025-07-07
Prepare changelog for v0.12.0 release

breedx-splk:


CHANGELOG.md
```diff
@@ -2,6 +2,39 @@
 
 ## Unreleased
```
bidetofevil, line 3:
```suggestion
## Version 0.12.0 (2025-07-08)
```

bidetofevil: This PR should update the version as per the suggestion or will that come in a different PR post release?

breedx-splk: Appreciate the thoroughness. The build automation for the release will create a follow-up PR that updates the version, and I think if we set it here it actually BREAKS the release. 😁  👍🏻

PR #1055, opened 2025-07-08
Cherry pick from release #1054

breedx-splk:
See the description and screenshot in #1054.

These modules fail to publish, due to missing platform dependency. This adds the platform dep and should fix this going forward. I think we should have a better way of doing this validation, before encountering it at release time... 🤔


bidetofevil (approved):
LGTM

PR #1075, opened 2025-07-17
Include the upstream components in our android bom

### LINKED ISSUE #1074: Include OpenTelemetry Java Instrumentation and OpenTelemetry Java Core in BOM
It would be great to include the BOMs for OpenTelemetry Java Instrumentation and OpenTelemetry Java Core in the BOM for this project, given that each release is built based on specific versions of these projects.

E.g. for https://github.com/open-telemetry/opentelemetry-android/releases/tag/v0.12.0
> This release is based on OpenTelemetry Java Instrumentation 2.17.0 and the OpenTelemetry Java Core (sdk/api/exporters) 1.51.0.

Sidenote, it seems no BOM was published for v0.12

breedx-splk:
Resolves #1074 (see discussion there).

I tested this by publishing locally, and it creates a new block in our bom pom:

```xml
<dependencyManagement>
    <dependencies>
      <dependency>
        <groupId>io.opentelemetry.instrumentation</groupId>
        <artifactId>opentelemetry-instrumentation-bom-alpha</artifactId>
        <version>2.17.1-alpha</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
      <dependency>
        <groupId>io.opentelemetry.instrumentation</groupId>
        <artifactId>opentelemetry-instrumentation-bom</artifactId>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
      <dependency>
        <groupId>io.opentelemetry</groupId>
        <artifactId>opentelemetry-bom</artifactId>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
    </dependencies>
  </dependencyManagement>
```

Note that explicit version information is not included with the `opentelemetry-instrumenation-bom` or the `opentelemetry-bom` dependencies, because those versions are resolved transiently from the topmost instrumentation-bom-alpha dependency...but I think it still makes sense to have them here, with `<scope>import</scope>`.

### DIFF
--- a/opentelemetry-android-bom/build.gradle.kts
+++ b/opentelemetry-android-bom/build.gradle.kts
@@ -11,10 +11,12 @@ dependencies {
     api(platform(libs.opentelemetry.platform.alpha))
     api(platform(libs.opentelemetry.platform))
     api(platform(libs.opentelemetry.core.bom))
-    rootProject.subprojects.forEach { subproject ->
-        if (subproject != project) {
-            subproject.plugins.withId("maven-publish") {
-                api(subproject)
+    constraints {
+        rootProject.subprojects.forEach { subproject ->
+            if (subproject != project) {
+                subproject.plugins.withId("maven-publish") {
+                    api(subproject)
+                }
             }
         }
     }


opentelemetry-android-bom/build.gradle.kts
```diff
@@ -5,13 +5,16 @@ plugins {
...
+    api(platform(libs.opentelemetry.platform.alpha))
+    api(platform(libs.opentelemetry.platform))
+    api(platform(libs.opentelemetry.core.bom))
+    rootProject.subprojects.forEach { subproject ->
+        if (subproject != project) {
+            subproject.plugins.withId("maven-publish") {
+                api(subproject)
             }
         }
     }
```
LikeTheSalad: I'm curious if we can keep this block inside `constraints {}` and only leave the platform dependencies out of it? If it still works, it should be preferable based on what they [explain here](https://docs.gradle.org/current/userguide/java_platform_plugin.html#sec:java_platform_separation) about using dependencies vs constraints.

breedx-splk: I tried a few permutations of this, but yeah, I agree with this. Lemme make that change. Appreciate it!

LikeTheSalad (approved):

PR #1076, opened 2025-07-17
DO NOT MERGE -- Test build cache performance

breedx-splk:
I'm adding this to see how it may impact build performance.


breedx-splk:
Well, it looks to be 3 minutes faster than the other PR I was comparing it to, so maybe that's just a fluke. Meh.

PR #1083, opened 2025-07-22
Only run the assembleRelease target to verify the demo app compiles

bidetofevil:
Reduce the scope of verification of the demo-app by only assembling the release version and bypass debug, lint, and other checks. Doing debug will be even faster but having it go through release and the APK optimizations (including desugaring) is a more comprehensive verification.

### DIFF
--- a/.github/workflows/pr-check.yaml
+++ b/.github/workflows/pr-check.yaml
@@ -22,7 +22,7 @@ jobs:
         run: ./gradlew check
       - name: build demo app
         working-directory: ./demo-app
-        run: ./gradlew check assemble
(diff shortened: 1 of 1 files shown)


LikeTheSalad (approved):

LikeTheSalad:
> Given how slow the build is, I'm ok with this, but agree that it's not as thorough. I will wait for @LikeTheSalad to chime in before merging. Thanks!!

I think this is fine for the demo app since we mostly just want to avoid breaking its compilation with the agent changes, and usually the most _breakable_ variant is the release one with the R8 stuff that goes on in there 👍

PR #1084, opened 2025-07-22
Improve perf of current network tests

bidetofevil:
This test was running a lot of redundant API versions, which all has a fixed cost to spin up for the first time it was also unnecessarily reinitializing a lot of mocks that aren't inspected.

This PR curates a set of API versions that are relevant (i.e. code contains checks that differentiate behavior), as well as clean up some of the redundancies that are impactful.

### DIFF
--- a/services/src/test/java/io/opentelemetry/android/internal/services/network/CurrentNetworkProviderTest.kt
+++ b/services/src/test/java/io/opentelemetry/android/internal/services/network/CurrentNetworkProviderTest.kt
@@ -10,6 +10,7 @@ import android.net.ConnectivityManager.NetworkCallback
 import android.net.Network
 import android.net.NetworkRequest
 import android.os.Build
+import androidx.test.ext.junit.runners.AndroidJUnit4
 import io.mockk.every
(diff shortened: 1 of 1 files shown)


breedx-splk (approved):
LGTM thanks!

LikeTheSalad (approved):
Thank you!

services/src/test/java/io/opentelemetry/android/internal/services/network/CurrentNetworkProviderTest.kt
```diff
@@ -39,10 +36,15 @@ internal class CurrentNetworkProviderTest {
             .subType("LTE")
             .build()
     private val noNetwork = CurrentNetwork.builder(NetworkState.NO_NETWORK_AVAILABLE).build()
+    private val mockConnectivityManager = mockk<ConnectivityManager>()
```
fractalwrench, line 39: I'm reasonably sure this mock is taking a long time to construct, but I'll investigate that in a separate changeset later this week.

fractalwrench (approved):

PR #1085, opened 2025-07-22
Update docs to reflect that desugaring is required for minSdk < 26

bidetofevil:
Desugaring is required universally for minSdk < 26 now because of the use of the Java 9 `Clock` implementation in OTel `1.50`. Updated the documentation to reflect that, although many places already reference 26 because of a Java SDK requirement that we never quite ran into.


breedx-splk (approved):
Thanks. I think this is also worth adding in the CHANGELOG, which we do at release time, but if it was in here we'd be sure not to miss it.

PR #1087, opened 2025-07-23
Include `service.version` resource attribute

breedx-splk:
The `AndroidResource` had a glaring omission in the form of missing the service version (app version). This prevents RU&M users from knowing which version of the app a user was using! :) Seems important to me.

This also migrates the `AndroidResource` java class to kotlin (object).

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/AndroidResource.kt
+++ b/core/src/main/java/io/opentelemetry/android/AndroidResource.kt
@@ -3,64 +3,68 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package io.opentelemetry.android;
+package io.opentelemetry.android
 
-import static io.opentelemetry.android.common.RumConstants.RUM_SDK_VERSION;
-import static io.opentelemetry.semconv.ServiceAttributes.SERVICE_NAME;
-import static io.opentelemetry.semconv.incubating.DeviceIncubatingAttributes.DEVICE_MANUFACTURER;
-import static io.opentelemetry.semconv.incubating.DeviceIncubatingAttributes.DEVICE_MODEL_IDENTIFIER;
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/opentelemetry/android/AndroidResource.kt
```diff
@@ -0,0 +1,77 @@
...
+private const val DEFAULT_APP_NAME = "unknown_service:android"
+
+object AndroidResource {
+    @JvmStatic
+    fun createDefault(application: Application): Resource {
+        val appName = readAppName(application)
+        val resourceBuilder =
+            Resource.getDefault().toBuilder().put(SERVICE_NAME, appName)
+        val appVersion = readAppVersion(application)
+        appVersion?.let { resourceBuilder.put(SERVICE_VERSION, it) }
```
breedx-splk, line 31: This is the only _real_ change. The rest shows new because of the java -> kotlin migration.

LikeTheSalad (approved):

bidetofevil (approved):
LGTM. But a warning that the API used to get the version is a binder call and can potentially be slow. We don't do this in the Embrace SDK for that reason, and ask folks to set it manually before, though they may be doing the same thing anyway, so for an SDK that always requires this, it it's probably the right thing to do.

The "binder call free" way of doing this involves a Gradle plugin, so it's probably out of the scope of ths.

breedx-splk:
> LGTM. But a warning that the API used to get the version is a binder call and can potentially be slow. 

Oh interesting, thanks for pointing that out. So this would potentially, but not necessarily, result in a strict mode violation? Do you have any references for this being binder/blocking/slow? I couldn't find anything in the 60 seconds I looked...

> The "binder call free" way of doing this involves a Gradle plugin, so it's probably out of the scope of ths.

Fair enough. In the interest of iterating, I'm going to merge and we can improve later if needed. Thanks!

PR #1092, opened 2025-07-24
Use setup-gradle GitHub action

fractalwrench:
## Goal

The [setup-gradle Github Action](https://github.com/gradle/actions/blob/main/docs/setup-gradle.md) offers better build performance on CI. It caches build state such as gradle JARs and output from build tasks between jobs which therefore avoids starting the build from scratch each time. I didn't spot any alternative solution that was doing this already.

If this approach is accepted I can raise a separate changeset to do this for the other CI workflows.

## Testing

Relying on the CI job. I don't have permissions to retry the job to test the cache, which may be worth doing to get an idea of the speed improvements.

### DIFF
--- a/.github/workflows/pr-check.yaml
+++ b/.github/workflows/pr-check.yaml
@@ -16,6 +16,8 @@ jobs:
         with:
           distribution: temurin
           java-version: 17
+      - name: Setup Gradle build action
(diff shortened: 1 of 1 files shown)


LikeTheSalad (approved):
Thank you!

breedx-splk (approved):
I'm excited about this -- we use it in the instrumentation builds as well. 🤞🏻

PR #1099, opened 2025-07-24
Fix v0.13.0 release

breedx-splk:
So the versions of these two are determined by gradle transitively from the top `opentelemetry-instrumentation-bom-alpha`. Their inclusion here causes the sonatype release process to fail validation (because they don't have explicit versions)....so let's just remove them. They shouldn't be necessary anyway.

And much like last time, I'll cherry pick this into main after the release.


bidetofevil (approved):
LGTM. If bomA includes bomB and we take in bomA, we probably shouldn't include bomB anyway

PR #1101, opened 2025-07-25
Updated ANR data model from span to log event

tonzhan2:
The ANR Instrumentation was sending out non zero length spans, despite claiming to send out zero length spans. This was due to the `start` and `end` calls by the instrumenter happening at slightly different times, causing a short duration.

This PR eliminates that and refactors the ANR Instrumentation to emit a log event in line with crashes, which makes more sense

Testing:
All builds, tests and checks
Verified log event output with demo app


instrumentation/anr/src/main/java/io/opentelemetry/android/instrumentation/anr/AnrWatcher.kt
```diff
@@ -51,14 +67,36 @@ internal class AnrWatcher(
...
+                .put(THREAD_NAME, mainThread.name)
+                .put(EXCEPTION_STACKTRACE, stackTraceToString(stackTrace))
+
+        for (extractor in additionalExtractors) {
+            extractor.onStart(attributesBuilder, Context.current(), stackTrace)
+        }
+
+        val eventBuilder = anrLogger.logRecordBuilder() as ExtendedLogRecordBuilder
+        eventBuilder
+            .setEventName("device.anr")
```
breedx-splk, line 90: We should explicitly call this out as a name change -- even if there isn't yet semconv around this. Specifically, the previous span name was called `ANR` and the new event name will be `device.anr`. It doesn't have to be in this PR, but I think we should follow this up with an effort to land a semantic convention for this. As part of that, I would also make the case that this probably belongs in the `app.` namespace, because it literally unpacks to "application not responding", which is more accurate than `device.`. Again, doesn't need to happen here. I just wanted to make sure this was noted.

breedx-splk (approved):
Thanks thanks!

PR #1102, opened 2025-07-25
Allow empty global attributes from Supplier at startup

### LINKED ISSUE #1094: globalAttributeSupplier doesn't get installed as expected
If a global attribute supplier is provided, but returns an empty list, the supplier does not get installed. This makes it difficult to dynamically add attributes later. As it stands, a dummy attribute has to be returned just to get the supplier installed which is just a dumb gross hack. The easiest way to fix this is to just remove the nonsensical isEmpty check and change the default to null.

https://github.com/open-telemetry/opentelemetry-android/blob/541754e3d8482bb2b35bf564b2c0eb7484791cbd/core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.java#L482-L489

https://github.com/open-telemetry/opentelemetry-android/blob/541754e3d8482bb2b35bf564b2c0eb7484791cbd/core/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java#L43-L46

https://github.com/open-telemetry/opentelemetry-android/blob/541754e3d8482bb2b35bf564b2c0eb7484791cbd/core/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java#L22

breedx-splk:
Resolves #1094.

We need to allow the case where a user sets up a `Supplier<Attributes>` that returns empty during build time. This allows the `GlobalAttributesAppender` to be created with a `Supplier` that might later return attribute data. Without this, the wiring of the `GlobalAttributesAppender` is simply skipped.

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
+++ b/core/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
@@ -5,6 +5,9 @@
 
 package io.opentelemetry.android.config;
 
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
 import io.opentelemetry.android.ScreenAttributesSpanProcessor;
 import io.opentelemetry.android.features.diskbuffering.DiskBufferingConfig;
 import io.opentelemetry.android.internal.services.network.CurrentNetworkProvider;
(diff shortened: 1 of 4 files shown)


LikeTheSalad (approved):
Cheers!

PR #1104, opened 2025-07-28
Add event attribute extractor

breedx-splk:
So the design of the `AttributesExtractor` class is tightly tied in with how the instrumentation api is designed, which is focused squarely on tracing and making spans. As a result, the `AttributesExtractor` has `onStart()` and `onEnd()` methods which don't really line up with point-in-time eventing. This leads to clumsy implementations that have a `Void` generic type and an empty/useless `onEnd()`. The `onStart()` api itself is also kinda gross, because it expects/requires implementations to just cause side effects on the `AttributesBuilder` that is passed to it (presumably for optimization/performance reasons?).

This introduces a new interface called `EventAttributesExtractor` that takes the parent context and a "subject" and produces `Attributes`. Callers/users of the `EventAttributesExtractor` can then take the resulting attributes and add those to the event being emitted.

### DIFF
--- a/instrumentation/crash/src/main/java/io/opentelemetry/android/instrumentation/crash/RuntimeDetailsExtractor.java
+++ b/instrumentation/crash/src/main/java/io/opentelemetry/android/instrumentation/crash/RuntimeDetailsExtractor.java
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package io.opentelemetry.android.common;
+package io.opentelemetry.android.instrumentation.crash;
 
 import static io.opentelemetry.android.common.RumConstants.BATTERY_PERCENT_KEY;
 import static io.opentelemetry.android.common.RumConstants.HEAP_FREE_KEY;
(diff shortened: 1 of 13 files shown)


instrumentation/crash/src/main/java/io/opentelemetry/android/instrumentation/crash/RuntimeDetailsExtractor.java
```diff
@@ -15,13 +15,14 @@
...
 import android.os.BatteryManager;
 import androidx.annotation.Nullable;
+import io.opentelemetry.android.instrumentation.common.EventAttributesExtractor;
+import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.api.common.AttributesBuilder;
-import io.opentelemetry.instrumentation.api.instrumenter.AttributesExtractor;
 import java.io.File;
 
 /** Represents details about the runtime environment at a time */
 public final class RuntimeDetailsExtractor<RQ, RS> extends BroadcastReceiver
```
breedx-splk, line 24: The only remaining usage of the `RuntimeDeatilsExtractor` was in the crash instrumentation, so I'm not sure why it was still hanging out in common? I moved it into the instrumentation module.

PR #1105, opened 2025-07-28
Migrate two classes and 1 test to kotlin

breedx-splk:
Just a pretty straightforward port.

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/SessionIdRatioBasedSampler.kt
+++ b/core/src/main/java/io/opentelemetry/android/SessionIdRatioBasedSampler.kt
@@ -32,7 +32,7 @@ class SessionIdRatioBasedSampler(
         name: String,
         spanKind: SpanKind,
         attributes: Attributes,
-        parentLinks: MutableList<LinkData?>,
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/opentelemetry/android/SessionIdRatioBasedSampler.kt
```diff
@@ -0,0 +1,53 @@
...
+    // SessionId uses the same format as TraceId, so we can reuse trace ID ratio sampler.
+    private val ratioBasedSampler: Sampler = Sampler.traceIdRatioBased(ratio)
+
+    override fun shouldSample(
+        parentContext: Context,
+        traceId: String,
+        name: String,
+        spanKind: SpanKind,
+        attributes: Attributes,
+        parentLinks: MutableList<LinkData?>,
```
fractalwrench: I think `List<LinkData>` could be used here rather than a mutable/nullable collection

fractalwrench (approved):
LGTM

LikeTheSalad (approved):

PR #1106, opened 2025-07-29
Specify maxParallelForks for JVM tests

fractalwrench:
## Goal

Configuring [`maxParallelForks` for JVM tests](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.testing.Test.html#org.gradle.api.tasks.testing.Test:maxParallelForks) allows test classes to run in parallel. This can improve test execution times particularly if test cases are distributed across lots of different source files. Generally speaking this shouldn't cause problems unless tests are accessing shared objects/resources, which doesn't seem to be the case in this project.


breedx-splk (approved):
Seems reasonable to me. I couldn't find good docs on how to find the number of cores for your runner, but some resources say it's 2 cores by default -- so this would still use a 2 parallel, which seems good (assuming the default is 1).

Issue #1071, opened 2025-07-15
feature request: shutdown sdk


MustafaHaddara:
I have a span processor that reads from some other resources in my application and there are situations in which that span processor needs to be re-initialized. This requires me to be able to re-initialize the OpenTelemetry SDK and pass in the new instance of SpanProcessor. 

However, the OpenTelemetry Android SDK is intentionally designed to be a singleton, so I cannot just call build() again, and there is no way to "tear down" or "reset" the SDK. I'd like to be able to shutdown an existing instance, make some configuration changes, and then create a new OpenTelemetry instance, without killing the existing process.

breedx-splk:
Thanks @MustafaHaddara for opening this. It's true that the current design intends to initialize the otel RUM SDK as early in the application lifecycle, and have it remain active and instrumenting for the duration of the application's run. As you point out, there is currently no way to shut it down or reinitialize things, and this is by design.

As discussed in SIG, one workaround would be to reinitialize the internals of your SpanProcessor as needed, while leaving the rest of the SDK active. 

We are interested in hearing from users about the importance of such a feature, and any other use cases that might benefit from such a thing. Thanks again!

jzwc:
Let me just note that generally re-initialization (because this, not just shutdown, is the goal, isn't it)  of this kind of SDK is quite complex and dangerous thing. If some instrumentation uses byte manipulation on Android or swizzling on iOS, does the uninstall also restore the original state? And even on easier note: many tracked events are asynchronous - what about a network request caught in the middle of the re-initialization?

Even if the uninstall opts for some "soft" version, the probability that the SDK will end up in some not well defined state is high, and the cost of doing this right would be rather high.

jzwc:
I generally think that most of the "re-initialization" things could be done "soft" by just  programmatically ignoring or filtering the detected signals. It does not remove the instrumentation load from the app, though. As for the case stated in the original post, implementing some additional business logic to the original span processor would do the job (at the cost of a bit higher processing cost, perhaps), wouldn't it?

breedx-splk:
> As for the case stated in the original post, implementing some additional business logic to the original span processor would do the job (at the cost of a bit higher processing cost, perhaps), wouldn't it?

Yes absolutely, and that was the approach we recommended. Create a stateful SpanProcessor that can change its behavior as needed.

PR #1109, opened 2025-07-29
OpenTelemetryRum.shutdown() and instrumentation uninstall

breedx-splk:
Related to #1071.

So this just starts the process of being able to uninstall instrumentation and shut down the sdk. It should not be considered a complete solution, because I only picked up a single instrumentation to see how it would feel to shut down. 

If we like this approach, I can create a tracking issue for the other instrumentations and we can knock those uninstalls out one-at-a-time.

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/OpenTelemetryRum.java
+++ b/core/src/main/java/io/opentelemetry/android/OpenTelemetryRum.java
@@ -15,12 +15,7 @@ import io.opentelemetry.sdk.logs.SdkLoggerProvider;
 import io.opentelemetry.sdk.metrics.SdkMeterProvider;
 import io.opentelemetry.sdk.trace.SdkTracerProvider;
 
-/**
- * Entrypoint for the OpenTelemetry Real User Monitoring library for Android.
- *
- * <p>This class is internal and is hence not for public use. Its APIs are unstable and can change
- * at any time.
- */
+/** Entrypoint for the OpenTelemetry Real User Monitoring library for Android. */
 public interface OpenTelemetryRum {
 
     /**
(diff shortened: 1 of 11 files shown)


core/src/main/java/io/opentelemetry/android/OpenTelemetryRum.java
```diff
@@ -15,12 +15,7 @@
 import io.opentelemetry.sdk.metrics.SdkMeterProvider;
 import io.opentelemetry.sdk.trace.SdkTracerProvider;
 
-/**
- * Entrypoint for the OpenTelemetry Real User Monitoring library for Android.
- *
- * <p>This class is internal and is hence not for public use. Its APIs are unstable and can change
- * at any time.
```
breedx-splk, line 22: I wasn't sure why we claimed this was internal....it's literally the main thing that users want to build. :)

breedx-splk:
The from disk background reader is probably still active and needs to be accounted for. What about "services"?

LikeTheSalad:
> The from disk background reader is probably still active and needs to be accounted for.

Agree, I haven't checked, though if it doesn't require many code changes, maybe we can do it as part of this PR?

> What about "services"?

We should also make them closeable, but I think that change requires its own PR. I can have a look at it next week, unless someone can tackle it before that, which also works for me :)

breedx-splk:
> > The from disk background reader is probably still active and needs to be accounted for.
> 
> Agree, I haven't checked, though if it doesn't require many code changes, maybe we can do it as part of this PR?

Please have another look @LikeTheSalad , I think I have the disk reading code shutting down now...but could use a confirmation. I didn't touch the service part, and agree it should be in a follow-up PR. ❤️

PR #1122, opened 2025-08-01
Bump compileSdk to 36

fractalwrench:
## Goal

Bumps the compileSdk to the latest version. I scanned the [release notes](https://developer.android.com/about/versions/16/features) and don't see anything that would prevent an upgrade. I also ran the example app & checks locally and everything passed.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -22,6 +22,6 @@ org.gradle.configuration-cache.parallel=true
 android.useAndroidX=true
 
 android.minSdk=21
-android.compileSdk=35
+android.compileSdk=36
 version=0.14.0
 group=io.opentelemetry.android
--- a/demo-app/build.gradle.kts
+++ b/demo-app/build.gradle.kts
@@ -13,7 +13,7 @@ localProperties.load(FileInputStream(rootProject.file("local.properties")))
 
 android {
     namespace = "io.opentelemetry.android.demo"
-    compileSdk = 35
+    compileSdk = 36
 
     defaultConfig {
         applicationId = "io.opentelemetry.android.demo"


gradle.properties
```diff
@@ -22,6 +22,6 @@ org.gradle.configuration-cache.parallel=true
 android.useAndroidX=true
 
 android.minSdk=21
-android.compileSdk=35
+android.compileSdk=36
```
LikeTheSalad, line 25: IIRC, we're not closed to bumping up this value, though I think we're not in a hurry either. My understanding is that it can cause consumer apps to get enforced to bump up their values too, which sometimes creates a chained reaction where then the AGP version, and consequently, the Android Studio version have to be upgraded as well, and ultimately, I know this is something that a lot of Android devs try to avoid as much as possible because it can sometimes cause other issues if for example there's a bug in the new AS version or the like.

Because of the above, and also because I'm not aware of a crucial benefit that we can get from bumping up this value as early as possible, I'm more of the idea of waiting until we have no choice, which usually happens when one of our androidx dependencies enforces us to do so (which was the case for version 35). In any case, if you're aware of an important reason why we should bump up this value as soon as possible, I think we can discuss if further.

fractalwrench: The general guidance on Android is to use the [latest `compileSdk` available](https://medium.com/androiddevelopers/picking-your-compilesdkversion-minsdkversion-targetsdkversion-a098a0341ebd). API changes to `compileSdk` are almost always additive & just used for compiling the library. Setting the `minSdk` to a higher version would certainly propagate and would be a trickier decision as it would force library consumers to update.

I'm pretty sure that a newer version of AGP/Gradle is only required for a project that's building an artifact from source. If an end-user is consuming an AAR from mavenCentral, bumping the `compileSdk` should not require any change in their project setup as far as I'm aware

LikeTheSalad: This is an example of the kind of issue I think we should avoid for our users, as much as we can: https://stackoverflow.com/questions/77271961/dependency-androidx-activityactivity1-8-0-requires-libraries-or-apps-that-de 

Usually, after that error message, people are forced to upgrade their compileSdk version value (although, interestingly enough, the most popular answer there seems to be about downgrading the problematic dependency), which can then cause another issue that looks like the following: https://stackoverflow.com/questions/73112894/we-recommend-using-a-newer-android-gradle-plugin-to-use-compilesdk-33 

The ideal choice after seeing that error, is to update their AGP version (although it seems like some people decide to 'fix' it by suppressing the error), but if they upgrade the AGP, then they might get an error like the following: https://stackoverflow.com/questions/79401176/project-is-using-an-incompatible-version-agp-8-8-0-even-though-its-up-to-date

Which means that they'd have to update their Android Studio version. I know for a fact, and also based on the answers on the above posts, that this isn't something devs are usually happy to do, in fact some people might even choose to downgrade libs to avoid that, which is quite dangerous in case the newer versions contain security patches.

> The general guidance on Android is to use the latest compileSdk available.

I think this guidance is expected of every library, so that's fair enough, but my understanding (unless I missed something) is that upgrading the compileSdk only brings some newer apis to become available at compile time, that's neat, though if we don't need those ones, and based on the issues we can bring to host apps by upgrading it, I still can't find a reason why we need to rush it.

fractalwrench: I'm reasonably sure that error message is caused when a library specifies a [`minCompileSdk` in its AAR metadata](https://developer.android.com/build/publish-library/prep-lib-release#set-up-aar-metadata), so wouldn't apply in this case. Perhaps if there's time in the SIG tomorrow we can discuss this

LikeTheSalad: Got it. Based on the description, it looks like so. I won't be able to attend tomorrow's SIG meeting, though, but I still think it should be discussed if possible, because if it ends up being that we won't cause any of the issues I mentioned above by bumping up `compileSdk`, so long as we don't touch `minCompileSdk`, then I don't see any issue with updating `compileSdk`.

bidetofevil: Just changing the `compileSdk` without changing `targetSdk` (which is set at the app level) shouldn't be a big deal. This will prevent apps from getting warnings down the road saying our SDK is compiled using an old version of Android, so it's probably a change we should make relatively soon after a new one drops.

LikeTheSalad: I also had some time today to run some local tests, and it matches what @fractalwrench mentioned about `minCompileSdk`, so since it doesn't seem to cause issues to end users, then I'm fine with making the change. Thanks for your patience, @fractalwrench.

PR #1123, opened 2025-08-01
Setup detekt static analysis plugin

fractalwrench:
## Goal

[Detekt](https://detekt.dev/) is a static analysis program for Kotlin that is pretty commonly used on Android. I've set it up on any module that is using the Android library conventions gradle script as I think it'd be beneficial for catching potential problems and maintaining consistent formatting style in Kotlin code.

Analysis can be triggered via `./gradlew detekt` or `./gradlew check`. I've suppressed all existing warnings in the project by using a baseline file, and any subsequent warnings would fail the build. The baseline can be added to by `./gradlew detektBaseline` which will regenerate the file.

If any of the default rules in the baseline seem annoying or too much of a false positive I'm pretty happy to update the PR, as specific rules can be [configured or even disabled](https://detekt.dev/docs/rules/style). I went with as close to the default as possible, but FWIW here are the modifications we found useful at Embrace: https://github.com/embrace-io/embrace-android-sdk/blob/main/config/detekt/detekt.yml


buildSrc/src/main/kotlin/otel.android-library-conventions.gradle.kts
```diff
@@ -56,6 +57,24 @@ tasks.withType<Test> {
...
+    autoCorrect = true
+
+    // overwrite default behaviour here, if needed
+    config.from(project.files("${project.rootDir}/config/detekt/detekt.yml"))
+    // suppress pre-existing issues on a per-project basis
+    baseline = project.file("${project.projectDir}/config/detekt/baseline.xml")
+}
+
+project.tasks.withType(Detekt::class.java).configureEach {
+    jvmTarget = "1.8"
```
LikeTheSalad: Do you think we could somehow reuse the `javaVersion` val defined earlier in this file?

buildSrc/src/main/kotlin/otel.android-library-conventions.gradle.kts
```diff
@@ -56,6 +57,24 @@ tasks.withType<Test> {
     useJUnitPlatform()
 }
 
+detekt {
+    buildUponDefaultConfig = true
+    autoCorrect = true
+
+    // overwrite default behaviour here, if needed
+    config.from(project.files("${project.rootDir}/config/detekt/detekt.yml"))
```
LikeTheSalad: Please let's use `rootProject.files("config/`... instead.

buildSrc/src/main/kotlin/otel.android-library-conventions.gradle.kts
fractalwrench: Yep - I'll update the changeset to use `targetJvm.target`.

buildSrc/src/main/kotlin/otel.android-library-conventions.gradle.kts
fractalwrench: Updated

LikeTheSalad: Cheers!

bidetofevil:
The idea is that we should fix or configure the ruleset so there are no XML files, but I'd take those rather than have the issues that those bring up.

Disabling rules that we don't care about enough to fix is always a good option

PR #1124, opened 2025-08-04
Setup binary compatibility validator

### LINKED ISSUE #217: Public API compatibility validator
As an SDK, we should avoid breaking changes, whenever possible obviously.

Sometimes we'd do them, but then we'd need to increase a minor or major version if following semver.

The goal of this issue is to keep track of the Public API.

To do that, we could use the [binary-compatibility-validator](https://github.com/Kotlin/binary-compatibility-validator) plugin.

This plugin generates a file with all the public methods, and every time you either increase or decrease the public API, this file is gonna be changed and you could see the changes during the PR review, if it was an intentional change, all good, if it's by mistake, then you'd know before shipping the change, it's a preventive approach rather than reactive.

[Example](https://github.com/PostHog/posthog-android/blob/main/posthog/api/posthog.api) of the generated file.
[Example](https://github.com/PostHog/posthog-android/blob/b507e70c52f008c2a9cddc0be9dd764ccf3f5f96/build.gradle.kts#L54-L56) of the setup.
[Example](https://github.com/PostHog/posthog-android/blob/b507e70c52f008c2a9cddc0be9dd764ccf3f5f96/Makefile#L20-L21) generating the file.
When you run eg `./gradlew check` but don't do `./gradlew apiDump` before`, the build is gonna fail if you increased/decreased the public API not intentionally.

Something to consider is also using the [explicitApi](https://kotlinlang.org/docs/whatsnew14.html#explicit-api-mode-for-library-authors) Kotlin mode.

This will force you to set all the visibility modifiers in methods, properties, classes, etc, it could be a bit of boilerplate (eg setting all the classes to `public` - it is public by default already), but it is more explicit and is better for identifying breaking changes proactively.

fractalwrench:
## Goal

This changeset adds Kotlin's [binary compatibility validator](https://github.com/Kotlin/binary-compatibility-validator) which generates a dump of API signatures and checks whether the project source code matches the known dump, thus allowing any API changes to be detected. The project works on Java & Kotlin files that are publicly exposed. `./gradlew apiCheck` runs the task, `./gradlew apiDump` regenerates the API dump file. The check will automatically run with `./gradlew check` so CI should catch any regressions/changes in the public API.

I've added this to any module that applies the `otel.android-library-conventions` gradle script and have generated a dump file for the existing APIs.

Closes #217

### DIFF
--- a/android-agent/api/android-agent.api
+++ b/android-agent/api/android-agent.api
@@ -1,0 +1,71 @@
+public final class io/opentelemetry/android/agent/OpenTelemetryRumInitializer {
+	public static final field INSTANCE Lio/opentelemetry/android/agent/OpenTelemetryRumInitializer;
+	public static final fun initialize (Landroid/app/Application;Ljava/lang/String;Ljava/util/Map;Lio/opentelemetry/android/agent/connectivity/EndpointConnectivity;Lio/opentelemetry/android/agent/connectivity/EndpointConnectivity;Lio/opentelemetry/android/agent/connectivity/EndpointConnectivity;Lio/opentelemetry/android/config/OtelRumConfig;Lio/opentelemetry/android/agent/session/SessionConfig;Lkotlin/jvm/functions/Function1;Lio/opentelemetry/android/instrumentation/common/ScreenNameExtractor;Lkotlin/jvm/functions/Function1;Lio/opentelemetry/android/instrumentation/common/ScreenNameExtractor;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/time/Duration;)Lio/opentelemetry/android/OpenTelemetryRum;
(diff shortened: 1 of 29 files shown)


test-common/api/test-common.api
LikeTheSalad, line 1: Can we ignore this test module?

fractalwrench: @LikeTheSalad I've disabled the check for the test-common module only by adding a list of ignored projects to the gradle script. Another approach would be to define a property in the `gradle.properties` of that module that can be used to determine whether the check should run or not.

LikeTheSalad: Thanks @fractalwrench, for now, since this is the only exception, I think it's fine to leave it that way.

LikeTheSalad (approved):

PR #1129, opened 2025-08-05
Remove READ_PHONE_STATE permission and update docs

bidetofevil:
Remove what the Android docs call a "dangerous" permission from the SDK manifest. This only affects one attribute - the network subtype for mobile connections (e.g. LTE, 5G, etc.), and SDK users can grant this permission in their own manifest and get this extra information. Otherwise, it won't be included.


LikeTheSalad (approved):
Cheers!

PR #1131, opened 2025-08-06
Group androidx.test

breedx-splk:
Supersedes #1113 and #1121. Also fixes a warning.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -106,6 +106,12 @@
       "groupName": "androidx.core"
     },
     {
+      "matchPackagePatterns": [
+        "^androidx.test"
+      ],
+      "groupName": "androidx.test"
+    },
+    {
       "matchPackagePrefixes": ["com.gradle.develocity"],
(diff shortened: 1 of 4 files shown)


breedx-splk:
> I'm not aware of any guarantee that libraries under androidx.test are released with the same versioning (see the [current released version matrix](https://developer.android.com/jetpack/androidx/releases/test) for an example of this). In practice I think the core/rule artifacts would nearly always be released each time so it might not be an issue.

Eek, ok, we should not merge this then. I will close it. Thank you!
I think I made a bad assumption based on the fact that there were two that came in nearly at the same time, and [the core one has a bunch of build failures](https://github.com/open-telemetry/opentelemetry-android/pull/1113) but now that I'm looking again it seems unrelated.

> I'm not super familiar with Renovate - is the behavior of `groupName` that it sets everything to the lowest or highest version when there are artifacts in the same group with different versions available?

I think it buffers (waits) until the things in the group can be released together. Which would mean that this change would effectively block renovate, which is not what we want.

PR #1133, opened 2025-08-07
Convert instrumentation code to kotlin

fractalwrench:
## Goal

Begins converting some of the instrumentation code to Kotlin. I've avoided API breaking changes by relying on the API validation added in #1124 which should detect any changes to the public API. The only change is functions/classes are final by default in Kotlin, which feels a reasonable change IMO.

The ultimate goal of these changes is to migrate from kapt to ksp. Doing this requires migrating away from `@AutoService` to [a Kotlin solution](https://github.com/ZacSweers/auto-service-ksp) that uses the same annotation, and as part of this, it'll be necessary to migrate these classes to Kotlin.


instrumentation/okhttp3-websocket/library/src/main/java/io/opentelemetry/instrumentation/library/okhttp/v3_0/websocket/internal/OkHttpWebsocketInstrumentation.kt
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+@file:Suppress("ktlint:standard:package-name")
```
breedx-splk, line 6: Is that because of the underscore?

instrumentation/android-log/library/src/main/java/io/opentelemetry/instrumentation/library/log/AndroidLogInstrumentation.kt
```diff
@@ -0,0 +1,24 @@
...
+import com.google.auto.service.AutoService
+import io.opentelemetry.android.instrumentation.AndroidInstrumentation
+import io.opentelemetry.android.instrumentation.InstallationContext
+import io.opentelemetry.instrumentation.library.log.internal.LogRecordBuilderCreator.configure
+
+@AutoService(AndroidInstrumentation::class)
+class AndroidLogInstrumentation : AndroidInstrumentation {
+    private companion object {
+        public const val INSTRUMENTATION_NAME = "android-log"
+    }
```
breedx-splk: I think remove the companion and just inline the string literal (like you did in the other instrumentation(s)). 👍🏻

instrumentation/anr/src/main/java/io/opentelemetry/android/instrumentation/anr/AnrInstrumentation.kt
```diff
@@ -0,0 +1,57 @@
...
+    fun setMainLooper(looper: Looper): AnrInstrumentation {
+        mainLooper = looper
+        return this
+    }
+
+    @VisibleForTesting
+    internal fun setScheduler(scheduler: ScheduledExecutorService): AnrInstrumentation {
+        this.scheduler = scheduler
+        return this
+    }
```
breedx-splk: Inconsistent use of the `@VisibleForTesting` annotation. Include it in both places probably? Alternatively (and maybe this is an improvement) -- you could take them from `var` to `val` by putting them as defaulted constructor args...

instrumentation/android-log/library/src/main/java/io/opentelemetry/instrumentation/library/log/AndroidLogInstrumentation.kt
fractalwrench: This is _technically_ a change to the public API as the constant was public in Java. I can still make the change for consistency if you want?

instrumentation/okhttp3-websocket/library/src/main/java/io/opentelemetry/instrumentation/library/okhttp/v3_0/websocket/internal/OkHttpWebsocketInstrumentation.kt
fractalwrench: Yep - which presumably isn't something that can be changed without breaking the API

instrumentation/android-log/library/src/main/java/io/opentelemetry/instrumentation/library/log/AndroidLogInstrumentation.kt
fractalwrench: I've made this change for consistency.

fractalwrench:
Updated according to review feedback.

instrumentation/android-log/library/src/main/java/io/opentelemetry/instrumentation/library/log/AndroidLogInstrumentation.kt
breedx-splk: Thanks yeah seems minor, and I can't imagine we had anyone using that....and those that were were misguided. :)

Issue #232, opened 2024-01-25
Consider removing READ_PHONE_STATE


breedx-splk:
So in our Splunk distro, a user raised an issue about [READ_PHONE_STATE](https://github.com/signalfx/splunk-otel-android/issues/733) permission being a security concern.  It exists in the manifest [here](https://github.com/open-telemetry/opentelemetry-android/blob/883804b558aa584b436b840fe472168f3b89c444/instrumentation/src/main/AndroidManifest.xml#L6).

I don't remember the exact details right now, but I think that maybe network detection doesn't work correctly on older versions of Android without it?

Curious if other folks know specifics, and if we should consider relaxing the requirement around this permission.

Relates to #49.

ber4444:
Just to clarify, will Open telemetry as well as Slunk RUM properly handle network subtype being null?

surbhiia:
@breedx-splk Please assign this ticket to me. I'll create a PR implementing what @marandaneto mentioned above as we discussed in the SIG meeting. Thanks!

fractalwrench:
From memory, this is now optional and the behavior is to use the permission if it is there, and otherwise gracefully fallback. Therefore closing this issue

PR #1147, opened 2025-08-12
Update network fetch - Use relevant APIs and permissions across different API Levels

surbhiia:
**Changes** done in this PR:
- Not limiting carrier details fetch to API 28+. Use right feature, permission checks and APIs as needed for different API levels to gather needed information.
- Update the SubType finding logic as well similarly. 
- Update the NetworkDetector logic to ensure correct APIs used for different supported Android API levels. 
- Remove READ_PHONE_STATE permission from this SDK’s modules. Added it to demo-app to show how it should be used. 
- Update test cases where necessary and add new tests
- Increased the maximum number of lines allowed by **detekt** to 140 as **spotlessApply** and detekt were conflicting, spotlessApply prefers simplified syntax (making the function one line with a = sign instead of the braces function syntax where application) over line length (ktlint default 100). So, there was no way to resolve this conflict other than  increasing the line length detekt allows

**Resolves issues** - #232 , #49 

**Tested** the changes on emulators for Android API levels - 33, 28, 23, 21 with and without permission grants.   

**Note**: On API 21 test with emulator, carrier details were not fetched.   Yet to figure this out, might need to test on a real device perhaps. 

**Remaining** (I'll update these in a separate PR) :
- Ensure warning logs are added where necessary when a feature or permission is missing. 
- Update Readme’s as necessary.


demo-app/src/main/java/io/opentelemetry/android/demo/MainActivity.kt
```diff
@@ -87,5 +91,16 @@ class MainActivity : ComponentActivity() {
             }
         }
         viewModel.sessionIdState.value = OtelDemoApplication.rum?.rumSessionId!!
+
+        if (ContextCompat.checkSelfPermission(this, Manifest.permission.READ_PHONE_STATE)
```
bidetofevil: Will this work if you have [READ_BASIC_PHONE_STATE](https://developer.android.com/reference/android/Manifest.permission#READ_BASIC_PHONE_STATE)? That's a much narrower and safer permission to request and works with API 33+.

common/src/main/java/io/opentelemetry/android/common/internal/features/networkattributes/data/Carrier.kt
```diff
@@ -5,18 +5,13 @@
...
-
 /**
  * This class is internal and not for public use. Its APIs are unstable and can change at any time.
  */
-@RequiresApi(api = Build.VERSION_CODES.P)
 data class Carrier
     @JvmOverloads
     constructor(
-        val id: Int = TelephonyManager.UNKNOWN_CARRIER_ID,
+        val id: Int = -1,
```
bidetofevil: Is there a way to avoid a magic number like this?

services/src/main/java/io/opentelemetry/android/internal/services/network/CarrierFinder.java
```diff
@@ -5,46 +5,115 @@
...
+            return pm.hasSystemFeature(PackageManager.FEATURE_TELEPHONY);
         }
-        return new Carrier(id, name, mobileCountryCode, mobileNetworkCode, isoCountryCode);
     }
 
-    private boolean validString(CharSequence str) {
-        return !(str == null || str.length() == 0);
+    /** Creates a default carrier object for when telephony is not available. */
+    @NonNull
+    private Carrier createUnknownCarrier() {
```
bidetofevil: If this is a constant, why not just create a static instance and return that?

services/src/main/java/io/opentelemetry/android/internal/services/network/NetworkUtils.kt
```diff
@@ -0,0 +1,59 @@
...
+import android.content.Context
+import android.content.pm.PackageManager
+import android.telephony.TelephonyManager
+import androidx.core.content.ContextCompat
+
+/**
+ * Shared utilities for network detection.
+ * This class is internal and not for public use. Its APIs are unstable and can change at any time.
+ */
+object NetworkUtils {
```
bidetofevil: This doesn't need to be an object - you can just add functions to a Kotlin file directly and have each be imported as needed.

services/src/main/java/io/opentelemetry/android/internal/services/network/SubTypeFinder.kt
```diff
@@ -0,0 +1,55 @@
...
+import android.os.Build
+import android.telephony.TelephonyManager
+import android.util.Log
+import androidx.annotation.RequiresApi
+import io.opentelemetry.android.common.RumConstants
+
+/**
+ * This class is internal and not for public use. Its APIs are unstable and can change at any time.
+ */
+class SubTypeFinder(
```
bidetofevil: Any reason why this has to be an object rather than just a function that that takes what's required as parameters?

services/src/main/java/io/opentelemetry/android/internal/services/network/detector/NetworkDetector.kt
```diff
@@ -21,22 +27,76 @@ interface NetworkDetector {
     companion object {
         @JvmStatic
         fun create(context: Context): NetworkDetector {
-            // TODO: Use ServiceManager to get the ConnectivityManager or similar (not yet managed/abstracted)
+            // TODO: Use ServiceManager to get the ConnectivityManager, TelephonyManager or similar (not yet managed/abstracted)
```
bidetofevil: This seems quite complicated given that you're adding more logic to this factory method. Why not just create an implementation class for the interface rather than already creating a new anonymous instance?

services/src/test/java/io/opentelemetry/android/internal/services/network/CarrierFinderTest.java
```diff
@@ -6,43 +6,211 @@
...
 import io.opentelemetry.android.common.internal.features.networkattributes.data.Carrier;
-import org.junit.jupiter.api.Test;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.robolectric.annotation.Config;
 
-class CarrierFinderTest {
+@RunWith(AndroidJUnit4.class)
+public class CarrierFinderTest {
```
bidetofevil, line 28: Since the original class is so small, why not convert to Kotlin first?

services/src/test/java/io/opentelemetry/android/internal/services/network/detector/NetworkDetectorTest.java
```diff
@@ -5,34 +5,205 @@
...
+import org.junit.Assert;
+import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
+import org.robolectric.Shadows;
 import org.robolectric.annotation.Config;
+import org.robolectric.shadows.ShadowNetworkInfo;
 
 @RunWith(AndroidJUnit4.class)
 public class NetworkDetectorTest {
```
bidetofevil, line 38: ditto Kotlin conversion

common/src/main/java/io/opentelemetry/android/common/internal/features/networkattributes/data/Carrier.kt
surbhiia: Actually `TelephonyManager.UNKNOWN_CARRIER_ID = -1` but it is only available from API 28+ but our class works for all API levels so used the actual value instead of the constant.

services/src/test/java/io/opentelemetry/android/internal/services/network/CarrierFinderTest.java
surbhiia: For this PR, due to less bandwidth right now, I've decided to only add any new files as kotlin files and kept the old ones as is. I'll take up the kotlin conversion sometime later if that seems okay. :)

services/src/main/java/io/opentelemetry/android/internal/services/network/SubTypeFinder.kt
surbhiia: Thought it'd be better/ cleaner to have it as a separate object like CarrierFinder, so we know what APIs / permissions go into finding this and can be tested as a separate unit as well.

common/src/main/java/io/opentelemetry/android/common/internal/features/networkattributes/data/Carrier.kt
bidetofevil: Ah ok. Maybe define a constant for that so you can add a comment to where it's defined so the magic number won't proliferate?

services/src/main/java/io/opentelemetry/android/internal/services/network/SubTypeFinder.kt
bidetofevil: What I mean is you can put this as a separate utility function rather than a class with that contains references, e.g.

```
fun findSubtype(context: Context, telephoneyManager: TelephoneyManager) : String {
  ...
}
```

In fact, you can probably check and cache the permission check so you only need to pass in the `TelephoneyManager` (or make it an extension function of it). You can still unity test this appropriately by doing that.

services/src/test/java/io/opentelemetry/android/internal/services/network/CarrierFinderTest.java
bidetofevil: Sure. The one advantage is that you won't have to add all the Java backwards compat annotations if all your callers are Kotlin, but that's fine too.

demo-app/src/main/AndroidManifest.xml
```diff
@@ -3,6 +3,7 @@
     xmlns:tools="http://schemas.android.com/tools">
 
     <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.READ_PHONE_STATE" />
```
fractalwrench, line 6: Personally I would default this feature to off and perhaps leave a comment in the manifest explaining how to switch it on for folks who do have interest in capturing this information. Not a strong opinion

services/api/services.api
```diff
@@ -52,6 +52,18 @@ public abstract interface class io/opentelemetry/android/internal/services/netwo
...
+	public static final field INSTANCE Lio/opentelemetry/android/internal/services/network/NetworkUtils;
+	public final fun getNetworkTypeName (I)Ljava/lang/String;
+	public final fun hasReadPhoneStatePermission (Landroid/content/Context;)Z
+	public final fun isValidString (Ljava/lang/CharSequence;)Z
+}
+
+public final class io/opentelemetry/android/internal/services/network/SubTypeFinder {
+	public fun <init> (Landroid/content/Context;Landroid/telephony/TelephonyManager;)V
+	public final fun get ()Ljava/lang/String;
+}
```
fractalwrench: Unless these need to form part of OpenTelemetry's public API, these symbols should be made non-public

services/src/main/java/io/opentelemetry/android/internal/services/network/detector/NetworkDetector.kt
```diff
@@ -21,22 +27,79 @@ interface NetworkDetector {
...
-                    telephonyManager,
-                    carrierFinder,
-                    context,
-                )
+            val telephonyManager =
+                context.getSystemService(Context.TELEPHONY_SERVICE) as TelephonyManager
+            val carrierFinder = CarrierFinder(context, telephonyManager)
+            val subTypeFinder = SubTypeFinder(context, telephonyManager)
+
+            return object : NetworkDetector {
```
fractalwrench: Consider separating this out from the source file containing the interface, given the complexity has grown somewhat

fractalwrench (commented):
Thanks for taking this on!

demo-app/src/main/AndroidManifest.xml
surbhiia: This is just added in the demo-app to showcase the behavior of network details capture. The app also requests this permission at run-time. For the SDK itself we're following the strategy you mentioned. Let me know if you meant otherwise - i.e that it should be removed from demo-app as well?

demo-app/src/main/java/io/opentelemetry/android/demo/MainActivity.kt
surbhiia: Thanks for pointing this out. API 33 onwards either READ_BASIC_PHONE_STATE or READ_PHONE_STATE should be granted for calling the APIs we're calling. I'll fix this.

demo-app/src/main/AndroidManifest.xml
fractalwrench: Yeah my opinion is that this shouldn't be on in the demo app either by default - I don't think most folks would want to capture this level of information in an Android app. Perhaps that could be spun into a separate issue though as I don't want to hold this up

bidetofevil: Agreed with @fractalwrench. This is a dangerous permission. I think `READ_BASIC_PHONE_STATE` is reasonable to add in the demo app so I'd go with that maybe?

surbhiia: @bidetofevil READ_BASIC_PHONE_STATE was added in API 33 but demo-app can be tested with lower APIs too as we support upto API 21. So, we either add both or add none. In my opinion, demo-app showcases features of this SDK, we should add it there to showcase an example about how to add these and what they're needed for.

LikeTheSalad: I don't see an issue with having it for the demo app, as it won't make the agent force anything on the consumer; it's just an example of how they can get all the data that the agent gathers automatically. The idea of making it opt-in/opt-out despite the permission being available is an interesting topic that I think should be handled as its own issue. For now, I think it's worth leaving the old permission as it won't cause confusion around people who might want to try the demo app with the min API level that we claim to support, as well as to avoid breaking our tests which are using a [API 29](https://github.com/open-telemetry/opentelemetry-android/blob/276e5b6a785b3aec30b6e3dfe27a95f2f15d4bd9/.github/workflows/codeql.yml#L67). (they should be using 21, now that I think about it, so we should probably lower it for our tests).

If the concern is that people might think that they must use `READ_PHONE_STATE` and it turns out that `READ_BASIC_PHONE_STATE` is enough, then I think we should explain it in a comment around the declaration of `READ_PHONE_STATE` and also in our docs.

PR #1149, opened 2025-08-13
Fix links in markdown

breedx-splk:
Some java classes converted to kotlin and broke the links in the readme. Follows #1133 and #1139.


fractalwrench (approved):
LGTM - thanks for catching this

breedx-splk:
> LGTM - thanks for catching this

I only caught it because the build caught it! 😅

PR #1155, opened 2025-08-14
Okhttp jvm android resolution

LikeTheSalad:
Supersedes #1137

### DIFF
--- a/buildSrc/src/main/kotlin/otel.android-dependency-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.android-dependency-conventions.gradle.kts
@@ -1,0 +1,11 @@
+configurations.all {
+    resolutionStrategy {
+        eachDependency {
+            if (requested.group == "com.squareup.okhttp3" && requested.name == "okhttp-jvm") {
+                useTarget("com.squareup.okhttp3:okhttp-android:${requested.version}")
+                because("choosing okhttp-android over okhttp-jvm")
+            }
+        }
+    }
+}
+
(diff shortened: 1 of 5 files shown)


buildSrc/src/main/kotlin/otel.android-dependency-conventions.gradle.kts
```diff
@@ -0,0 +1,11 @@
+configurations.all {
+    resolutionStrategy {
+        eachDependency {
+            if (requested.group == "com.squareup.okhttp3" && requested.name == "okhttp-jvm") {
+                useTarget("com.squareup.okhttp3:okhttp:${requested.version}")
```
breedx-splk, line 5: I think this is fine, but did we consider being explicit here and instead using `okhttp-android`?

LikeTheSalad: I'm fine either way as long as we are consistent across the whole project, and right now, we're declaring it as `okhttp` in our [toml file](https://github.com/open-telemetry/opentelemetry-android/blob/1450d81ad68a29628b56dcff68f5948e70e90844/gradle/libs.versions.toml#L35). The reason I think it's better to keep consistency is because it can avoid us issues in the future in case we define some dependency rules around `okhttp`, for example, expecting those to cover `okhttp-android`, and probably at first Gradle doesn't know that those are the same for this project and don't apply everything consistently. There are ways to tell Gradle that 2 or more dependencies are essentially the same by declaring custom [dependency capabilities](https://docs.gradle.org/current/userguide/component_capabilities.html#sec:declaring-capabilities-external-modules), but that route seems to be unnecessarily overcomplicating things. So if you like, I can change it here to `okhttp-android` and then also change it in our toml file to keep it the same everywhere.

breedx-splk: Ok let's stick with this for now and if we have a good reason later to make it `-android` (explicitly) then we can change it then. This should be fine as-is and does match what's in the toml. Thanks!

PR #1158, opened 2025-08-14
Fix build warnings

fractalwrench:
## Goal

Fixes some additional build warnings I spotted when running `./gradlew check`, mostly related to deprecated symbols that need to be invoked from test code.

### DIFF
--- a/instrumentation/anr/src/main/java/io/opentelemetry/android/instrumentation/anr/AnrWatcher.kt
+++ b/instrumentation/anr/src/main/java/io/opentelemetry/android/instrumentation/anr/AnrWatcher.kt
@@ -74,10 +74,12 @@ internal class AnrWatcher(
     }
 
     private fun emitAnrEvent(stackTrace: Array<StackTraceElement>) {
+        @Suppress("DEPRECATION")
+        val id = mainThread.id
         val attributesBuilder =
             Attributes
                 .builder()
-                .put(THREAD_ID, mainThread.id)
+                .put(THREAD_ID, id)
                 .put(THREAD_NAME, mainThread.name)
                 .put(EXCEPTION_STACKTRACE, stackTraceToString(stackTrace))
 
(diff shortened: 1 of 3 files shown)


instrumentation/anr/src/main/java/io/opentelemetry/android/instrumentation/anr/AnrWatcher.kt
```diff
@@ -73,6 +73,7 @@ internal class AnrWatcher(
         }
     }
 
+    @Suppress("DEPRECATION") // suppress deprecation for id
```
breedx-splk: Should we switch the implementation to use `threadId` instead then?

services/src/test/java/io/opentelemetry/android/internal/services/periodicwork/PeriodicWorkTest.kt
```diff
@@ -27,6 +27,7 @@ class PeriodicWorkTest {
         service = PeriodicWork()
     }
 
+    @Suppress("DEPRECATION") // suppress deprecation for id
```
breedx-splk: Should we change the impl instead to use `threadId`?

instrumentation/anr/src/main/java/io/opentelemetry/android/instrumentation/anr/AnrWatcher.kt
fractalwrench: Yeah that sounds reasonable, we'd need to add an API version check as that function was only added in API 36.

fractalwrench: I updated this with a version check but ran into a build failure due to the animal sniffer gradle plugin, so have left this as-is for now. The error message was:

> Task :instrumentation:anr:animalsnifferDebug FAILED
[Undefined reference] io.opentelemetry.android.instrumentation.anr.(AnrWatcher.kt:80)

I assume that perhaps animal sniffer is configured to use an older Android API that doesn't contain the `threadId` function that was added in 36?

breedx-splk (approved):

PR #1169, opened 2025-08-18
Attempt to fix the build by disabling incremental kotlin compile 

breedx-splk:
Our CI builds for `main` are currently failing the (snapshot) publish step ([example 1](https://github.com/open-telemetry/opentelemetry-android/actions/runs/17046150939/job/48322387165), [example 2](https://github.com/open-telemetry/opentelemetry-android/actions/runs/17046169176/job/48322451764)). 

This seems to be caused by a bug in gradle ([see this thread](https://youtrack.jetbrains.com/projects/KT/issues/KT-79047/Gradle-compileKotlin-fails-with-configuration-cache)). It should be fixed in the next version which is imminent, but until then let's get our build fixed.

I've also disabled it for our release process builds, because we anticipate releasing this week can can't necessarily wait for the fix.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -29,7 +29,9 @@ jobs:
         working-directory: ./demo-app
         run: ./gradlew check assemble
       - name: publish snapshot
-        run: ./gradlew publishToSonatype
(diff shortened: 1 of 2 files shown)


fractalwrench (approved):
LGTM

PR #1172, opened 2025-08-19
Revamp crash instrumentation

fractalwrench:
## Goal

This changeset fully converts the crash instrumentation to Kotlin and adds appropriate test coverage. I've attempted to keep the implementation broadly the same but have removed Java concepts such as `Consumer` in favour of more idiomatic Kotlin. I will add comments inline that detail decisions I've taken during this conversion.

There are a few remaining problems to address in this area of the codebase: namely the use of `OpenTelemetrySdk` rather than an interface, and also the fact there's no sensible limit for truncation of stackframes. It'd be great to discuss the approach as part of this PR but IMO we should action them in a subsequent PR.

## Testing

I replaced the majority of the test coverage as it was using a large number of mocks - I've replaced these with fakes instead. I'm very curious to get feedback on this approach as I think it'd be preferable to write tests in this way throughout the codebase rather than relying on mocks.

Essentially, each test case creates an instance of the default `OpenTelemetry` with a fake object configured for `LogRecordExporter`. The test case asserts a log record was received for each crash that happened & therefore allows the crash functionality to be tested through most parts of the SDK it will interact with.

At Embrace we extend this approach further by asserting that logs/spans match a known JSON output. If folks are happy with the general approach I've taken to testing, I can abstract the test code further so that it can be used by the other instrumentation modules.

Finally, the crash tests seem faster for me locally. Previously each test case took ~300ms to run locally, whereas now they take a few milliseconds each.


instrumentation/crash/api/crash.api
```diff
@@ -1,14 +1,6 @@
 public final class io/opentelemetry/android/instrumentation/crash/CrashDetails {
-	public static fun create (Ljava/lang/Thread;Ljava/lang/Throwable;)Lio/opentelemetry/android/instrumentation/crash/CrashDetails;
-	public fun equals (Ljava/lang/Object;)Z
```
fractalwrench, line 3: These seemed unused so I removed them

instrumentation/crash/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashReporter.kt
```diff
@@ -0,0 +1,72 @@
...
+import io.opentelemetry.semconv.ExceptionAttributes.EXCEPTION_TYPE
+import io.opentelemetry.semconv.incubating.ThreadIncubatingAttributes.THREAD_ID
+import io.opentelemetry.semconv.incubating.ThreadIncubatingAttributes.THREAD_NAME
+import java.util.concurrent.TimeUnit
+
+internal class CrashReporter(
+    additionalExtractors: List<EventAttributesExtractor<CrashDetails>>,
+) {
+    private val extractors: List<EventAttributesExtractor<CrashDetails>> =
+        additionalExtractors.toList()
```
fractalwrench, line 24: Added defensive copy to avoid potential concurrent modification

instrumentation/crash/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashReporter.kt
```diff
@@ -0,0 +1,72 @@
...
+        val throwable = crashDetails.cause
+        val thread = crashDetails.thread
+        val attributesBuilder =
+            builder()
+                .put(THREAD_ID, thread.id)
+                .put(THREAD_NAME, thread.name)
+                .put(EXCEPTION_MESSAGE, throwable.message)
+                .put(
+                    EXCEPTION_STACKTRACE,
+                    throwable.stackTraceToString(),
```
fractalwrench, line 55: Used a Kotlin stdlib function for this rather than using printwriters directly

instrumentation/crash/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashReporterInstrumentation.kt
```diff
@@ -23,8 +23,10 @@ class CrashReporterInstrumentation : AndroidInstrumentation {
     }
 
     override fun install(ctx: InstallationContext) {
-        addAttributesExtractor(RuntimeDetailsExtractor.create<CrashDetails, Any>(ctx.application))
+        addAttributesExtractor(RuntimeDetailsExtractor.create(ctx.application))
         val crashReporter = CrashReporter(additionalExtractors)
+
+        // TODO avoid using OpenTelemetrySdk methods, only use the ones from OpenTelemetry api.
```
fractalwrench, line 29: This is a pre-existing issue, I moved the comment

instrumentation/crash/src/test/java/io/opentelemetry/android/instrumentation/crash/CrashReportIntegrationTest.kt
```diff
@@ -0,0 +1,233 @@
...
+        assertEquals("value2", attrs["key2"])
+    }
+
+    @Test
+    fun `test stackoverflow behavior`() {
+        val exc = generateLargeStacktraceException()
+        val log = simulateUncaughtException(exc)
+        val stacktrace = exc.stackTraceToString()
+
+        // TODO: future: truncate stackframes at some reasonable limit
```
fractalwrench, line 111: We need to truncate stacktraces at some reasonable limit e.g. 500-1000. Production stacktraces can in rare circumstances be much larger than this with repeated frames

LikeTheSalad (approved):
This is great, thank you 🙏 

> If folks are happy with the general approach I've taken to testing, I can abstract the test code further so that it can be used by the other instrumentation modules.

I'm happy with going with this approach. I still think we need to use mocks in some places, though, fakes can be very useful for data-related objects, but maybe for some others with lots of business logic, it might be tiresome to create and maintain fake impls in many places. But generally speaking, I like the idea of using fakes more, so I'm up for it.

Also, I think we should create an issue around the comment you added to truncate large stacktraces, to keep track of it.

instrumentation/crash/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashDetails.kt
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.android.instrumentation.crash
+
+/** A class representing all the details of an application crash. */
+class CrashDetails internal constructor(
```
breedx-splk, line 9:
```suggestion
data class CrashDetails internal constructor(
```

Not sure how much this matters....but maybe slightly more idiomatic?

instrumentation/crash/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashReporter.kt
```diff
@@ -0,0 +1,72 @@
...
+
+    private fun processCrash(
+        openTelemetry: OpenTelemetrySdk,
+        crashDetails: CrashDetails,
+    ) {
+        val logger = openTelemetry.sdkLoggerProvider.loggerBuilder("io.opentelemetry.crash").build()
+        val throwable = crashDetails.cause
+        val thread = crashDetails.thread
+        val attributesBuilder =
+            builder()
```
breedx-splk: This is a rare case where I think the static import actually slightly hurts readability, because there's not enough context to know what's being built. Prefer the `Attributes.builder()` to be explicit.

breedx-splk (approved):
This seems cool to me, thanks for taking this on!

Regarding the new test framework being developed here, we should give some care to try and not duplicate stuff that might already exist in the upstream java repo(s). For example, please make sure you're looking at `OpenTelemetryRule` and friends...which already provide a way to get and do assertions against telemetry data.

instrumentation/crash/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashDetails.kt
fractalwrench: If this was internal I'd agree, however data classes in publicly exposed APIs should generally be avoided as Kotlin generates [various functions](https://jakewharton.com/public-api-challenges-in-kotlin/) that can easily lead to API breakages. @breedx-splk I'm assuming that `CrashDetails` forms part of the publicly exposed API here?

instrumentation/crash/src/test/java/io/opentelemetry/android/instrumentation/crash/CrashReportIntegrationTest.kt
fractalwrench: Opened #1195

fractalwrench:
Thanks for the reviews all! I have addressed/commented on the feedback, opened an issue for truncating large stacktraces, and have updated the test case to use the existing `OpenTelemetryRule`.

PR #1174, opened 2025-08-20
Update network monitor README.md

surbhiia:
A quick PR to update network monitor README to mention permission requirement without going into too much details. Let me know if it looks good! :) 

For the main README, for a future consolidated update, we can also consider including a description of all the general attributes (including network attributes) that are added by default to all logs and spans.


LikeTheSalad (approved):
Thank you!

PR #1175, opened 2025-08-20
Refactor jank to use events 

### LINKED ISSUE #904: Migrate slow rendering instrumentation from spans to events
See [this comment](https://github.com/open-telemetry/opentelemetry-android/pull/903#discussion_r2007090688).

The slow rendering instrumentation has always used zero-duration spans to send data. We should migrate these to events proper.

breedx-splk:
Resolves #904.

Also attempts to address the prototype implementation in https://github.com/open-telemetry/semantic-conventions/pull/2552.  

This was done in a way that allows us to continue emitting the zero-duration span, but to mark it as deprecated for a release and remove it in the next release.


instrumentation/slowrendering/src/main/java/io/opentelemetry/android/instrumentation/slowrendering/SlowRenderListener.kt
```diff
@@ -0,0 +1,106 @@
...
+    constructor(jankReporter: JankReporter, pollInterval: Duration) : this(
+        jankReporter,
+        Executors.newScheduledThreadPool(1),
+        Handler(startFrameMetricsLoop()),
+        pollInterval,
+    )
+
+    // the returned future is very unlikely to fail
+    fun start() {
+        executorService.scheduleWithFixedDelay(
```
cleverchuk, line 42: Always thought that this one executes once. Just had to read the Java doc this time to realize why it feels that way.

instrumentation/slowrendering/src/main/java/io/opentelemetry/android/instrumentation/slowrendering/SlowRenderingInstrumentation.kt
```diff
@@ -61,14 +71,19 @@ class SlowRenderingInstrumentation : AndroidInstrumentation {
...
-                ctx.openTelemetry.getTracer("io.opentelemetry.slow-rendering"),
-                slowRenderingDetectionPollInterval,
-            )
+        val logger = ctx.openTelemetry.logsBridge.get("app.jank")
+        var jankReporter: JankReporter = EventsJankReporter(logger, SLOW_THRESHOLD_MS / 1000.0)
+        jankReporter = jankReporter.combine(EventsJankReporter(logger, FROZEN_THRESHOLD_MS / 1000.0))
+
+        if (useDeprecatedSpan) {
+            val tracer = ctx.openTelemetry.getTracer("io.opentelemetry.slow-rendering")
+            jankReporter = jankReporter.combine(SpanBasedJankReporter(tracer))
```
cleverchuk, line 89: Was wondering why the function is necessary and this explains it. Nice!

instrumentation/slowrendering/src/main/java/io/opentelemetry/android/instrumentation/slowrendering/SpanBasedJankReporter.kt
```diff
@@ -0,0 +1,70 @@
...
+import android.util.Log
+import android.util.SparseIntArray
+import io.opentelemetry.android.common.RumConstants
+import io.opentelemetry.api.trace.Span
+import io.opentelemetry.api.trace.Tracer
+import java.time.Instant
+
+internal const val SLOW_THRESHOLD_MS = 16
+internal const val FROZEN_THRESHOLD_MS = 700
+
```
cleverchuk, line 16: I'm assuming these ☝️  where taken from some Android doc?

instrumentation/slowrendering/src/main/java/io/opentelemetry/android/instrumentation/slowrendering/EventsJankReporter.kt
```diff
@@ -0,0 +1,58 @@
...
+import io.opentelemetry.api.common.Attributes
+import io.opentelemetry.api.incubator.logs.ExtendedLogRecordBuilder
+import io.opentelemetry.api.logs.Logger
+
+// TODO: Replace with semconv constants
+internal val FRAME_COUNT: AttributeKey<Long> = AttributeKey.longKey("app.jank.frame_count")
+internal val PERIOD: AttributeKey<Double> = AttributeKey.doubleKey("app.jank.period")
+internal val THRESHOLD: AttributeKey<Double> = AttributeKey.doubleKey("app.jank.threshold")
+
+internal class EventsJankReporter(
```
cleverchuk: !nit, How about singular `EventJankReporter`

cleverchuk (approved):
LGTM

instrumentation/slowrendering/src/main/java/io/opentelemetry/android/instrumentation/slowrendering/SpanBasedJankReporter.kt
breedx-splk: Yeah, those both come from https://developer.android.com/topic/performance/vitals/render

instrumentation/slowrendering/src/main/java/io/opentelemetry/android/instrumentation/slowrendering/SlowRenderingInstrumentation.kt
breedx-splk: Yeah, might be overkill tho. We might consider removing it after we're down to just the events...

instrumentation/slowrendering/src/main/java/io/opentelemetry/android/instrumentation/slowrendering/EventsJankReporter.kt
breedx-splk: Yes probably.

instrumentation/slowrendering/src/main/java/io/opentelemetry/android/instrumentation/slowrendering/JankReporter.kt
```diff
@@ -0,0 +1,39 @@
...
+
+import android.util.SparseIntArray
+
+/**
+ * Responsible for sending telemetry. This is a temporary class that we can remove
+ * after the jank semconv becomes more stable.
+ */
+internal fun interface JankReporter {
+    fun reportSlow(
+        durationToCountHistogram: SparseIntArray,
```
bidetofevil: FYI, the perf improvement hack folks used to do with SparseArrays are no longer required. No need to change it here if you don't want, but we can probably change them all to `Map` and rely on the default `HashMap` implementation.

instrumentation/slowrendering/src/main/java/io/opentelemetry/android/instrumentation/slowrendering/EventsJankReporter.kt
```diff
@@ -0,0 +1,58 @@
...
+        durationToCountHistogram: SparseIntArray,
+        periodSeconds: Double,
+        activityName: String,
+    ) {
+        var frameCount: Long = 0
+        for (i in 0 until durationToCountHistogram.size()) {
+            val durationMillis = durationToCountHistogram.keyAt(i)
+            if ((durationMillis / 1000.0) > threshold) {
+                val count = durationToCountHistogram.get(durationMillis)
+                Log.d(
```
bidetofevil: Non-blocking: I'd probably pull this out of the loop if not entirely removed. It'll make logcat very noisy and cause needless strings to be created and thrown away at runtime.

instrumentation/slowrendering/src/main/java/io/opentelemetry/android/instrumentation/slowrendering/SlowRenderListener.kt
```diff
@@ -0,0 +1,106 @@
...
+import java.util.concurrent.TimeUnit
+
+@RequiresApi(api = Build.VERSION_CODES.N)
+internal class SlowRenderListener(
+    private val jankReporter: JankReporter,
+    private val executorService: ScheduledExecutorService,
+    private val frameMetricsHandler: Handler,
+    private val pollInterval: Duration,
+) : DefaultingActivityLifecycleCallbacks {
+    private val activities: ConcurrentMap<Activity, PerActivityListener> = ConcurrentHashMap()
```
bidetofevil, line 31: In case this callback is removed before it has a chance to process the deregistration of the listener for a particular activity, I'd use a `WeakHashMap` and `WeakReference` keys here as we don't want this to be the only reference preventing the GC of a dead Activity.

bidetofevil (approved):
LGTM. Nothing blocking. The leak in the case of a removal of the `SlowRenderListener` before clean up should be straightforward, but it's theoretical and if you don't think it'll happen, don't worry about it.

instrumentation/slowrendering/src/main/java/io/opentelemetry/android/instrumentation/slowrendering/EventsJankReporter.kt
LikeTheSalad: +1

instrumentation/slowrendering/src/main/java/io/opentelemetry/android/instrumentation/slowrendering/SlowRenderListener.kt
LikeTheSalad: I agree we should avoid keeping hard references to Activities.

instrumentation/slowrendering/src/main/java/io/opentelemetry/android/instrumentation/slowrendering/JankReporter.kt
breedx-splk: Yeah I have no idea why it was set up that way long ago. I've made the change to map and it simplified quite a few things. Thanks for the suggestion.

instrumentation/slowrendering/src/main/java/io/opentelemetry/android/instrumentation/slowrendering/EventsJankReporter.kt
breedx-splk: Well, hopefully it doesn't happen that often....but famous last words. 😄 
I've found it useful to see these in the output, so I'm going to wrap a flag around it and turn it off by default. Thanks for the idea.

instrumentation/slowrendering/src/main/java/io/opentelemetry/android/instrumentation/slowrendering/SlowRenderListener.kt
breedx-splk: Hmmm, you really got me thinking about this. 😅 It would be quite undesirable indeed to leak Activity references. I've looked at it from a few different angles and have just concluded that it's complicated.

I'm not sure _how_ it could happen, other than via `SlowRenderingInstrumentation.uninstall()` which both unregisters the listener, shuts down the `SlowRenderListener` which removes all the frame listeners and clears the activity map, and then it nulls its own reference...which should allow it to be GCd anyway. Agreed?

So if I reason through that I think we're already covered...unless you have another way that this listener might get deregistered?

breedx-splk: Because it's been this way for a LONG time, and In the interest of moving forward, I'll open an issue for this and we can tackle it in a follow-up effort.

breedx-splk: Ok https://github.com/open-telemetry/opentelemetry-android/issues/1192
I think we might also have a problem with referencing the wrong `Window` instance if/when an `Activity` changes it. Please have a look.

LikeTheSalad: Got it. I didn't realize we've been doing it this way all this time. I think we can move forward with this for the purposes of migrating to the new spec, though we should definitely try to address that new issue asap.

LikeTheSalad (approved):

PR #1179, opened 2025-08-21
Convert Android log substituation function to Kotlin

bidetofevil:
Convert this to Kotlin and add detekt overrides


config/detekt/detekt.yml
```diff
@@ -2,6 +2,19 @@
...
+  LongParameterList:
+    ignoreDataClasses: true
+  TooManyFunctions:
+    thresholdInFiles: 20
+    thresholdInClasses: 20
+    thresholdInInterfaces: 20
+    thresholdInObjects: 20
+    thresholdInEnums: 20
+    ignoreOverridden: true
+
```
breedx-splk: Maybe we should discuss these before configuring them this way at the top level? Can you do baseline instead, like we have done for the other instrumentations until then?

instrumentation/android-log/library/src/main/java/io/opentelemetry/instrumentation/library/log/AndroidLogSubstitutions.kt
```diff
@@ -0,0 +1,267 @@
...
+import android.util.Log
+import io.opentelemetry.api.common.AttributeKey
+import io.opentelemetry.api.common.Attributes
+import io.opentelemetry.api.logs.Severity
+import io.opentelemetry.instrumentation.library.log.internal.LogRecordBuilderCreator.createLogRecordBuilder
+import io.opentelemetry.instrumentation.library.log.internal.LogRecordBuilderCreator.getTypeName
+import io.opentelemetry.instrumentation.library.log.internal.LogRecordBuilderCreator.printStacktrace
+import io.opentelemetry.semconv.ExceptionAttributes
+
+object AndroidLogSubstitutions {
```
breedx-splk, line 17: internal?

config/detekt/detekt.yml
bidetofevil: Right. There are pretty arbitrary. I can pull them out and just ignore any violations for now.

instrumentation/android-log/library/src/main/java/io/opentelemetry/instrumentation/library/log/AndroidLogSubstitutions.kt
bidetofevil: I can do that. It was just public before but internal makes more sense.

instrumentation/android-log/library/src/main/java/io/opentelemetry/instrumentation/library/log/AndroidLogSubstitutions.kt
```diff
@@ -0,0 +1,267 @@
...
+    val tagKey: AttributeKey<String> = AttributeKey.stringKey("android.log.tag")
+
+    @JvmStatic
+    fun substitutionForVerbose(
+        tag: String?,
+        message: String,
```
marandaneto, line 23: not sure if this is for compatibility, but i usually do the mandatory params first, and the optionals after.
usually the afters have a default value like `tag: String? = null` so we dont have to pass anything at all

config/detekt/detekt.yml
breedx-splk: Cool, I'd appreciate that. Whoops, also forgot to discuss at SIG. We can add that for next time.

instrumentation/android-log/library/src/main/java/io/opentelemetry/instrumentation/library/log/AndroidLogSubstitutions.kt
bidetofevil: Actually, never mind - this needs to be public. The code that this replaces (i.e. logcat methods) will be in other modules that won't have access to this at runtime. Even the test, which is in a different module, fails to compile as a result. So I'll keep it as public

instrumentation/android-log/library/src/main/java/io/opentelemetry/instrumentation/library/log/AndroidLogSubstitutions.kt
bidetofevil: The parameter order of these methods are supposed to mirror what it's replacing (e.g. `Log.v`). This is the code that consumes it:

```
MemberSubstitution.relaxed()
                            .method(
                                    is(
                                            Log.class.getDeclaredMethod(
                                                    "v", String.class, String.class)))
                            .replaceWith(
                                    AndroidLogSubstitutions.class.getDeclaredMethod(
                                            "substitutionForVerbose", String.class, String.class))
```

So this is never really called as a Kotlin method directly, so the ergonomics of it isn't relevant.

bidetofevil:
Anytthing else needs to be done here for merging? I hope the answer to the parameter order above is sufficient?

breedx-splk (approved):
LGTM thank you!

PR #1189, opened 2025-08-26
Drop LOLLIPOP support and update robolectric

breedx-splk:
Supersedes #1186.

The [latest robolectric drops support for LOLLIPOP](https://github.com/open-telemetry/opentelemetry-android/pull/1186#issuecomment-3224891849), which I think makes a lot of sense these days. So let's go ahead and drop support for it as well (since we can't test for it after updating). It's less than 1% of the install base according to wikipedia.

Remember that we've deprecated Volley and will be dropping support for it in a few releases, but in the meantime, somehow this change also required apache http core to be on the classpath at test runtime. 🤷🏻

### DIFF
--- a/demo-app/src/main/java/io/opentelemetry/android/demo/OtelDemoApplication.kt
+++ b/demo-app/src/main/java/io/opentelemetry/android/demo/OtelDemoApplication.kt
@@ -7,6 +7,7 @@ package io.opentelemetry.android.demo
 
 import android.annotation.SuppressLint
 import android.app.Application
+import android.os.StrictMode
 import android.util.Log
 import io.opentelemetry.android.OpenTelemetryRum
 import io.opentelemetry.android.agent.OpenTelemetryRumInitializer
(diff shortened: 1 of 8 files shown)


demo-app/src/main/java/io/opentelemetry/android/demo/OtelDemoApplication.kt
```diff
@@ -27,6 +28,17 @@ class OtelDemoApplication : Application() {
     @SuppressLint("RestrictedApi")
     override fun onCreate() {
         super.onCreate()
+        StrictMode.setThreadPolicy(
```
LikeTheSalad: Was this change intended for this PR?

gradle/libs.versions.toml
```diff
@@ -30,6 +30,7 @@ androidx-core = "androidx.core:core:1.17.0"
 androidx-annotation = "androidx.annotation:annotation:1.9.1"
 androidx-lifecycle-process = "androidx.lifecycle:lifecycle-process:2.9.2"
 androidx-preference-ktx = "androidx.preference:preference-ktx:1.2.1"
+apache-httpcore = "org.apache.httpcomponents:httpcore:4.4.16"
```
LikeTheSalad, line 34: Maybe it's worth adding a comment/todo to remove it once we no longer have Volley?

LikeTheSalad (approved):
I left a couple of comments, though overall it looks good to me.

bidetofevil:
Are there any upstream users that support Android 5? Is there anyway to find out?

demo-app/src/main/java/io/opentelemetry/android/demo/OtelDemoApplication.kt
breedx-splk: whoops, no, that was leftover cruft from a different thing. I'll remove it.

breedx-splk:
> Are there any upstream users that support Android 5? Is there anyway to find out?

What do you mean by this? Are you asking if there are users of the java core sdk that are on Android 5/Lollipop? Or are you asking if there are users of `opentelemetry-android` on Android 5?

Either way, I'm not sure this is something we can realistically answer.

fractalwrench (approved):
LGTM

PR #1198, opened 2025-09-02
Initializer instrumentation config

LikeTheSalad:
Enhances the RUM initializer configuration for the default instrumentations by:
- Using a more Kotlin idiomatic approach.
- Reducing the initializer function set of arguments, making it easier to read and extend.

This is an example of how users will be able to configure instrumentations after these changes:

```kotlin
rum = OpenTelemetryRumInitializer.initialize(
    //...
    instrumentations = {
        activity {
            screenNameExtractor(/*some screen name extractor*/)
        }
        slowRenderingReporter {
            detectionPollInterval(10.seconds)
            enableVerboseDebugLogging()
        }
    }
)
```

### DIFF
--- a/android-agent/api/android-agent.api
+++ b/android-agent/api/android-agent.api
@@ -4,22 +4,22 @@ public final class io/opentelemetry/android/agent/OpenTelemetryRumInitializer {
 	public static synthetic fun initialize$default (Landroid/app/Application;Ljava/lang/String;Ljava/util/Map;Lio/opentelemetry/android/agent/connectivity/EndpointConnectivity;Lio/opentelemetry/android/agent/connectivity/EndpointConnectivity;Lio/opentelemetry/android/agent/connectivity/EndpointConnectivity;Lio/opentelemetry/android/config/OtelRumConfig;Lio/opentelemetry/android/agent/session/SessionConfig;Lkotlin/jvm/functions/Function1;ILjava/lang/Object;)Lio/opentelemetry/android/OpenTelemetryRum;
 }
 
-public final class io/opentelemetry/android/agent/OpenTelemetryRumInitializer$ActivityLifecycleConfiguration {
-	public final fun screenNameExtractor (Lio/opentelemetry/android/instrumentation/common/ScreenNameExtractor;)V
-	public final fun tracerCustomizer (Lkotlin/jvm/functions/Function1;)V
+public final class io/opentelemetry/android/agent/OpenTelemetryRumInitializer$ActivityLifecycleConfiguration : io/opentelemetry/android/agent/OpenTelemetryRumInitializer$ScreenLifecycleConfigurable {
+	public fun screenNameExtractor (Lio/opentelemetry/android/instrumentation/common/ScreenNameExtractor;)V
+	public fun tracerCustomizer (Lkotlin/jvm/functions/Function1;)V
 }
 
-public final class io/opentelemetry/android/agent/OpenTelemetryRumInitializer$AnrReporterConfiguration {
-	public final fun addAttributesExtractor (Lio/opentelemetry/android/instrumentation/common/EventAttributesExtractor;)V
+public final class io/opentelemetry/android/agent/OpenTelemetryRumInitializer$AnrReporterConfiguration : io/opentelemetry/android/agent/OpenTelemetryRumInitializer$WithEventAttributes {
+	public fun addAttributesExtractor (Lio/opentelemetry/android/instrumentation/common/EventAttributesExtractor;)V
 }
 
-public final class io/opentelemetry/android/agent/OpenTelemetryRumInitializer$CrashReporterConfiguration {
-	public final fun addAttributesExtractor (Lio/opentelemetry/android/instrumentation/common/EventAttributesExtractor;)V
+public final class io/opentelemetry/android/agent/OpenTelemetryRumInitializer$CrashReporterConfiguration : io/opentelemetry/android/agent/OpenTelemetryRumInitializer$WithEventAttributes {
+	public fun addAttributesExtractor (Lio/opentelemetry/android/instrumentation/common/EventAttributesExtractor;)V
 }
 
-public final class io/opentelemetry/android/agent/OpenTelemetryRumInitializer$FragmentLifecycleConfiguration {
-	public final fun screenNameExtractor (Lio/opentelemetry/android/instrumentation/common/ScreenNameExtractor;)V
-	public final fun tracerCustomizer (Lkotlin/jvm/functions/Function1;)V
+public final class io/opentelemetry/android/agent/OpenTelemetryRumInitializer$FragmentLifecycleConfiguration : io/opentelemetry/android/agent/OpenTelemetryRumInitializer$ScreenLifecycleConfigurable {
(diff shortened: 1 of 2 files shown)


android-agent/api/android-agent.api
```diff
@@ -1,7 +1,50 @@
...
+	public static final fun initialize (Landroid/app/Application;Ljava/lang/String;Ljava/util/Map;Lio/opentelemetry/android/agent/connectivity/EndpointConnectivity;Lio/opentelemetry/android/agent/connectivity/EndpointConnectivity;Lio/opentelemetry/android/agent/connectivity/EndpointConnectivity;Lio/opentelemetry/android/config/OtelRumConfig;Lio/opentelemetry/android/agent/session/SessionConfig;Lkotlin/jvm/functions/Function1;)Lio/opentelemetry/android/OpenTelemetryRum;
+	public static synthetic fun initialize$default (Landroid/app/Application;Ljava/lang/String;Ljava/util/Map;Lio/opentelemetry/android/agent/connectivity/EndpointConnectivity;Lio/opentelemetry/android/agent/connectivity/EndpointConnectivity;Lio/opentelemetry/android/agent/connectivity/EndpointConnectivity;Lio/opentelemetry/android/config/OtelRumConfig;Lio/opentelemetry/android/agent/session/SessionConfig;Lkotlin/jvm/functions/Function1;ILjava/lang/Object;)Lio/opentelemetry/android/OpenTelemetryRum;
+}
+
+public final class io/opentelemetry/android/agent/OpenTelemetryRumInitializer$ActivityLifecycleConfiguration {
+	public fun <init> ()V
```
fractalwrench: Can the constructors on these be made invisible if this forms a public API, given it doesn't make sense for a library user to construct these objects themeselves?

android-agent/api/android-agent.api
```diff
@@ -1,7 +1,50 @@
...
+}
+
+public final class io/opentelemetry/android/agent/OpenTelemetryRumInitializer$AnrReporterConfiguration {
+	public fun <init> ()V
+	public final fun addAttributesExtractor (Lio/opentelemetry/android/instrumentation/common/EventAttributesExtractor;)V
+}
+
+public final class io/opentelemetry/android/agent/OpenTelemetryRumInitializer$CrashReporterConfiguration {
+	public fun <init> ()V
+	public final fun addAttributesExtractor (Lio/opentelemetry/android/instrumentation/common/EventAttributesExtractor;)V
```
fractalwrench: Is it worth considering some sort of interface for this function, and a separate one for `tracerCustomizer`, given they're reused in several places and presumably it makes sense to keep the signature the same?

android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
```diff
@@ -130,62 +110,103 @@ object OpenTelemetryRumInitializer {
...
-        anrAttributesExtractors: List<EventAttributesExtractor<Array<StackTraceElement>>>,
-        crashAttributesExtractors: List<EventAttributesExtractor<CrashDetails>>,
-        networkChangeAttributesExtractors: List<NetworkAttributesExtractor>,
-        slowRenderingDetectionPollInterval: Duration?,
-    ) {
-        val activityLifecycleInstrumentation =
-            getInstrumentation<ActivityLifecycleInstrumentation>()
-        if (activityTracerCustomizer != null) {
-            activityLifecycleInstrumentation?.setTracerCustomizer(activityTracerCustomizer)
+    class InstrumentationConfiguration {
```
fractalwrench: DslMarker might be worth investigating here: https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-dsl-marker/

Right now it's possible to nest the DSL in a way that doesn't make sense e.g.:

```
instrumentations = {
      activity {
          anrReporter {
              fragment {

              }
          }
      }
  }
```

DslMarker could make that fail at compile-time & therefore help IDE discoverability.

LikeTheSalad: This is a great point. I've just added a dslmarker to avoid it.

android-agent/api/android-agent.api
LikeTheSalad: Good catch, I've just hid those constructors.

android-agent/api/android-agent.api
LikeTheSalad: Do you mean an interface that can be shared not only across these agent configs, but also with the underlying instrumentation classes? - And regarding the `tracerCustomizer` one, do you mean having an interface with that function, or having an interface for the function's parameter and avoiding using a higher-order function (like we do with `ScreenNameExtractor`, for example)?

fractalwrench: Here is an example of what I was suggesting for the agent configs. Individual configs such as `ActivityLifecycleConfiguration` could implement the `TracerCustomizer` interface. `screenNameExtractor` and `addAttributesExtractor` are potential candidates for this pattern too.

```kotlin
class ActivityLifecycleConfiguration: TracerCustomizer {
      private val activityLifecycleInstrumentation: ActivityLifecycleInstrumentation by lazy {
          getInstrumentation()
      }

      override fun tracerCustomizer(value: (Tracer) -> Tracer) {
          activityLifecycleInstrumentation.setTracerCustomizer(value)
      }

      fun screenNameExtractor(value: ScreenNameExtractor) {
          activityLifecycleInstrumentation.setScreenNameExtractor(value)
      }
  }

  class FragmentLifecycleConfiguration: TracerCustomizer {
      private val fragmentLifecycleInstrumentation: FragmentLifecycleInstrumentation by lazy {
          getInstrumentation()
      }

      override fun tracerCustomizer(value: (Tracer) -> Tracer) {
          fragmentLifecycleInstrumentation.setTracerCustomizer(value)
      }

      fun screenNameExtractor(value: ScreenNameExtractor) {
          fragmentLifecycleInstrumentation.setScreenNameExtractor(value)
      }
  }

interface TracerCustomizer {
    fun tracerCustomizer(value: (Tracer) -> Tracer)
}
```

I hadn't considered doing the same for the `AndroidInstrumentation` subclasses but it feels like a similar pattern there might be worth exploring?

I don't consider this feedback blocking as it's beyond the scope of the original PR

fractalwrench (approved):
LGTM

android-agent/api/android-agent.api
LikeTheSalad: Got it. Thanks for the details. It's always good to avoid repeating things, so I've created a couple of interfaces for the common methods.

LikeTheSalad:
> I'd be curious to know how users can currently enable/disable individual instrumentations, and whether this API is a good place for doing it?

I missed this yesterday, but yes, the support to disable instrumentations was added [here](https://github.com/open-telemetry/opentelemetry-android/pull/883) not too long ago, and I think this new API could be a good place to use it, maybe by creating an interface with a `disable()` or `suppress()` method to be implemented by all of the config types. IIRC, we discussed at that time to add support for it for the agent's default instrumentations, but it seems like we lost track of that work, so I've just created [this issue](https://github.com/open-telemetry/opentelemetry-android/issues/1206) to keep track of it and add it later.

breedx-splk (approved):
Thanks! This is really nice and expressive on the user-side of things.

PR #1201, opened 2025-09-02
Makes Services class use interfaces rather than concrete implementations

fractalwrench:
## Goal

This changes `Services` to so that an explicit interface is available for each of its properties rather than just a concrete type. This will ultimately allow the use of mocking to be reduced by creating fakes that use the interface instead (and hopefully speed up the tests, as my gut feel is that mock instantiation is taking up most of the test run time).

I'll tackle replacing some of the mocks used by `Services` with fakes in a subsequent PR. I assume we would probably need to create a new module that contains test fakes, given that `Services` is used across various modules.


breedx-splk:
Sorry @fractalwrench I think I upset this when I merged the services close PR. Please give a rebase. 🙇🏻

fractalwrench:
@breedx-splk no problem - I've rebased this.

LikeTheSalad (approved):
I'm a bit nervous about creating too many fakes, which we'd need to keep track of when changing the interfaces. But right now, I cannot see that it might actually become an issue, and services are used in a lot of places, so it might make sense that mocking them could impact the tests. So, I'm up for giving it a try.

fractalwrench:
@LikeTheSalad I understand the nervousness and felt the same when we went through an effort to replace mocks with fakes in Embrace's codebase. I've personally found it much more useful to have test fakes break at a compile-time than have mocks break at run-time (often in a tricky to understand way). We also ran into perf issues when using mocks on certain Android framework classes that were reduced by adding interfaces in an appropriate places.

I'm definitely happy to take it slow with fakes at first & see how folks feel about this approach after a bit of time has passed.

PR #1205, opened 2025-09-03
Add more warning logs in Network detection 

surbhiia:
Add warning logs in Network detection logic when permission or feature check fails.

### DIFF
--- a/services/src/main/java/io/opentelemetry/android/internal/services/network/CarrierFinder.java
+++ b/services/src/main/java/io/opentelemetry/android/internal/services/network/CarrierFinder.java
@@ -37,7 +37,7 @@ public class CarrierFinder {
         if (!hasTelephonyFeature(context)) {
             Log.w(
                     RumConstants.OTEL_RUM_LOG_TAG,
-                    "Cannot determine network carrier details: missing required telephony feature.");
+                    "Cannot determine carrier details: telephony feature missing.");
             return null;
         }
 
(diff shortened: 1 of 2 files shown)


breedx-splk:
These will get emitted every time there is a network change -- do we want to be that verbose, or should we consider some logger that only logs it once?

surbhiia:
> These will get emitted every time there is a network change -- do we want to be that verbose, or should we consider some logger that only logs it once?

I have now made them slightly less verbose as possible. There isn't a logical top level place where we can put this instead such that they are not logged on every network change. I think it's beneficial to log them every time as permissions can be changed later too by the user so these logs will help.

services/src/main/java/io/opentelemetry/android/internal/services/network/CarrierFinder.java
```diff
@@ -35,13 +35,22 @@ public CarrierFinder(@NonNull Context context, @NonNull TelephonyManager telepho
...
         }
 
         try {
-            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P
-                    && hasPhoneStatePermission(context)) {
-                return getCarrierPostApi28();
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+                if (hasPhoneStatePermission(context)) {
+                    return getCarrierPostApi28();
+                } else {
```
breedx-splk, line 48: Prefer omitting the redundant `else`.

breedx-splk (approved):

PR #1210, opened 2025-09-08
Convert ScreenNameExtractor to kotlin

fractalwrench:
## Goal

Converts `ScreenNameExtractor` to Kotlin and separates the interface from its implementation. I've additionally added some extra test cases to confirm that the androidx fragment is supported when extracting screen names, as that's what most folks will be using these days.


instrumentation/common-api/src/main/java/io/opentelemetry/android/instrumentation/annotations/RumScreenName.kt
```diff
@@ -0,0 +1,17 @@
...
+ */
+
+package io.opentelemetry.android.instrumentation.annotations
+
+/**
+ * This annotation can be used to customize the `screen.name` attribute for an instrumented
+ * Fragment or Activity.
+ */
+@Retention(AnnotationRetention.RUNTIME)
+@Target(AnnotationTarget.CLASS)
```
marandaneto, line 13: is that compatible with the former `ElementType.TYPE`?

fractalwrench: Yes. There are a few subtle differences according to the documentation but I don't think they would apply for the most common use-cases, which I'd assume would just be a class or interface.

(`AnnotationTarget.CLASS`)[https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.annotation/-annotation-target/-c-l-a-s-s/] will allow a class, interface, or object. The [docs for `ElementType.TYPE`](https://developer.android.com/reference/java/lang/annotation/ElementType?hl=en#TYPE) state that it will allow a class, interface, enum, or record declaration.

LikeTheSalad (approved):
LGTM!

instrumentation/common-api/src/main/java/io/opentelemetry/android/instrumentation/annotations/RumScreenName.kt
```diff
@@ -0,0 +1,17 @@
...
+annotation class RumScreenName(
+    /** Return the customized screen name.  */
+    val value: String,
+)
```
breedx-splk, line 17: Do we know if migrating this annotation from java to kotlin will break existing java users?

breedx-splk (commented):
This is great, thanks. I would like for us to understand if the annotation change is breaking so that we can call it out.

instrumentation/common-api/src/main/java/io/opentelemetry/android/instrumentation/annotations/RumScreenName.kt
fractalwrench: I don't think this will break existing Java users. There's no breaking change in the binary compatibility validator & the annotation target is equivalent to Java's `ElementType`. Additionally I tested this out locally with a Java class that was annotated with `RumScreenName` and the customised screen name was extracted correctly.

breedx-splk: Works for me, thank you again!

PR #1211, opened 2025-09-09
Setup code coverage tool

fractalwrench:
## Goal

Whilst code coverage isn't a perfect metric it is a useful tool for highlighting when parts of a codebase are missing tests entirely. [Kover](https://github.com/Kotlin/kotlinx-kover/) provides code coverage reports for both Kotlin/Java source code and allows generating either a HTML or XML file.

In this changeset I've applied Kover to all production modules and created an aggregate report that can be accessed with `./gradlew koverHtmlReport`. This report produces a HTML file that can be accessed at `build/reports/kover/html`.

If there was desire to setup code coverage tracking it would be possible to create an XML report using `./gradlew koverXmlReport` and upload it to a service such as codecov (what we use at Embrace). For now I feel it's just useful having this as an option to highlight code coverage locally rather than tracking things continuously.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -16,6 +16,7 @@ buildscript {
 plugins {
     id("otel.spotless-conventions")
     alias(libs.plugins.publishPlugin)
+    id("org.jetbrains.kotlinx.kover")
 }
 
 extra["java_version"] = JavaVersion.VERSION_1_8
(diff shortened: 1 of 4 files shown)


breedx-splk:
Having coverage is great! I think we'd want to be conservative about gating builds on it, but we definitely use `codecov` in the java core repo for PRs to help ensure that coverage isn't significantly declining ([here's an example](https://github.com/open-telemetry/opentelemetry-java/pull/7639#issuecomment-3253156477)) and I think it's nice to demonstrate project maturity by having a badge on the main readme (like java core does). 🤘🏻

breedx-splk (approved):
Love it!

Issue #1187, opened 2025-08-26
Define an updated set of rules in detekt to minimize explicitly ignoring violations


bidetofevil:
The default set of rules and settings used by `detekt` to find code smells is fairly arbitrary (e.g. parameter list length, # of methods in a class, etc.). I find that they are generally pretty conservative, and it makes sense to bump things up if we find that we are ignoring a lot of violations consistently.

Someone should do an initial pass of what's reasonable and propose it so the group can decide if it's reasonable. The goal is to have a reasonable set of rules that we don't need to ignore (i.e. via annotations of in baseline.xml) for the most part.

bidetofevil:
I'll propose a set of rules in a PR and the set of associated fixes and `baseline.xml` cleanup that will hopefully reset the baseline so touching files won't result in a bunch of new violations.

fractalwrench:
@breedx-splk I think this one can be closed now

PR #1212, opened 2025-09-09
Propose custom detekt rules, fix code, and clean up baseline

bidetofevil:
Added a basic set of overrides to the default detekt rule set and removed all the exceptions in the `baseline.xml` files that we don't intend to keep for the long term. This addresses issue #1187.

The gist of the overriding rules set is mostly to raise the threshold of what is considered too complex when it comes to class sizes, function parameters, etc., and well as be less strict with some of the stylistic faux pas's like using "magic numbers" or comments with specific tags. It is a subset of what we use at Embrace, designed to be conservative so as to not preemptively allow behaviours that we don't have and could be considered problematic. Take a look at the diff for `detekt.xml` for more information.

The rule of thumb going forward should be that additions to the `baseline.xml` files should be scrutinized before being added to. It should fall into one of three boxes:

1. The behaviour is OK and we should change the ruleset to allow that
2. The behaviour in this instance is OK (e.g. comment contains a URL that is way too long but that should be there), and we should add it to the `baseline.xml`
3. The behaviour is not OK and the code should be fixed before we merge it.

1 should require a the sign off of a maintainer, which is OK since we need that to merge anyway.

### DIFF
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -7,7 +7,7 @@ complexity:
     constructorThreshold: 15
     functionThreshold: 31
   NestedBlockDepth:
-    threshold: 10
+    threshold: 6
   TooManyFunctions:
     thresholdInFiles: 31
     thresholdInClasses: 31


common/config/detekt/baseline.xml
```diff
@@ -1,7 +1,4 @@
 <?xml version="1.0" ?>
 <SmellBaseline>
   <ManuallySuppressedIssues/>
-  <CurrentIssues>
-    <ID>ForbiddenComment:RumConstants.kt$RumConstants.Events$// TODO: Use the semconv when available</ID>
-  </CurrentIssues>
```
breedx-splk, line 6: So with this change, we can remove this file now entirely can't we?

core/config/detekt/baseline.xml
```diff
@@ -1,19 +1,4 @@
...
-    <ID>ImplicitDefaultLocale:SessionIdRatioBasedSampler.kt$SessionIdRatioBasedSampler$String.format( "SessionIdRatioBased{traceIdRatioBased:%s}", ratioBasedSampler.description, )</ID>
-    <ID>LongParameterList:DiskBufferingConfig.kt$DiskBufferingConfig.Companion$( enabled: Boolean = false, maxCacheSize: Int = DEFAULT_MAX_CACHE_SIZE, maxFileAgeForWriteMillis: Long = TimeUnit.SECONDS.toMillis(30), minFileAgeForReadMillis: Long = TimeUnit.SECONDS.toMillis(33), maxFileAgeForReadMillis: Long = TimeUnit.HOURS.toMillis(18), maxCacheFileSize: Int = MAX_CACHE_FILE_SIZE, debugEnabled: Boolean = false, signalsBufferDir: File? = null, )</ID>
-    <ID>MagicNumber:DiskBufferingConfig.kt$DiskBufferingConfig.Companion$18</ID>
-    <ID>MagicNumber:DiskBufferingConfig.kt$DiskBufferingConfig.Companion$30</ID>
-    <ID>MagicNumber:DiskBufferingConfig.kt$DiskBufferingConfig.Companion$33</ID>
-    <ID>MagicNumber:DiskBufferingConfig.kt$DiskBufferingConfig.Companion$5</ID>
-    <ID>MagicNumber:DiskManager.kt$DiskManager$3</ID>
-    <ID>MagicNumber:SignalFromDiskExporter.kt$SignalFromDiskExporter$5</ID>
-  </CurrentIssues>
```
breedx-splk, line 18: Same question -- can we remove this file now?

core/src/main/java/io/opentelemetry/android/SessionIdRatioBasedSampler.kt
```diff
@@ -47,6 +47,7 @@ class SessionIdRatioBasedSampler(
 
     override fun getDescription(): String =
         String.format(
+            Locale.ENGLISH,
             "SessionIdRatioBased{traceIdRatioBased:%s}",
             ratioBasedSampler.description,
```
breedx-splk: Would it be more idiomatic and simpler just to use a string template instead of `.format()`?

common/config/detekt/baseline.xml
bidetofevil: We can. I can do that instead of trimming them.

core/src/main/java/io/opentelemetry/android/SessionIdRatioBasedSampler.kt
bidetofevil: Yeah actually! I'll make the change.

instrumentation/android-instrumentation/src/main/java/io/opentelemetry/android/instrumentation/internal/AndroidInstrumentationLoaderImpl.kt
```diff
@@ -28,7 +28,7 @@ class AndroidInstrumentationLoaderImpl : AndroidInstrumentationLoader {
 
     @Throws(IllegalStateException::class)
     fun registerForTest(instrumentation: AndroidInstrumentation) {
-        if (instrumentation::class.java in instrumentations) {
+        check(instrumentation::class.java !in instrumentations) {
             throw IllegalStateException("Instrumentation with type '${instrumentation::class.java}' already exists.")
         }
```
breedx-splk: I think you don't need to throw, just return a string from the closure.

```suggestion
fun registerForTest(instrumentation: AndroidInstrumentation) {
    check(instrumentation::class.java !in instrumentations) {
        "Instrumentation with type '${instrumentation::class.java}' already exists."
    }
```

instrumentation/common-api/src/test/java/io/opentelemetry/android/instrumentation/common/ScreenNameExtractorTest.kt
```diff
@@ -42,6 +42,13 @@ internal class ScreenNameExtractorTest {
         Assertions.assertEquals("squarely", name)
     }
 
+    @Test
+    fun testAnnotatedFragment() {
+        val fragment = AnnotatedFragment()
+        val name = ScreenNameExtractor.DEFAULT.extract(fragment)
+        Assertions.assertEquals("bumpity", name)
```
breedx-splk: Might be a nice opportunity to leverage the static import for these as well.

config/detekt/detekt.yml
```diff
@@ -2,9 +2,38 @@
...
+    constructorThreshold: 15
+    functionThreshold: 31
+  NestedBlockDepth:
+    threshold: 10
+  TooManyFunctions:
+    thresholdInFiles: 31
+    thresholdInClasses: 31
+    thresholdInInterfaces: 31
+    thresholdInObjects: 31
+    thresholdInEnums: 31
```
breedx-splk, line 16: Generous! 😁

config/detekt/detekt.yml
```diff
@@ -2,9 +2,38 @@
 build:
   maxIssues: 0
 
+complexity:
+  LongParameterList:
+    constructorThreshold: 15
+    functionThreshold: 31
+  NestedBlockDepth:
+    threshold: 10
```
breedx-splk: I'd like to see this MUCH lower....like 3...but that feels aggressive. :) Can we do 5?

instrumentation/android-instrumentation/src/main/java/io/opentelemetry/android/instrumentation/internal/AndroidInstrumentationLoaderImpl.kt
bidetofevil: Ugh, of course. I was relying on the tests passing, but apparently this wrong way of doing it also gets past detekt lol

config/detekt/detekt.yml
bidetofevil: random

instrumentation/common-api/src/test/java/io/opentelemetry/android/instrumentation/common/ScreenNameExtractorTest.kt
bidetofevil: ugh, copy and paste dumbness

config/detekt/detekt.yml
bidetofevil: nested block depth? yeah lets see what I can put without having to refactor lol

bidetofevil: `ComposeTapTargetDetector` has blocks that go 5 deep, so I'm setting it at 6 for now. If you feel strongly, I can refactor and get this down to 3 or 4.

config/detekt/detekt.yml
breedx-splk: prime!

PR #1227, opened 2025-09-17
Prepare 0.15.0 release

breedx-splk:


CHANGELOG.md
```diff
@@ -2,7 +2,23 @@
 
 ## Unreleased
 
-- Drop support for Lollipop and API level 21.
+### ⚠️⚠️ Breaking changes
+
+- Drop support for Lollipop and API level 21.
```
fractalwrench: Small nit around the wording: from the linked PR it seems like we've just stopped testing against that API level, rather than anything that would actively break users on Lollipop. If we have broken compatibility with Lollipop then it'd be good to state what the minimum supported version is.

breedx-splk: Oh this is a good point...but without the test automation in place it's really hard to claim compatibility with a legacy version (Lollipop or otherwise). I don't think we actually know of or claim a lower limit do we? 

I'm happy to rework the wording here to make it clearer, but do we think this should still be considered a breaking change?

LikeTheSalad: Makes sense, we still state that we support api 21, and I'm not aware of any reason why we would need to drop it right now. So this wording is confusing. If we must mention that we updated our tests, we should probably be more explicit, e.g., "Drop API level 21 Robolectric tests", wdyt, @breedx-splk?

breedx-splk: Yeah ok that sounds good. I'll also open an issue to track the disconnect between test automation min version and what api we claim to support. Thanks for the input!

LikeTheSalad (approved):

PR #1236, opened 2025-09-19
Run codecov on pushes to main

fractalwrench:
## Goal

Runs codecov when commits are pushed to the main branch. This should allow a base commit to be generated, allowing comparison between the coverage on other branches.


breedx-splk (approved):
Lovely! Thank you.

PR #1238, opened 2025-09-19
Add Incubating annotation to indicate unstable APIs

fractalwrench:
## Goal

Adds the `Incubating` annotation which [enforces `RequiresOptIn`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-requires-opt-in/) on any symbol with the annotation. This results in a compiler warning for any consumer of symbols from the library that they must suppress to get rid of the warnings.

I've only added this to `SessionConfig` to demonstrate how the annotation works. `OpenTelemetryRumInitializer` has opted in to the incubating API as it uses the `SessionConfig` symbol.

I haven't added the annotation to anywhere else on the API surface - that's probably something that should be discussed as part of the 1.0.0 roadmap talks. Personally I feel like adding this annotation to any part of the public API surface folks aren't confident will remain the same would be good if it meant that 1.0.0 shipped sooner as a result.

### DIFF
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/Incubating.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/Incubating.kt
@@ -1,0 +1,23 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.android
+
+/**
+ * Marks an API as incubating. Incubating APIs can be subject to breaking change without warning
+ * between versions, and require an opt-in annotation to prevent compiler warnings.
+ *
+ * This annotation is intended for use on opentelemetry-android's API only and should not be used
+ * by library consumers on their own code.
+ */
+@RequiresOptIn(level = RequiresOptIn.Level.WARNING)
+@Retention(AnnotationRetention.BINARY)
+@Target(
+    AnnotationTarget.CLASS,
+    AnnotationTarget.FUNCTION,
+    AnnotationTarget.PROPERTY,
+    AnnotationTarget.TYPEALIAS,
+)
+annotation class Incubating
--- a/android-agent/api/android-agent.api
+++ b/android-agent/api/android-agent.api
@@ -1,3 +1,6 @@
+public abstract interface annotation class io/opentelemetry/android/Incubating : java/lang/annotation/Annotation {
+}
+
 public final class io/opentelemetry/android/agent/OpenTelemetryRumInitializer {
 	public static final field INSTANCE Lio/opentelemetry/android/agent/OpenTelemetryRumInitializer;
 	public static final fun initialize (Landroid/app/Application;Ljava/lang/String;Ljava/util/Map;Lio/opentelemetry/android/agent/connectivity/EndpointConnectivity;Lio/opentelemetry/android/agent/connectivity/EndpointConnectivity;Lio/opentelemetry/android/agent/connectivity/EndpointConnectivity;Lio/opentelemetry/android/config/OtelRumConfig;Lio/opentelemetry/android/agent/session/SessionConfig;Lkotlin/jvm/functions/Function1;)Lio/opentelemetry/android/OpenTelemetryRum;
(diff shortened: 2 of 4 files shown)


fractalwrench:
I've added a comment to the annotation's docs stating that it should only be used on opentelemetry-android's APIs.

@LikeTheSalad I hadn't given the location of this symbol much thought. Ideally I guess it'd be in a core/common module that is accessible by pretty much every module where we'd want to use it, for artefacts that we otherwise consider largely stable. I'm not familiar enough with the project structure to be sure on where that should be: possibly in `common` or `core`? I agree that versioning the whole artifact as alpha is a reasonable way of avoiding the need to annotate _everything_

android-agent/src/main/kotlin/io/opentelemetry/android/Incubating.kt
```diff
@@ -0,0 +1,23 @@
...
+ */
+@RequiresOptIn(level = RequiresOptIn.Level.WARNING)
+@Retention(AnnotationRetention.BINARY)
+@Target(
+    AnnotationTarget.CLASS,
+    AnnotationTarget.FUNCTION,
+    AnnotationTarget.PROPERTY,
+    AnnotationTarget.TYPEALIAS,
+)
+annotation class Incubating
```
marandaneto, line 23: why dont we use https://github.com/androidx/androidx/blob/androidx-main/annotation/annotation-experimental/src/main/java/androidx/annotation/RequiresOptIn.kt instead?

marandaneto: IDEs have better integration with those annotations (provides warnings etc)
Not sure this works for custom annotations

fractalwrench: IDE/compiler warnings should work the same for a custom annotation. I'd be happy with either solution but have a slight preference for a custom annotation as it avoids relying on an external dependency.

marandaneto: i suspect we already depend on the annotations lib indirectly due to https://github.com/open-telemetry/opentelemetry-android/blob/8e97eb97bd60210ae494d4253844b90c510319e9/core/build.gradle.kts#L68
its a transitive dependency isnt it?

marandaneto: https://mvnrepository.com/artifact/androidx.core/core/1.17.0 check compile dependencies

marandaneto: https://kotlinlang.org/docs/opt-in-requirements.html#require-opt-in-to-use-api
TIL `RequiresOptIn`, so Kotlin does the job, and I believe the IDE respects the `RequiresOptIn` then.
I thought the IDE wouldn't know how/when to warn just because you declared a new custom annotation

fractalwrench: You're right, it's already included as a transitive dependency. I still lean towards creating a custom annotation as we could control if/when there are any breaking changes to this annotation, but am happy to go with majority opinion.

LikeTheSalad: Could you confirm if AS provides warnings out of the box with the custom annotation, @fractalwrench? I don't have a super strong opinion on this, though I agree it would be convenient to use a custom one.

LikeTheSalad, line 23: > I've added a comment to the annotation's docs stating that it should only be used on opentelemetry-android's APIs.
> 
> @LikeTheSalad I hadn't given the location of this symbol much thought. Ideally I guess it'd be in a core/common module that is accessible by pretty much every module where we'd want to use it, for artefacts that we otherwise consider largely stable. I'm not familiar enough with the project structure to be sure on where that should be: possibly in `common` or `core`? I agree that versioning the whole artifact as alpha is a reasonable way of avoiding the need to annotate _everything_

Thanks for the clarification, @fractalwrench. I'd suggest keeping it simple then and leaving the annotation here where you added it, but making it `internal`. I also think we shouldn't annotate symbols from `-alpha` versioned artifacts, and since `android-agent` seems to be the only candidate for stable so far, it should be fine to leave the annotation here. If we need to expand this to other modules in the future, we can always move it without issues, as it'd be internal, so nobody should reference it anyway.

fractalwrench: Yep, I can confirm it will show up as an IDE warning and a compiler warning when building on the command line.
 Screenshot 2025-09-26 at 16 58 31

fractalwrench: @LikeTheSalad I'm reasonably sure the annotation needs to be public, as folks will need to access it from outside its original module if they want to opt-out with `@OptIn(Incubating::class)`

breedx-splk: @marandaneto I think in general we like where this is going, and in the interest of progress I think we should merge. Please feel free to open an issue or drop something on the SIG agenda for next week if you would like to continue this discussion. I hope you understand. 👍🏻

PR #1239, opened 2025-09-19
Convert services unit tests to kotlin

fractalwrench:
## Goal

Converts the unit tests in the `services` module from Java to Kotlin. I've kept the logic the same and have made a couple of style tweaks from the auto-conversion.


breedx-splk (approved):

breedx-splk:
Thanks thanks!

Issue #1127, opened 2025-08-04
documentation request: Exporter delegating


Doohl:
### Issue

There is a lack of documentation anywhere on the exporter delegate chain. We could benefit from some documentation somewhere that explains that the default delegate chain.

### Notes 

Taken from @LikeTheSalad (CNCF Slack):

The current delegate order is:

```
BufferDelegating -> DiskBuffering -> OtlpHttp[Span | LogRecord]Exporter
```

- BufferDelegating is an in-memory exporter set to collect telemetry before the agent is fully initialized, which is only used for that time window. See [this PR](https://github.com/open-telemetry/opentelemetry-android/pull/709) for more details.
    - Example: https://github.com/open-telemetry/opentelemetry-android/blob/main/core/src/main/java/io/opentelemetry/android/export/BufferDelegatingLogExporter.kt
- DiskBuffering is the first in the chain, so that everything is stored in disk first.
- OtlpHttp*Exporter is the one where the actual export happens eventually in one of the disk read operations.

Additionally, people can add their own exporter layers using the `add*ExporterCustomizer` builder methods. In which case they decide what's the last exporter in the chain, as DiskBuffering will be on top of whatever is added this way.

marcschaeferger:
@LikeTheSalad Can be closed due to merging the PR #1243

LikeTheSalad:
> [@LikeTheSalad](https://github.com/LikeTheSalad) Can be closed due to merging the PR [#1243](https://github.com/open-telemetry/opentelemetry-android/pull/1243)

Thank you!

PR #1243, opened 2025-09-21
docs(exporter-chain): add exporter chain overview with default

marcschaeferger:
This pull request introduces documentation for the new Exporter Chain feature in the OpenTelemetry Android RUM SDK. The main goal is to explain how the SDK handles asynchronous exporter initialization, buffering, and customization, ensuring reliable telemetry export even during startup or offline scenarios.

**Documentation Additions:**

* Added a new `EXPORTER_CHAIN.md` document detailing the exporter delegation chain, including in-memory and disk buffering, asynchronous initialization, and customization hooks.
* Updated `README.md` to reference the new Exporter Chain feature and documentation, highlighting its purpose and usage.

## Motivation
Fills a gap in documentation around exporter-chain internals, configuration, and extension hooks that was request in [Issue #1127](https://github.com/open-telemetry/opentelemetry-android/issues/1127)
Addresses lack of centralized explanation of buffering layers, initialization order, and customization options.

### DIFF
--- a/docs/EXPORTER_CHAIN.md
+++ b/docs/EXPORTER_CHAIN.md
@@ -1,0 +1,180 @@
+# Exporter Delegate Chain (Spans / Logs / Metrics)
+
+Status: draft
+
+This document explains the default exporter delegation chain used by the OpenTelemetry Android RUM SDK and how you can customize or extend it with your own exporter layers.
+
+## Goals
+
+1. Avoid ANRs and StrictMode violations during SDK initialization by deferring expensive exporter setup off the main thread.
+2. Ensure early telemetry (signals produced before exporters finish initializing) is not lost.
+3. Provide optional durable (disk) buffering for offline / startup scenarios.
+4. Allow applications to insert custom exporters (filtering, redaction, fan‑out, encryption, etc.).
+
+## High-Level Flow
+
+When you call `OpenTelemetryRum.builder()...build()`:
+
+1. Lightweight in‑memory buffer exporters are installed immediately so spans, logs, and metrics can be accepted.
+2. Actual exporter initialization (logging / OTLP + optional disk storage wiring) runs asynchronously on a background executor.
+3. Once ready, buffered signals are flushed to the real exporters and all future exports are delegated directly.
+4. If disk buffering is enabled, signals flow through disk first, then are later read back and exported to the network exporter.
+
+## Default Chain (No Custom Exporters, Disk Buffering Enabled)
+
+For each signal type (Span / LogRecord / Metric):
+
+```text
+BufferDelegating*Exporter  -->  *ToDiskExporter  -->  (Original default exporter)
+    (in-memory buffer)           (writes batch          (e.g. LoggingSpanExporter,
+                                 files to storage)       SystemOutLogRecordExporter,
+                                                         LoggingMetricExporter)
+```text
+
+Later, a periodic scheduler reads batches from disk and replays them on the original exporter via:
+
+```text
+SignalFromDiskExporter -> *FromDiskExporter -> Original exporter
+```
+
+(Where `*` stands for `Span`, `LogRecord`, or `Metric`.)
+
+## Default Chain (Disk Buffering Disabled)
+
+```text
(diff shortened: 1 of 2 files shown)


docs/EXPORTER_CHAIN.md
```diff
@@ -0,0 +1,180 @@
...
+BufferDelegating*Exporter  -->  (Original default exporter)
+```
+
+No disk layer is inserted and no scheduled reader is enabled.
+
+## Components
+
+### BufferDelegating[Span|Log|Metric]Exporter
+
+Internal in‑memory temporary exporters created immediately. They hold up to 5,000 items (per signal type) produced before the real delegate is attached. If the buffer fills, additional items are dropped with a warning log (`The <type> buffer was filled before export delegate set...`). After `setDelegate(...)` is invoked, the buffered data is exported, optional pending `flush()` / `shutdown()` calls are honored, and the buffer is cleared.
```
LikeTheSalad: This is true, though I think it could help adding that, once `setDelegate()` is called and the buffered data is exporter, no more data will get buffered in disk from that point onwards, as it will be delegated right away to the "Original exporter" (or the *ToDiskExporter in case disk buffering is enabled).

docs/EXPORTER_CHAIN.md
```diff
@@ -0,0 +1,180 @@
...
+  .addLogRecordExporterCustomizer(exp -> SpanAttributeRedactingLogExporter.wrap(exp))
+  .addMetricExporterCustomizer(exp -> myMetricsFanOut(exp))
+  .build();
+```
+
+Customizer semantics:
+
+* Each customizer receives the current exporter and must return the exporter to use next.
+* Multiple customizers compose in order of registration (first added runs first in the chain).
+* Disk buffering (if enabled) is inserted AFTER your customizers and BEFORE buffering to disk.
```
LikeTheSalad: I'm finding this sentence hard to follow. We're talking about customizers here, it's true that the customizers are applied in the order they are registered; however, in the end if all customizers wrap the exporter provided by their predecessor, it means that the last exporter (wrapper) will be the first in the chain of exporting. So, even though disk buffering exporters are added after the ones provided by customizers, the data will first pass through them, and then it will be delegated from disk to the other custom exporters.

In fact, disk buffering isn't added via an exporter customizer, all the customizers are evaluated first to create the "Original exporter", then if disk buffering is enabled, the disk buffering exporters would wrap that exporter, so that the data is stored in disk first, and then the disk buffering exporter is wrapped by the bufferdelegating* ones to buffer the first items in memory.

In summary, maybe a more practical way to describe the role of disk buffering and BufferDelegating* exporters in relation to exporters provided via customizers would be something like:

_"Once an exporter is built and has gone through all the registered customizers, the end result will be wrapped by a disk buffering exporter (if enabled) and then finally wrapped by a BufferDelegating* one"_

docs/EXPORTER_CHAIN.md
```diff
@@ -0,0 +1,180 @@
...
+1. Buffered data is exported
+2. A flush is issued if it was pending
+3. A shutdown is issued if it was pending
+4. Pending futures complete with the real delegate result
+
+## Failure and Backpressure Characteristics
+
+* Buffer Overflow: If more than 5,000 signals of a type are produced before delegate attachment, newer signals beyond capacity are dropped (a warning is logged). Consider reducing startup emission volume or initializing earlier if this occurs.
+* Disk Layer: If disk initialization fails, the SDK logs an error and proceeds WITHOUT disk buffering (the chain reverts to memory buffer -> original exporter). The scheduled disk reader is disabled in this case.
+* From-Disk Export Failures: Batches that fail to export remain on disk (subject to age / size pruning rules governed by `DiskBufferingConfig`).
```
LikeTheSalad, line 143: It's also probably worth adding another item for "To-Disk Export failures", which happens when, even though disk buffering is enabled and successfully initialized, for some reason the data cannot be stored on disk when created. The current behavior in those cases is to try to delegate that data to the "original exporter" right away.

marcschaeferger:
@LikeTheSalad Applied all your Suggestion

Issue #1024, opened 2025-06-24
Add `instrumentation-all` package


cleverchuk:
An `instrumentation-all` package will make it easy for users to try-out what we have to offer without having to chase down individual instrumentation packages. This package will contain all available instrumentation and any new instrumentation that's added.

breedx-splk:
One challenge that [@LikeTheSalad brought up in #1254](https://github.com/open-telemetry/opentelemetry-android/pull/1254#pullrequestreview-3280455140) is that some instrumentations provide build-time bytecode weaving, which won't be covered by a single `instrumentation-all` module.  Just something to consider.

PR #1254, opened 2025-09-26
feat(instrumentation-all): introduce aggregate dependency bundling all instrumentation ([#1024](https://github.com/marcschaeferger/opentelemetry-android/issues/1024))

marcschaeferger:
## Pull Request: Add `instrumentation:all` Gradle Module  

**Issue:** [#1024 – Add instrumentation-all package](https://github.com/open-telemetry/opentelemetry-android/issues/1024)  

---

### Summary
This PR adds a new Gradle module **`instrumentation:all`** that aggregates all end-user Android instrumentation libraries into a single dependency for quick evaluation and experimentation. It is designed to help users explore OpenTelemetry Android's full feature set without having to manage multiple individual dependencies. 

Issue requested an "`instrumentation-all`" package to lower the onboarding friction when evaluating OpenTelemetry Android.  
By offering a single aggregate module, developers can quickly test features for demos, Proof-of-Concepts, and exploratory instrumentation without manually tracking individual modules, making adoption and initial evaluation significantly simpler.  

All Suggestions and changes are welcome ;-)

---

### What's Included
- **New module:** `instrumentation:all`
- **Dependencies:**  
  - *Instrumentation:*
    - activity
    - anr
    - crash
    - fragment
    - network
    - slowrendering
    - sessions
    - startup
    - view-click
    - compose:click  
  - *Libraries:*  
    - okhttp3  
    - okhttp3-websocket  
    - httpurlconnection  
    - android-log  
  - *Foundational:*  
    - common-api  
    - android-instrumentation  
- **Intentional exclusions:** Agent and testing variants (`*:agent`, `*:testing`).  
- **Documentation:**  
  - New **README** inside the `instrumentation:all` module explaining usage, scope, production caveats, and maintenance expectations.  
  - Update to **root README** with a brief mention and pointer to `instrumentation:all` (full install instructions omitted, per guidance) and also fixing small markdown violation.

---

✅ **Closes:** #1024

### DIFF
--- a/instrumentation/all/build.gradle.kts
+++ b/instrumentation/all/build.gradle.kts
@@ -7,7 +7,7 @@ description = "OpenTelemetry Android aggregate: all available instrumentation fe
 
 android {
     namespace = "io.opentelemetry.android.instrumentation.all"
-    
+
     defaultConfig {
         // No consumer rules; this module just aggregates others.
     }


breedx-splk:
Hmmm, do we think that the existing of build-time bytecode instrumentation somehow defeats the utility+convenience of this module, though? Some users will just take the runtime instrumentation or use the wrappers, and it also shouldn't prevent users who want to use the bytebuddy gradle plugin from also adding it, right?

fractalwrench:
A bit of a tangent, but I think a gradle plugin that performs auto-instrumentation should be a long-term goal of the project. It would make it much simpler to install instrumentation at the appropriate place.

LikeTheSalad:
> Hmmm, do we think that the existing of build-time bytecode instrumentation somehow defeats the utility+convenience of this module, though?

I think it depends on the user's expectations. If they look at the `instrumentation-all` dependency as a one-liner to get all instrumentations fully working, then I think it does more harm than good, as we'll get a lot of false-positive bug issues opened due to the confusion.

> and it also shouldn't prevent users who want to use the bytebuddy gradle plugin from also adding it, right?

That's true, but adding the bytebuddy plugin alone won't make the instrumentations work. To make a single instrumentation work (that relies on bytecode weaving), there has to be a combination of the bytebuddy plugin + agent dependency + runtime dependency. And my understanding is that this `instrumentation-all` dependency only provides the runtime dependency. So to make everything work, they'll have to add the butebuddy plugin, plus each one of the `byteBuddy` dependencies from every instrumentation that relies on bytecode weaving.

PR #1265, opened 2025-09-29
Convert unit tests to kotlin

fractalwrench:
## Goal

Converts some additional unit tests to Kotlin.


instrumentation/common-api/src/main/java/io/opentelemetry/android/instrumentation/common/EventAttributesExtractor.kt
```diff
@@ -15,7 +15,7 @@ import io.opentelemetry.context.Context
  * at a point in time. The resulting Attributes will all be added to the event
  * that is being generated.
  */
-interface EventAttributesExtractor<SUBJECT> {
+fun interface EventAttributesExtractor<SUBJECT> {
```
breedx-splk, line 18: oh this is cool, never thought about it!

breedx-splk (approved):
LGTM thank you!

PR #1266, opened 2025-09-29
Convert fragment/activity test cases to kotlin

fractalwrench:
## Goal

Converts the activity/fragment lifecycle test cases to Kotlin.


breedx-splk:
Even if this is just conversions, this PR is starting to hit a size/review size challenge.

fractalwrench:
My bad - I hadn't realised it had got quite that big. I'll close it and split it into separate PRs for the activity/fragment conversions, which should be more manageable?

Issue #1257, opened 2025-09-26
DEVELOPER FEEDBACK: android-agent 1.x stable APIs


breedx-splk:
Hello Android developers! We encourage you to try this project in anger and to specifically provide us with feedback about the use of the `android-agent` and the `OpenTelemetryRumInitializer` APIs.

Our intention is to mark these APIs "stable" with a 1.x release in the near future, and your feedback will help inform this decision.  ❤️ 

Please keep this in mind:

* The design of the `android-agent` is not to solve every odd/bespoke/specific use case. We do hope it will satisfy 90% of most common usages.
* Where more complicated, nonstandard configurations are required, the APIs in the `core` module MAY be used. Those APIs remain subject to change. We hope we won't need to discuss those (much?) here. Let's focus on the agent.

What do you like? What would make using it easier? Do you have a use case that's not covered that you think should be?

Please leave us a comment below. Thanks!

svrnm:
Can I suggest that having https://opentelemetry.io/docs/platforms/client-apps/android/ filled with content to be a 1.0 requirement?:-) cc @open-telemetry/docs-approvers

fractalwrench:
I agree that making the `android-agent` module stable in the near future is a good goal, and the general idea should be to produce a very minimal API that can have features added on later. I have a few additional suggestions (some of which might have already come up in the SIG meetings)

1. All public APIs should be in Kotlin (I notice that `OtelRumConfig` is currently in Java, and there are probably a few others)
2. We should probably assess the public API via the api dump files, and reduce the visibility of anything that isn't needed, either on a class or symbol level (e.g. I notice public constructors)
3. The `android-agent` module exposes several dependencies as `api` rather than `implementation`. Do we count those as part of the public API, or could they be hidden?
4. There are several concrete implementations exposed as part of the public API, e.g. `SessionStorage.InMemory`. Is it worth hiding these behind some interface/enum/alternative approach that would buy us flexibility to change the implementation in future?
5. We should document some strategy for API stability _after_ 1.0. I think it's inevitable that we'll be stuck with at least one API that doesn't quite feel right, or needs to change due to requirements shifting, so it'd be good to explicitly mention how breaking changes would be avoided whilst allowing the API surface to evolve

LikeTheSalad:
[...]

Issue #668, opened 2024-10-29
SDK global debug mode configuration


marandaneto:
Re https://github.com/open-telemetry/opentelemetry-android/pull/664#discussion_r1820593520
Relates to https://github.com/open-telemetry/opentelemetry-android/issues/597

Eg:

`Log(tag, level, message, args)`

You can configure the `minLevel` for filtering such as `error` during SDK init.

```
debugConfig {
	enabled: false,
    minLevel: error
}
```

We'd call Log.log('DiskBufferingConfiguration', ErrorLevel.Error, 'The error is: %s', error) and this will call `Logcat` internally.
If `enabled` is enabled and if `ErrorLevel` is >= `minLevel`, we'd call logcat, otherwise not.

PR #1272, opened 2025-09-30
Removing OTelRumConfig from initializer

### LINKED ISSUE #991: Enabling disk-buffering in the agent by default
This is a proposal to enable disk-buffering out of the box for the [agent initialization](https://github.com/open-telemetry/opentelemetry-android/blob/32fd8e1681dde7a0fc91381c047320bed0946c8e/android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt#L58). The agent initialization is supposed to contain all of our "opinions" on what parts of core/RUM should be enabled and how they should be configured so that the agent behaves well for most of the end-user use cases without having to write much code.

Storing signals on disk is crucial to avoid data loss, especially in mobile environments, so it makes sense to me that this feature should be enabled by default as part of the [agent initialization](https://github.com/open-telemetry/opentelemetry-android/blob/32fd8e1681dde7a0fc91381c047320bed0946c8e/android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt#L58). There might be reasons why disk buffering should be disabled in some cases, although my understanding is that those would be exceptional ones. Therefore, should we decide to enable it by default in the agent, we should include a config option to optionally disable it.

I'm looking forward to any opinions on this proposal before creating a PR to apply the changes to enable it by default, in case I missed something important that requires it to be kept disabled by default.

LikeTheSalad:
Related to #1257 
Closes #991

### DIFF
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
@@ -48,8 +48,8 @@ object OpenTelemetryRumInitializer {
      * @param spanEndpointConnectivity Span-specific endpoint configuration.
      * @param logEndpointConnectivity Log-specific endpoint configuration.
      * @param metricEndpointConnectivity Metric-specific endpoint configuration.
-     * @param globalAttributes Configures the set of global attributes to emit with every span and event.
      * @param sessionConfig The session configuration, which includes inactivity timeout and maximum lifetime durations.
+     * @param globalAttributes Configures the set of global attributes to emit with every span and event.
      * @param diskBuffering Configures the disk buffering feature.
      * @param instrumentations Configurations for all the default instrumentations.
      */


fractalwrench:
Does #1273 cover everything that was previously possible to configure via `OtelRumConfig`, or are there some APIs that we're happy to jettison as part of this change?

LikeTheSalad:
> Does #1273 cover everything that was previously possible to configure via `OtelRumConfig`, or are there some APIs that we're happy to jettison as part of this change?

The latter. My goal with these changes is to make the initializer API as lean as possible so that we can add stuff later as needed, and to minimize deprecations in the near future, in case some of the APIs we expose right away don't hold up to real-world use cases that people can find later. OtelRumConfig includes several low-level config options that I don't think it would be wise to release upfront, though some of them are needed for basic usage afaik, such as providing global attributes, which I think we can add via the initializer's parameters. Right now, I just added `globalAttributes` (and enabled disk buffering by default), though I'm open to adding any other config option from OtelRumConfig right now that someone else deems necessary for basic usage.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
```diff
@@ -72,13 +73,18 @@ object OpenTelemetryRumInitializer {
...
-        rumConfig: OtelRumConfig = OtelRumConfig(),
         sessionConfig: SessionConfig = SessionConfig.withDefaults(),
+        globalAttributes: (() -> Attributes)? = null,
         instrumentations: (InstrumentationConfiguration.() -> Unit)? = null,
     ): OpenTelemetryRum {
         instrumentations?.let { configure ->
             InstrumentationConfiguration().configure()
         }
+        val rumConfig = OtelRumConfig()
+        rumConfig.setDiskBufferingConfig(DiskBufferingConfig.create(enabled = true))
```
breedx-splk: At minimum, I think we need to expose a way to disable disk buffering and a way to turn on/off debugging for it (although that could be tied to a broader debug setting if one existed).

LikeTheSalad: Fair enough 👍 

I was thinking that it might be nice to reuse the same enabling/disabling config mechanism proposed for instrumentations [here](https://github.com/open-telemetry/opentelemetry-android/pull/1273) for diskbuffering as well. If you agree, then I'd like to hold this PR until that other one is merged, to do a rebase here and use the same config interface.

breedx-splk: Yeah, I like that approach and think it would apply nicely here too. If you would like to do it as a follow-up PR, please open a tracking issue so that we don't lose this. Otherwise, this PR should be good. ❤️

LikeTheSalad: Thank you. The changes aren't that much, so I've pushed them to this PR. I've added a DSL-like config for disk buffering that, for now, only allows enabling/disabling it, like so:

```kotlin
OpenTelemetryRumInitializer.initialize(
    diskBuffering = {
        enabled(false) // true by default
    }
)
```

Regarding the debug mode, in case we think it's helpful, I think it'd be better to have a broader debug setting. I was about to create an issue for it until I found [this one](https://github.com/open-telemetry/opentelemetry-android/issues/668), which I think is what we're looking for. FWIW, currently the debug mode for disk buffering means that some info level logs are only printed when debug = true, however, in my recent changes to disk buffering, I removed that flag and changed the log levels to a finer scope, which will make the logs only show up when the host app's JUL config is set to show those logs, which I guess will be when the app is running in "debug mode" in general, but it won't be the default. This change aligns disk buffering's logs with the rest of OTel Java related repos. If we wanted to provide a "debug mode", I think we need to decide what to do with all the OTel Java related repos' logs too, so it seems to me that this would require discussing it further, probably in a SIG meeting.

fractalwrench (approved):
LGTM

marandaneto (approved):

PR #1273, opened 2025-09-30
Config for disabling default instrumentations

### LINKED ISSUE #1206: Adding config to suppress default agent instrumentations via the RUM initializer
This is to extend the RUM initializer config to allow suppressing a specific instrumentation from the ones provided by default within the agent.

LikeTheSalad:
Closes #1206
Related to #1257 

Allows to disable instrumentations like so:

```kotlin
val rum = OpenTelemetryRumInitializer.initialize(
    instrumentations = {
        activity { 
            disable()
        }
     }
)
```

### DIFF
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -57,6 +57,7 @@ public final class io/opentelemetry/android/SessionIdRatioBasedSampler : io/open
 
 public class io/opentelemetry/android/config/OtelRumConfig {
 	public fun <init> ()V
+	public fun allowInstrumentation (Ljava/lang/String;)Lio/opentelemetry/android/config/OtelRumConfig;
 	public fun disableInstrumentationDiscovery ()Lio/opentelemetry/android/config/OtelRumConfig;
 	public fun disableNetworkAttributes ()Lio/opentelemetry/android/config/OtelRumConfig;
 	public fun disableScreenAttributes ()Lio/opentelemetry/android/config/OtelRumConfig;


fractalwrench (approved):
LGTM. I have a very minor preference for syntax like `{ enabled = false }` instead but don't particularly mind as long as there's a way to disable things.

breedx-splk:
> LGTM. I have a very minor preference for syntax like `{ enabled = false }` instead but don't particularly mind as long as there's a way to disable things.

I think I agree, but it's a very minor preference. One downside of that approach is that it also allows for `{ enabled = true }` which is the default state...so effectively a no-op. The approach here at least prevents that, so I don't think it's worth changing. Curious about what other people think too.

fractalwrench:
>One downside of that approach is that it also allows for { enabled = true } which is the default state

That does become an advantage if we ever ship instrumentation that is off by default. I don't think that's currently the case for anything but seems a plausible option? So we'd either need this or an `enable()` API too

breedx-splk:
> > One downside of that approach is that it also allows for { enabled = true } which is the default state
> 
> That does become an advantage if we ever ship instrumentation that is off by default. I don't think that's currently the case for anything but seems a plausible option? So we'd either need this or an `enable()` API too

Yeah, fair point. The `{ enabled = x }` approach is more complete/symmetric.

LikeTheSalad:
> Yeah, fair point. The { enabled = x } approach is more complete/symmetric.

It seems like we've decided to allow enabling and disabling with the same API, so I've applied those changes, but there are some caveats that I had to solve when working on the changes:

* The syntax `enabled = true/false` would require a field for the common interface, which would in turn require overriding its `getter` accessor, which is not needed for this use case, as we only need to set a boolean. So I went with an `enabled(Boolean)` function instead.
* The core API involved (OtelRumConfig) doesn't currently allow "enabling" or "unsuppressing" instrumentations, only suppressing them, as by default, instrumentations are gotten via SPIs, so as long as they are in the runtime classpath (and not in the `suppressed` list), they'll get used. So in order to fully implement the `enabled(Boolean)` function, I had to add a way to "unsuppress" instrumentations into `OtelRumConfig`, for when the value provided is `true`.

I don't have strong opinions on this topic. I mention these details in case any of you have some further feedback on the recent changes. @breedx-splk @fractalwrench

fractalwrench (approved):
LGTM.

PR #1275, opened 2025-10-01
Add functional interfaces to support config DSL

breedx-splk:
Found two places where a functional interface slightly improves the writability of `OpenTelemetryRumInitializer` configuration.


fractalwrench (approved):
LGTM

PR #1278, opened 2025-10-02
Hide symbols from public api

fractalwrench:
## Goal

This changeset hides symbol from the public API exposed in `agent-api`. `SessionStorage` and `SessionIdGenerator` don't appear to be usable for an external consumer of the library, so IMO it makes sense to hide the concepts for now. I additionally separated out the implementations from the interfaces.

### DIFF
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionIdGenerator.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionIdGenerator.kt
@@ -5,18 +5,6 @@
 
 package io.opentelemetry.android.agent.session
 
-import io.opentelemetry.api.trace.TraceId
-import java.util.Random
-
-interface SessionIdGenerator {
+internal fun interface SessionIdGenerator {
     fun generateSessionId(): String
-
-    object DEFAULT : SessionIdGenerator {
-        override fun generateSessionId(): String {
-            val random = Random()
-            // The OTel TraceId has exactly the same format as a RUM SessionId, so let's re-use it here,
-            // rather than re-inventing the wheel.
-            return TraceId.fromLongs(random.nextLong(), random.nextLong())
-        }
-    }
 }
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionIdTimeoutHandler.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionIdTimeoutHandler.kt
@@ -5,6 +5,7 @@
 
 package io.opentelemetry.android.agent.session
 
+import io.opentelemetry.android.Incubating
 import io.opentelemetry.android.internal.services.applifecycle.ApplicationStateListener
 import io.opentelemetry.sdk.common.Clock
 import kotlin.time.Duration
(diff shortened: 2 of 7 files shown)


android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionIdTimeoutHandler.kt
```diff
@@ -34,6 +35,7 @@ internal class SessionIdTimeoutHandler(
     private var state = State.FOREGROUND
 
     // for testing
+    @OptIn(Incubating::class)
```
LikeTheSalad, line 38: Is this needed, given that the constructor is internal?

fractalwrench: Yes - `@Incubating` propagates so that any usage of an annotated symbol will require an opt-in, regardless of visibility. This suppresses a couple of compiler warnings that got introduced in the PR that added it. The alternative is to remove the annotation from `SessionConfig`

android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionIdGenerator.kt
```diff
@@ -5,18 +5,6 @@
 
 package io.opentelemetry.android.agent.session
 
-import io.opentelemetry.api.trace.TraceId
-import java.util.Random
-
-interface SessionIdGenerator {
+internal fun interface SessionIdGenerator {
```
breedx-splk, line 8: Oh I think this was public on purpose. There have been users who previously wanted to provide their own session id values....

android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionStorage.kt
```diff
@@ -7,18 +7,8 @@ package io.opentelemetry.android.agent.session
 
 import io.opentelemetry.android.session.Session
 
-interface SessionStorage {
+internal interface SessionStorage {
```
breedx-splk, line 10: Same comment -- I think this was visible because some users wanted to be able to provide their own storage mechanism?

android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionIdGenerator.kt
fractalwrench: I don't see any public-facing API that allows setting a custom `SessionIdGenerator` or `SessionStorage` - what's the expected usage?

LikeTheSalad: It's not currently something users can provide via the initializer. The options we have are either hide them until needed, as proposed with these changes, or expose them right away, maybe via the `SessionConfig` object or the like. I'm fine either way tbh, though if we expose them right away, it's probably good to mark them all as incubating. Wdyt, @breedx-splk?

breedx-splk: Oh it just sounds like I'm mistaken, or maybe I'm thinking about the `setSessionProvider()` method on the OTRB, which is still there for folks to use if they want to bypass the agent and use core directly.

PR #1279, opened 2025-10-02
Convert OpenTelemetryRum interface to kotlin

fractalwrench:
## Goal

Converts the `OpenTelemetryRum` interface to Kotlin so that we can get to a point where all APIs in the `core` module are Kotlin.

## Testing

Largely relied on existing testing and API checks, although I added a small unit test for the noop scenario.


core/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
```diff
@@ -0,0 +1,137 @@
...
+     * instance is using for instrumentation.
+     */
+    val openTelemetry: OpenTelemetry
+
+    /**
+     * Get the client session ID associated with this instance of the RUM instrumentation library.
+     * Note: this value will change throughout the lifetime of an application instance, so it is
+     * recommended that you do not cache this value, but always retrieve it from here when needed.
+     */
+    fun getRumSessionId(): String
```
breedx-splk, line 30: Isn't it more consistent to also implement this as a `val` like the `openTelemetry` above?

core/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
```diff
@@ -0,0 +1,137 @@
...
+     *
+     * @param eventName The name of the event to emit.
+     * @param body The body of the event, typically containing additional data.
+     * @param attributes The attributes associated with the event, providing metadata.
+     */
+    fun emitEvent(
+        eventName: String,
+        body: String,
+        attributes: Attributes,
+    )
```
breedx-splk, line 45: I think all 3 of these `emitEvent` methods can be collapsed into a single method with defaults... `body = ""` and `attributes = Attributes.empty()`...probably with `@JvmOverloads` or whatever for backwards compatibility?

core/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
fractalwrench: Semantically I felt it was better to implement as a function as it indicates the value can change through an application lifecycle, whereas the `opentelemetry` property cannot change. I can change it if that's still desired?

core/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
fractalwrench: I've collapsed this into a single method with defaults. Unfortunately `@JvmOverloads` isn't allowed on functions within interfaces, so folks would need to use named parameters if they wanted to omit `body` but supply `attributes`.

core/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
LikeTheSalad: +1 on the consistency. I actually prefer having all functions on interfaces, even though there's no reason why not using `val` for Kotlin interfaces, maybe I'm just old 😅, though I do like the idea of keeping consistency in general.

LikeTheSalad (approved):
Thanks! Added a small comment on one of Jason's, but I'm fine with whatever option you guys decide to go with.

core/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
fractalwrench: I've updated the interface so that it's implemented as `getOpenTelemetry()` rather than as a property.

PR #1280, opened 2025-10-02
Remove all usages of ExtendedLog*

breedx-splk:
This includes:

* `ExtendedLogger`
* `ExtendedLogRecordBuilder`
* `ExtendedLogRecord`
* `ExtendedLogRecordData`

We originally used these while the event name field on logs was incubating, but it's stable now and these extended usages are superfluous.

### DIFF
--- a/demo-app/src/main/java/io/opentelemetry/android/demo/OtelDemoApplication.kt
+++ b/demo-app/src/main/java/io/opentelemetry/android/demo/OtelDemoApplication.kt
@@ -8,14 +8,15 @@ package io.opentelemetry.android.demo
 import android.annotation.SuppressLint
 import android.app.Application
 import android.util.Log
+import io.opentelemetry.android.Incubating
 import io.opentelemetry.android.OpenTelemetryRum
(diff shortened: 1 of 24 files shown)


demo-app/src/main/java/io/opentelemetry/android/demo/OtelDemoApplication.kt
```diff
@@ -81,9 +82,11 @@ class OtelDemoApplication : Application() {
         }
 
         fun eventBuilder(scopeName: String, eventName: String): LogRecordBuilder {
-            val logger = rum?.openTelemetry?.logsBridge?.loggerBuilder(scopeName)?.build()
-            var builder: ExtendedLogRecordBuilder = logger?.logRecordBuilder() as ExtendedLogRecordBuilder
-            return builder.setEventName(eventName)
+            if(rum == null){
```
fractalwrench:
```suggestion
            if (rum == null) {
```

fractalwrench (approved):
LGTM

marandaneto (approved):
noice

LikeTheSalad (approved):

PR #1281, opened 2025-10-02
feat(instrumentation): Add FirstDraw spans to activity instrumentation

### LINKED ISSUE #1143: First draw instrumentation
It would be nice to have signals on how long it takes for Activities (and Fragments?) to become interactive for the user. The [Android vitals](https://developer.android.com/topic/performance/vitals/launch-time#time-full) documentation specs out two vitals:
- [TTID](https://developer.android.com/topic/performance/vitals/launch-time#time-initial) - Time to initial display (AKA "time to first draw"), which is the time it takes to display the first frame of the UI
- [TTFD](https://developer.android.com/topic/performance/vitals/launch-time#time-full) - Time to full display, which is the time it takes for the app to become fully interactive for the user.

TTFD depends on the app reporting when a component's fully drawn (all async fetches resolved, etc): https://developer.android.com/reference/androidx/activity/ComponentActivity#reportFullyDrawn()

TTID is much easier to automatically instrument within our existing Activity lifecycle. We can rely on [ViewTreeObserver.addOnDrawListener](https://developer.android.com/reference/android/view/ViewTreeObserver#addOnDrawListener(android.view.ViewTreeObserver.OnDrawListener)) to report on Activity first draw. We can remove the listener after the first draw to avoid reports on subsequent draws, since the first draw is the most important.

Doohl:
fixes #1143

## What is this change?

This PR adds instrumentation to track the initial draw time of Android activities. A new `FirstDraw` span is created as a child of the activity creation span (e.g., `AppStart` or `Created`), measuring the time from activity creation until the first frame is rendered on screen.

The implementation:
- Hooks into Android's `ViewTreeObserver.OnDrawListener` to detect when the first draw occurs
- Handles Android API < 26 bug where draw listeners weren't properly merged into the view tree observer
- Captures screen complexity metrics (view node count and depth) when the draw completes

### 1. Screen view complexity attributes

Two new attributes are added to the `FirstDraw` span to capture screen complexity:

- `screen.view.nodes` (long): Total count of all View nodes in the view hierarchy
- `screen.view.depth` (long): Maximum depth of the view hierarchy

We believe these metrics will add some value if users try to debug long `FirstDraw` spans. Feedback here would be appreciated, though!

### 2. FirstDraw value

`FirstDraw` spans aren't a perfect signal for when Activities become 'interactable'. Rather, this can help users identify issues with the UI pipeline. For that, users would have to manually instrument spans or events that fire off when the app or activity enters a state where the end user can actually start interacting with the app.

### 3. FirstDraw span lifecycle

`FirstDraw` spans are children of the `Created` spans. However, `FirstDraw` span can end long after the parent span has ended. This can create a confusing tracing experience, but we probably don't want to change the definition of the existing `Created` span.

**Does it make sense to keep FirstDraw as a child of Created?**

## How was this tested?

Tested some scenarios with the demo app (API versions 28 and 30) to validate the correctness of the instrumentation. If you have any ideas how this could be better tested, feel free to opine!

image

image

Raw trace: https://www.codebin.cc/code/cmg9yjmxb0001jz037mtuj2v0:HXhjtG6XV3Vhfh8Ewy6r1S6KfapJtmgM1S1beucxBmpG

### DIFF
--- a/instrumentation/activity/src/test/java/io/opentelemetry/android/instrumentation/activity/ActivityTracerTest.kt
+++ b/instrumentation/activity/src/test/java/io/opentelemetry/android/instrumentation/activity/ActivityTracerTest.kt
@@ -230,6 +230,62 @@ class ActivityTracerTest {
         assertEquals("squarely", span.attributes.get(RumConstants.SCREEN_NAME_KEY))
     }
 
+    @Test
+    fun endInitialDrawSpan_whenSpanExists() {
+        val trackableTracer =
+            ActivityTracer
+                .builder(Mockito.mock(Activity::class.java))
+                .setTracer(tracer)
+                .setAppStartupTimer(appStartupTimer)
+                .setActiveSpan(activeSpan)
+                .build()
+
+        trackableTracer.startActivityCreation()
+        trackableTracer.endInitialDrawSpan()
+        trackableTracer.endActiveSpan()
+
+        val spans = otelTesting.spans
+        assertEquals(2, spans.size)
+        assertEquals("FirstDraw", spans[0].name)
+        assertEquals("Created", spans[1].name)
+    }
+
+    @Test
+    fun endInitialDrawSpan_whenSpanDoesNotExist() {
+        val trackableTracer =
+            ActivityTracer
+                .builder(Mockito.mock(Activity::class.java))
+                .setTracer(tracer)
+                .setAppStartupTimer(appStartupTimer)
+                .setActiveSpan(activeSpan)
+                .build()
+
+        trackableTracer.endInitialDrawSpan()
+
+        assertEquals(0, otelTesting.spans.size)
+    }
+
+    @Test
+    fun endInitialDrawSpan_multipleCallsOnlyEndsOnce() {
+        val trackableTracer =
+            ActivityTracer
+                .builder(Mockito.mock(Activity::class.java))
+                .setTracer(tracer)
+                .setAppStartupTimer(appStartupTimer)
+                .setActiveSpan(activeSpan)
+                .build()
+
+        trackableTracer.startActivityCreation()
+        trackableTracer.endInitialDrawSpan()
+        trackableTracer.endInitialDrawSpan()
+        trackableTracer.endActiveSpan()
+
+        val spans = otelTesting.spans
+        assertEquals(2, spans.size)
+        assertEquals("FirstDraw", spans[0].name)
+        assertEquals("Created", spans[1].name)
+    }
+
     private val singleSpan: SpanData
         get() {
             val generatedSpans =


LikeTheSalad (changes requested):
Thank you for your contribution. These changes seem to be related to this [semconv PR](https://github.com/open-telemetry/semantic-conventions/pull/2831) where I've posted some concerns about things such as the type of signal to use for these kinds of events, as well as the utility of certain attributes such as `nodes` and `depth` followed by a lot of questions on what to do with them around certain scenarios. I'd like to get to a consensus over there first before moving forward with code changes.

bidetofevil:
The problem with detecting the first frame to be drawn is that, as you mentioned, it doesn't mean the Activity is ready to be consumed, it also doesn't necessarily mean the view tree is complete. If we are dealing with a Compose-based Activity, a minimal view tree will first be drawn and the rest will be filled in after the first frame is delivered.

This means that the page complexity attribute being logged may be inaccurate - or at least not represent what you think it does. (I think there are further issues with using that metric in the first place, but one thing at a time 😅 )

I think tracking this event offers good utility, but we should probably properly contextualize it, and perhaps not add in extra data that may be noisy or misleading.

marandaneto:
i think the user has to tell when the activity is ready to be used (vs fully drawn etc)
maybe we could inspire or use https://developer.android.com/reference/kotlin/androidx/activity/FullyDrawnReporter#reportFullyDrawn()

marandaneto:
https://github.com/square/papa/tree/main also has a few things related to what we do here
but its more about the app's launch and not per activity i think (worth checking)

Doohl:
> The problem with detecting the first frame to be drawn is that, as you mentioned, it doesn't mean the Activity is ready to be consumed, it also doesn't necessarily mean the view tree is complete.

Agree, but that isn't really the purpose of tracking this event. As it stands the best way to track when an Activity is ready to interact / consume (ie a "Time to interactive") would be to track the Activity [`reportFullyDrawn`](https://developer.android.com/reference/androidx/activity/ComponentActivity#reportFullyDrawn()). That is a separate thing altogether.

This PR is concerned with the [TimeToInitialDisplay](https://developer.android.com/topic/performance/vitals/launch-time#time-initial) vital.

> If we are dealing with a Compose-based Activity, a minimal view tree will first be drawn and the rest will be filled in after the first frame is delivered.

Yeah InitialDraw doesn't work too well in the context of Compose Activities. Open to ideas.

> This means that the page complexity attribute being logged may be inaccurate - or at least not represent what you think it does. (I think there are further issues with using that metric in the first place, but one thing at a time 😅 )

The complexity / depth attributes are removed! I think they aren't really needed for this new type of telemetry tbh

> I think tracking this event offers good utility, but we should probably properly contextualize it, and perhaps not add in extra data that may be noisy or misleading.

Agreed. I've removed it.

> i think the user has to tell when the activity is ready to be used (vs fully drawn etc)
maybe we could inspire or use

Yeah, the dev has to instrument something that can broadcast when an app / activity is ready to interact with. But we can automatically instrument, at least, when the Activity is ready to draw or has begun drawing.

LikeTheSalad:
Hi @Doohl 

Again, thank you for taking the time to make this contribution and for your patience. I'm following up here based on [this comment](https://github.com/open-telemetry/semantic-conventions/pull/2831#discussion_r2451173947) from the related semconv PR. I've covered a lot of details there that seem relevant to this work, and, in a nutshell, I'd like to try and see if this PR can help make progress in the semconv one. Please take a look at that comment for more details.

Before diving deep into the implementation details that you propose here, I'd like to make sure we're all on the same page in terms of what the expected outcome is that you'd like to achieve. I know that the idea behind the semconv PR is to define a platform-agnostic span, which I think would be great, but for practical purposes, I'd like to get a better understanding of what it means specifically for Android. So, I'd like to use a visualization that relies on Android-specific terms to see if that helps to get a better understanding overall:

Screenshot 2025-10-22 at 14 33 29

The image shows 3 possible scenarios (used to be 4, but then I realized that 3 is enough) that can be covered with a span. Which scenario do you think better covers the outcome that you'd expect from this implementation?

instrumentation/activity/README.md
```diff
@@ -38,6 +38,18 @@ This instrumentation produces the following telemetry:
...
+
+* Type: Span
+* Name: `FirstDraw`
+* Description: On activity PreCreated or Created (pre API 29) callback, a span will be created
+  to represent the time that the UI pipeline took to render that Activity's first frame. The span ends
+  at the first draw of the window
+  [DecorView](https://developer.android.com/reference/android/view/Window#getDecorView())
+* Attributes:
+  * `activity.name`:  name of activity
+  * `screen.name`:  name of screen
```
limhjgrace, line 51: suggestion: We've recently aligned on using `app.screen.name` for the attribute key, could we update to use that? Related PR: https://github.com/open-telemetry/semantic-conventions/pull/2744/files

instrumentation/activity/README.md
```diff
@@ -38,6 +38,18 @@ This instrumentation produces the following telemetry:
   * `screen.name`:  name of screen
   * `last.screen.name`:  name of screen, only when span contains the `activityPostResumed` event.
 
+### First Draw
+
+* Type: Span
+* Name: `FirstDraw`
```
limhjgrace, line 44: suggestion: I'm currently proposing `app.screen.time_to_first_draw` in https://github.com/open-telemetry/semantic-conventions/pull/2831. PR is still open so it may change but I don't think we should be using Pascal case regardless. I know that existing spans e.g. `AppStart` are in Pascal case but I recall both Jason and Hanson saying that this is something that needs to be fixed in the OTel Android SDK since it doesn't align with the usual casing convention of OTel.

PR #1283, opened 2025-10-03
feat: Add CpuAttributesSpanAppender

### LINKED ISSUE #1142: Monitoring relative cpu utilization for client spans
Current spans and events don't provide enough context to tell you, for example, why an AppStart or activity Created / Resumed span is taking a long time. Currently the only way you could reliably figure out why a span has elevated duration is by instrumenting additional sub-spans yourself.

In a very generic sense, any of your application's spans could be delayed for the following reasons:
1. You have complex or deeply-nested view hierarchies that bog down the Android UI pipeline [[Android App quality article]](https://developer.android.com/topic/performance/rendering/optimizing-view-hierarchies).
    * In reality this would mostly affect draw times, but that's a separate issue altogether.
2. There are expensive operations happening in the process / UI thread that are slowing things down. The easiest way to get a measure on this is to monitor CPU performance (through a `CPU utilization` gauge or metric).

We could benefit from a SpanProcessor that can sample relative, average CPU utilization for the duration of a span and append the following attributes:
- `process.cpu.avg_utilization` - The relative, average CPU utilization for the app process during the span
- `process.cpu.elapsed_time_start_millis` - The elapsed CPU time at the start of this span
- `process.cpu.elapsed_time_end_millis` - The elapsed CPU time at the end of this span

**How do we do this?**

The calculation is very straightforward:

```
cpuUtilizationAvg = 100 * (spanElapsedCpuDuration / spanDuration) / numCores
```

*Calculations, overall strategy, taken from a [Lyft Engineering article](https://eng.lyft.com/monitoring-cpu-performance-of-lyfts-android-applications-4e36fafffe12) on this subject.*

We would also need to take advantage of the experimental ExtendedSpanProcessor introduced in an upstream PR (https://github.com/open-telemetry/opentelemetry-java/pull/6367) to append the new attributes *before* a span ends via the `onEnding` callback provided in the new SpanProcessor type.

Doohl:
fixes #1142

## What is this change?

This PR implements a new SpanProcessor (`CpuAttributesSpanAppender`) that appends the following attributes to spans:
- `process.cpu.avg_utilization` - The relative, average CPU utilization for the app process during the span
- `process.cpu_elapsed_time_start` - The elapsed CPU time at the start of this span (in millis)
- `process.cpu_elapsed_time_end` - The elapsed CPU time at the end of this span (in millis)

## How was this tested?

Tested manually with the demo app. Here is an example of the telemetry generated:

```json
{
    "data": [
        {
            "traceID": "1b720aae66b3959b06eb5335247ada5b",
            "spans": [
                {
                    "traceID": "1b720aae66b3959b06eb5335247ada5b",
                    "spanID": "e53cbd939dce77af",
                    "operationName": "Created",
                    "references": [],
                    "startTime": 1759468632716000,
                    "duration": 46198,
                    "tags": [
                        {
                            "key": "otel.library.name",
                            "type": "string",
                            "value": "io.opentelemetry.lifecycle"
                        },
                        {
                            "key": "activity.name",
                            "type": "string",
                            "value": "AstronomyShopActivity"
                        },
                        {
                            "key": "last.screen.name",
                            "type": "string",
                            "value": "MainActivity"
                        },
                        {
                            "key": "network.connection.type",
                            "type": "string",
                            "value": "wifi"
                        },
                        {
                            "key": "process.cpu.avg_utilization",
                            "type": "float64",
                            "value": 37.5
                        },
                        {
                            "key": "process.cpu.elapsed_time_end",
                            "type": "int64",
                            "value": 1827
                        },
                        {
                            "key": "process.cpu.elapsed_time_start",
                            "type": "int64",
                            "value": 1758
                        },
                        {
                            "key": "screen.name",
                            "type": "string",
                            "value": "AstronomyShopActivity"
                        },
                        {
                            "key": "session.id",
                            "type": "string",
                            "value": "43e73022b738dfbeb9aaa2674806fb38"
                        },
                        {
                            "key": "toolkit",
                            "type": "string",
                            "value": "jetpack compose"
                        },
                        {
                            "key": "span.kind",
                            "type": "string",
                            "value": "internal"
                        },
                        {
                            "key": "internal.span.format",
                            "type": "string",
                            "value": "otlp"
                        }
                    ],
                    "logs": [
                        {
                            "timestamp": 1759468632716095,
                            "fields": [
                                {
                                    "key": "event",
                                    "type": "string",
                                    "value": "activityPreCreated"
                                }
                            ]
                        },
                        {
                            "timestamp": 1759468632724703,
                            "fields": [
                                {
                                    "key": "event",
                                    "type": "string",
                                    "value": "activityCreated"
                                }
                            ]
                        },
                        {
                            "timestamp": 1759468632759650,
                            "fields": [
                                {
                                    "key": "event",
                                    "type": "string",
                                    "value": "activityPostCreated"
                                }
                            ]
                        },
                        {
                            "timestamp": 1759468632760196,
                            "fields": [
                                {
                                    "key": "event",
                                    "type": "string",
                                    "value": "activityPreStarted"
                                }
                            ]
                        },
                        {
                            "timestamp": 1759468632760275,
                            "fields": [
                                {
                                    "key": "event",
                                    "type": "string",
                                    "value": "activityStarted"
                                }
                            ]
                        },
                        {
                            "timestamp": 1759468632760738,
                            "fields": [
                                {
                                    "key": "event",
                                    "type": "string",
                                    "value": "activityPostStarted"
                                }
                            ]
                        },
                        {
                            "timestamp": 1759468632760962,
                            "fields": [
                                {
                                    "key": "event",
                                    "type": "string",
                                    "value": "activityPreResumed"
                                }
                            ]
                        },
                        {
                            "timestamp": 1759468632760985,
                            "fields": [
                                {
                                    "key": "event",
                                    "type": "string",
                                    "value": "activityResumed"
                                }
                            ]
                        },
                        {
                            "timestamp": 1759468632762171,
                            "fields": [
                                {
                                    "key": "event",
                                    "type": "string",
                                    "value": "activityPostResumed"
                                }
                            ]
                        }
                    ],
                    "processID": "p1",
                    "warnings": null
                }
            ],
            "processes": {
                "p1": {
                    "serviceName": "OpenTelemetryDemoApp",
                    "tags": [
                        {
                            "key": "device.manufacturer",
                            "type": "string",
                            "value": "Google"
                        },
                        {
                            "key": "device.model.identifier",
                            "type": "string",
                            "value": "sdk_gphone64_arm64"
                        },
                        {
                            "key": "device.model.name",
                            "type": "string",
                            "value": "sdk_gphone64_arm64"
                        },
                        {
                            "key": "os.description",
                            "type": "string",
                            "value": "Android Version 16 (Build BP22.250221.010 API level 36)"
                        },
                        {
                            "key": "os.name",
                            "type": "string",
                            "value": "Android"
                        },
                        {
                            "key": "os.type",
                            "type": "string",
                            "value": "linux"
                        },
                        {
                            "key": "os.version",
                            "type": "string",
                            "value": "16"
                        },
                        {
                            "key": "rum.sdk.version",
                            "type": "string",
                            "value": "0.16.0-alpha-SNAPSHOT"
                        },
                        {
                            "key": "service.version",
                            "type": "string",
                            "value": "1.0"
                        },
                        {
                            "key": "telemetry.sdk.language",
                            "type": "string",
                            "value": "java"
                        },
                        {
                            "key": "telemetry.sdk.name",
                            "type": "string",
                            "value": "opentelemetry"
                        },
                        {
                            "key": "telemetry.sdk.version",
                            "type": "string",
                            "value": "1.54.1"
                        }
                    ]
                }
            },
            "warnings": null
        }
    ],
    "total": 0,
    "limit": 0,
    "offset": 0,
    "errors": null
}
```

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/CpuAttributesSpanAppender.kt
+++ b/core/src/main/java/io/opentelemetry/android/CpuAttributesSpanAppender.kt
@@ -1,0 +1,57 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.android
+
+import android.os.Process
+import io.opentelemetry.android.common.RumConstants
+import io.opentelemetry.context.Context
+import io.opentelemetry.sdk.trace.ReadWriteSpan
+import io.opentelemetry.sdk.trace.ReadableSpan
+import io.opentelemetry.sdk.trace.internal.ExtendedSpanProcessor
+
+/**
+ * A [SpanProcessor] that uses the experimental ExtendedSpanProcessor API to append OS process
+ * cpu statistics into span attributes. We establish 'cpu utilization average' to be:
+ *
+ *   cpuUtilizationAvg = 100 * (cpuTimeMs / spanDurationMs) / number of CPU cores
+ *     * cpuTimeMs is the time in milliseconds that the app process has taken in active CPU
+ *       time
+ *     * spanDurationMs is the total running time in milliseconds that the span has been active
+ *       for
+ */
+class CpuAttributesSpanAppender(
+    private val cpuCores: Int = Runtime.getRuntime().availableProcessors(),
+) : ExtendedSpanProcessor {
+    override fun isStartRequired(): Boolean = true
+
+    override fun onEnd(span: ReadableSpan) {}
+
+    override fun isEndRequired(): Boolean = false
+
+    override fun isOnEndingRequired(): Boolean = true
+
+    override fun onStart(
+        parentContext: Context,
+        span: ReadWriteSpan,
+    ) {
+        val cputime = Process.getElapsedCpuTime()
+        span.setAttribute(RumConstants.CPU_ELAPSED_TIME_START_KEY, Process.getElapsedCpuTime())
+    }
+
+    override fun onEnding(span: ReadWriteSpan) {
+        val startCpuTime =
+            span.getAttribute(RumConstants.CPU_ELAPSED_TIME_START_KEY) ?: return
+        val endCpuTime = Process.getElapsedCpuTime()
+        val cpuTimeMs = (endCpuTime - startCpuTime).toDouble()
+        val spanDurationMs = (span.latencyNanos / 1_000_000).toDouble()
+
+        if (spanDurationMs > 0) {
+            val cpuUtilization = (cpuTimeMs / spanDurationMs) * 100.0 / cpuCores.toDouble()
+            span.setAttribute(RumConstants.CPU_AVERAGE_KEY, cpuUtilization)
+        }
+        span.setAttribute(RumConstants.CPU_ELAPSED_TIME_END_KEY, endCpuTime)
+    }
+}
--- a/core/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
+++ b/core/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
@@ -7,6 +7,7 @@ package io.opentelemetry.android.config;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import io.opentelemetry.android.CpuAttributesSpanAppender;
 import io.opentelemetry.android.ScreenAttributesSpanProcessor;
 import io.opentelemetry.android.features.diskbuffering.DiskBufferingConfig;
 import io.opentelemetry.android.internal.services.network.CurrentNetworkProvider;
@@ -27,6 +28,7 @@ public class OtelRumConfig {
     private boolean generateSdkInitializationEvents = true;
     private boolean includeScreenAttributes = true;
     private boolean discoverInstrumentations = true;
+    private boolean includeCpuAttributes = true;
     private DiskBufferingConfig diskBufferingConfig = DiskBufferingConfig.create();
     private final List<String> suppressedInstrumentations = new ArrayList<>();
 
@@ -66,9 +68,23 @@ public class OtelRumConfig {
         return this;
     }
 
+    /**
+     * Disables the cpu attributes for spans. See {@link CpuAttributesSpanAppender} for more
+     * information. Default = true.
+     */
+    public OtelRumConfig disableCpuAttributes() {
+        includeCpuAttributes = false;
+        return this;
+    }
+
     /** Returns true if runtime network attributes are enabled, false otherwise. */
     public boolean shouldIncludeNetworkAttributes() {
         return includeNetworkAttributes;
+    }
+
+    /** Returns true if cpu attributes are enabled, false otherwise */
+    public boolean shouldIncludeCpuAttributes() {
+        return includeCpuAttributes;
     }
 
     /**
--- a/core/src/test/java/io/opentelemetry/android/OpenTelemetryRumBuilderTest.java
+++ b/core/src/test/java/io/opentelemetry/android/OpenTelemetryRumBuilderTest.java
@@ -508,6 +508,31 @@ public class OpenTelemetryRumBuilderTest {
                                 .build());
     }
 
+    @Test
+    public void verifyCpuAttributesSpanProcessor() {
+        createAndSetServiceManager();
+        OtelRumConfig config = buildConfig();
+
+        OpenTelemetryRum rum =
+                OpenTelemetryRum.builder(application, config)
+                        .addTracerProviderCustomizer(
+                                (tracerProviderBuilder, app) ->
+                                        tracerProviderBuilder.addSpanProcessor(
+                                                SimpleSpanProcessor.create(spanExporter)))
+                        .build();
+
+        rum.getOpenTelemetry().getTracer("test").spanBuilder("test span").startSpan().end();
+
+        await().atMost(Duration.ofSeconds(30))
+                .untilAsserted(
+                        () -> {
+                            List<SpanData> spans = spanExporter.getFinishedSpanItems();
+                            assertThat(spans).hasSize(1);
+                            assertThat(spans.get(0).getAttributes().asMap())
+                                    .containsKey(longKey("process.cpu.elapsed_time_start"));
+                        });
+    }
+
     private static Services createAndSetServiceManager() {
         Services services = mock(Services.class);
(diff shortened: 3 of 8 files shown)


core/src/main/java/io/opentelemetry/android/config/OtelRumConfig.java
```diff
@@ -66,11 +68,25 @@ public OtelRumConfig disableNetworkAttributes() {
...
     }
 
+    /**
+     * Disables the cpu attributes for spans. See {@link CpuAttributesSpanAppender} for more
+     * information. Default = true.
+     */
+    public OtelRumConfig disableCpuAttributes() {
+        includeCpuAttributes = false;
+        return this;
+    }
```
breedx-splk, line 78: Until we can establish that there is a broad(er) desire for these attributes, I think it's safer for this to be opt-in (default disabled). Can we flip this around?

core/src/main/java/io/opentelemetry/android/CpuAttributesSpanAppender.kt
```diff
@@ -0,0 +1,57 @@
...
+ * cpu statistics into span attributes. We establish 'cpu utilization average' to be:
+ *
+ *   cpuUtilizationAvg = 100 * (cpuTimeMs / spanDurationMs) / number of CPU cores
+ *     * cpuTimeMs is the time in milliseconds that the app process has taken in active CPU
+ *       time
+ *     * spanDurationMs is the total running time in milliseconds that the span has been active
+ *       for
+ */
+class CpuAttributesSpanAppender(
+    private val cpuCores: Int = Runtime.getRuntime().availableProcessors(),
```
breedx-splk, line 26: Guess what? This value is not guaranteed to stay the same for the life of the runtime! I know, it's wild.  https://developer.android.com/reference/java/lang/Runtime#availableProcessors()

core/src/main/java/io/opentelemetry/android/CpuAttributesSpanAppender.kt
```diff
@@ -0,0 +1,57 @@
...
+        val cputime = Process.getElapsedCpuTime()
+        span.setAttribute(RumConstants.CPU_ELAPSED_TIME_START_KEY, Process.getElapsedCpuTime())
+    }
+
+    override fun onEnding(span: ReadWriteSpan) {
+        val startCpuTime =
+            span.getAttribute(RumConstants.CPU_ELAPSED_TIME_START_KEY) ?: return
+        val endCpuTime = Process.getElapsedCpuTime()
+        val cpuTimeMs = (endCpuTime - startCpuTime).toDouble()
+        val spanDurationMs = (span.latencyNanos / 1_000_000).toDouble()
```
breedx-splk, line 49: I tend to favor existing fluent utility classes instead of manually doing the math with conversion factors.

```suggestion
        val spanDurationMs = TimeUnit.NANOSECONDS.toMillis(span.latencyNanos).toDouble()
```

core/src/test/java/io/opentelemetry/android/OpenTelemetryRumBuilderTest.java
```diff
@@ -508,6 +508,31 @@ public void verifyGlobalAttrsForLogs() {
...
+
+        rum.getOpenTelemetry().getTracer("test").spanBuilder("test span").startSpan().end();
+
+        await().atMost(Duration.ofSeconds(30))
+                .untilAsserted(
+                        () -> {
+                            List<SpanData> spans = spanExporter.getFinishedSpanItems();
+                            assertThat(spans).hasSize(1);
+                            assertThat(spans.get(0).getAttributes().asMap())
+                                    .containsKey(longKey("process.cpu.elapsed_time_start"));
```
breedx-splk, line 532: I think it would also help to 
* verify that the other 2 attributes are also present
* verify that the end time is greater than or equal the start time

core/src/main/java/io/opentelemetry/android/CpuAttributesSpanAppender.kt
breedx-splk: And yeah, unfortunately, this uncertainty can make the data harder to interpret without also sending along how many cores were used in the average calculation. For example, if the app is running along for some time with 8 cores, and then the number of cores drops to 4, the per-core reported "cpu average" reported value will double considerably and the user will have no indication about why.

To address this, one _could_ imagine also sending the number of cores at span end and then the computation isn't required (it can be done in the backend, and the computed average is redundant). 

This also opens up some other slightly stupider edge cases which are probably not worth thinking too deeply about....like what if the core count changes mid-span (hint: there's no trivial way of knowing!)...or the fact that you can't guarantee that every cpu/core is running with the same performance/frequency...etc.

This makes me think of an interesting feature possibility tho -- polling the cpu count and generating an event when that number changes. 🤔

breedx-splk (commented):
Thanks @Doohl for the contribution. I think this is a curious/interesting feature, and I'm excited to hear what others think about it and how users will leverage this data.

I have left a few comments, but think that this could proceed with some changes.

core/src/main/java/io/opentelemetry/android/CpuAttributesSpanAppender.kt
marandaneto: > Guess what? This value is not guaranteed to stay the same for the life of the runtime! I know, it's wild. developer.android.com/reference/java/lang/Runtime#availableProcessors()

TIL

bidetofevil: Yeah, we had to dig into the native layer to get consistent info about cores. It's arguably not worth it...

bidetofevil:
Android phone are similar to multi-tenant systems in that the app is not responsible for scheduling, and thus have no control on the cores it gets assigned or whether or not they are throttled by the OS. As well, the main thread is often blocked by binder calls, which means it's just waiting for the OS to respond, which would affect average utilization

Given that, these metrics will be missing key aspects to contextualize why the numbers are the way they are, which would result in pretty noisy data.

BTW, what is the plan of how this data will be consumed, and what problems do you think it'll find?

breedx-splk:
@Doohl This has been sitting here for a couple months now. Are you able to come back to it? Thanks!

PR #1284, opened 2025-10-03
Use Context rather than Application

fractalwrench:
## Goal

Uses `Context` rather than `Application` for initializing the SDK as originally discussed in #1257.

`Context` is a less specific type than `Application` which allows folks to initialize the SDK from places other than the `Application` subclass. Potential use-cases include:

- Jetpack Startup library
- ContentProvider initialization
- Initializing telemetry in an activity after explicit user opt-in

I do think that initializing from an `Application` subclass remains a sensible default & should be the recommended installation approach, but using `Context` gives the option for users to initialize in those use-cases if they wish.

### DIFF
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
@@ -6,6 +6,7 @@
 package io.opentelemetry.android.agent
 
 import android.app.Application
+import android.content.Context
 import io.opentelemetry.android.Incubating
 import io.opentelemetry.android.OpenTelemetryRum
 import io.opentelemetry.android.agent.dsl.DiskBufferingConfigurationSpec
@@ -39,6 +40,32 @@ object OpenTelemetryRumInitializer {
         globalAttributes: (() -> Attributes)? = null,
         diskBuffering: (DiskBufferingConfigurationSpec.() -> Unit)? = null,
         configuration: (OpenTelemetryConfiguration.() -> Unit) = {},
+    ): OpenTelemetryRum =
+        initialize(
+            context = application,
+            globalAttributes = globalAttributes,
+            diskBuffering = diskBuffering,
+            configuration = configuration,
+        )
+
+    /**
+     * Opinionated [OpenTelemetryRum] initialization.
+     *
+     * @param context Your android app's application context. This should be from your Application
+     * subclass or an appropriate context that allows retrieving the application context. If you
+     * supply an inappropriate context (e.g. from attachBaseContext) then instrumentation relying
+     * on activity lifecycle callbacks will not function correctly.
+     * @param globalAttributes Configures the set of global attributes to emit with every span and event.
+     * @param diskBuffering Configures the disk buffering feature.
+     * @param configuration Type-safe config DSL that controls how OpenTelemetry
+     * should behave.
+     */
+    @JvmStatic
+    fun initialize(
+        context: Context,
+        globalAttributes: (() -> Attributes)? = null,
+        diskBuffering: (DiskBufferingConfigurationSpec.() -> Unit)? = null,
+        configuration: (OpenTelemetryConfiguration.() -> Unit) = {},
     ): OpenTelemetryRum {
         val rumConfig = OtelRumConfig()
         val cfg = OpenTelemetryConfiguration(rumConfig)
@@ -57,9 +84,19 @@ object OpenTelemetryRumInitializer {
                 cfg.sessionConfig.backgroundInactivityTimeout,
                 cfg.sessionConfig.maxLifetime,
             )
+
+        // ensure we're using the Application Context to prevent potential leaks.
+        // if context.applicationContext is null (e.g. called from within attachBaseContext),
+        // fallback to the supplied context.
+        val ctx =
+            when (context) {
+                is Application -> context
+                else -> context.applicationContext ?: context
+            }
+
         return OpenTelemetryRum
-            .builder(application, cfg.rumConfig)
-            .setSessionProvider(createSessionProvider(application, sessionConfig))
+            .builder(ctx, rumConfig)
+            .setSessionProvider(createSessionProvider(ctx, sessionConfig))
             .addSpanExporterCustomizer {
                 OtlpHttpSpanExporter
                     .builder()
@@ -82,11 +119,11 @@ object OpenTelemetryRumInitializer {
     }
 
     private fun createSessionProvider(
-        application: Application,
+        context: Context,
         sessionConfig: SessionConfig,
     ): SessionProvider {
         val timeoutHandler = SessionIdTimeoutHandler(sessionConfig)
-        Services.get(application).appLifecycle.registerListener(timeoutHandler)
+        Services.get(context).appLifecycle.registerListener(timeoutHandler)
         return SessionManager.create(timeoutHandler, sessionConfig)
     }
 }
--- a/core/src/test/java/io/opentelemetry/android/AndroidResourceTest.kt
+++ b/core/src/test/java/io/opentelemetry/android/AndroidResourceTest.kt
@@ -5,7 +5,7 @@
 
 package io.opentelemetry.android
 
-import android.app.Application
+import android.content.Context
 import android.content.pm.ApplicationInfo
 import android.os.Build
 import io.mockk.MockKAnnotations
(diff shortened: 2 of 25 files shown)


android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
```diff
@@ -79,9 +80,17 @@ object OpenTelemetryRumInitializer {
         instrumentations?.let { configure ->
             InstrumentationConfiguration(rumConfig).configure()
         }
+
+        // ensure we're using the Application Context to prevent potential leaks
+        val ctx =
+            when (context) {
+                is Application -> context
+                else -> context.applicationContext
```
marandaneto:
```suggestion
                else -> context.applicationContext ?: context
```

applicationContext can be nullable in very specific cases

fractalwrench: @marandaneto out of curiosity, what are those specific cases?

marandaneto: if you init the SDK from a very specific lifecycle that is before the `applicationContext` is fully initialized.
i dont recall the name of the lifeyccle method, but the context instance type is `ContextImpl` and then the `applicationContext` is still null

marandaneto: found it, if you init the SDK from `attachBaseContext`, this will happen

LikeTheSalad: I think this relates to my comment [here](https://github.com/open-telemetry/opentelemetry-android/issues/1257#issuecomment-3365213213), where I was suggesting using `Context.getApplicationContext()` for cases where we might need an Application-specific functionality, such as listening to activities' lifecycle callbacks. 

The fact that there's a risk of users initializing the agent using a Context object that won't provide the application context could mean that some specific features won't be able to initialize.

fractalwrench: IMO it is reasonable to document that as a caveat of this approach, specifically that if you attempt to initialize within `attachBaseContext` that lifecycle callbacks won't work. At Embrace we cast applicationContext to `Application` when setting lifecycle callbacks & anecdotally haven't experienced customers running into this issue. I do know that some folks want to initialize their SDK earlier than the `Application` lifecycle and I view it as a tradeoff that allows them to capture early telemetry that might be more important to them than other telemetry that might become harder to capture because of the change.

marandaneto: +1 on documenting but there's no reason not to fallback to `context` if `applicationContext` is null
if later on `ctx` isnt castable to `Application`, you just skip installing the features that wont work without it, thats how i've done many times before.

marandaneto: the current approach will assign `null` to `ctx` and then you cannot do almost anything if the SDK is init in the eg `attachBaseContext` lifecycle, so no telemetry at all.

fractalwrench: I've altered the code with your suggestion and tweaked the docs to indicate what could happen if a library consumer passed in the wrong context

instrumentation/android-instrumentation/src/main/java/io/opentelemetry/android/instrumentation/InstallationContext.kt
```diff
@@ -6,11 +6,14 @@
...
 
-data class InstallationContext(
-    val application: Application,
+class InstallationContext(
+    val context: Context,
     val openTelemetry: OpenTelemetry,
     val sessionProvider: SessionProvider,
-)
+) {
+    val application: Application = context as Application
```
marandaneto: if that fails, its gonna throw

services/src/main/java/io/opentelemetry/android/internal/services/Services.kt
```diff
@@ -52,7 +54,7 @@ class Services internal constructor(
     }
 
     override val visibleScreenTracker: VisibleScreenTracker by lazy {
-        VisibleScreenTrackerImpl(application)
+        VisibleScreenTrackerImpl(context as Application)
```
marandaneto: same as https://github.com/open-telemetry/opentelemetry-android/pull/1284/files#r2405700380

android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
LikeTheSalad: Got it. Thanks for the code docs update. If we're going to go with this approach, I think it's important enough to document it thoroughly, probably in the markdown files too, and also printing warning logs when there's a problem where the application is not available, because otherwise I'm concerned that the kind of issues that can arise from this won't be easy to spot. For example, a user might initialize the agent with a context that has no way to access the application object, while at the same time, they add the activity/fragment instrumentations. Then they'd compile their app, and it would all run without apparent issues, but they wouldn't see telemetry on their dashboards, or worse, they would see _some_ telemetry, making them think that everything's working for quick checks, without realizing that some aren't because they might not know what to look out for unless they do a thorough e2e test and know exactly what to expect out of each instrumentation/feature. These kinds of sneaky issues are the worst for me, because there are no clear clues for the affected people.

My preferred approach to this, considering that we need to provide an initializer with `Context` instead of `Application` which in theory sounds useful but I'm not sure if in practice someone will need it, would be:

- Changing the initializer to receive a Context instead of Application, this will be the "unsafe initializer".
- Creating an initializer overload function that receives an Application and all it does is to call the "main initializer", this will be the "safe initializer".
- Marking the "unsafe initializer" as "Incubating" or adding any other kind of annotation that would create a lint warning, so that people know that they better provide an application when calling the initializer if they want the lint warning to go away, otherwise it'll be clear that they are willing to take responsibility for calling the "unsafe one".

fractalwrench:
Thanks all for the detailed feedback. I've tweaked the implementation so that it avoids typecasts throwing errors - the SDK would now log a warning during initialization in the unlikely event an `Application` context couldn't be obtained. I've added some docs to an overload of `initialize` that explains some of the background behind this.

core/src/test/java/io/opentelemetry/android/AndroidResourceTest.kt
```diff
@@ -47,7 +47,7 @@ internal class AndroidResourceTest {
                 labelRes = 12345
             }
 
-        every { app.applicationInfo } returns appInfo
+        every { app.applicationContext.applicationInfo } returns appInfo
```
breedx-splk: Kinda surprised that this is still passing. With the change to the `AndroidResource` to only use the context and not application, shouldn't this test also change to reflect that? The test probably shouldn't use Application at all.

instrumentation/android-instrumentation/src/main/java/io/opentelemetry/android/instrumentation/InstallationContext.kt
```diff
@@ -6,11 +6,14 @@
 package io.opentelemetry.android.instrumentation
 
 import android.app.Application
+import android.content.Context
 import io.opentelemetry.android.session.SessionProvider
 import io.opentelemetry.api.OpenTelemetry
 
-data class InstallationContext(
-    val application: Application,
+class InstallationContext(
```
breedx-splk, line 13: Why strip the data class nature here?

fractalwrench: `InstallationContext` is part of the public API of the android-instrumentation module. Use of data classes in library APIs is [generally discouraged](https://kotlinlang.org/docs/api-guidelines-backward-compatibility.html#avoid-using-data-classes-in-your-api) in Kotlin because copy/componentN functions are synthesised that make it easy to introduce breaking changes between versions.

core/src/test/java/io/opentelemetry/android/AndroidResourceTest.kt
fractalwrench: `Application` is just a subtype of `Context` which is why this still works. I've tweaked the test to use a `Context` mock instead so that it's less specific.

instrumentation/android-instrumentation/src/main/java/io/opentelemetry/android/instrumentation/InstallationContext.kt
breedx-splk: Ok thanks for that clarification. Makes sense.

PR #1285, opened 2025-10-03
Adding support for otel.stable property

LikeTheSalad:
This is to enable a similar mechanism to the one used in OTel Java Contrib, for example, where a module that has the property `otel.stable=true` doesn't get the `-alpha` suffix on its version. This will allow us to mark the `android-agent` as stable in the future.

### DIFF
--- a/buildSrc/src/main/kotlin/otel.publish-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.publish-conventions.gradle.kts
@@ -5,8 +5,23 @@ plugins {
     id("signing")
 }
 
-version = project.version.toString().replaceFirst("(-SNAPSHOT)?$".toRegex(), "-alpha$1")
+// If this module is not marked as stable (the default state) then append "-alpha" to its version.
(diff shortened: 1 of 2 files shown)


fractalwrench (approved):
LGTM

buildSrc/src/main/kotlin/otel.publish-conventions.gradle.kts
```diff
@@ -5,8 +5,23 @@ plugins {
     id("signing")
 }
 
-version = project.version.toString().replaceFirst("(-SNAPSHOT)?$".toRegex(), "-alpha$1")
```
breedx-splk, line 8: Interesting that we can just reference `version` below and not the full `project.version`. I guess this works somehow tho.

LikeTheSalad: Version is a special project property, so it's not needed for it, but custom ones are less straightforward to get/set 😅

PR #1289, opened 2025-10-06
Demonstrate type-safe DSL for initializing OpenTelemetry

fractalwrench:
## Goal

This PR demonstrates what a [type-safe DSL](https://kotlinlang.org/docs/type-safe-builders.html) could look like for initializing OpenTelemetry, as originally discussed in #1257. It's not in a mergeable state - I'd like to get feedback before proceeding further with any changes, then split the work up into smaller changesets. After the changes in this PR, initializing opentelemetry-android would look something like the following:

```
fun exampleUsage(application: Application) {
    OpenTelemetryRumInitializer.initialize(application) {
        httpExport {
            baseHeaders = mapOf("key" to "value")
            spans {
                url = "http://localhost:4318"
                headers = mapOf("key" to "override")
            }
        }
        instrumentations {
            anrReporter {
                enabled(true)
                addAttributesExtractor { _, _ ->
                    Attributes.empty()
                }
            }
        }
        session {
            maxLifetime = 10.minutes
        }
    }
}
```

IMO this has a few main advantages over the current API:

1. Only one parameter is required to configure the SDK. If we need to add additional configuration, it will not be a breaking change to add an extra parameter to the `initialize` function, rather we can just add an extra function to the DSL
2. Configuration can be entirely declarative. This means implementation details that construct concrete types such as `HttpEndpointConnectivity` don't need to be exposed
3. Symbols have been moved to a separate package rather than being declared in `OpenTelemetryRumInitializer`. If configuration grows substantially in future this prevents a scenario where a single source file needs to contain $LARGE_NUMBER lines-of-code

### What else is needed if we want to land this change?

IMO the following things would be useful:

1. Splitting up the changeset into smaller PRs
2. Adding KDoc for every function that forms part of the DSL
3. Providing a DSL that replaces/allows `OtelRumConfig` to be configured

### DIFF
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
@@ -8,33 +8,17 @@ package io.opentelemetry.android.agent
 import android.app.Application
 import io.opentelemetry.android.Incubating
 import io.opentelemetry.android.OpenTelemetryRum
-import io.opentelemetry.android.OpenTelemetryRumBuilder
-import io.opentelemetry.android.agent.connectivity.EndpointConnectivity
-import io.opentelemetry.android.agent.connectivity.HttpEndpointConnectivity
+import io.opentelemetry.android.agent.dsl.OpenTelemetryConfiguration
 import io.opentelemetry.android.agent.session.SessionConfig
 import io.opentelemetry.android.agent.session.SessionIdTimeoutHandler
 import io.opentelemetry.android.agent.session.SessionManager
-import io.opentelemetry.android.config.OtelRumConfig
-import io.opentelemetry.android.instrumentation.AndroidInstrumentation
-import io.opentelemetry.android.instrumentation.AndroidInstrumentationLoader
-import io.opentelemetry.android.instrumentation.activity.ActivityLifecycleInstrumentation
-import io.opentelemetry.android.instrumentation.anr.AnrInstrumentation
-import io.opentelemetry.android.instrumentation.common.EventAttributesExtractor
-import io.opentelemetry.android.instrumentation.common.ScreenNameExtractor
-import io.opentelemetry.android.instrumentation.crash.CrashDetails
-import io.opentelemetry.android.instrumentation.crash.CrashReporterInstrumentation
-import io.opentelemetry.android.instrumentation.fragment.FragmentLifecycleInstrumentation
-import io.opentelemetry.android.instrumentation.network.NetworkAttributesExtractor
-import io.opentelemetry.android.instrumentation.network.NetworkChangeInstrumentation
-import io.opentelemetry.android.instrumentation.slowrendering.SlowRenderingInstrumentation
 import io.opentelemetry.android.internal.services.Services
 import io.opentelemetry.android.session.SessionProvider
-import io.opentelemetry.api.trace.Tracer
+import io.opentelemetry.api.common.Attributes
 import io.opentelemetry.exporter.otlp.http.logs.OtlpHttpLogRecordExporter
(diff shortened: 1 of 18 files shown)


android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OpenTelemetryConfiguration.kt
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * Configures individual instrumentations.
+     */
+    internal val instrumentations: InstrumentationConfiguration =
+        InstrumentationConfiguration(rumConfig)
+
+    /**
+     * Configures how OpenTelemetry should export telemetry over HTTP.
+     */
+    fun httpExport(action: HttpExportConfiguration.() -> Unit) {
```
LikeTheSalad, line 32: I love this, because it leaves room for adding a `grpcExport` config in the future if someone asks for grpc support!

android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
```diff
@@ -8,97 +8,61 @@ package io.opentelemetry.android.agent
...
-            ),
-        metricEndpointConnectivity: EndpointConnectivity =
-            HttpEndpointConnectivity.forMetrics(
-                endpointBaseUrl,
-                endpointHeaders,
-            ),
-        rumConfig: OtelRumConfig = OtelRumConfig(),
-        sessionConfig: SessionConfig = SessionConfig.withDefaults(),
-        instrumentations: (InstrumentationConfiguration.() -> Unit)? = null,
+        openTelemetryConfiguration: (OpenTelemetryConfiguration.() -> Unit) = {},
```
LikeTheSalad, line 36: nit: maybe we should name this param `configuration`.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
```diff
@@ -8,97 +8,61 @@ package io.opentelemetry.android.agent
...
-     * @param sessionConfig The session configuration, which includes inactivity timeout and maximum lifetime durations.
-     * @param instrumentations Configurations for all the default instrumentations.
+     * @param openTelemetryConfiguration Type-safe config DSL that controls how OpenTelemetry
+     * should behave.
      */
     @Suppress("LongParameterList")
     @JvmStatic
     fun initialize(
         application: Application,
-        endpointBaseUrl: String,
```
LikeTheSalad, line 58: I really like how much cleaner the initializer looks after these changes.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/HttpExportConfiguration.kt
```diff
@@ -0,0 +1,57 @@
...
+            ).let(::toEndpointConfiguration)
+
+    internal val metricsConfig: EndpointConfiguration =
+        HttpEndpointConnectivity
+            .forMetrics(
+                baseUrl,
+                baseHeaders,
+            ).let(::toEndpointConfiguration)
+
+    fun spans(action: EndpointConfiguration.() -> Unit) {
```
LikeTheSalad, line 43: I'm guessing this is not necessary to use all the time, but only when a user would like to have a different endpoint config for a specific type of signal?

fractalwrench: Exactly. The default behaviour would be to specify `baseUrl` and optionally `baseHeaders`. If folks wanted to override config for specific endpoints, that's where `spans()` and other functions come into play.

fractalwrench:
Closing as this is mostly addressed now in other PRs.

PR #1290, opened 2025-10-06
Demo app - Fix coordinates of local packages

breedx-splk:
I was helping someone on slack and noticed that the coordinates we have substitutions for in the demo app no longer match what we publish. I think this was just an oversight from some time ago...but I think this could be seriously confusing for some users.

I also added a note about where to find the substitutions, because there is indeed a bit of gradle black magic there.

### DIFF
--- a/demo-app/build.gradle.kts
+++ b/demo-app/build.gradle.kts
@@ -70,9 +70,11 @@ dependencies {
 
     coreLibraryDesugaring(libs.desugarJdkLibs)
 
-    implementation("io.opentelemetry.android:instrumentation-compose-click")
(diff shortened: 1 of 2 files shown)


bidetofevil (approved):
LGTM. Can we define these coordinates in a places like the TOML file so we don't have to make sure the hardcoded names always match?

breedx-splk:
> LGTM. Can we define these coordinates in a places like the TOML file so we don't have to make sure the hardcoded names always match?

Hmmm. I like the thought. Even if the names existed in the toml and we could use them, it would be a separate effort/concern to make sure that those names match the published coordinates. I don't anticipate much thrash around this stuff.

PR #1291, opened 2025-10-07
Extract dsl configuration to separate package

fractalwrench:
## Goal

This PR works towards a [type-safe DSL](https://kotlinlang.org/docs/type-safe-builders.html) for initializing OpenTelemetry, as originally discussed in #1257, by extracting the existing configuration into a separate package. This should make it easier to maintain in future if the configuration grows in size, rather than containing everything in one file.


instrumentation/android-instrumentation/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentationLoader.kt
```diff
@@ -41,7 +41,7 @@ interface AndroidInstrumentationLoader {
          * Convenience method for [AndroidInstrumentationLoader.getByType].
          */
         @JvmStatic
-        fun <T : AndroidInstrumentation> getInstrumentation(type: Class<out T>): T? = get().getByType(type)
+        fun <T : AndroidInstrumentation> getInstrumentation(type: Class<out T>): T = get().getByType(type)!!
```
fractalwrench, line 44: I tweaked this to force non-null types and inlined the version in `OpenTelemetryRumInitializer`, given that every call in there was already doing this.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/instrumentation/CanBeEnabledAndDisabled.kt
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.android.agent.dsl.instrumentation
+
+internal interface CanBeEnabledAndDisabled {
+    fun enabled(enabled: Boolean)
```
breedx-splk, line 9: Not necessarily for this PR, but it gives the reader pause to see this interface name and then a single method that only has the word `enable`d in it. 

I think default no-arg methods `enable()` and `disable()` that call `enabled(bool)` could help with that. For another PR tho.

PR #1292, opened 2025-10-07
Fw/dsl params

fractalwrench:
## Goal

Builds on #1291 by continuing to convert `initialize` to use a type-safe DSL. The last commit contains the changes for this PR specifically - the first commit contains code that can be reviewed separately in #1291.

I've converted everything apart from `OtelRumConfig` to use a type-safe DSL called `OpenTelemetryConfiguration`. `OtelRumConfig` will be converted in a future PR that I've yet to complete. An example of how this looks in practice can be found below:

```
fun exampleUsage(application: Application) {
    OpenTelemetryRumInitializer.initialize(application) {
        httpExport {
            baseHeaders = mapOf("key" to "value")
            spans {
                url = "http://localhost:4318"
                headers = mapOf("key" to "override")
            }
        }
        instrumentations {
            anrReporter {
                enabled(true)
                addAttributesExtractor { _, _ ->
                    Attributes.empty()
                }
            }
        }
        session {
            maxLifetime = 10.minutes
        }
    }
}
```


android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
```diff
@@ -30,47 +28,21 @@ object OpenTelemetryRumInitializer {
...
-                endpointHeaders,
-            ),
-        metricEndpointConnectivity: EndpointConnectivity =
-            HttpEndpointConnectivity.forMetrics(
-                endpointBaseUrl,
-                endpointHeaders,
-            ),
-        sessionConfig: SessionConfig = SessionConfig.withDefaults(),
         globalAttributes: (() -> Attributes)? = null,
         diskBuffering: (DiskBufferingConfigurationSpec.() -> Unit)? = null,
```
LikeTheSalad, line 40: Maybe a discussion for another PR, though I'm curious about the decision process to decide which parameters are left on the initializer method's arguments, vs which ones are placed into the `Configuration` dsl?

fractalwrench: @LikeTheSalad I'd like to move basically every parameter apart from `Application` onto the `Configuration` DSL. I wanted to keep the PR size small particularly as other folks had PRs in flight, and should be able to open another PR towards the end of the week that moves the rest.

LikeTheSalad: Sounds good to me, thanks for the clarification.

fractalwrench: I've moved the rest over in #1313

PR #1313, opened 2025-10-16
Finishes altering initialize to use a type-safe DSL

fractalwrench:
## Goal

Builds on #1292 by using a type-safe DSL rather than individual parameters. The last commit contains the changes for this PR specifically. Essentially, I have moved the `globalAttributes` and `diskBuffering` parameters onto `OpenTelemetryConfiguration`.

I also hid some values that are no longer publicly accessed from the public API by making them internal, and made some constructors internal too.

### DIFF
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
@@ -9,16 +9,12 @@ import android.app.Application
 import android.content.Context
 import io.opentelemetry.android.Incubating
 import io.opentelemetry.android.OpenTelemetryRum
-import io.opentelemetry.android.agent.dsl.DiskBufferingConfigurationSpec
 import io.opentelemetry.android.agent.dsl.OpenTelemetryConfiguration
 import io.opentelemetry.android.agent.session.SessionConfig
 import io.opentelemetry.android.agent.session.SessionIdTimeoutHandler
 import io.opentelemetry.android.agent.session.SessionManager
-import io.opentelemetry.android.config.OtelRumConfig
-import io.opentelemetry.android.features.diskbuffering.DiskBufferingConfig
 import io.opentelemetry.android.internal.services.Services
 import io.opentelemetry.android.session.SessionProvider
-import io.opentelemetry.api.common.Attributes
 import io.opentelemetry.exporter.otlp.http.logs.OtlpHttpLogRecordExporter
 import io.opentelemetry.exporter.otlp.http.metrics.OtlpHttpMetricExporter
 import io.opentelemetry.exporter.otlp.http.trace.OtlpHttpSpanExporter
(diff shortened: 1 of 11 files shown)


android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
```diff
@@ -29,22 +25,16 @@ object OpenTelemetryRumInitializer {
      * Opinionated [OpenTelemetryRum] initialization.
      *
      * @param application Your android app's application object.
-     * @param globalAttributes Configures the set of global attributes to emit with every span and event.
-     * @param diskBuffering Configures the disk buffering feature.
      * @param configuration Type-safe config DSL that controls how OpenTelemetry
      * should behave.
      */
     @JvmStatic
     fun initialize(
```
LikeTheSalad: Doesn't have to be on this PR, but I was thinking it's probably better to remove this `initialize(Application,...)` function. The initial idea that I mentioned was so that we could mark the `initialize(Context,...)` one with some sort of annotation that would cause lint warning messages, making it clear to users that they should instead pass an Application whenever possible. However, I just realized that the `initialize(Context,...)` one isn't marked with anything that would raise lint warnings; thus, the existence of the 2 doesn't make much sense, since Application extends Context, and it seems like the application context not being retrievable is an edge case as well. I still would like to make clear that if this were to cause an issue, it probably won't be easy to spot, as it'd be a silent one, so the more logs we add around this, the better.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/HttpExportConfiguration.kt
```diff
@@ -13,6 +13,12 @@ class HttpExportConfiguration internal constructor() {
      * Global URL for HTTP export requests.
      */
     var baseUrl: String = ""
+        set(value) {
+            field = value
+            spansConfig.url = HttpEndpointConnectivity.forTraces(value).getUrl()
```
LikeTheSalad: By setting these in this function, how likely is it that something like this could happen:

```kotlin
initialize(app) {
  httpExport {
      spans {
          url = "http://localhost:4318"
      }
  }
}
```

Would that override the span's url in this case?

android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
fractalwrench: Sounds good. I've removed the function with the `Application` parameter.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/HttpExportConfiguration.kt
fractalwrench: I've added some unit test coverage for this part of the DSL and it confirmed the implementation as-is wasn't respecting `baseUrl` or `baseHeaders` properly, leading to an error being thrown when launching the demo application. I've fixed the implementation so that's no longer an issue, and baseUrl doesn't override the span's URL.

LikeTheSalad: Awesome, thank you! 🙏

LikeTheSalad (approved):
Cheers! 🚀

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/EndpointConfiguration.kt
```diff
@@ -7,6 +7,6 @@ package io.opentelemetry.android.agent.dsl
 
 @OpenTelemetryDslMarker
 class EndpointConfiguration internal constructor(
-    var url: String,
+    var url: String?,
```
marandaneto: can we make this non nullable? feels wrong allowning a nullable url for an endpoint configuration

marandaneto (approved):

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/EndpointConfiguration.kt
fractalwrench: Currently `baseUrl` in `HttpExportConfiguration` is non-null, whereas this is nullable.

I could make this non-null but we'd probably need to check at the point where the URL is constructed whether the URL is valid. I guess checking whether `url` is an empty string, and if not falling back to `baseUrl` would be a good approach. I'll change that unless folks object

fractalwrench: I've updated the implementation to make this property non-nullable.

breedx-splk:
Trying to force easycla to retrigger

breedx-splk:
🤞🏻 easycla don't fail us now....

PR #1317, opened 2025-10-17
Use JDK 21

fractalwrench:
## Goal

Uses JDK 21 to build the project on CI, bumping from 17. This gets the project closer to the latest LTS (AGP doesn't quite support it yet). I noticed there are a couple more warnings about targeting 1.8 bytecode when building the project, but functionally everything seems to work ok.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -20,7 +20,7 @@ jobs:
         uses: actions/setup-java@dded0888837ed1f317902acf8a20df0ad188d165 # v5.0.0
         with:
           distribution: temurin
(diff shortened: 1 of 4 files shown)


LikeTheSalad (commented):
> AGP doesn't quite support it yet

This makes me quite nervous tbh. Is there any reason why we shouldn't wait until AGP adds proper support?

fractalwrench:
@LikeTheSalad to clarify my ambiguous statement, AGP doesn't support building with JDK 25 apart from in 9.0.0-alpha. AGP does support building with JDK 21 and has done for some time

PR #1326, opened 2025-10-21
Update contrib version and fix diskbuffering

breedx-splk:
Supersedes #1321

Hopefully I got the API usage more or less correct, and looking to @LikeTheSalad to keep me honest here.

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/features/diskbuffering/SignalFromDiskExporter.kt
+++ b/core/src/main/java/io/opentelemetry/android/features/diskbuffering/SignalFromDiskExporter.kt
@@ -6,9 +6,11 @@
 package io.opentelemetry.android.features.diskbuffering
 
 import androidx.annotation.WorkerThread
-import io.opentelemetry.contrib.disk.buffering.LogRecordFromDiskExporter
-import io.opentelemetry.contrib.disk.buffering.MetricFromDiskExporter
-import io.opentelemetry.contrib.disk.buffering.SpanFromDiskExporter
+import io.opentelemetry.contrib.disk.buffering.storage.SignalStorage
+import io.opentelemetry.sdk.common.CompletableResultCode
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/opentelemetry/android/features/diskbuffering/SignalFromDiskExporter.kt
```diff
@@ -65,11 +62,23 @@ class SignalFromDiskExporter
...
+
+        private fun <T> export(
+            storage: SignalStorage<T>,
+            exporter: (Collection<T>) -> CompletableResultCode,
+        ): Boolean {
+            var rc = false
+            val iter = storage.iterator()
+            while (iter.hasNext()) {
+                val result = exporter(iter.next()).join(exportTimeoutInMillis, TimeUnit.MILLISECONDS)
+                // TODO: What to do if a random export in the middle here just fails?
```
LikeTheSalad, line 75: Calling `next()` from the iterator will delete the current item, so if a network export fails, I'd say we should stop the loop and try again later. Most likely, it'd be due to network connectivity issues anyway, which should come and go often, so chances are it could be fixed by the next try.

LikeTheSalad (approved):
Thank you for tackling this 🙏 I had an idea that it'd be quite painful, especially the test changes.

Just left a comment on your open question.

core/src/main/java/io/opentelemetry/android/features/diskbuffering/SignalFromDiskExporter.kt
breedx-splk: I'll create an issue to track this and we can pick that up in a follow-up PR. Thanks for the feedback. 👍🏻

breedx-splk: https://github.com/open-telemetry/opentelemetry-android/issues/1335

Issue #932, opened 2025-04-08
Screen rotation/orientation instrumentation


breedx-splk:
I don't think there's semconv for it yet, but we should think about adding instrumentation for screen orientation change events. This could be especially important with the context of click events and the screen coordinate system.

Other considerations could be given to other changes of resolution and screen size/shape?

bidetofevil:
Good instrumentation for RUM to have. These could trigger crashes and other unexpected behaviour due to the recreation of the Activity. At the very least, tracking that they happened (as an event) and the new orientation would be useful

RiyazurRazak:
@breedx-splk I'm interested in working on this task if it's still available?.

breedx-splk:
Awesome @RiyazurRazak, thanks. I'll assign you.

PR #1333, opened 2025-10-23
Add Screen-orientation instrumentation module

RiyazurRazak:
Resolves https://github.com/open-telemetry/opentelemetry-android/issues/932

### DIFF
--- a/instrumentation/screen-orientation/src/main/java/io/opentelemetry/android/instrumentation/screenorientation/ScreenOrientationDetector.kt
+++ b/instrumentation/screen-orientation/src/main/java/io/opentelemetry/android/instrumentation/screenorientation/ScreenOrientationDetector.kt
@@ -34,7 +34,8 @@ internal class ScreenOrientationDetector(
     }
 
     private fun emitLog(orientation: String) {
-        logger.logRecordBuilder()
+        logger
+            .logRecordBuilder()
             .setEventName(EVENT_NAME)
             .setAttribute(SCREEN_ORIENTATION, orientation)
             .emit()


RiyazurRazak:
- [x] Api Dump

android-agent/build.gradle.kts
```diff
@@ -16,7 +16,7 @@ dependencies {
     implementation(project(":services"))
     implementation(libs.opentelemetry.exporter.otlp)
 
-    // Default instrumentations:
+    // region Default instrumentations:
```
breedx-splk: We haven't used this convention elsewhere. Is it helpful? Let's omit this unless it's providing tangible benefit.

RiyazurRazak: Updated this to take advantage of Android Studio’s code folding feature, as this section tends to expand. I’ll roll it back if it doesn’t fit.

RiyazurRazak:
Thanks for the feedback, @breedx-splk. I’ll update the Event structure accordingly and remove the extractors.

android-agent/build.gradle.kts
RiyazurRazak: Done

instrumentation/screen-orientation/.gitignore
LikeTheSalad, line 1: I think we don't need this file.

instrumentation/screen-orientation/src/main/java/io/opentelemetry/android/instrumentation/screenorientation/ScreenOrientationInstrumentation.kt
```diff
@@ -0,0 +1,37 @@
...
+                .logsBridge
+                .loggerBuilder("io.opentelemetry.$name")
+                .build()
+        val applicationContext = ctx.context.applicationContext
+        detector = ScreenOrientationDetector(applicationContext, logger)
+        applicationContext.registerComponentCallbacks(detector)
+    }
+
+    override fun uninstall(ctx: InstallationContext) {
+        if (!::detector.isInitialized) return
```
LikeTheSalad: If we need to check whether some field is initialized or not, I prefer to make them nullable instead of resorting to the `isInitialized` check.

instrumentation/screen-orientation/README.md
```diff
@@ -0,0 +1,24 @@
...
+* Type: Event
+* Name: `device.screen_orientation`
+* Description: An event representing a change in screen orientation.
+* Attributes:
+    * `screen.orientation` - Represents the changed screen orientation
+
+## Installation
+
+This instrumentation comes with the [android agent](../../android-agent) out of the box, so
+if you depend on it, you don't need to do anything else to install this instrumentation.
```
LikeTheSalad: Please add the dependency coordinates, in case someone isn't using the agent. [This is an example](https://github.com/open-telemetry/opentelemetry-android/blob/main/instrumentation/activity/README.md?plain=1#L48) of how we've added it to other instrumentations.

PR #1334, opened 2025-10-23
Convert AttributeModifyingSpanExporter to Kotlin

fractalwrench:
## Goal

Converts `AttributeModifyingSpanExporter` to Kotlin and refactors the test to avoid unnecessary mocks.

As part of this change I did alter the signature so that a single function is responsible for modifying attribute values by checking the passed in key matches.


core/src/main/java/io/opentelemetry/android/export/AttributeModifyingSpanExporter.kt
```diff
@@ -0,0 +1,41 @@
...
+import io.opentelemetry.sdk.common.CompletableResultCode
+import io.opentelemetry.sdk.trace.data.SpanData
+import io.opentelemetry.sdk.trace.export.SpanExporter
+
+/**
+ * A SpanExporter that is configured to modify some of its attributes at export time.
+ */
+class AttributeModifyingSpanExporter(
+    private val delegate: SpanExporter,
+    private val spanAttributeReplacements: (key: AttributeKey<*>, value: Any?) -> Any?,
```
marandaneto: should we create an interface instead? i find it friendlier than a function def.

core/src/main/java/io/opentelemetry/android/export/AttributeModifyingSpanExporter.kt
```diff
@@ -0,0 +1,41 @@
...
+        val originals = span.attributes.asMap()
+        val modified =
+            originals.mapValues { entry ->
+                spanAttributeReplacements(entry.key, entry.value)
+            }
+
+        val builder = Attributes.builder()
+        modified.forEach {
+            @Suppress("UNCHECKED_CAST")
+            builder.put(it.key as AttributeKey<Any>, it.value)
```
marandaneto, line 35:
```suggestion
            builder.put(it.key as AttributeKey<*>, it.value)
```

not sure if it'd work but thats how its defined L19

fractalwrench: Unfortunately that won't work due to all the overloads of `put` making the type ambiguous. This is equivalent to the previous Java implementation which casted to `AttributeKey<Object>`.

core/src/main/java/io/opentelemetry/android/export/AttributeModifyingSpanExporter.kt
fractalwrench: I've updated this to use a typealias for readability

core/src/main/java/io/opentelemetry/android/export/SpanDataModifier.java
```diff
@@ -132,7 +132,18 @@ public SpanExporter build() {
...
+                            delegate,
+                            (attributeKey, value) -> {
+                                //noinspection unchecked
+                                Function<Object, Object> function =
+                                        (Function<Object, Object>)
+                                                spanAttributeReplacements.get(attributeKey);
+                                if (function != null) {
+                                    return function.apply(value);
+                                } else {
+                                    return value;
```
breedx-splk, line 144: nit: Prefer omitting the redundant else (and yeah, if it was kotlin it would be even cleaner to return from the `if` I suppose).

fractalwrench: I don't see how this else is redundant as the lambda requires a return statement if `function` is null. Have I missed some syntax that would simplify this in Java?

LikeTheSalad: I think he's talking about writing it like this:

```java
if (function != null) {
   return function.apply(value);
}
return value;
```

Maybe it sounded like omitting the "whole else", which might be confusing.

Issue #1345, opened 2025-10-27
For android, replace the usage of `Thread.id` with the compat method


atulgpt:
In `AnrWatcher`, `CrashReporter`, `CrashReportIntegrationTest` & `PeriodicWorkTest` we are using the `Thread.getId()` which is deprecated. 

I suggest to make a utility function in `common` module of the sdk in the package `io.opentelemetry.android.common.internal.utils`

```kt
package io.opentelemetry.android.common.internal.utils

import android.os.Build

inline val Thread.threadIdCompat: Long
    get() =
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.BAKLAVA) {
            threadId()
        } else {
            @Suppress("DEPRECATION")
            id
        }
```

and then replace all the usages in the above mentioned files

PR #1346, opened 2025-10-28
Use `threadId()` under the `BAKLAVA` android version check

atulgpt:
Fixes https://github.com/open-telemetry/opentelemetry-android/issues/1345

### DIFF
--- a/common/api/common.api
+++ b/common/api/common.api
@@ -102,3 +102,7 @@ public final class io/opentelemetry/android/common/internal/tools/time/SystemTim
 	public fun getCurrentTimeMillis ()J
 }
 
+public final class io/opentelemetry/android/common/internal/utils/AndroidUtilsKt {
+	public static final fun getThreadIdCompat (Ljava/lang/Thread;)J
+}
+


common/src/main/java/io/opentelemetry/android/common/internal/utils/AndroidUtils.kt
```diff
@@ -0,0 +1,22 @@
...
+
+package io.opentelemetry.android.common.internal.utils
+
+import android.os.Build
+import androidx.annotation.RequiresApi
+
+val Thread.threadIdCompat: Long
+    get() =
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.BAKLAVA) {
+            threadIdApi36
```
fractalwrench, line 14: Is there a reason this needs to be a separate property or could the function call be inlined?

atulgpt: For `this needs to be a separate property` yes as `RequiresApi` annotation needs to be present in a function(it can't be present in a expression). Also this annotation is required to suppress the animalsniffer

`could the function call be inlined` `threadIdCompat` can be inlined but then I need to make below function `@PublishedApi` like below

```kt
inline val Thread.threadIdCompat: Long
    get() =
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.BAKLAVA) {
            threadIdApi36
        } else {
            @Suppress("DEPRECATION")
            id
        }

@get:RequiresApi(api = Build.VERSION_CODES.BAKLAVA)
@PublishedApi
internal val Thread.threadIdApi36: Long
    get() = threadId()
```

`internal val Thread.threadIdApi36: Long` can't be inlined as animalsniffer plugin somehow was getting suppressed. I have raised the issue https://github.com/mojohaus/animal-sniffer/issues/311

breedx-splk (approved):
Looks good to me, thanks for your contribution!

Issue #412, opened 2024-06-04
Add build-time auto-instrumentation for @WithSpan


breedx-splk:
Would be nice if we could leverage the `@WithSpan` attribute from the upstream instrumentation repo and have build-time bytecode weaving to support it in user code.

Start by looking at `WithSpanInstrumentation` ([link](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/opentelemetry-extension-annotations-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/extensionannotations/WithSpanInstrumentation.java)).

gilgoldzweig:
I know this is an old issue. If I'm not mistaken, Android doesn't support the Java agent needed to automate this process; however, I think we can get something very close by building a compiler plugin, in Kotlin, this can probably get done by using the K2/IR plugin, which will allow us to modify the code not at the bytecode level but at the IR level. 
I can probably help with it and write the plugin if there is enough interest!

breedx-splk:
Thanks @gilgoldzweig yeah that's interesting. We are using bytebuddy gradle plugin elsewhere, so doing it at the IR level with a K2 plugin would diverge from this. I'm not against that, but it does make things a bit more complicated and it would be cool if we had a strategy for consolidating or being eventually consistent in our use of build-time tools. Specifically, I wonder if there's room/desire for us to migrate our existing bytecode instrumentation to the K2/IR level. I know @LikeTheSalad has a lot of experience as well with gradle plugins, and I'm curious what he might think. 

I don't have a good sense of how much work it would be to build/support a K2 plugin for this, but maybe a rough draft PR could help show this. Thanks again!

gilgoldzweig:
That sounds interesting as well, I was just suggesting because that would lift a good chunk of our reliance of java. 
However, I'm interested in contributing in general so if there are different avenues you prefer to focus on, I'm open for suggestions!

fractalwrench:
My understanding is that K2 plugins won't run on Java so auto-instrumentation wouldn't work in any app code containing Java. As a rough guess I'd assume that >80% of the average Android app codebase is in Kotlin these days, but there will be legacy code written in Java that is still kicking around.

I'd be interested in discussing K2/WebObject ASM/ByteBuddy usage more generally. In an ideal world we would pick auto-instrumentation tooling that operates at build-time and runs on both Kotlin/Java code. Perhaps a separate issue would be better?

[...]

PR #1348, opened 2025-10-29
Draft: Add instrumentation for otel span annotations

Kleszczu98:
Example byteBuddy implementation of auto-instrumentation of the following annotations: 
- Annotations: [LINK](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation-annotations)
- Documentation: [LINK](https://opentelemetry.io/docs/zero-code/java/agent/annotations/)

https://github.com/open-telemetry/opentelemetry-android/issues/412


breedx-splk:
@Kleszczu98 hi. The title of this still says it's a draft PR, but it's not marked as such. Please remove the word Draft from the title if it's ready for review, or mark it truly as draft by clicking "Convert to draft". You'll will also need to sign the Contributor License Agreement (CLA).

Thanks!

instrumentation/span-annotation/agent/src/main/kotlin/io/opentelemetry/instrumentation/agent/spanannotation/advice/constructor/AddingSpanAttributesConstructorAdvice.kt
LikeTheSalad, line 1: Are we aware of a use case that requires tracing a constructor? It sounds a bit of an edge case to me. If there's no foreseeable usage right now, I'd prefer to leave it for a future PR.

instrumentation/span-annotation/agent/src/main/kotlin/io/opentelemetry/instrumentation/agent/spanannotation/advice/method/AddingSpanAttributesMethodAdvice.kt
```diff
@@ -0,0 +1,22 @@
...
+import net.bytebuddy.asm.Advice
+import java.lang.reflect.Method
+
+object AddingSpanAttributesMethodAdvice {
+
+    @JvmStatic
+    @Advice.OnMethodEnter(suppress = Throwable::class)
+    fun onEnter(
+        @Advice.AllArguments args: Array<Any?>,
+        @Advice.Origin method: Method
```
LikeTheSalad: I'd like to follow best practices from the upstream instrumentation, such as [avoiding using \@Advice.Origin Method](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/contributing/writing-instrumentation-module.md#avoid-using-adviceorigin-method). In this case it seems like we can just get the method name by using `@Advice.Origin("#m") String methodName`.

instrumentation/span-annotation/agent/src/main/kotlin/io/opentelemetry/instrumentation/agent/spanannotation/advice/method/AddingSpanAttributesMethodAdvice.kt
```diff
@@ -0,0 +1,22 @@
...
+import io.opentelemetry.instrumentation.library.spanannotation.HelperFunctions
+import net.bytebuddy.asm.Advice
+import java.lang.reflect.Method
+
+object AddingSpanAttributesMethodAdvice {
+
+    @JvmStatic
+    @Advice.OnMethodEnter(suppress = Throwable::class)
+    fun onEnter(
+        @Advice.AllArguments args: Array<Any?>,
```
LikeTheSalad, line 17: I'm not faimilar with `@WithSpan`, though I'm curious why it might need all the method's arguments?

instrumentation/span-annotation/agent/src/main/kotlin/io/opentelemetry/instrumentation/agent/spanannotation/advice/constructor/AddingSpanAttributesConstructorAdvice.kt
Kleszczu98: Ok, i will remove the constructor implementation and focus on just the normal method case to simplify the PR.

instrumentation/span-annotation/agent/src/main/kotlin/io/opentelemetry/instrumentation/agent/spanannotation/advice/method/AddingSpanAttributesMethodAdvice.kt
Kleszczu98: I found the `@AddingSpanAttributes` annotation but it seems i might've misunderstood it's function. 
It is not mentioned in the linked [documentation](https://opentelemetry.io/docs/zero-code/java/agent/annotations/) so perhaps I should remove this as well and stick to `@WithSpan` and `@SpanAttribute` ? It covers all the most common use cases already.

LikeTheSalad: I see. I also don't see it in the docs so it should be fine to leave it out, at least for now. I'll cc @breedx-splk in case he's more familiar with it to make sure we're not missing something important.

DavidGrath:
@Kleszczu98 , good day,
It seems this PR has aged quite a bit.
Do you still want to continue working on this? We'd be glad to have this merged once it's complete

breedx-splk:
stale. if work on this continues, please reopen. thanks!

PR #1356, opened 2025-11-03
Convert `OtelRumConfig` to kotlin

fractalwrench:
## Goal

Converts `OtelRumConfig` to Kotlin. I've attempted to maintain API backwards compatibility with this change as much as possible & there shouldn't be a change in the API dump other than the addition of the `final` modifier.


core/src/test/java/io/opentelemetry/android/config/OtelRumConfigTest.kt
```diff
@@ -16,62 +16,58 @@ class OtelRumConfigTest {
...
     @Test
     fun `setting a Supplier that returns empty attributes is fine`() {
         val config = OtelRumConfig()
         config.setGlobalAttributes { Attributes.empty() }
         assertThat(config.hasGlobalAttributes()).isTrue() // It might return some Attributes later
-        assertThat(config.globalAttributesSupplier.get().isEmpty).isTrue()
-    }
-
-    @Test
-    fun `setting a Supplier that returns null attributes is fine`() {
```
fractalwrench, line 63: I deleted this test case (and the ability to pass null attributes) as keeping it would mean that `getGlobalAttributesSupplier()` would need to invoke the supplier, check whether the value was null, and then return the supplier again. I assumed removing the ability to pass null was preferable but don't have strong opinions if others disagree

PR #1357, opened 2025-11-03
Add nullability annotations to OpenTelemetryRumBuilder

fractalwrench:
## Goal

Adds nullability annotations to `OpenTelemetryRumBuilder`. This indicates the nullability of return values and parameters, and is a helpful pre-requisite of converting this class to Kotlin.


LikeTheSalad (commented):
I don't have a problem with this change, though it seems like this should already be accomplished by this [package annotation](https://github.com/open-telemetry/opentelemetry-android/blob/main/core/src/main/java/io/opentelemetry/android/package-info.java)?

fractalwrench:
Ah, completely missed that. I'll close this PR given it's redundant

Issue #1247, opened 2025-09-23
Replace all mockito usages with mockk


LikeTheSalad:
Once all the unit tests written in Java are migrated to Kotlin, we should keep only the mocking framework mockk, and remove mockito.

PR #1362, opened 2025-11-05
Remove the usage of mockito with mockk

atulgpt:
Part of https://github.com/open-telemetry/opentelemetry-android/issues/1247

### DIFF
--- a/core/src/test/java/io/opentelemetry/android/export/FilteringSpanExporterTest.kt
+++ b/core/src/test/java/io/opentelemetry/android/export/FilteringSpanExporterTest.kt
@@ -6,7 +6,6 @@
 package io.opentelemetry.android.export
 
 import io.mockk.every
-import io.mockk.junit5.MockKExtension
 import io.mockk.mockk
 import io.mockk.slot
 import io.opentelemetry.android.export.TestSpanHelper.span
@@ -19,10 +18,8 @@ import io.opentelemetry.sdk.trace.export.SpanExporter
 import org.assertj.core.api.Assertions.assertThat
 import org.assertj.core.api.ThrowingConsumer
 import org.junit.jupiter.api.Test
-import org.junit.jupiter.api.extension.ExtendWith
 import java.util.function.Predicate
 
-@ExtendWith(MockKExtension::class)
 internal class FilteringSpanExporterTest {
     @Test
     fun filter() {


core/src/test/java/io/opentelemetry/android/OpenTelemetryRumBuilderTest.kt
```diff
@@ -484,6 +477,7 @@ class OpenTelemetryRumBuilderTest {
             ).hasSeverity(Severity.FATAL3)
     }
 
+    @Ignore("Earlier with mockito cacheDir was null which was causing this TC to pass")
```
atulgpt, line 480: Since earlier we were using Mockito, which in was returning null for `CacheStorage` -> `cacheDir` even when `cacheDir` is marked as null. In case of mockk returning null for nonnullable field is not possible. So for now I have ignored this TC. 

Later this can be fixed by changing the folder structure little(but that can have some more changes in other TCs as well).

So I am not doing those changes here in this PR

atulgpt:
Also two TC files were in Java, so I didn't convert those as of now. We can do the conversion in other PR

core/src/test/java/io/opentelemetry/android/export/FilteringSpanExporterTest.kt
```diff
@@ -16,16 +20,11 @@ import org.assertj.core.api.Assertions.assertThat
...
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.extension.ExtendWith
-import org.mockito.ArgumentCaptor
-import org.mockito.Captor
-import org.mockito.Mockito
-import org.mockito.junit.jupiter.MockitoExtension
 import java.util.function.Predicate
 
-@ExtendWith(MockitoExtension::class)
+@ExtendWith(MockKExtension::class)
```
fractalwrench: Is this required if nothing is annotated with a mockk annotation?

core/src/test/java/io/opentelemetry/android/export/FilteringSpanExporterTest.kt
```diff
@@ -16,16 +20,11 @@ import org.assertj.core.api.Assertions.assertThat
...
 internal class FilteringSpanExporterTest {
-    @Captor
-    private lateinit var spansCaptor: ArgumentCaptor<MutableCollection<SpanData>>
+    private val spansCaptor = slot<MutableCollection<SpanData>>()
```
fractalwrench: Personally I'd favour `lateinit var` and a setup method here, _or_ an annotation. My understanding is that this will only get initialized once for the entire source file, so could lead to problems with resetting state

core/src/test/java/io/opentelemetry/android/export/FilteringSpanExporterTest.kt
atulgpt: No it is not required

core/src/test/java/io/opentelemetry/android/export/FilteringSpanExporterTest.kt
atulgpt: There is not annotation so I inlined the usages

core/src/test/java/io/opentelemetry/android/export/FilteringSpanExporterTest.kt
atulgpt: Removed

core/src/test/java/io/opentelemetry/android/export/FilteringSpanExporterTest.kt
fractalwrench: I don't see inlining as a problem but currently this won't reset any state between test cases on the `spansCaptor` property

atulgpt: So after moving the slot to the `@Test` methods this will not be a problem. I think this is what you are implying. Right?

fractalwrench: There are two options:

1. Move the instantiation of slot within individual test methods
2. Change the property to `lateinit var` and instantiate in a setup method

This is to avoid interactions with the slot in one test case affecting those in other test cases

core/src/test/java/io/opentelemetry/android/export/FilteringSpanExporterTest.kt
breedx-splk: are you sure? :)

core/src/test/java/io/opentelemetry/android/export/FilteringSpanExporterTest.kt
atulgpt: Hi, I have already moved the instantiation of slot withing the individual tests(in the first commit). So this comment should be fine

core/src/test/java/io/opentelemetry/android/export/FilteringSpanExporterTest.kt
atulgpt: Ahh.. I missed that earlier 😅. Made the changes to remove this

atulgpt: There should be some rule to prevent this automatically in the PR

Issue #1128, opened 2025-08-05
Implement uninstall() for instrumentations


breedx-splk:
This is a follow-up from #1109, which added the ability for instrumentations (and other core components) to be uninstalled at shutdown time. 

This will serve as an umbrella/status issue to track the progress of implementing `uninstall()` on all the instrumentation modules. If an instrumentation does not actually need to implement `uninstall()` because it does not have any resources to free or listeners to unregister etc then we should leave a comment to that extent and we can mark it off.

## instrumentations needing uninstall()

* [x] activity
* [ ] android-instrumentation
* [ ] android-log
* [ ] anr
* [ ] build
* [ ] common-api
* [ ] compose
* [ ] crash
* [x] fragment
* [ ] httpurlconnection
* [ ] network
* [ ] okhttp3
* [ ] okhttp3-websocket
* [ ] sessions
* [x] slowrendering (part of #1109)
* [ ] startup
* [ ] view-click
* [X] ~volley~ _will not do: deprecated_

amishhhaaaa:
@breedx-splk had a doubt, some files dont have listeners, I might be wrong, but do we also have to add the listener files for executor services?

breedx-splk:
> [@breedx-splk](https://github.com/breedx-splk) had a doubt, some files dont have listeners, I might be wrong, but do we also have to add the listener files for executor services?

I don't understand what you're asking.

breedx-splk:
Oops, this should not have been closed!

PR #1369, opened 2025-11-08
Fix/1128/activity fragment uninstall

atulgpt:
Fixes https://github.com/open-telemetry/opentelemetry-android/issues/1128 for activity and fragment

### DIFF
--- a/instrumentation/activity/src/main/java/io/opentelemetry/android/instrumentation/activity/ActivityLifecycleInstrumentation.kt
+++ b/instrumentation/activity/src/main/java/io/opentelemetry/android/instrumentation/activity/ActivityLifecycleInstrumentation.kt
@@ -23,7 +23,6 @@ class ActivityLifecycleInstrumentation : AndroidInstrumentation {
     private val startupTimer: AppStartupTimer by lazy { AppStartupTimer() }
     private var screenNameExtractor: ScreenNameExtractor = DefaultScreenNameExtractor
     private var tracerCustomizer: (Tracer) -> Tracer = { it }
-    private var tracer: Tracer? = null
     private var startupLifecycle: Application.ActivityLifecycleCallbacks? = null
     private var activityLifecycle: Application.ActivityLifecycleCallbacks? = null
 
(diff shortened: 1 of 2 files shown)


instrumentation/activity/src/main/java/io/opentelemetry/android/instrumentation/activity/startup/AppStartupTimer.java
```diff
@@ -39,7 +39,7 @@ public class AppStartupTimer {
     // accessed only from UI thread
     private boolean isStartedFromBackground = false;
 
-    public Span start(Tracer tracer) {
+    public Span start(@NonNull Tracer tracer) {
```
breedx-splk: Not a blocking request, but I think our preference would be to not sprinkle `@NonNull` around every parameter. Rather, we should use the `@ParametersAreNonnullByDefault` package annotation, like in `package-info.java` in the `httpurlconnection` package, for example.

Alternatively, this problem goes away when we convert this to kotlin.

instrumentation/activity/src/main/java/io/opentelemetry/android/instrumentation/activity/ActivityLifecycleInstrumentation.kt
```diff
@@ -34,10 +38,33 @@ class ActivityLifecycleInstrumentation : AndroidInstrumentation {
     }
 
     override fun install(ctx: InstallationContext) {
-        startupTimer.start(ctx.openTelemetry.getTracer(INSTRUMENTATION_SCOPE))
+        tracer =
```
breedx-splk: I'm not convinced that we need to hold on to the tracer, just to null it out in `uninstall`. Can you explain why you think this is necessary, or remove the field? Thanks!

instrumentation/activity/src/main/java/io/opentelemetry/android/instrumentation/activity/startup/AppStartupTimer.java
atulgpt: Removed it.. I think when I was doing something around this code, kotlin make the param nullable as there was no annotation that's why I put one annotation.

instrumentation/activity/src/main/java/io/opentelemetry/android/instrumentation/activity/ActivityLifecycleInstrumentation.kt
atulgpt: Removed it. Yes it was not required

breedx-splk (approved):

breedx-splk:
Thanks for the assist!

PR #1372, opened 2025-11-10
Convert OpenTelemetryRumBuilder to kotlin

fractalwrench:
## Goal

Converts `OpenTelemetryRumBuilder` from Java to Kotlin. I've tried to retain the API as close as possible to the existing API dump, and have avoided converting symbols such as `BiFunction` to their Kotlin equivalents for now (although I think this is worth pursuing in future).

I've mostly relied on `OpenTelemetryRumBuilderTest` to alert of any breaking changes this might have introduced.


breedx-splk (commented):
LGTM

PR #1373, opened 2025-11-11
fix: add `openTelemetry` getter to `OpenTelemetryRum` to fix breaking API change

beekhc:
I'm not sure if it was intentional, but when `OpenTelemetryRum` was ported from Java to Kotlin, existing Kotlin code that used `.openTelemetry` to access `getOpenTelemetry()` broke, because that getter had previously been added by the automatic Kotlin wrapper around the Java. This change adds an extension with a readonly property to fix those cases.

The property cannot be added the class itself, because then the automatic _Java_ wrapper around the Kotlin would add a `getOpenTelemetry()` method that would conflict with the existing one.

An alternative would be to _replace_ the `getOpenTelemetry()` method with the readonly property. But I think that would break existing Kotlin code that explicitly calls `getOpenTelemetry()`. So I chose the safer option.

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
+++ b/core/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
@@ -20,7 +20,7 @@ interface OpenTelemetryRum {
      * Get a handle to the instance of the OpenTelemetry API that this
      * instance is using for instrumentation.
      */
-    fun getOpenTelemetry(): OpenTelemetry
+    val openTelemetry: OpenTelemetry
 
     /**
      * Get the client session ID associated with this instance of the RUM instrumentation library.
@@ -104,17 +104,3 @@ interface OpenTelemetryRum {
         fun noop(): OpenTelemetryRum = NoopOpenTelemetryRum
     }
 }
-
-/**
- * Property accessor for [OpenTelemetryRum.getOpenTelemetry]. Provided for backward compatibility
- * with code that accessed this as a property when the class was in Java.
- */
-val OpenTelemetryRum.openTelemetry: OpenTelemetry
-    get() = getOpenTelemetry()
-
-/**
- * Property accessor for [OpenTelemetryRum.getRumSessionId]. Provided for backward compatibility
- * with code that accessed this as a property when the class was in Java.
- */
-val OpenTelemetryRum.rumSessionId: String
-    get() = getRumSessionId()
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -50,11 +50,6 @@ public final class io/opentelemetry/android/OpenTelemetryRumBuilder {
 	public fun setSessionProvider (Lio/opentelemetry/android/session/SessionProvider;)Lio/opentelemetry/android/OpenTelemetryRumBuilder;
 }
 
-public final class io/opentelemetry/android/OpenTelemetryRumKt {
-	public static final fun getOpenTelemetry (Lio/opentelemetry/android/OpenTelemetryRum;)Lio/opentelemetry/api/OpenTelemetry;
-	public static final fun getRumSessionId (Lio/opentelemetry/android/OpenTelemetryRum;)Ljava/lang/String;
-}
-
 public final class io/opentelemetry/android/SdkPreconfiguredRumBuilder {
 	public final fun addInstrumentation (Lio/opentelemetry/android/instrumentation/AndroidInstrumentation;)Lio/opentelemetry/android/SdkPreconfiguredRumBuilder;
 	public final fun build ()Lio/opentelemetry/android/OpenTelemetryRum;
--- a/core/src/main/java/io/opentelemetry/android/NoopOpenTelemetryRum.kt
+++ b/core/src/main/java/io/opentelemetry/android/NoopOpenTelemetryRum.kt
@@ -10,7 +10,7 @@ import io.opentelemetry.api.common.Attributes
 import io.opentelemetry.api.trace.TraceId
 
 internal object NoopOpenTelemetryRum : OpenTelemetryRum {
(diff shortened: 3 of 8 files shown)


core/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
```diff
@@ -104,3 +104,17 @@ interface OpenTelemetryRum {
         fun noop(): OpenTelemetryRum = NoopOpenTelemetryRum
     }
 }
+
+/**
+ * Property accessor for [OpenTelemetryRum.getOpenTelemetry]. Provided for backward compatibility
```
LikeTheSalad: We haven't yet made this API stable, so there's no need to keep backwards compatibility. If I understand correctly from your changes, they seem to aim at providing some sort of syntactic sugar when used via Kotlin sources. Is that correct?

beekhc: Hmm, in that case, then I'd propose we replace the `getOpenTelemetry()` function with an `openTelemetry` property, since that is more idiomatic Kotlin. For example, in Android Studio, when calling Java code where both syntaxes are valid, the IDE actively recommends using the `.openTelemetry` syntax with a "Use of getter method instead of property access syntax" warning.

The only code that wouldn't continue working is Kotlin code with `.getOpenTelemetry()`, which used to give a warning anyway. Java code could continue to call `.getOpenTelemetry()`, since that's part of the auto-generated wrapper.

Would you be open to that change?

LikeTheSalad: I see what you mean.

I'm mostly up for consistency, so if we go with that option, then I think we should convert all the getter functions to `val`. I just recalled having a similar conversation [over here](https://github.com/open-telemetry/opentelemetry-android/pull/1279#discussion_r2400119334) when this interface was turned into Kotlin, where we decided to go with getters.

The thing is that, if we convert everything to `val`s, then @fractalwrench's point arises of it not being intuitive in some cases, which I agree with, given that "accessing a field" tends to be perceived as accessing something immutable, wich is not the case for `rumSessionId`. This is the kind of issue that's prone to bikeshedding, though, so I'd like to try and provide some options that I'm fine with, which would hopefully make it easier to come to a conclusion, after getting some input from @fractalwrench or @breedx-splk, and yourself, @beekhc, to help us getting an idea of users expectations for this API.

1. We turn everything to `val`. The `rumSessionId` will change internally, and probably some users will get confused by it. I guess the question here is, how much is that a concern for regular Kotlin users?
2. We leave everything as getters. My question here is, is it really that bad, from a regular Kotlin user's perspective, to call getter functions as functions instead of fields?
3. We mix `val`s and getter functions depending on the needs (as @fractalwrench originally proposed). This won't keep things consistent, which I'd like to do as much as possible, but it's not like I'm willing to die on this hill either, because is more of a nice to have for me. So if the above options cause too much trouble, I'm fine with going with this one.

fractalwrench: My 2 cents is that if a property doesn't change over the process lifecycle accessors should be implemented as a `val`, but if it could change, then it should be implemented as a function. To me, a function semantically suggests that the value can change whereas a `val` _usually_ indicates it won't.

Happy to go with majority opinion

beekhc: I hear you about bike-shedding. I didn't see where this had been discussed before; thank you for the link. I will gladly accept whatever decision the maintainers make.

As for which option to go with, I would mostly defer to the official [kotlin coding conventions](https://kotlinlang.org/docs/coding-conventions.html#functions-vs-properties), which say:

> Prefer a property over a function when the underlying algorithm:
> * Does not throw.
> * Is cheap to calculate (or cached on the first run).
> * Returns the same result over invocations _if the object state hasn't changed_.

[Emphasis mine.]

I think the decision for `rumSessionId` hinges on the interpretation of that italicized part. _Personally_, I think these all warrant being properties. But like I said, I don't wanna bike-shed, so let me know, and either I'll update the PR or we can close it.

LikeTheSalad: Thank you for your patience, @beekhc 

I think it's very helpful to guide ourselves by the official Kotlin conventions, I wasn't aware of it, so thank you for providing the link.

Based on it, I agree that it sounds like mutable properties should be accessed via functions (though they mention some conditional state situation, which is not too clear what they mean by it, i.e. if it's a state that changed from within the object or from the outside, via a setter for example).

I don't have a strong opinion for this particular issue, though I agree with @fractalwrench that accessing a property gives the impression that you're getting something immutable, plus, according to the Kotlin docs, it should also be cheap to calculate or cached, and `rumSessionId` doesn't match those requirements. [This is the underlying function](https://github.com/open-telemetry/opentelemetry-android/blob/4c512e0ff8497edb622d798160e362dd0d8d7082/android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionManager.kt#L38) called from our default getRumSessionId impl.

Having said so, I still would like to provide a nice user experience, as much as possible, so given that property accessors are an important feature for regular Kotlin devs, I think we could at least turn `getOpenTelemetry` into a `val`, as @fractalwrench originally suggested, because it's immutable. And from then on, we can keep some sort of consistency in the form of "`val` by default unless it changes, if it changes, make it a function" which is somewhat backed up by the Kotlin convention docs, which we can link to as well, in case a similar question arises in the future.

If it sounds good for you, @beekhc, I think we can proceed with making`getOpenTelemetry` a val, but leaving `getRumSessionId` as is.

beekhc: Done.

fractalwrench (approved):
LGTM

Issue #1364, opened 2025-11-06
`Clock.getDefault()` isn’t configurable; add platform-aware selection & override hooks


Sabaev:
**Component(s):** `sdk`, `common`, `bootstrap`  
**Environment:** Cross-platform (JVM & Android)

## Summary
`Clock.getDefault()` effectively hardwires `System.nanoTime()`. On Android this ignores deep sleep, leading to incorrect elapsed-time semantics for features like timeouts and background durations. There’s no easy global way to swap the default clock per platform.

## Expected behavior
The SDK should choose a platform-appropriate default clock and allow an explicit override:
- **JVM/server:** `System.nanoTime()` is fine for monotonic elapsed durations.
- **Android:** prefer `SystemClock.elapsedRealtimeNanos()` (includes deep sleep).

## Actual behavior
Users cannot globally configure the default clock. Each component must be hand-wired (if possible), which is fragile and easy to miss.

## Proposed change
Introduce a lightweight, platform-aware `ClockProvider` SPI with discovery (optimized for Android, similar in spirit to kotlinx.coroutines’ [FastServiceLoader](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jvm/src/internal/FastServiceLoader.kt)):

atulgpt:
Hi @Sabaev in the PR https://github.com/open-telemetry/opentelemetry-android/pull/1381 I have created a `AndroidClock` which for android replaces the `Clock.getDefault()` usage with `AndroidClock.INSTANCE` usage. This should fix the handling in session timeout and app startup calculation

Though the usages from java side will not be fixed and in that case your proposed solution of service loader can be used

fractalwrench:
`SystemClock.elapsedRealtimeNanos()` is not suitable as it returns the [time since boot](https://developer.android.com/reference/android/os/SystemClock#elapsedRealtimeNanos()), rather than the wall-clock time. `System.nanoTime()` is suitable for [elapsed durations](https://developer.android.com/reference/java/lang/System#nanoTime()) but doesn't provide a wall-clock time.

[...]

PR #1381, opened 2025-11-11
Add `AndroidClock` which overrides `nanoTime` with `SystemClock.elaps…

atulgpt:
…edRealtimeNanos()`

Fixes https://github.com/open-telemetry/opentelemetry-android/issues/1364

But doesn't make it overridable

### DIFF
--- a/android-agent/build.gradle.kts
+++ b/android-agent/build.gradle.kts
@@ -5,6 +5,11 @@ plugins {
 
 android {
     namespace = "io.opentelemetry.android.agent"
+
+    testOptions {
+        unitTests.isReturnDefaultValues = true
+        unitTests.isIncludeAndroidResources = true
+    }
 }
 
 dependencies {
--- a/common/build.gradle.kts
+++ b/common/build.gradle.kts
@@ -11,6 +11,11 @@ android {
     defaultConfig {
         consumerProguardFiles("consumer-rules.pro")
     }
+
+    testOptions {
+        unitTests.isReturnDefaultValues = true
+        unitTests.isIncludeAndroidResources = true
+    }
 }
 
 dependencies {
--- a/common/src/main/java/io/opentelemetry/android/common/internal/tools/time/AndroidClock.kt
+++ b/common/src/main/java/io/opentelemetry/android/common/internal/tools/time/AndroidClock.kt
@@ -10,8 +10,13 @@ import io.opentelemetry.sdk.common.Clock
 import javax.annotation.concurrent.ThreadSafe
 
 @ThreadSafe
-class AndroidClock : Clock by Clock.getDefault() {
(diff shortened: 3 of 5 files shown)


fractalwrench:
I think we can improve this using the suggestions stated here: https://github.com/open-telemetry/opentelemetry-android/issues/1364#issuecomment-3526746961

atulgpt:
Hi @fractalwrench the definition of `nanoTime` is as below(taken from `io.opentelemetry.sdk.common.Clock`)
```
  /**
   * Returns a time measurement with nanosecond precision that can only be used to calculate elapsed
   * time.
   */
```

so this API is only suitable for calculating the elapsed time. So do we need to do what you are suggesting?

fractalwrench:
`nanoTime` does not guarantee any correlation to wall-clock time. OTLP requires a [timestamp](https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-timestamp) - i.e. elapsed nanoseconds since the Unix epoch.

atulgpt:
`OTLP requires a [timestamp](https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-timestamp) - i.e. elapsed nanoseconds since the Unix epoch.` but for this there are other functions in `io.opentelemetry.sdk.common.Clock` which are `now` and `now(boolean highPrecision)` both returns timestamp in nano since epoch. 

That's why I was asking do we want to change the behavior of the third method `nanoTime` which is documented to use only for elapsed time?

fractalwrench:
I wouldn't want to accept this PR as-is because `elapsedRealTimeNanos()` does not give elapsed nanoseconds since the Unix epoch. I've suggested an alternative approach in this comment: https://github.com/open-telemetry/opentelemetry-android/issues/1364#issuecomment-3526746961

@atulgpt you're welcome to open a PR with that approach, or equally I'd be happy to give it a go. LMK your preference and I'll assign the issue accordingly

atulgpt:
> I wouldn't want to accept this PR as-is because `elapsedRealTimeNanos()` does not give elapsed nanoseconds since the Unix epoch. I've suggested an alternative approach in this comment: [#1364 (comment)](https://github.com/open-telemetry/opentelemetry-android/issues/1364#issuecomment-3526746961)
> 
> @atulgpt you're welcome to open a PR with that approach, or equally I'd be happy to give it a go. LMK your preference and I'll assign the issue accordingly

Yes I would like to try

atulgpt:
Hi @fractalwrench I have updated the code. PTAL Thanks

marandaneto:
replaced by https://github.com/open-telemetry/opentelemetry-android/pull/1487 i think?

fractalwrench:
Closing as #1487 implemented this functionality

PR #1383, opened 2025-11-12
DNM - Convert OpenTelemetryRumBuilder into Kotlin

bidetofevil:
Conversion of the OTRB into Kotlin - for when we can make API breaking changes, which this is

This is not meant to be merged now, but something for later


fractalwrench:
This duplicates #1372, which should retain source compatibility with the previous API

bidetofevil:
Haha, you beat me to it! Nice.

PR #1387, opened 2025-11-13
Move OpenTelemetryRum to a new API module

bidetofevil:
Move the public interface `OpenTelemetryRum` to a new module that all other modules can depend on, and move `core` as a dependency exported transitively via the `android-agent` module. That way, most folks won't be confused with having the choice of using `OpenTelemetryRumInitializer`, which is the preferred, streamlined supported option vs `OpenTelemetryRumBuilder`, the supposedly internal object that we don't consider part of our API.

Folks who need to figure that directly can still do so by directly including the `core` module, but that should not be available by default.

At the same time I changed all the module dependencies to `implementation` as there is no reason to export public symbols of internal modules to the consuming app. Instead, the only transitive dependency that is exported via `android-agent` is the new module, `agent-api`, so `OpenTelemetryRum` is still available. Internally, modules that depend on `agent-api` will pull it its dependencies too, which are the OTel ones.

Note: this is technically an API change in that the static functions on `OpenTelemetryRum` are now on a new object inside `core` called `RumBuilder`, but since this was supposed to NOT be part of the public API, this should be OK.

### DIFF
--- a/core/src/test/java/io/opentelemetry/android/SdkPreconfiguredRumBuilderTest.kt
+++ b/core/src/test/java/io/opentelemetry/android/SdkPreconfiguredRumBuilderTest.kt
@@ -37,7 +37,7 @@ class SdkPreconfiguredRumBuilderTest {
                 }
             }
 
-        val builder = OpenTelemetryRum.builder(context, sdk, config, sessionProvider)
+        val builder = RumBuilder.builder(context, sdk, config, sessionProvider)
         builder.addInstrumentation(fooInstrumentation)
         builder.addInstrumentation(sessionInstrumentation)
 


instrumentation/screen-orientation/build.gradle.kts
```diff
@@ -18,7 +18,6 @@ android {
 }
 
 dependencies {
-    api(project(":instrumentation:android-instrumentation"))
-    api(platform(libs.opentelemetry.platform.alpha))
-    api(libs.opentelemetry.api)
+    implementation(project(":instrumentation:android-instrumentation"))
+    api(project(":agent-api"))
```
fractalwrench: Some instrumentation modules transitively expose this dependency by adding it as `api`, whereas some don't add the dependency at all. Is this truly part of the public API of each of these instrumentation modules or could it be considered an implementation detail that should be added with `implementation`?

demo-app/src/main/java/io/opentelemetry/android/demo/OtelDemoApplication.kt
```diff
@@ -48,26 +46,6 @@ class OtelDemoApplication : Application() {
         } catch (e: Exception) {
             Log.e(TAG, "Oh no!", e)
         }
-
-        // This is needed to get R8 missing rules warnings.
-        initializeOtelWithGrpc()
-    }
-
-    // This is not used but it's needed to verify that our consumer proguard rules cover this use case.
```
fractalwrench, line 56: Does this still need verification?

android-agent/build.gradle.kts
```diff
@@ -8,24 +8,25 @@ android {
...
-    api(project(":instrumentation:activity"))
-    api(project(":instrumentation:anr"))
-    api(project(":instrumentation:crash"))
-    api(project(":instrumentation:fragment"))
-    api(project(":instrumentation:network"))
-    api(project(":instrumentation:slowrendering"))
-    api(project(":instrumentation:startup"))
-    api(project(":instrumentation:sessions"))
-    api(project(":instrumentation:screen-orientation"))
+    implementation(project(":instrumentation:activity"))
```
fractalwrench, line 21: How should folks configure individual instrumentation? For example, if someone wanted to configure `AnrInstrumentation` what would that look like, or is that something that should be answered further down the line?

core/src/main/java/io/opentelemetry/android/RumBuilder.kt
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+package io.opentelemetry.android
+
+import android.content.Context
+import io.opentelemetry.android.config.OtelRumConfig
+import io.opentelemetry.android.session.SessionProvider
+import io.opentelemetry.sdk.OpenTelemetrySdk
+
+object RumBuilder {
```
fractalwrench, line 13: I'm a fan of breaking this out from the interface into a separate symbol, but preferred the name `OpenTelemetryRum.builder` compared to `RumBuilder`

agent-api/api/agent-api.api
```diff
@@ -0,0 +1,8 @@
+public abstract interface class io/opentelemetry/android/OpenTelemetryRum {
```
fractalwrench, line 1: How do folks construct instances of this interface? Ideally there should be some way of initializing from the API module IMO.

instrumentation/screen-orientation/build.gradle.kts
bidetofevil: That's actually true. It doesn't make sense for any instrumentation module to just be included in a project that doesn't have access to the agent, which will include this API by default. I'll change it

demo-app/src/main/java/io/opentelemetry/android/demo/OtelDemoApplication.kt
bidetofevil: I feel like this is a workaround for some classes that should really just have `dontwarn` entries in the proguard files. Regardless, it's not a problem anymore, at least in the building of the demo app, likely because `core` isn't exported by default and included in this app.

android-agent/build.gradle.kts
bidetofevil: If they are using the Initializer, then through the DSL via `InstrumentationConfiguration.anrReporter { }` 😅 

All the instrumentation modules are pretty explicitly tied to the agent and the common instrumentation API, and until that changes, this will just work. To change that relationship, well, that will include changing this as well as other things.

core/src/main/java/io/opentelemetry/android/RumBuilder.kt
bidetofevil: So like an extension function on the `OpenTelemetryRum` interface called `builder()` that returns an `object`? I kind of just stuck this here and am open to other ways of referencing this.

agent-api/api/agent-api.api
bidetofevil: `OpenTelemetryRumInitializer.initialize()` and `OpenTelemetryRumBuilder.build()`

The `agent-api` module is basically just that - `android-agent` is effectively the public implementation that depends on `core` to do the hard work for `agent-api`, and it itself exposes some additional API like the `OpenTelemetryRumInitializer`. There could more work done in this area to consolidate the true public API to be in this module, and have `android-agent` just be an implementation, but that's for later.

core/src/main/java/io/opentelemetry/android/RumBuilder.kt
LikeTheSalad: The extension idea sounds good to me 👍

fractalwrench: My main objection is that `OpenTelemetryRumBuilder` (or similar) was a more specific name than `RumBuilder`. Hanson noted there were some implementation problems on making this an extension function (i.e. it's not possible to add a static function on `OpenTelemetryRum`) so I'd personally be happy with the existing object, renamed

bidetofevil: I think we can stick it in `OpenTelemetryRumBuilder` once that class becomes Kotlin (#1372), but the current state is as good as any right now.

I can raise another PR once both this one and that one are merged to clean it up

bidetofevil: (The extension doesn't really make sense and kind of makes it clunky after I did it - there are really just static factory functions that should be put in an appropriate place and extending one of the parameters don't add a tonne of value IMO)

bidetofevil:
There's some build errors after the merge. I'll fix that now.

PR #1389, opened 2025-11-14
feat: Add customizers for processors and readers

agrognetti:
This change introduces new customization points in `OpenTelemetryRumBuilder` for more advanced configuration of the OTel SDK.

It is now possible to:
* Customize the `BatchSpanProcessorBuilder`, `BatchLogRecordProcessorBuilder`, and `PeriodicMetricReaderBuilder` before the components are built.
* Wrap or replace the final `SpanProcessor`, `LogRecordProcessor`, and `MetricReader` instances.

This allows for intercepting and modifying telemetry signals without needing to add extra processors or readers via the provider customizers.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -38,7 +38,7 @@ jobs:
         run: touch demo-app/local.properties
 
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@0499de31b99561a6d14a36a5f662c2a54f91beee # v4.31.2
+        uses: github/codeql-action/init@014f16e7ab1402f30e7c3329d33797e7948572db # v4.31.3
         with:
           languages: java, actions
           # using "linked" helps to keep up with the latest Kotlin support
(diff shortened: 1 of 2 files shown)


breedx-splk:
Hey @agrognetti thanks for this PR, and so sorry that it's taken this long to reply to you. We are in the process of releasing our first stable 1.0.0 version, and that effort has grabbed quite a bit of our attention. 

It looks like your changes have a conflict now, largely due to the `OpenTelemetryRumBuilder` have been migrated to kotlin from java. Sorry that got changed out from under you...but this really cannot be merged (or reviewed) in its current state now.

I think this PR is probably a welcome change for advanced users who use the core module (via `OpenTelemetryRumBuilder`) directly, but I think it could also be helpful for us to hear about your specific use cases. What, specifically, are you hoping to accomplish with these customizers that could not be handled as easily with the existing mechanisms? 

It would be helpful for us to understand this, because we intend for the `OpenTelemetryRumBuilder` to really only be used in advanced cases and not required for most users. Most users should just use the `OpenTelemetryRumInitializer`. When additional advanced customizations are needed, the `OpenTelemetryRumBuilder` should already provide a way to do virtually all programmatic SDK customizations that are available...but I get the impression that you wanted these to be easier? I worry slightly that these additional extension points might encourage users to use this, instead of the initializer.

Your feedback is helpful, so thanks in advance!

breedx-splk:
Another small piece of feedback -- even tho this PR isn't that huge, it might be helpful to break this into 3 separate PRs, each with their specific customization responsibility. Not required, just an idea to help things flow more easily.

PR #1394, opened 2025-11-17
Chore/migrate to kt

atulgpt:
Migrating one TC file to kt so that mockito can be removed
This was mentioned in https://github.com/open-telemetry/opentelemetry-android/pull/1362#issuecomment-3495345400

This will unblock https://github.com/open-telemetry/opentelemetry-android/issues/1247


instrumentation/slowrendering/src/test/java/io/opentelemetry/android/instrumentation/slowrendering/SlowRenderListenerTest.kt
```diff
@@ -0,0 +1,289 @@
...
+
+        testInstance.onActivityResumed(activity)
+
+        Mockito
+            .verify(activity.window)
+            .addOnFrameMetricsAvailableListener(
+                activityListenerCaptor.capture(),
+                ArgumentMatchers.eq(frameMetricsHandler),
+            )
+        Assert.assertEquals(
```
breedx-splk: We prefer the assertj style assertions, can you modify this to use assertj please?

instrumentation/slowrendering/src/test/java/io/opentelemetry/android/instrumentation/slowrendering/SlowRenderListenerTest.kt
```diff
@@ -0,0 +1,289 @@
...
+import org.junit.runner.RunWith
+import org.mockito.Answers
+import org.mockito.ArgumentCaptor
+import org.mockito.ArgumentMatchers
+import org.mockito.Captor
+import org.mockito.Mock
+import org.mockito.Mockito
+import org.mockito.junit.MockitoJUnit
+import org.mockito.junit.MockitoRule
+import org.mockito.stubbing.Answer
```
breedx-splk: The PR and the issue you referenced in the PR description both make mention of our desire to move away from mockito in favor of mockk. Would it be possible for you to do that conversion as part of this PR as well? Thanks!

instrumentation/okhttp3/testing/src/androidTest/java/io/opentelemetry/instrumentation/library/okhttp/v3_0/InstrumentationTest.kt
```diff
@@ -0,0 +1,186 @@
...
+        span.makeCurrent().use { ignored ->
+            val client =
+                OkHttpClient
+                    .Builder()
+                    .addInterceptor(
+                        Interceptor { chain: Interceptor.Chain? ->
+                            val currentSpan = Span.current().spanContext
+                            Assert.assertEquals(
+                                span.spanContext.traceId,
+                                currentSpan.traceId,
```
breedx-splk: It was already structured this way, so you're not making it worse or anything, but any time I see assertions inside of callback/async/conditional code I'm leery, because there can be unexpected cases where the callback (in this case an interceptor) doesn't get invoked at all and the test still passes without the assertion being invoked.

If we're going to keep that around, then I think there also needs to be a side effect that we can also assert on to ensure that the assertion was invoked (I've frequently used countdown latches for such a thing).

instrumentation/okhttp3/testing/src/androidTest/java/io/opentelemetry/instrumentation/library/okhttp/v3_0/InstrumentationTest.kt
```diff
@@ -54,37 +54,39 @@ class InstrumentationTest {
...
                     .addInterceptor(
-                        Interceptor { chain: Interceptor.Chain? ->
+                        Interceptor { chain: Interceptor.Chain ->
                             val currentSpan = Span.current().spanContext
-                            Assert.assertEquals(
-                                span.spanContext.traceId,
-                                currentSpan.traceId,
-                            )
-                            chain!!.proceed(chain.request())
+                            Assertions
```
breedx-splk: Let's use the static import for these, because it reads nicer.

instrumentation/slowrendering/src/test/java/io/opentelemetry/android/instrumentation/slowrendering/SlowRenderListenerTest.kt
```diff
@@ -254,19 +213,19 @@ class SlowRenderListenerTest {
             }.collect(Collectors.toList())
```
breedx-splk: Now that we're in kotlin, using `Stream` for this should be unnecessary. You can do something more like 

```kotlin
listOf(
    5L, 11L, 101L, // slow
    701L, // frozen
    17L, // slow
    17L, // slow
    16L, 11L)
.map { it.milliseconds.inWholeNanoseconds }
```

PR #1399, opened 2025-11-18
Add example to initialize agent to the README

bidetofevil:
Add example for initialization using the `OpenTelemetryRumInitializer` to the main README

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,53 @@ dependencies {
 }
 ```
 
+## Agent Initialization
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -43,6 +43,53 @@ dependencies {
...
+            context = context,
+            configuration = {
+                httpExport {
+                    baseUrl = "http://10.0.2.2:4318"
+                    baseHeaders = mapOf("foo" to "bar")
+                }
+                instrumentations {
+                    activity {
+                        enabled(true)
+                    }
```
breedx-splk, line 72: Maybe a comment that explains that this wouldn't really be needed because it's enabled by default?

breedx-splk:
We can add the comment later.

PR #1400, opened 2025-11-18
Add docs for Kotlin First policy

bidetofevil:
Add some documentation about the project's policy of being Kotlin First

### DIFF
--- a/docs/KOTLIN_FIRST.md
+++ b/docs/KOTLIN_FIRST.md
@@ -1,17 +1,27 @@
 # Kotlin-First Policy
 
-The OpenTelemetry Android Agent was initially built when Java was the predominant language for Android development. Over the years, the platform has shifted towards being one that is primarily Kotlin, so this project will follow suit. This means that we consider the primary use case of this project to be apps written in Kotlin by folks who are familiar with Android development in Kotlin.
+The OpenTelemetry Android Agent was initially built when Java was the predominant language for Android development. Over the years, the 
+platform has shifted towards being one that is primarily Kotlin, so this project will follow suit. This means that we consider the primary 
+use case of this project to be apps written in Kotlin by folks who are familiar with Android development in Kotlin.
 
-In practice, we want our API and code to be structured with the Kotlin language and idioms in mind. This will allow the project to feel more native and be more readable to modern Android developers, thereby reduce one layer of friction of adopting OpenTelemetry: the Java-ness of it all.
+In practice, we want our API and code to be structured with the Kotlin language and idioms in mind. This will allow the project to feel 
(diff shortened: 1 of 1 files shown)


docs/KOTLIN_FIRST.md
```diff
@@ -0,0 +1,17 @@
+# Kotlin-First Policy
+
+The OpenTelemetry Android Agent was initially built when Java was the predominant language for Android development. Over the years, the platform has shifted towards being one that is primarily Kotlin, so this project will follow suit. This means that we consider the primary use case of this project to be apps written in Kotlin by folks who are familiar Android development in Kotlin.
```
breedx-splk:
```suggestion
The OpenTelemetry Android Agent was initially built when Java was the predominant language for Android development. Over the years, the platform has shifted towards being one that is primarily Kotlin, so this project will follow suit. This means that we consider the primary use case of this project to be apps written in Kotlin by folks who are familiar with Android development in Kotlin.
```

docs/KOTLIN_FIRST.md
```diff
@@ -0,0 +1,17 @@
...
+
+## OpenTelemetry API Compatibility
+
+OpenTelemetry Android Agent is currently built on the OTel Java SDK and ecosystem. It also exports the OTel Java API for recording OTel signals. This will be the case until official Kotlin versions are released and stabilized under the OpenTelemetry org, which is currently under development.
+
+## Java Compatibility
+
+Despite this policy, the Agent API should still be usable from Java, but they will not be optimized for it. For instance, a method that uses default arguments in Kotlin will require Java users to pass them in explicitly if no reasonable overloads can be provided.
+
+As the API evolves, we will endeavor to preserve Java compile and runtime compatibility as much as possible. But we will only consider a change "breaking" if it affects the Kotlin use-case. As such, we advise that all calls to the Agent API are made from Kotlin if possible to prevent syntactic changes from breaking your app's build.
```
breedx-splk: So I'd like to see this changed so that we explain that we do NOT make any efforts to preserve Java compatibility, and the Java use case is "unsupported" -- which means that issues should not be filed and we will not accept PRs to address Java language compatibility issues at the API or configuration layers.

bidetofevil: Even stronger stance. OK! I take @LikeTheSalad's thumbs up on his agreement.

bidetofevil:
> Thanks @bidetofevil, this helps! I had a few small comments.
> 
> Also, can you please add hard line breaks in the markdown? It makes it easier to review and add suggestions without having to suggest the whole darn paragraph. 🙏🏻 We should have a markdown linter one day that helps with this (other projects have it), but alas we do not yet.

I was hoping spotless would do this for me. Alas....

PR #1408, opened 2025-11-19
Feature/compose click name from modifiers

Mateusz-Stasielowicz:
This PR enchances the compose-click module for better detection of the name of tapped compose element.

It introduces two ways of detecting name of the tapped element:

1. `Modifier.opentelemetry` - introduces a custom Modifier (similar solution is used by other telemetry SDKs like Datadog and Sentry. Developers can use this modifier to mark a composable element and give it a name. Then the compose-click module will be able to easily find such tapped element and retrieve a name from it.
2. `Modifier.testTag` - handling as a fallback to read the TestTag value of tapped element. TestTag is often used for Unit tests and therefore it is useful to take it during composable element name discovery.

### DIFF
--- a/instrumentation/compose/click/src/main/kotlin/io/opentelemetry/instrumentation/compose/click/ComposeTapTargetDetector.kt
+++ b/instrumentation/compose/click/src/main/kotlin/io/opentelemetry/instrumentation/compose/click/ComposeTapTargetDetector.kt
@@ -138,7 +138,7 @@ internal class ComposeTapTargetDetector(
                 className = modifier::class.qualifiedName
             }
             val testTag = modifier.getTestTag()
-            if (!testTag.isNullOrBlank()) {
+            if (testTag != null) {
                 return testTag
             }
         }
--- a/instrumentation/compose/click/src/test/kotlin/io/opentelemetry/instrumentation/compose/click/ComposeInstrumentationTest.kt
+++ b/instrumentation/compose/click/src/test/kotlin/io/opentelemetry/instrumentation/compose/click/ComposeInstrumentationTest.kt
@@ -97,6 +97,74 @@ internal class ComposeInstrumentationTest {
 
     @Test
     fun capture_compose_click() {
+        val motionEvent =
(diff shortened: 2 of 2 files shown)


fractalwrench:
Hey @Mateusz-Stasielowicz - as this is marked as draft I'm assuming it doesn't need review yet. Please comment here or mark it as ready for review when the time comes.

instrumentation/compose/click/src/main/kotlin/io/opentelemetry/instrumentation/compose/click/OpentelemetryModifier.kt
```diff
@@ -0,0 +1,33 @@
...
+import androidx.compose.ui.semantics.semantics
+
+/**
+ * An opentelemetry Modifier that allows to mark a composable element as traceable.
+ * When a composable element that has this Modifier will be tapped, then the [name] from this
+ * Modifier will be taken as the element name
+ *
+ * @param name name of the tapped element
+ */
+fun Modifier.opentelemetry(name: String): Modifier =
```
marandaneto, line 20:
```suggestion
fun Modifier.openTelemetry(name: String): Modifier =
```

not sure if we do this everywhere, but i always prefer camel case

instrumentation/compose/click/src/main/kotlin/io/opentelemetry/instrumentation/compose/click/TestTagFinder.kt
```diff
@@ -0,0 +1,42 @@
...
+    if ("androidx.compose.ui.platform.TestTagElement" == modifier::class.qualifiedName) {
+        try {
+            val testTagField = modifier::class.java.getDeclaredField(TEST_TAG_FIELD_NAME)
+            testTagField.isAccessible = true
+            val testTag = testTagField.get(modifier) as String
+            if (testTag.isNotEmpty()) {
+                return testTag
+            }
+        } catch (_: Exception) {
+        }
```
marandaneto, line 39: will that work if the app is minified?

breedx-splk:
@Mateusz-Stasielowicz please feel free to reopen if/when you can come back to this. Thanks again!

PR #1413, opened 2025-11-20
Force session instrumentation to be installed first when available.

### LINKED ISSUE #1396: `session.start` doesn't get fired when `ActivityLifecycleInstrumentation` appears before `SessionInstrumentation`
This is because, when `ActivityLifecycleInstrumentation` `install` gets called before which starts the `startupTimer` timer, and at that time     override fun addObserver(observer: SessionObserver) { is not populated in [`SessionManager`](https://github.com/open-telemetry/opentelemetry-android/blob/main/android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionManager.kt)

When `SessionInstrumentation` gets its `install` and then add the observers, it already became too late as now we don't call `onSessionEnded` and `onSessionStarted`

One related issue(though I think it doesn't mention this issue): https://github.com/open-telemetry/opentelemetry-android/issues/781

breedx-splk:
Resolves #1396


fractalwrench (approved):
LGTM

LikeTheSalad (approved):

Issue #1311, opened 2025-10-15
Feature Request: compose/click module could use a TestTag semantic property as a fallback to set clickable element name


Mateusz-Stasielowicz:
A proposal to use the `SemanticsProperties.TestTag`  as a fallback in `compose/click` module to discover clicked node name and send it as app.widget.name.

Currently only `SemanticsActions.OnClick` and `SemanticsProperties.ContentDescription` are discovered on clickable composable element to get the name of clicked widget. 

In many cases the `TestTag` is used on the clickable elements for further testing purposes and it would be nice to have this also as an option in open-telemetry to not duplicate multiple semantic properties with the same value.

code snippet
```
val testTagSemanticsConfiguration = getOrNull(SemanticsProperties.TestTag)
if (testTagSemanticsConfiguration.isNullOrBlank().not()) {
    return testTagSemanticsConfiguration
}
```

breedx-splk:
Thanks @Mateusz-Stasielowicz. Just to be clear -- you want this as a fallback and not the primary, correct? In other words, continue using the existing`SemanticsProperties.OnClick` and `SemanticsProperties.ContentDescription`, but also check for the `SemanticsProperties.TestTag` before resorting to the class name? Do I have that right?

Mateusz-Stasielowicz:
> Thanks [@Mateusz-Stasielowicz](https://github.com/Mateusz-Stasielowicz). Just to be clear -- you want this as a fallback and not the primary, correct? In other words, continue using the existing`SemanticsProperties.OnClick` and `SemanticsProperties.ContentDescription`, but also check for the `SemanticsProperties.TestTag` before resorting to the class name? Do I have that right?

Yes that was the first idea. I will revisit the requirement on my side. Need to check if the `TestTag` could the primary option. 
I proposed to use the TestTag as a fallback to keep the backward compatibility, but as current state of the `opentelemetry-android` is alpha, it could be also used as the primary one.

Mateusz-Stasielowicz:
proposed implementation: https://github.com/open-telemetry/opentelemetry-android/pull/1415

PR #1415, opened 2025-11-20
Feature/compose click node name from modifier

Mateusz-Stasielowicz:
This PR enchances the compose-click module for better detection of the name of tapped compose element.

It introduces two ways of detecting name of the tapped element:

1. `Modifier.opentelemetry` - introduces a custom Modifier (similar solution is used by other telemetry SDKs like Datadog and Sentry. Developers can use this modifier to mark a composable element and give it a name. Then the compose-click module will be able to easily find such tapped element and retrieve a name from it.
2. `Modifier.testTag` - handling as a fallback to read the TestTag value of tapped element. TestTag is often used for Unit tests and therefore it is useful to take it during composable element name discovery.

### DIFF
--- a/instrumentation/compose/click/api/click.api
+++ b/instrumentation/compose/click/api/click.api
@@ -9,3 +9,7 @@ public final class io/opentelemetry/instrumentation/compose/click/ComposeLayoutN
 	public static final field VIEW_CLICK_EVENT_NAME Ljava/lang/String;
 }
 
+public final class io/opentelemetry/instrumentation/compose/click/OpentelemetryModifierKt {
+	public static final fun opentelemetry (Landroidx/compose/ui/Modifier;Ljava/lang/String;)Landroidx/compose/ui/Modifier;
+}
+
--- a/instrumentation/compose/click/src/main/kotlin/io/opentelemetry/instrumentation/compose/click/ComposeTapTargetDetector.kt
+++ b/instrumentation/compose/click/src/main/kotlin/io/opentelemetry/instrumentation/compose/click/ComposeTapTargetDetector.kt
@@ -67,17 +67,15 @@ internal class ComposeTapTargetDetector(
     ): LayoutNode? {
         val queue = LinkedList<LayoutNode>()
         queue.addFirst(owner.root)
-        var target: LayoutNode? = null
-
         while (queue.isNotEmpty()) {
             val node = queue.removeFirst()
             if (node.isPlaced && hitTest(node, x, y)) {
-                target = node
+                return node
             }
 
             queue.addAll(node.zSortedChildren.asMutableList())
         }
-        return target
+        return null
     }
 
     private fun isValidClickTarget(node: LayoutNode): Boolean {
@@ -85,6 +83,9 @@ internal class ComposeTapTargetDetector(
             val modifier = info.modifier
             if (modifier is SemanticsModifier) {
                 with(modifier.semanticsConfiguration) {
+                    if (contains(OpentelemetrySemanticsPropertyKey)) {
+                        return true
+                    }
                     if (contains(SemanticsActions.OnClick)) {
                         return true
                     }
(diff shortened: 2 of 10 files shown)


breedx-splk:
Relates to #1311.

Does it resolve it? Seems like it probably does....

instrumentation/compose/click/src/main/kotlin/io/opentelemetry/instrumentation/compose/click/ComposeTapTargetDetector.kt
```diff
@@ -67,24 +67,25 @@ internal class ComposeTapTargetDetector(
...
             val node = queue.removeFirst()
             if (node.isPlaced && hitTest(node, x, y)) {
-                target = node
+                return node
             }
 
             queue.addAll(node.zSortedChildren.asMutableList())
         }
-        return target
+        return null
```
breedx-splk, line 78: At first glance, this appears to be a semantic change. Previously, the code would keep adding children and walking the queue until it found the "last" one that matched -- which I think means that it potentially got more specific. So was this change intentional, and if so, why?

instrumentation/compose/click/src/test/kotlin/androidx/compose/ui/platform/TestTagElement.kt
```diff
@@ -0,0 +1,14 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package androidx.compose.ui.platform
+
+import androidx.compose.ui.Modifier
+
+// Test-only stand-in for the private Compose TestTagElement.
+// Must have a private field named `tag` so reflection in the production code finds it.
+class TestTagElement(
```
breedx-splk, line 12: This is interesting. Are users supposed to be able to consume this, because I don't think that this would be published in any consumable artifacts. Maybe it should just be in the `main` src tree, under some `test` package, rather than test source?

instrumentation/compose/click/src/test/kotlin/io/opentelemetry/instrumentation/compose/click/ComposeClickEventGeneratorTest.kt
```diff
@@ -284,7 +286,7 @@ internal class ComposeClickEventGeneratorTest {
         }
 
         every { nodeList[0].zSortedChildren } returns mutableVectorOf(nodeList[1], nodeList[2])
-        every { nodeList[1].zSortedChildren } returns mutableVectorOf(nodeList[4], nodeList[3])
+        every { nodeList[1].zSortedChildren } returns mutableVectorOf(nodeList[3], nodeList[4])
```
cleverchuk, line 289: wondering what triggered this change because the idea is that 3 is higher than 4 in z-order and swapping them means 4 is now on top and the assertion on lines `194-199` should fail?

instrumentation/compose/click/api/click.api
```diff
@@ -9,3 +9,7 @@ public final class io/opentelemetry/instrumentation/compose/click/ComposeLayoutN
 	public static final field VIEW_CLICK_EVENT_NAME Ljava/lang/String;
 }
 
+public final class io/opentelemetry/instrumentation/compose/click/OpentelemetryModifierKt {
```
fractalwrench, line 12: Could this be internal to avoid exposing it to anybody using this module, or does it need to be public?

instrumentation/compose/click/src/main/kotlin/io/opentelemetry/instrumentation/compose/click/ComposeTapTargetDetector.kt
Mateusz-Stasielowicz: Yes this change was intentional. For me it seems to be a bug to take the last node. 

In my test application when a compose component consists of multiple nodes (i.e. i'm using some UI components from some 3rd party library), this code caused to take the node from the library, but not the node that I define in my application.

In my opinion it should get the top most node.

instrumentation/compose/click/src/test/kotlin/androidx/compose/ui/platform/TestTagElement.kt
Mateusz-Stasielowicz: This is for internal unit test of opentelemetry-android compose-click module

instrumentation/compose/click/api/click.api
Mateusz-Stasielowicz: It has to be public as this is meant to be used on jetpack compose elements defined by application

instrumentation/compose/click/src/test/kotlin/io/opentelemetry/instrumentation/compose/click/ComposeClickEventGeneratorTest.kt
Mateusz-Stasielowicz: Let me check the unit tests again

breedx-splk: @Mateusz-Stasielowicz did you have a chance to check?

PR #1419, opened 2025-11-22
Session Manager Concurrency Fix

gregory-at-cvs:
PR #1: Session Manager Concurrency Fix
## Description

The changes fix a thread-safety issue in SessionManager where concurrent calls to `getSessionId()` 
could create duplicate sessions during session timeout/expiration.

## Problem
The existing code had a documented TODO acknowledging this race condition:
```kotlin
// TODO FIXME: This is not threadsafe -- if two threads call getSessionId()
// at the same time while timed out, two new sessions are created
```

## Solution:

1. Changed session from mutable variable to AtomicReference<Session>
2. Implemented compare-and-set for atomic session updates
3. Only one thread successfully creates new session; others use the newly created one
4. Extracted observer notification into separate method for clarity
5. Added  concurrency tests

## Related PRs
This is part 1 of 3 in a session management enhancement:

This PR: Thread-safety fix (foundation)
[PR 2: Session infrastructure](https://github.com/open-telemetry/opentelemetry-android/pull/1420) (depends on this)
[PR 3: Session telemetry integration](https://github.com/open-telemetry/opentelemetry-android/pull/1421) (depends on [PR 2](https://github.com/open-telemetry/opentelemetry-android/pull/1420))
**Review order:** This PR should be reviewed and merged first.

## Type of Change
[x] Bug fix (non-breaking change which fixes an issue)
[ ] New feature
[ ] Breaking change
[x] Documentation update

## Checklist
[x] Code follows project style guidelines (spotless applied)
[x] Self-review completed
[x] Added tests that prove the fix is effective
[x] New and existing tests pass locally
[x] Documentation updated where applicable

## Additional Context
This uses the atomic compare-and-set pattern commonly used for thread-safe operations in Java/Kotlin.

Part of a larger effort: We're contributing session management enhancements to bring Android SDK to parity with iOS (spans/logs) and beyond (metrics). This PR is the foundation - fixing a known thread-safety issue before building on it.

Related information: https://opentelemetry.io/docs/specs/semconv/general/session/

## Related PRs

This is **part 1 of 3** in a comprehensive session management enhancement:
- **This PR:** Thread-safety fix (foundation)
- [PR 2: Session infrastructure (depends on this)](https://github.com/open-telemetry/opentelemetry-android/pull/1420)
- [PR 3: Session telemetry integration ](https://github.com/open-telemetry/opentelemetry-android/pull/1421)(depends on PR 2)

**Review order:** This PR should be reviewed and merged first, as PRs 2 and 3

### DIFF
--- a/android-agent/src/test/kotlin/io/opentelemetry/android/agent/session/SessionManagerTest.kt
+++ b/android-agent/src/test/kotlin/io/opentelemetry/android/agent/session/SessionManagerTest.kt
@@ -159,7 +159,7 @@ internal class SessionManagerTest {
         val sessionId =
             SessionManager(
                 timeoutHandler = timeoutHandler,
-                maxSessionLifetime = MAX_SESSION_LIFETIME.hours
+                maxSessionLifetime = MAX_SESSION_LIFETIME.hours,
             )
 
         // When - access session ID twice, should be same
@@ -203,15 +203,16 @@ internal class SessionManagerTest {
         val sessionIdCount = AtomicInteger(0)
 
         // When - multiple threads access session concurrently after timeout
-        val params = AddSessionIdsParameters(
-            numThreads,
-            executor,
-            firstLatch,
-            lastLatch,
-            sessionManager,
-            sessionIds,
-            sessionIdCount
-        )
+        val params =
+            AddSessionIdsParameters(
+                numThreads,
+                executor,
+                firstLatch,
+                lastLatch,
+                sessionManager,
+                sessionIds,
+                sessionIdCount,
+            )
         addSessionIdsAcrossThreads(params)
 
         val isCountZero = lastLatch.await(SESSION_AWAIT_SECONDS, TimeUnit.SECONDS)
@@ -245,15 +246,16 @@ internal class SessionManagerTest {
         val sessionIdCount = AtomicInteger(0)
 
         // When - multiple threads access session with timeout handler indicating timeout
-        val params = AddSessionIdsParameters(
-            numThreads,
-            executor,
-            firstLatch,
-            lastLatch,
-            sessionManager,
-            sessionIds,
-            sessionIdCount
-        )
+        val params =
+            AddSessionIdsParameters(
+                numThreads,
+                executor,
+                firstLatch,
+                lastLatch,
+                sessionManager,
+                sessionIds,
+                sessionIdCount,
+            )
         addSessionIdsAcrossThreads(params)
 
         val isCountZero = lastLatch.await(SESSION_AWAIT_SECONDS, TimeUnit.SECONDS)
@@ -283,15 +285,16 @@ internal class SessionManagerTest {
         val sessionIdCount = AtomicInteger(0)
 
         // When - multiple threads access session concurrently
-        val params = AddSessionIdsParameters(
-            numThreads,
-            executor,
-            firstLatch,
-            lastLatch,
-            sessionManager,
-            sessionIds,
-            sessionIdCount
-        )
+        val params =
+            AddSessionIdsParameters(
+                numThreads,
+                executor,
+                firstLatch,
+                lastLatch,
+                sessionManager,
+                sessionIds,
+                sessionIdCount,
+            )
         addSessionIdsAcrossThreads(params)
 
         val isCountZero = lastLatch.await(SESSION_AWAIT_SECONDS, TimeUnit.SECONDS)
@@ -354,9 +357,7 @@ internal class SessionManagerTest {
         assertThat(thirdSessionId).isNotEqualTo(secondSessionId)
     }
 
-    private fun addSessionIdsAcrossThreads(
-        params: AddSessionIdsParameters
-    ) {
+    private fun addSessionIdsAcrossThreads(params: AddSessionIdsParameters) {
         repeat(params.numThreads) {
             params.executor.submit {
                 try {


android-agent/src/test/kotlin/io/opentelemetry/android/agent/session/SessionManagerTest.kt
```diff
@@ -37,28 +48,42 @@ internal class SessionManagerTest {
 
     @Test
     fun valueValid() {
+        // Verifies that generated session IDs are valid 32-character hex strings
```
breedx-splk: Rather than having these comments in the test methods, maybe we can just rename the methods themselves to be more accurate. So something like

```suggestion
fun `verify session id hex string format`() {
```

and similar for the rest of them. It's a small balance between being too verbose and too general/vague though.

android-agent/src/test/kotlin/io/opentelemetry/android/agent/session/SessionManagerTest.kt
```diff
@@ -37,28 +48,42 @@ internal class SessionManagerTest {
...
-                maxSessionLifetime = 4.hours,
+                maxSessionLifetime = MAX_SESSION_LIFETIME.hours,
             )
         val sessionId = sessionManager.getSessionId()
-        assertThat(sessionId).isNotNull()
-        assertThat(sessionId).hasSize(32)
-        assertThat(Pattern.compile("[a-f0-9]+").matcher(sessionId).matches()).isTrue()
+
+        // Then
+        assertAll(
```
breedx-splk: We don't generally use `assertAll` for tests like this....can you explain what the value is? Does it just give the potential to have multiple failures or something? Doesn't seem that valuable...and I'd drop it to match the rest of the repo style.

android-agent/src/test/kotlin/io/opentelemetry/android/agent/session/SessionManagerTest.kt
```diff
@@ -125,18 +169,170 @@ internal class SessionManagerTest {
...
         every { timeoutHandler.hasTimedOut() } returns true
 
+        // Then - should create new session
         assertThat(value).isNotEqualTo(sessionId.getSessionId())
         verify(exactly = 3) { timeoutHandler.bump() }
     }
+
+    @Test
+    fun `concurrent access during timeout should create only one new session`() {
+        // Verifies that concurrent access during session timeout creates exactly one new session
```
breedx-splk: Similar to my suggestion above, I think the descriptive name of this method makes the comment redundant.

```suggestion
    fun `concurrent access during timeout should create only one new session`() {
```

android-agent/src/test/kotlin/io/opentelemetry/android/agent/session/SessionManagerTest.kt
```diff
@@ -125,18 +169,170 @@ internal class SessionManagerTest {
...
+        clock.advance(5, TimeUnit.HOURS)
+
+        val numThreads = 10
+        val executor = Executors.newFixedThreadPool(numThreads)
+        val latch = CountDownLatch(numThreads)
+        val sessionIds = mutableSetOf<String>()
+        val sessionIdCount = AtomicInteger(0)
+
+        // When - multiple threads access session concurrently after timeout
+        repeat(numThreads) {
```
breedx-splk: I love this test, thanks for putting the time in to write this up!

One small nit is that these tasks get submitted serially to the thread pool, and it's possible for one or more of the tasks to be finished before the others have even been created, let alone started.

At the expense of test complexity, the race condition could be made considerably tighter by using a second `CountDownLatch(numThreads)` that the tasks themselves count on and _also_ wait on before getting the session id. Hope that makes sense.

android-agent/src/test/kotlin/io/opentelemetry/android/agent/session/SessionManagerTest.kt
```diff
@@ -125,18 +169,170 @@ internal class SessionManagerTest {
...
+
+        assertThat(latch.await(5, TimeUnit.SECONDS)).isTrue()
+        executor.shutdown()
+
+        // Then - verify the expected number of session IDs
+        assertAll(
+            "Race condition validation",
+            { assertThat(sessionIdCount.get()).isEqualTo(numThreads) },
+            { assertThat(sessionIds).isNotEmpty() },
+            { assertThat(sessionIds.size).isLessThanOrEqualTo(numThreads) },
```
breedx-splk: Why is less than valid here? If the timeout is always true, shouldn't the number of sessions be equal to the number of threads?

android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionManager.kt
```diff
@@ -23,51 +24,58 @@ internal class SessionManager(
...
     }
 
     override fun getSessionId(): String {
-        // value will never be null
-        var newSession = session
+        val currentSession = session.get()
 
-        if (sessionHasExpired() || timeoutHandler.hasTimedOut()) {
+        // Check if we need to create a new session.
+        return if (sessionHasExpired() || timeoutHandler.hasTimedOut()) {
```
breedx-splk: Can we now please modify the signature of `sessionHasExpired()` to take the `currentSession` as a parameter, rather than having it reach back into the atomic? I think it will be easier to reason about that way, and it reduces the chance of the method looking at a different value.

gregory-at-cvs: Good eyes and points. I'll make the change.

android-agent/src/test/kotlin/io/opentelemetry/android/agent/session/SessionManagerTest.kt
gregory-at-cvs: Sounds good 👍   Yeah, the balance can become a fine line. :)

android-agent/src/test/kotlin/io/opentelemetry/android/agent/session/SessionManagerTest.kt
gregory-at-cvs: I find its usage to be valuable in getting the complete success or failure information in one run  to avoid repeated runs.  Every condition will get run even if one fails. That way cycles are reduced debugging/traiging issues. 

I will certainly make the change to keep consistency with the existing codebase.

android-agent/src/test/kotlin/io/opentelemetry/android/agent/session/SessionManagerTest.kt
gregory-at-cvs: Thanks. I'm glad you caught that.
I'll add in the change.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionManager.kt
gregory-at-cvs: Done

android-agent/src/test/kotlin/io/opentelemetry/android/agent/session/SessionManagerTest.kt
gregory-at-cvs: I fixed those up.

android-agent/src/test/kotlin/io/opentelemetry/android/agent/session/SessionManagerTest.kt
gregory-at-cvs: Done

android-agent/src/test/kotlin/io/opentelemetry/android/agent/session/SessionManagerTest.kt
gregory-at-cvs: In addition to adding the second latch, I realized there was an additional opportunity for reusability across tests that I hadn't taken care of yet. I made that change as well.

android-agent/src/test/kotlin/io/opentelemetry/android/agent/session/SessionManagerTest.kt
gregory-at-cvs: You're right to question that. 👍 Thanks. I should have removed that one earlier. It's done now.

android-agent/src/test/kotlin/io/opentelemetry/android/agent/session/SessionManagerTest.kt
gregory-at-cvs: Fixed

gregory-at-cvs:
Hey @breedx-splk 👋 
The PR is ready for you to review again.

breedx-splk (commented):
Thanks again for this fix!

breedx-splk (approved):

PR #1429, opened 2025-11-25
Fix dependencies for release

breedx-splk:
_Note: This is a PR into the v1.0.x release branch._

In #1387, we somehow overlooked the removal of all the `api(platform(bom))` dependencies, which [breaks the release process](https://github.com/open-telemetry/opentelemetry-android/actions/runs/19681990167/job/56378008928#step:9:1913). I hypothesized previously (when I discovered this in the past) that the updated sonatype publishing step has some additional checks to ensure that independent artifact versions have a complete pom -- one that allows for clear, unambiguous resolution of all dependencies with version information.

Once this merges and the release is complete, we can cherry pick this back to the main branch.

### DIFF
--- a/agent-api/build.gradle.kts
+++ b/agent-api/build.gradle.kts
@@ -3,6 +3,8 @@ plugins {
     id("otel.publish-conventions")
 }
 
+description = "OpenTelemetry android agent api"
+
 android {
     namespace = "io.opentelemetry.android.agent.api"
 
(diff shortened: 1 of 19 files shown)


bidetofevil (approved):
LGTM. If the BOM can't be transitively imported, we should add it with a script or something that brings in all the deps each module should have. We can do that in main later.

PR #1436, opened 2025-12-01
chore: add license scanning url to clomonitor config

thompson-tomo:
This adds a link to the fossa scan reports to satisfy the license scan check for clo monitor as identified by https://clomonitor.io/projects/cncf/open-telemetry#opentelemetry-android.


breedx-splk:
There is an example of this here: https://github.com/cncf/clomonitor/blob/7dd0fddb8418839e4aea3547463fad8911c3165d/docs/metadata/.clomonitor.yml#L10

I couldn't easily find a doc that describes the config yaml mappings into the respective features, but this looks sane. 🤷🏻

thompson-tomo:
Yeap that is the example I followed and just checked clo monitor and it is being recognised.

Issue #1434, opened 2025-11-29
we are using `opentelemetry-sdk-testing` in our otel plugins but we are not adding the dpendency of `opentelemetry-platform-alpha`


atulgpt17:
In the files, `otel.android-app-conventions.gradle.kts` and `otel.android-library-conventions.gradle.kts` we are using below dependency

https://github.com/open-telemetry/opentelemetry-android/blob/fe31da77d73eb6689c1bb488dc258a2d59faa0e4/buildSrc/src/main/kotlin/otel.android-library-conventions.gradle.kts#L96

but we have not added the below platform which provides those dependency(currently I think this is not failing in other modules as we are adding this platform manually)
`androidTestImplementation(platform(libs.findLibrary("opentelemetry-platform-alpha").get()))`

PR #1438, opened 2025-12-02
Add platform dependency

atulgpt17:
Fixes https://github.com/open-telemetry/opentelemetry-android/issues/1434

### DIFF
--- a/buildSrc/src/main/kotlin/otel.android-app-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.android-app-conventions.gradle.kts
@@ -45,6 +45,7 @@ dependencies {
     androidTestImplementation(libs.findLibrary("androidx-test-core").get())
     androidTestImplementation(libs.findLibrary("androidx-test-rules").get())
     androidTestImplementation(libs.findLibrary("androidx-test-runner").get())
+    androidTestImplementation(platform(libs.findLibrary("opentelemetry-platform-alpha").get()))
     androidTestImplementation(libs.findLibrary("opentelemetry-sdk-testing").get())
     coreLibraryDesugaring(libs.findLibrary("desugarJdkLibs").get())
 }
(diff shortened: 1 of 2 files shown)


breedx-splk:
Hey @atulgpt17 it looks like you submitted this from your work account instead of your personal account (@atulgpt ) maybe? In any case, we require the CLA to be signed...so please either sign it from this account or close this PR and open it with your other account (which has already signed the CLA).

Thanks!

breedx-splk:
Thanks @atulgpt17. This feels like a good idea, but I can't determine if the absence of these platform test deps was actually causing problems? If so, can you please explain what you observed or what was broken? The build and tests do indeed seem fine without them, and the dependencies seem to be resolved just fine before....so?

atulgpt17:
Hi @breedx-splk in our current modules, problem is not arising as for each module either we are adding `api(platform(libs.opentelemetry.platform.alpha))` separately in `build.gradle.kts`(which is fine in itself) or adding `implementation(project(":common"))` which has `api(platform(libs.opentelemetry.platform.alpha))` so that version is also satisfied

But let's say you create a module which doesn't have either of those, but you have added the convention plugin `id("otel.android-library-conventions")` then when you will run your UT then build will fail with the below error
```
FAILURE: Build failed with an exception.
* What went wrong:
Configuration cache state could not be cached: field `__dependencies__` of task `:testLib:generateDebugUnitTestStubRFile` of type `com.android.build.gradle.internal.res.GenerateLibraryRFileTask`: error writing value of type 'org.gradle.api.internal.file.collections.DefaultConfigurableFileCollection'
> Could not resolve all files for configuration ':testLib:debugUnitTestRuntimeClasspath'.
   > Could not find io.opentelemetry:opentelemetry-sdk-testing:.
     Required by:
         project ':testLib'
```
(above note the `:.` this is coming as in the `testLib` module there is no way to get the version info)

Ideally if we are adding one dependency `"opentelemetry-sdk-testing"` in the convention plugin then we should either specify it's version or add some bom from where Gradle can determine it's version in that convention plugin so that plugin can become self sufficient

breedx-splk:
@atulgpt17 thanks for the clear explanation, that makes sense. The platform dependency declarations are spread around so much, I think that maybe consolidating them into the conventions probably makes sense, but we should do that in a separate follow-up PR. Thanks for catching this.

atulgpt17:
Hi @breedx-splk I have added the cleanup PR at https://github.com/open-telemetry/opentelemetry-android/pull/1453

PTAL if this is why you mean by `The platform dependency declarations are spread around so much, I think that maybe consolidating them into the conventions probably makes sense,`

PR #1441, opened 2025-12-03
Another suggestion for contributing: Single purpose PRs

breedx-splk:
Sometimes contributors might be inclined to submit a single PR that does several things in one go. This suggestion here is to make PRs do a single clear thing.


fractalwrench (approved):
LGTM

PR #1453, opened 2025-12-04
Consolidate otel alpha dependency in one place

atulgpt17:


atulgpt17:
Hi, I tried simplifying the dependency by moving it into the convention plugin, so it doesn’t need to be declared every time, as mentioned in the comment. I agree it should be api, but I’d like to confirm with @breedx-splk  whether this is the right approach.
If so, I’ll switch implementation to api in buildSrc/src/main/kotlin/otel.android-library-conventions.gradle.kts.
However, this would mean the otel alpha dependency will always be exposed to any library using the convention plugin.

android-agent/build.gradle.kts
```diff
@@ -9,7 +9,6 @@ android {
 
 dependencies {
     api(project(":agent-api"))
-    api(platform(libs.opentelemetry.platform.alpha)) // Required for sonatype publishing
```
breedx-splk, line 12: I don't think we can merge this, due to the problems that I've now encountered multiple times during the release process. Namely, sonatype does some additional validations that will fail if this platform dependency isn't there.

breedx-splk (changes requested):
I like the idea of consolidating these, but I am afraid that the `api` -> `implementation` might break something...and I'm not entirely convinced (yet) that this change will be compatible with the sonatype verifications that happen during release time.

Maybe we can try this after the 1.x release settles down, but for now I'd like to sleep on it for a bit.

PR #1455, opened 2025-12-04
Add API level to resource

### LINKED ISSUE #389: Add api level to resource
The semantic conventions include an Android specific attribute called `android.os.api_level`:

https://github.com/open-telemetry/semantic-conventions/blob/c98600cf91251effcb4eba4129711ecf45c3d4bf/docs/resource/android.md

This attribute is "recommended" so we should include it in the AndroidResource.

fractalwrench:
## Goal

Fixes #389 by adding the API level as a resource attribute.

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/AndroidResource.kt
+++ b/core/src/main/java/io/opentelemetry/android/AndroidResource.kt
@@ -11,6 +11,7 @@ import io.opentelemetry.sdk.resources.Resource
 import io.opentelemetry.semconv.ServiceAttributes.SERVICE_NAME
 import io.opentelemetry.semconv.ServiceAttributes.SERVICE_VERSION
 import io.opentelemetry.semconv.TelemetryAttributes.TELEMETRY_SDK_VERSION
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/opentelemetry/android/AndroidResource.kt
```diff
@@ -36,6 +37,7 @@ object AndroidResource {
             .put(DEVICE_MODEL_IDENTIFIER, Build.MODEL)
             .put(DEVICE_MANUFACTURER, Build.MANUFACTURER)
             .put(OS_NAME, "Android")
+            .put(AndroidIncubatingAttributes.ANDROID_OS_API_LEVEL, Build.VERSION.SDK_INT.toString())
```
breedx-splk: We static import all the other constants, so I think it's probably fine to static import this one as well, for readability.

breedx-splk (approved):
Looks good. Thanks for making the test better as well!

core/src/main/java/io/opentelemetry/android/AndroidResource.kt
fractalwrench: Updated to use a static import.

Issue #1460, opened 2025-12-05
We should run detekt with type resolution currently we don't run


atulgpt17:
Ideally we should replace `./gradlew check` with `./gradlew check -x detekt detektDebug detektDebugUnitTest detektDebugAndroidTest` this will enable type resoultion and run rules which required type resolution(though it will slightly increase the PR run time)

This will catch issues like https://github.com/open-telemetry/opentelemetry-android/pull/1452

Doc for type resolution: https://detekt.dev/docs/gettingstarted/type-resolution/

PR #1463, opened 2025-12-05
Add detekt type resolution related tasks

atulgpt17:
Fix the issues which were getting ignored as we were running the detekt w/o type resolution
Type resolution: https://detekt.dev/docs/gettingstarted/type-resolution/

Fixes https://github.com/open-telemetry/opentelemetry-android/issues/1460

### DIFF
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -39,3 +39,6 @@ style:
     active: false
   UnusedImports:
     active: true
+  VarCouldBeVal:
+    active: true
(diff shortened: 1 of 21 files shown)


config/detekt/detekt.yml
```diff
@@ -39,3 +39,6 @@ style:
     active: false
   UnusedImports:
     active: true
+  VarCouldBeVal:
```
atulgpt17, line 42: this was already enabled, I add this explicitly to add the `ignoreAnnotated`

breedx-splk: Are there any examples of this? We could leave this out until needed?

Issue #1461, opened 2025-12-05
In `WindowCallbackWrapper` we are not overriding the default method


atulgpt17:
This is required so that we can pass the default `Window.Callback` to delegated property `callback`

fractalwrench:
This looks like an issue for both the Compose/View instrumentation as `Callback` is a Java interface which means interface delegation doesn't work for default methods in Kotlin. The fix would be to add explicit methods that delegate.

PR #1464, opened 2025-12-05
Add overridden method for default

atulgpt17:
Fixes https://github.com/open-telemetry/opentelemetry-android/issues/1461


breedx-splk (approved):
Nice catch, thanks!

PR #1465, opened 2025-12-05
Add carrier information in case of WiFi connection with SIM card.

### LINKED ISSUE #1435: Network carrier information missing for WiFi connections despite SIM card availability
## Problem Statement

Currently, carrier attributes (`network.carrier.name`, `network.carrier.mcc`, `network.carrier.mnc`, `network.carrier.icc`) are only included in network spans when the device is actively connected via cellular network. However, SIM card information is available via `TelephonyManager` regardless of the active connection type (WiFi, VPN, Ethernet, etc.).

This means that when a device with a SIM card is connected via WiFi, valuable carrier information is lost, even though it's technically available and could provide useful context for telemetry data.

## Expected Behavior

When a device has a SIM card installed, carrier attributes should be included in network spans regardless of the active connection type (WiFi, VPN, Ethernet, or Cellular). This would provide consistent carrier information across all network connection types.

## Current Behavior

- ✅ Carrier attributes are included when connected via cellular network
- ❌ Carrier attributes are missing when connected via WiFi (even with SIM card present)
- ❌ Carrier attributes are missing when connected via VPN/Ethernet (even with SIM card present)

## Proposed Solution

Modify the `buildNetwork()` method in `NetworkDetectorImpl.kt` to also retrieve carrier information using `CarrierFinder`, similar to how `buildCellularNetwork()` does it.

**Current implementation:**
```
private fun buildNetwork(networkState: NetworkState) = CurrentNetwork(networkState)
```
**Proposed implementation:**
```
private fun buildNetwork(networkState: NetworkState): CurrentNetwork {
    val carrier = carrierFinder.get()
    return CurrentNetwork(
        state = networkState,
        carrier = carrier,
    )
}
```
## Technical Details

- SIM card information is accessible via `TelephonyManager` regardless of active network type
- The `CarrierFinder` class already handles permission checks and API level differences
- This change would be backward compatible (carrier will be `null` if no SIM card or permissions are missing)
- No additional permissions are required beyond what's already needed for cellular connections

## Use Case

This is particularly useful for:
- Mobile apps that primarily use WiFi but need carrier information for analytics/telemetry
- Enterprise applications tracking device context across different network types
- Consistent telemetry data regardless of connection type

## Additional Context

The `CarrierFinder` class already exists and handles edge cases (missing permissions, no telephony feature, etc.), so this change would leverage existing infrastructure.

## Related Files

- `services/src/main/java/io/opentelemetry/android/internal/services/network/detector/NetworkDetectorImpl.kt`
- `services/src/main/java/io/opentelemetry/android/internal/services/network/CarrierFinder.java`

Hemant21KG:
Fixed : https://github.com/open-telemetry/opentelemetry-android/issues/1435

### DIFF
--- a/instrumentation/network/src/main/java/io/opentelemetry/android/instrumentation/network/internal/detector/NetworkDetectorImpl.kt
+++ b/instrumentation/network/src/main/java/io/opentelemetry/android/instrumentation/network/internal/detector/NetworkDetectorImpl.kt
@@ -76,7 +76,13 @@ internal class NetworkDetectorImpl(
     /**
      * Builds a network for non-cellular networks.
      */
-    private fun buildNetwork(networkState: NetworkState) = CurrentNetwork(networkState)
+    private fun buildNetwork(networkState: NetworkState): CurrentNetwork {
+        val carrier = carrierFinder.get()
+        return CurrentNetwork(
+            state = networkState,
+            carrier = carrier,
+        )
+    }
 
     /**
      * Builds a cellular network with carrier and subtype information.
--- a/instrumentation/network/src/test/java/io/opentelemetry/android/instrumentation/network/internal/detector/NetworkDetectorTest.kt
+++ b/instrumentation/network/src/test/java/io/opentelemetry/android/instrumentation/network/internal/detector/NetworkDetectorTest.kt
@@ -67,6 +67,8 @@ class NetworkDetectorTest {
         every { connectivityManager.getNetworkCapabilities(network) } returns networkCapabilities
         every { networkCapabilities.hasTransport(any()) } returns false // default
         every { telephonyManager.simOperatorName } returns "JibroCom" // default
+        every { telephonyManager.simOperator } returns "JibroComSimOperator" // default
+        every { telephonyManager.simCountryIso } returns "JibroComSimCountryIso" // default
 
         // Mock telephony feature as available by default
         every { packageManager.hasSystemFeature(PackageManager.FEATURE_TELEPHONY) } returns true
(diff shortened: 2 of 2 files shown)


breedx-splk:
@Hemant21KG I owe you a huge apology! I have no good/reasonable explanation for why this PR was left unreviewed and and without comment for so long. 

This is a nice contribution and I think we would like to proceed with it, but right now it will need a rebase. If you can do that, that would be awesome. Alternatively, feel free to create a new PR or give us the go ahead to recreate this on your behalf if you don't have time to come back to it.

Once again, I'm so sorry that it took this long. This is not a normal thing here, it's pretty embarrassing, and I'm sincerely sorry. Hopefully we can work it out and get this merged.

Hemant21KG:
@breedx-splk Thanks for following up, and no worries at all. I’ve rebased the branch and updated the PR.

Please let me know if there’s anything else that needs to be addressed. Looking forward to getting this merged.

Thanks again for taking the time to review it!

instrumentation/network/src/main/java/io/opentelemetry/android/instrumentation/network/internal/detector/NetworkDetectorImpl.kt
```diff
@@ -76,7 +76,13 @@ internal class NetworkDetectorImpl(
...
      * Builds a network for non-cellular networks.
      */
-    private fun buildNetwork(networkState: NetworkState) = CurrentNetwork(networkState)
+    private fun buildNetwork(networkState: NetworkState): CurrentNetwork {
+        val carrier = carrierFinder.get()
+        return CurrentNetwork(
+            state = networkState,
+            carrier = carrier,
+        )
+    }
```
DavidGrath, line 85: @breedx-splk , I'm still thinking of that edge case where the WiFi could be degraded and both cellular and WiFi are simultaneously active. Would there be a need to actually include that or is this okay as it is?

Hemant21KG: Good point. I think this is okay as-is for now since we're only capturing the current active network

Hemant21KG: @DavidGrath

DavidGrath: True. I think it's okay. I just didn't want it to slip by since I remembered a recent discussion about it. Non blocking

DavidGrath (approved):

instrumentation/network/src/main/java/io/opentelemetry/android/instrumentation/network/internal/detector/NetworkDetectorImpl.kt
breedx-splk: Yeah, I think this relates to #1772 but isn't exactly the same thing. Until that's addressed, I think this is still an improvement. 👍🏻

Issue #845, opened 2025-02-20
Integrate Android PANS Metrics with OpenTelemetry


cforce:
Android's **Per-Application Network Selection (PANS)** feature enables routing app-specific traffic through OEM-paid networks, allowing better control over data costs and connectivity management. To improve observability, we need to consume PANS metrics and send them as OpenTelemetry metrics for further analysis.  

- Implement a service to collect **PANS-related metrics** from system services (e.g., `netstats dumpsys`, `connectivity dumpsys`).  
- Extract key metrics, including:  
  - **Per-app network usage** (Data transmitted via `OEM_PAID` and `OEM_PRIVATE` networks)  
  - **App-to-network mappings** (Which apps are using which OEM networks)  
  - **Network preference changes** (Updates to `setOemNetworkPreference`)  
  - **Connectivity issues or errors**  
- Convert these metrics into **OpenTelemetry format** with appropriate labels such as:  
  - `app_package_name`  
  - `network_type` (OEM_PAID, OEM_PRIVATE)  
  - `data_transmitted_bytes`  
  - `timestamp`  
- Use **OpenTelemetry SDK** to send metrics to the configured backend (e.g., **Prometheus, Jaeger, or an OTLP collector**).  
- Ensure proper **error handling and logging** for debugging connectivity issues.  

#### **Considerations**  
- Utilize Android **`netstats` dumpsys** for collecting network usage metrics per application.  
- Implement an **Android Service** that periodically fetches metrics and pushes them to OpenTelemetry.  
- Use the **OpenTelemetry Android SDK** for sending telemetry data.  
- Ensure compatibility with **Android Automotive OS** and PANS APIs.  
- Provide **persistence across reboots** to reapply network preferences.  
- Optimize for **battery efficiency** and minimize resource usage.  

#### **References**  
- [Android Automotive PANS API](https://source.android.com/docs/automotive/connectivity?authuser=1#call-pans-

namanONcode:
@breedx-splk assign this to me

namanONcode:
@breedx-splk plan to implement this a new instrumentation in module at instrumentation/pans/ that implements Android Per-Application Network Selection (PANS) metrics collection. The service will integrate with OpenTelemetry metrics to track per-app network usage, network preferences, and changes. It will use the existing PeriodicWork service for battery-efficient collection and follow established project patterns including auto-service registration, Kotlin-first implementation, and dependency injection.

namanONcode:
[...]

PR #1467, opened 2025-12-07
feat(pans): integrate PANS metrics collection with OpenTelemetry

namanONcode:
This pull request introduces a new Android instrumentation module for PANS (Per-Application Network Selection) within the OpenTelemetry project. The changes add all the core classes, Gradle configuration, permissions, and Proguard rules necessary to support the collection and reporting of per-app network metrics, network preference changes, and network availability on Android devices.

**Key changes include:**

### Core functionality and data models

* Added core data classes for PANS metrics, including `PANSMetrics`, `AppNetworkUsage`, `PreferenceChange`, and `NetworkAvailability`, along with helper functions to build OpenTelemetry attributes for these metrics (`PANSMetrics.kt`).
* Introduced wrapper classes for Android's `ConnectivityManager` and `NetworkStatsManager` to safely and conveniently collect network state, availability, and per-app network usage statistics (`ConnectivityManagerWrapper.kt`, `NetStatsManager.kt`). [[1]](diffhunk://#diff-d35df9dbecc5ced9f26c3535bbb67cbb0e5a783d55b9802499d5c2023e10fb5bR1-R122) [[2]](diffhunk://#diff-fe50c37c2be3a0c3a2514be066fd07a6739665ade8c5635eea0201c7f76ca28dR1-R139)

### Module setup and configuration

* Created a new Gradle build configuration for the PANS module, including dependencies, plugins, Jacoco coverage setup, and custom coverage reporting tasks (`build.gradle.kts`).
* Added a Proguard consumer rules file to ensure all PANS instrumentation classes are kept and named correctly for consumers (`consumer-rules.pro`).
* Defined the required Android permissions in the module's `AndroidManifest.xml` to allow network stats collection and monitoring (`PACKAGE_USAGE_STATS`, `ACCESS_NETWORK_STATE`, `INTERNET`).

### API exposure

* Generated the public API file for the new module, exposing the main classes and their methods for use by other components or consumers (`pans.api`).

### DIFF
--- a/instrumentation/pans/api/pans.api
+++ b/instrumentation/pans/api/pans.api
@@ -1,0 +1,88 @@
+public final class io/opentelemetry/android/instrumentation/pans/AppNetworkUsage {
+	public fun <init> (Ljava/lang/String;ILjava/lang/String;JJLio/opentelemetry/api/common/Attributes;)V
+	public final fun component1 ()Ljava/lang/String;
+	public final fun component2 ()I
+	public final fun component3 ()Ljava/lang/String;
+	public final fun component4 ()J
+	public final fun component5 ()J
+	public final fun component6 ()Lio/opentelemetry/api/common/Attributes;
+	public final fun copy (Ljava/lang/String;ILjava/lang/String;JJLio/opentelemetry/api/common/Attributes;)Lio/opentelemetry/android/instrumentation/pans/AppNetworkUsage;
+	public static synthetic fun copy$default (Lio/opentelemetry/android/instrumentation/pans/AppNetworkUsage;Ljava/lang/String;ILjava/lang/String;JJLio/opentelemetry/api/common/Attributes;ILjava/lang/Object;)Lio/opentelemetry/android/instrumentation/pans/AppNetworkUsage;
+	public fun equals (Ljava/lang/Object;)Z
+	public final fun getAttributes ()Lio/opentelemetry/api/common/Attributes;
+	public final fun getBytesReceived ()J
+	public final fun getBytesTransmitted ()J
+	public final fun getNetworkType ()Ljava/lang/String;
+	public final fun getPackageName ()Ljava/lang/String;
+	public final fun getUid ()I
+	public fun hashCode ()I
+	public fun toString ()Ljava/lang/String;
+}
+
+public final class io/opentelemetry/android/instrumentation/pans/NetworkAvailability {
+	public fun <init> (Ljava/lang/String;ZILio/opentelemetry/api/common/Attributes;)V
+	public synthetic fun <init> (Ljava/lang/String;ZILio/opentelemetry/api/common/Attributes;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public final fun component1 ()Ljava/lang/String;
+	public final fun component2 ()Z
+	public final fun component3 ()I
+	public final fun component4 ()Lio/opentelemetry/api/common/Attributes;
+	public final fun copy (Ljava/lang/String;ZILio/opentelemetry/api/common/Attributes;)Lio/opentelemetry/android/instrumentation/pans/NetworkAvailability;
+	public static synthetic fun copy$default (Lio/opentelemetry/android/instrumentation/pans/NetworkAvailability;Ljava/lang/String;ZILio/opentelemetry/api/common/Attributes;ILjava/lang/Object;)Lio/opentelemetry/android/instrumentation/pans/NetworkAvailability;
+	public fun equals (Ljava/lang/Object;)Z
+	public final fun getAttributes ()Lio/opentelemetry/api/common/Attributes;
+	public final fun getNetworkType ()Ljava/lang/String;
+	public final fun getSignalStrength ()I
+	public fun hashCode ()I
+	public final fun isAvailable ()Z
+	public fun toString ()Ljava/lang/String;
+}
+
+public final class io/opentelemetry/android/instrumentation/pans/PANSMetrics {
+	public fun <init> ()V
+	public fun <init> (Ljava/util/List;Ljava/util/List;Ljava/util/List;)V
+	public synthetic fun <init> (Ljava/util/List;Ljava/util/List;Ljava/util/List;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public final fun component1 ()Ljava/util/List;
+	public final fun component2 ()Ljava/util/List;
+	public final fun component3 ()Ljava/util/List;
+	public final fun copy (Ljava/util/List;Ljava/util/List;Ljava/util/List;)Lio/opentelemetry/android/instrumentation/pans/PANSMetrics;
+	public static synthetic fun copy$default (Lio/opentelemetry/android/instrumentation/pans/PANSMetrics;Ljava/util/List;Ljava/util/List;Ljava/util/List;ILjava/lang/Object;)Lio/opentelemetry/android/instrumentation/pans/PANSMetrics;
+	public fun equals (Ljava/lang/Object;)Z
+	public final fun getAppNetworkUsage ()Ljava/util/List;
+	public final fun getNetworkAvailability ()Ljava/util/List;
+	public final fun getPreferenceChanges ()Ljava/util/List;
+	public fun hashCode ()I
+	public fun toString ()Ljava/lang/String;
+}
+
+public final class io/opentelemetry/android/instrumentation/pans/PansInstrumentation : io/opentelemetry/android/instrumentation/AndroidInstrumentation {
+	public static final field Companion Lio/opentelemetry/android/instrumentation/pans/PansInstrumentation$Companion;
+	public fun <init> ()V
+	public fun getName ()Ljava/lang/String;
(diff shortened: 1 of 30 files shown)


breedx-splk:
Related to #845.

namanONcode:
@breedx-splk  yes it is pull request for this > Related to #845.

breedx-splk (changes requested):
Well, it's another very large GenAI created PR. We strongly prefer that PR size be reduced to facilitate thorough reviews. In the current state (>8000 lines), this simply not reasonable. 

There's also just a lot going on in this PR. Lots of logging (you won't see this level of logging elsewhere in instrumentation). Weird triplicate tests. Some bespoke gradle foo that you won't find elsewhere in this repo. Jacoco. A Meter named logger (we intentionally don't have any other instrumentation [yet] that emits metrics from Android). Comments that just mirror already clear variable names. No mention of nor consideration given to semantic conventions. Missing README that describes the instrumentation (all other modules have this). Etc. etc. etc.

I think it's all just a little too much. 

If you'd like to take another crack at this, please see about reducing the PR size, for starters. That will allow you to get actionable feedback that you can iterate on. Think about small changes that move this feature forward, rather than an epic torrent of code that tries to do it all in one go.

bidetofevil:
Maybe I'm missing something, but this doesn't appear to do anything? `NetStatsManager` always returns an empty list, and network data usage can't be obtained through a platform API like that.

Also, the agent is initialized for an application and has access to a subset of data running in that process. It's not possible for it to collect data about other processes and how much data they use.

Unless I'm missing some earlier discussion, I don't think this does (or can do) what is claimed.

bidetofevil (changes requested):
This should not be merged

namanONcode:
@bidetofevil wait i will fix this is a small bug i have identified> This should not be merged

bidetofevil:
This is not a small bug. There is no way to get network usage information through the Android Platform API. You need to basically count the bytes going through the network connections and I don't see a facility to do that.

namanONcode:
> Maybe I'm missing something, but this doesn't appear to do anything? `NetStatsManager` always returns an empty list, and network data usage can't be obtained through a platform API like that.
> 
> Also, the agent is initialized for an application and has access to a subset of data running in that process. It's not possible for it to collect data about other processes and how much data they use.
> 
> Unless I'm missing some earlier discussion, I don't think this does (or can do) what is claimed.

@breedx-splk @bidetofevil fixes apply please review i will improve

cforce:
There are fundamental two  approaches.
a.) only counting bytes for same app/isolated process - might work if all network relevant clients are instrumented which collect and aggregate  the traffic. Does not need right for platform level but only have a limited  use case.
b.) something like a dedicated application with system/privileged app access which can be run as platform stack for the complete android runtime delivery metrics for all apps
As  a regular app i cannot run the queries from an isolated process; permissions won’t apply there, but what about a system/privileged app?
Itshall be able can access NetworkStatsManager for device/user/UID summaries and aggregate usage for all UIDs - including those of isolated processes - provided the querying code runs in a non‑isolated privileged process and you broker results via Binder. Maybe  manifest  with permissions like below would make it

```
<manifest ... >
  <uses-permission android:name="android.permission.READ_NETWORK_USAGE_STATS"/>
  <uses-permission android:name="android.permission.PACKAGE_USAGE_STATS"/>
  <!-- Optional / platform-dependent -->
  <uses-permission android:name="android.permission.READ_PRIVILEGED_PHONE_STATE"/>

  <application android:usesCleartextTraffic="false" ... >
    <service
      android:name=".NetworkStatsService"
      android:exported="true"
      android:permission="android.permission.BIND_PRIVILEGED_SERVICE">
      <intent-filter>
        <action android:name="com.example.netstats.ACTION_BIND"/>
      </intent-filter>
    </service>
  </application>
</manifest>

```

Another approach:
As since Android 9 (kernel ≥ 4.9), Android counts traffic per UID and interface using eBPF filters data is stored in eBPF maps (e.g., traffic_uid_stats_map, traffic_tag_stats_map, traffic_iface_stats_map) pinned under /sys/fs/bpf. netd’s TrafficController and NetworkStatsService read from these maps to provide usage statistics. 
A native, privileged daemon (system component) reads pinned eBPF maps via the Android BPF library (bpf_obj_get()), aggregates counters, and exposes results over Binder to apps. The daemon is allowed via SELinux policy and runs under an appropriate service UID/group (often co‑located with netd or with similar capabilities).
see https://source.android.com/docs/core/data/ebpf-traffic-monitor

fractalwrench:
IMO supporting instrumentation specifically for a system/privileged app within opentelemetry-android is an edge-case. It's therefore not something I'd personally support adding to the project, since each feature adds complexity and maintenance burden and I'm not convinced a large number of folks would use this telemetry.

cforce:
opentelemetry-android on an edge-device .. is a very important use case for IOT .
Also i think that it can be  seen a s matter of configuration/customization because  opentelemetry shall also support system apps and why not allow to use the capabilities then. The hostmetric collection on the otelcollector does exactly the same on bare linux

namanONcode:
@bidetofevil @cforce got it i will try to implement this for edge devices i think need some more dependency and new system design for make it low latency on edge devices so sending metrics i will think about this but this opentelemetry is for andrioid not the main one

namanONcode:
> IMO supporting instrumentation specifically for a system/privileged app within opentelemetry-android is an edge-case. It's therefore not something I'd personally support adding to the project, since each feature adds complexity and maintenance burden and I'm not convinced a large number of folks would use this telemetry.

Yeah u are right

namanONcode:
> IMO supporting instrumentation specifically for a system/privileged app within opentelemetry-android is an edge-case. It's therefore not something I'd personally support adding to the project, since each feature adds complexity and maintenance burden and I'm not convinced a large number of folks would use this telemetry.

@bidetofevil @fractalwrench @breedx-splk  is there anything change to implement ?

namanONcode:
@bidetofevil @fractalwrench @breedx-splk  is there anything change to implement ?

breedx-splk:
> @bidetofevil @fractalwrench @breedx-splk is there anything change to implement ?

This AI assisted PR is really large and is going to be a challenge for us to review in a reasonable time, plus we've had holidays in many places.

I asked earlier if you might be able to find a way to approach this incrementally, rather than one large singular dump of 8k+ lines of code. Can you give that some consideration? Maybe we can do this piece-at-a-time and make iterative changes that build to a solution in a way that's more reviewable?

namanONcode:
> > @bidetofevil @fractalwrench @breedx-splk is there anything change to implement ?
> 
> This AI assisted PR is really large and is going to be a challenge for us to review in a reasonable time, plus we've had holidays in many places.
> 
> I asked earlier if you might be able to find a way to approach this incrementally, rather than one large singular dump of 8k+ lines of code. Can you give that some consideration? Maybe we can do this piece-at-a-time and make iterative changes that build to a solution in a way that's more reviewable?

@breedx-splk  I know but this also a big feature we are talking about PANS metrics so if we are doing PR this in smaller chunks it can break system cause this is a core feature we are implementing so its obvious take time to review this PR 


Otherwise i will try my best to break down this pr

bidetofevil:
I don't think that instrumentation that only works for privileged apps that would rarely be granted should be included in this project, but be built as a separate package outside of this project. This is not intended to be a monolith that contains all the instrumentation under the sun.

namanONcode:
Yeah u can make a separate package with separate api call this with a event listener @breedx-splk @bidetofevil @fractalwrench  what about this idea ? > I don't think that instrumentation that only works for privileged apps that would rarely be granted should be included in this project, but be built as a separate package outside of this project. This is not intended to be a monolith that contains all the instrumentation under the sun.

fractalwrench:
I do not think this instrumentation belongs in this project & think it's a better fit as custom instrumentation included in your own codebase.

If other folks would find this useful however, I would encourage you to give the issue a thumbs up. This will indicate the popularity of the feature.

namanONcode:
> I do not think this instrumentation belongs in this project & think it's a better fit as custom instrumentation included in your own codebase.
> 
> If other folks would find this useful however, I would encourage you to give the issue a thumbs up. This will indicate the popularity of the feature.

So make it like feature if its in use it can be enabled disable by default
because this feature is need for pans metric collection and can improve logging for android apps i also seen personaly pans has bug sometime delays if there is metric collection option there will be easy debug also this feature will very useful for application related to file share kind of stuff

namanONcode:
@breedx-splk @bidetofevil @cforce @fractalwrench any suggestion to implement or anything else otherwise this pr going to close automaticly

breedx-splk:
Honestly, between the massive size and the questionable utility, I haven't been inspired to look into this much. We should decide if we actually want to continue with PANS instrumentation or if some other entity might own/host this.

@cforce opened #845 originally, might be nice if they could get involved too?

cforce:
@namanONcode i think you shall fork this as a standalone application reusing the android sdk as dependency to build a privileged system application which can collect extended network metrics on system level. The collector written in go and no support for listening to android bound interfaces does not offer such features-so this be of use for others

PR #1470, opened 2025-12-09
Session Management: Added core infrastructure, etc.

gregory-at-cvs:
### Changes

These changes are a part of a larger set of changes to foundational infrastructure for adding session tracking across all telemetry signals. They add core session management infrastructure to SDK and implemented session tracking with automatic session ID injection into spans, logs, and events. Also, they include SessionManager, SessionProvider with previous session ID support, processors for automatic attribute injection, and extension functions for manual session ID management.

This is the first of several PRs coming to add the session and previous session ids to telemetry data.  After these changes are merged in, the upcoming PRs will tie this work in with:
1. ANR and Crash Instrumentation
2. Activity and Fragment Lifecycle Instrumentation
3. View Click and Compose Click Instrumentation
4. Network, WebSocket, Slow Rendering Instrumentation
5. Startup Instrumentation
6. Demo app updates

### Details
Here are some details of the changes:
1. Added SessionProvider.getPreviousSessionId() to track session transitions
2. Added SessionProvider.NO_SESSION_ID constant for explicit empty session handling
3. New SessionIdSpanAppender and SessionIdLogRecordAppender processors automatically add session IDs to all telemetry
4. Extension functions setSessionIdentifiersWith() for explicit session ID injection
5. SessionManager implementation with configurable timeouts (default: 15min background, 4hr max)
7. Only adds session.id attribute when session is active (prevents empty values in telemetry)

### Benefits
The benefits include:
1. Changes are additions only
2. Existing instrumentations are unchanged - automatic session ID coverage via processors
3. Session correlation with session.previous_id attribute enables flow analysis across session boundaries
4. Compliant with OpenTelemetry semantic conventions
5. Configurable session lifecycle (background timeout, max lifetime)


gregory-at-cvs:
Hi @fractalwrench and @breedx-splk 👋 

As we discussed, this PR was broken out from the other two and has the core infrastructure changes, needed by upcoming PRs, for the session management changes to add session ids across telemetry signals.

session/src/main/kotlin/io/opentelemetry/android/session/SessionProvider.kt
```diff
@@ -5,16 +5,44 @@
 
 package io.opentelemetry.android.session
 
+/**
```
fractalwrench, line 8: Thanks for adding docs!

android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionConfig.kt
```diff
@@ -9,11 +9,63 @@ import kotlin.time.Duration
...
+ *   ends and a new one will be created when the app becomes active again.
+ *
+ * @param maxLifetime maximum duration a session can remain active regardless of app activity.
+ *   Default is 4 hours, meaning even if the app stays in foreground continuously, sessions will
+ *   rotate every 4 hours to prevent unbounded session growth and ensure fresh session context.
+ *
+ * @see io.opentelemetry.android.agent.session.SessionManager
+ * @see io.opentelemetry.android.session.SessionProvider
+ */
+data class SessionConfig(
```
fractalwrench: `android-agent` is part of the public API and [data classes are discouraged](https://kotlinlang.org/docs/api-guidelines-backward-compatibility.html#avoid-using-data-classes-in-your-api) from public APIs in Kotlin. Additionally we've been moving towards using a DSL on `OpenTelemetryRumInitializer`. `SessionConfiguration` should be considered the entrypoint and this class is effectively an implementation detail (although that has changed recently)

android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/factory/SessionManagerFactory.kt
```diff
@@ -0,0 +1,47 @@
...
+     *
+     * This implementation creates a [SessionManager] with a timeout handler that is
+     * registered with the application lifecycle service.
+     *
+     * @param application the Android application instance.
+     * @param sessionConfig the configuration for session management.
+     * @return a session manager instance.
+     */
+    override fun createSessionProvider(
+        application: Application,
```
fractalwrench: IMO `application` seems like an implementation detail of the fact that `Services` requires context. Do you think it makes sense to remove it as a parameter from the original function, and then if it is required it can be passed into the constructor for this class?

android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/factory/SessionManagerFactory.kt
```diff
@@ -0,0 +1,47 @@
...
+ * - A timeout handler configured from the session config
+ * - Application lifecycle integration for timeout management
+ *
+ * This class is open to allow custom implementations for specialized use cases.
+ *
+ * @see SessionProviderFactory
+ * @see SessionManager
+ * @see SessionConfig
+ */
+open class SessionManagerFactory : SessionProviderFactory {
```
fractalwrench: I'd suggest making this `internal` so that it isn't exposed, as I'd consider this an implementation detail. As long as the `SessionProviderFactory` type is public and there's a way to set it on the `OpenTelemetryConfiguration` DSL, that covers the use-case of providing custom implementations.

core/api/core.api
```diff
@@ -11,6 +11,9 @@ public final class io/opentelemetry/android/BuildConfig {
 	public fun <init> ()V
 }
 
+public abstract interface annotation class io/opentelemetry/android/Incubating : java/lang/annotation/Annotation {
```
fractalwrench: This should probably stay in agent-api. If we do remove it, that can be discussed as a separate PR

session/src/main/kotlin/io/opentelemetry/android/session/SessionProvider.kt
```diff
@@ -5,16 +5,44 @@
...
+     *
+     * @return the previous session ID string, or [NO_SESSION_ID] if there is no previous session.
+     */
+    fun getPreviousSessionId(): String
+
     companion object {
+        /**
+         * Constant representing the absence of a session ID.
+         */
+        const val NO_SESSION_ID = ""
```
fractalwrench, line 36: This raises an interesting question on what the attribute value should be if no session is set. I didn't spot any guidance in the [semantic conventions docs](https://opentelemetry.io/docs/specs/semconv/registry/attributes/session/#session-id) but my natural inclination would be to avoid adding an empty string. I guess that also raises the question of whether we should return `String` or `String?` from `getSessionId()`.

core/src/test/java/io/opentelemetry/android/OpenTelemetryRumBuilderTest.kt
```diff
@@ -154,10 +152,6 @@ class OpenTelemetryRumBuilderTest {
                     .hasName("test span")
                     .hasResource(resource)
                     .hasAttributesSatisfyingExactly(
-                        OpenTelemetryAssertions.equalTo(
-                            SessionIncubatingAttributes.SESSION_ID,
-                            sessionId,
-                        ),
```
fractalwrench, line 160: Is this assertion no longer required, or is it covered elsewhere by some other test case?

fractalwrench (commented):
Thanks for taking a stab at this @gregory-at-cvs! I've had an initial look at this and left some comments inline. If I'm taking a while to respond please feel free to tag me - there's a lot of issues + PRs to keep up with.

session/src/main/kotlin/io/opentelemetry/android/session/SessionProvider.kt
gregory-at-cvs: You're welcome.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionConfig.kt
gregory-at-cvs: Hey @fractalwrench 👋 

### Data Class vs Regular Class
Oh, how'd I miss that one. 🤦 😆 Thanks for the reminder/heads up. I changed it to a regular class.

### DSL vs Public Access
Thanks for the context about the move to use the DSL. I want to make sure we're aligned on the architectural use cases here.

I see two distinct initialization patterns in the Android ecosystem:

#### 1.) Simple Apps
 Simple apps: Use OpenTelemetryRumInitializer.initialize() with the DSL
```kotlin
OpenTelemetryRumInitializer.initialize(context) {
    session {
       backgroundInactivityTimeout = 30.minutes ,
       maxLifetime = 4.hours    
    }
}
```

#### 2.) Complex Apps
Complex apps: Use OpenTelemetryRum.builder() directly when they need:

1. Custom provider builders (pre-configured processors, samplers, resource attributes)
2. Dependency injection frameworks (Dagger/Hilt/Koin)
3. Library wrappers with their own configuration API

They need direct access to SessionConfig because the DSL isn't accessible within their builder flow.

### Example
Here's a real world example with Dagger/Hilt:

```kotlin
@Module
@InstallIn(SingletonComponent::class)
object TelemetryModule {
    @Provides
    @Singleton
    fun provideOpenTelemetryRum(
        application: Application,
        loggerBuilder: SdkLoggerProviderBuilder, // Pre-configured with custom processors
        meterBuilder: SdkMeterProviderBuilder,   // Pre-configured with custom readers
    ): OpenTelemetryRum {
        val sessionConfig = SessionConfig(
            backgroundInactivityTimeout = 30.minutes,
            maxLifetime = 4.hours
        )
        val sessionProvider = SessionManagerFactory(application).createSessionProvider(sessionConfig)
        
        return OpenTelemetryRum.builder(application, otelConfig)
            .addLoggerProviderCustomizer { _, _ -> loggerBuilder }
            .addMeterProviderCustomizer { _, _ -> meterBuilder }
            .setSessionProvider(sessionProvider)
            .build()
    }
}
```

Just to note, the SessionManagerFactory needs to remain public for the same reason. Apps using 
this approach need to construct the SessionProvider outside of the initializer's static context.

### Going Forward

I'm seeing SessionConfig as analogous to other public configuration classes like DiskBufferingConfig and OtelRumConfig — it's configuration data, not implementation logic. The actual implementation (SessionManager, SessionIdTimeoutHandler) remains internal. This design approach enables the best of both worlds:
1. The documentation recommends using the simpler approaches for a lot of use cases
2. The SDK provides flexibility for use cases where the simpler approach doesn't fit

Is this approach a good compromise?

android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/factory/SessionManagerFactory.kt
gregory-at-cvs: It is and I made the change.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/factory/SessionManagerFactory.kt
gregory-at-cvs: Please see my explanation here as to why it needs to be public: https://github.com/open-telemetry/opentelemetry-android/pull/1470#discussion_r2722668598

core/api/core.api
gregory-at-cvs: Initially, Thought it needed to be moved for some of the access; however, I revisited it and switched it back.

session/src/main/kotlin/io/opentelemetry/android/session/SessionProvider.kt
gregory-at-cvs: Typically, I like using this otherwise to provide clarity and avoid the use of null, which it's usage can lead to increased risk in frameworks, SDKs, and libraries. I revisited it here and changed it back.

core/src/test/java/io/opentelemetry/android/OpenTelemetryRumBuilderTest.kt
gregory-at-cvs: This assertion is no longer needed because the test uses a noop session provider that returns an empty session ID, and the logic now skips adding empty session attributes.

gregory-at-cvs:
Hey @fractalwrench 👋 

Thanks for bearing on the late response and changes through the holidays. I've made changes and left responses. I think there's a caching issue in CI as there is a KSP issue reported; however, the project builds successfully locally. Could you trigger a rebuild in CI? I don't think I have permissions to run the workflow.

gregory-at-cvs:
Hey @LikeTheSalad 👋 

Heads up: The label can be removed. I responded to Jamie and made changes.

gregory-at-cvs:
> Hey @fractalwrench 👋
> 
> Thanks for bearing on the late response and changes through the holidays. I've made changes and left responses. I think there's a caching issue in CI as there is a KSP issue reported; however, the project builds successfully locally. Could you trigger a rebuild in CI? I don't think I have permissions to run the workflow.

Heads up @fractalwrench , all of the PR checks have passed.

android-agent/api/android-agent.api
```diff
@@ -105,3 +102,29 @@ public final class io/opentelemetry/android/agent/dsl/instrumentation/SlowRender
 	public fun enabled (Z)V
 }
 
+public final class io/opentelemetry/android/agent/session/SessionConfig {
```
fractalwrench, line 105: IMO it'd be preferable to use the existing `SessionConfiguration` in the DSL API rather than using concrete classes such as `SessionConfig` and `SessionManagerFactory`. E.g. something like this should be possible:

```kotlin
OpenTelemetryRumInitializer.initialize(ctx) {
    sessionConfig {
        backgroundInactivityTimeout = 30.minutes
    }
    sessionProvider = MyCustomSessionProvider()
}
```

android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/factory/SessionManagerFactory.kt
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * This class is open to allow custom implementations for specialized use cases.
+ *
+ * @param application the Android application instance used to access platform services.
+ * @param clock the clock instance to use for time-based operations. Defaults to [Clock.getDefault].
+ * @see SessionProviderFactory
+ * @see SessionManager
+ * @see SessionConfig
+ */
+open class SessionManagerFactory(
```
fractalwrench, line 32: I'm not a fan of exposing concrete types as part of the public API, particularly if they're `open` like here. IMO the default implementation should be `internal` and not visible to an end-user

fractalwrench (commented):
Thanks for updating. I took another look at this and think a few changes to the symbols made public in `android-agent` would be good, particularly around the DSL on `OpenTelemetryRumInitializer.initialize`. I understand there was some discussion about those being needed for `OpenTelemetryRumBuilder` so I think it'd be good to get input from @breedx-splk and @LikeTheSalad given that the repo is now stable & we have to maintain whatever symbol is added to the `android-agent` module.

I do think it would be beneficial to split this PR up further into even smaller chunks. The maximum size diff I'd usually want to review is about 500 lines. IMO it'll go quicker and be less painful for everybody involved if we can find parts of this PR that can be broken off and approved quickly. `SessionExtensions.kt` might be a good first candidate?

android-agent/api/android-agent.api
gregory-at-cvs: Got it. So  you're suggesting adding `sessionProvider` as a property to `OpenTelemetryConfiguration` so users can inject custom providers via DSL. Is that correct? If yes, then something like this:

```kotlin
OpenTelemetryRumInitializer.initialize(ctx) {
    sessionProvider = MyCustomSessionProvider()
}
```
That works for injection, but **_there's still a creation question_** of how do DI users create the `SessionProvider` in the first place?

**For instance, when DI is used:**
```kotlin
@Provides
fun provideSessionProvider(app: Application): SessionProvider {
    // Need to create a SessionProvider here
    // If SessionConfig and SessionManagerFactory are internal, what do I use?
    return ???
}

@Provides
fun provideOpenTelemetryRum(
    app: Application,
    sessionProvider: SessionProvider
): OpenTelemetryRum {
    return OpenTelemetryRumInitializer.initialize(app) {
        sessionProvider = sessionProvider // The DSL suggestion works for this part
    }
}
```

**Here are the options I see for creating the SessionProvider:**
1. **Keep SessionManagerFactory public:**
Users can do

```kotlin
SessionManagerFactory(app).createSessionProvider(config)
```
2. **Add public factory method:** e.g., 
```kotlin
SessionProvider.create(app, timeout, maxLifetime)
```
3. **Users implement SessionProvider from scratch:**
They rebuild SessionManager's logic _(~200 lines of concurrent code, session storage, timeout handling, lifecycle integration)_

Which approach makes the most sense to you? I'm leaning heavily toward option 1 (public factory) since it matches how DiskBufferingConfiguration works and keeps the actual implementation internal.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/factory/SessionManagerFactory.kt
gregory-at-cvs: I think it relates to how we're thinking about the SDK's architecture.  These are good discussions to have. 😄 

### Libraries vs SDKs vs Frameworks

From my perspective, this SDK's design fits that of a framework and not a library. That distinction matters because:

**Libraries** tend to give you code to use as-is without customization
**Frameworks** provide building blocks you extend
**SDKs** combine both, providing framework capabilities plus tooling

This SDK already leans into the framework side:
1. `OpenTelemetryRum.builder()` lets you customize at every layer
2. Config classes like `DiskBufferingConfiguration` and `OtelRumConfig` are public
3. Extension points throughout (customizers for providers, exporters, propagators)

Making `SessionManagerFactory` public and `open` follows that same pattern. It makes the SDK extensible by design.

### On Your DSL Suggestion

Adding `sessionProvider` as a property to `OpenTelemetryConfiguration` would solve the **injection** part:

```kotlin
OpenTelemetryRumInitializer.initialize(ctx) {
    sessionProvider = myCustomProvider
} // But there's still a creation question - how do DI users create that `SessionProvider` in the first place if `SessionManagerFactory` is internal?

@Provides
fun provideSessionProvider(app: Application, config: AppConfig): SessionProvider {
    return ??? // Need a way to create this with custom config
} // That's why the factory needs to be public.
```

### The Codebase Already Works This Way

`OtelRumConfig` is public so teams can create it in their DI containers and pull values from Firebase Remote Config, feature flags, or their own config services. 

`SessionManagerFactory` needs to be public for the same reason. Teams need to instantiate it in DI with session timeouts from their config sources, not hardcoded values.

The `open` modifier lets teams subclass the factory to add custom behavior (analytics integration, custom logging) while reusing the default implementation. Without `open`, they'd duplicate the timeout handler creation, lifecycle registration, and session manager wiring.

**What Remains Hidden**

The complicated stuff still stays internal:
- `SessionManager` - internal
- `SessionIdTimeoutHandler` - internal  
- `SessionStorage` - internal
- All the threading, concurrency, lifecycle wiring - internal

Only the **extension points** are public (factory and config). The implementation stays locked down.

This is how frameworks like Dagger, OkHttp, and others work. They expose factories because customization is part of the public contract.

This is the architectural pattern I'm seeing throughout the SDK. If you have a different approach for how teams should create SessionProviders in DI scenarios, I'd like to understand that.

gregory-at-cvs:
> Thanks for updating. I took another look at this and think a few changes to the symbols made public in `android-agent` would be good, particularly around the DSL on `OpenTelemetryRumInitializer.initialize`. I understand there was some discussion about those being needed for `OpenTelemetryRumBuilder` so I think it'd be good to get input from @breedx-splk and @LikeTheSalad given that the repo is now stable & we have to maintain whatever symbol is added to the `android-agent` module.
> 
> I do think it would be beneficial to split this PR up further into even smaller chunks. The maximum size diff I'd usually want to review is about 500 lines. IMO it'll go quicker and be less painful for everybody involved if we can find parts of this PR that can be broken off and approved quickly. `SessionExtensions.kt` might be a good first candidate?

Hey @fractalwrench 👋

You're welcome and thanks for the feedback. I addressed the individual comments.

On the PR size, I completely understand wanting smaller PRs for easier review. I think this one is actually scoped right for infrastructure work. The reason is that session management is one cohesive system: SessionManager generates session IDs, the timeout handler watches app lifecycle, the processors automatically inject session IDs into telemetry, and the extensions provide manual control when needed.

These aren't separate features, but interdependent parts of one feature. It can be difficult to understand the timeout handler without seeing SessionManager. The processors don't work without the extensions.

**This matches what's been merged here:**  
[PR #145 (Disk Buffering)](https://github.com/open-telemetry/opentelemetry-android/pull/145) was **679 lines** with similar infrastructure:  services, managers, and config working together.

Splitting into many PRs would mean tracking architecture across multiple reviews, can't validate end-to-end behavior until the last PR, and more time context-switching than reviewing.

This PR is already organized for review, with clean module boundaries, tests, and clear documentation.

Let me know if you'd still like it split. I can do it, but I think the review will actually be harder, not easier.

fractalwrench:
Happy to wait and see what the other maintainers' opinions are on whether splitting the PR down is worth the effort. I'm newer to this project so have less context on what has been accepted in the past, my opinions is based largely at conventions at Embrace & previous companies I've worked at.

Regardless, it'd probably be good to get someone else's eyes on this PR as I've already given it a first pass.

gregory-at-cvs:
> Happy to wait and see what the other maintainers' opinions are on whether splitting the PR down is worth the effort. I'm newer to this project so have less context on what has been accepted in the past, my opinions is based largely at conventions at Embrace & previous companies I've worked at.
> 
> Regardless, it'd probably be good to get someone else's eyes on this PR as I've already given it a first pass.

That makes sense and I appreciate you being open about the context. 

@breedx-splk @LikeTheSalad - I would love your perspectives on the PR structure and the API design questions when you get a chance. Thanks!

core/src/main/java/io/opentelemetry/android/ktx/SessionExtensions.kt
```diff
@@ -0,0 +1,81 @@
...
+import io.opentelemetry.api.trace.Span
+import io.opentelemetry.sdk.logs.ReadWriteLogRecord
+import io.opentelemetry.semconv.incubating.SessionIncubatingAttributes
+
+/**
+ * Adds the session identifiers from the [sessionProvider] to this [Span]'s attributes.
+ * @param sessionProvider the provider that supplies the session identifiers.
+ * @return this [Span] for which the session identifiers have been added.
+ */
+fun Span.setSessionIdentifiersWith(sessionProvider: SessionProvider): Span = setSessionIdentifiersWith(sessionProvider, ::setAttribute)
```
LikeTheSalad, line 23: It seems like what these extensions do is to add session and previous session id attributes for an individual signal. Sessions should be present in all spans and logs, so I don't see the point of creating tools to attach them individually. Ideally, the work of adding these attrs to signals should be done in the processor, where we intercept all signals and enrich them if needed. This is also already [done for spans](https://github.com/open-telemetry/opentelemetry-android/blob/495ef25dc7b4798c0684d35b1be34e489170adcc/core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.kt#L527) and [logs](https://github.com/open-telemetry/opentelemetry-android/blob/495ef25dc7b4798c0684d35b1be34e489170adcc/core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.kt#L549) in the rum builder, which is used by the initializer function, so it should already be addressed in all scenarios.

Also, I don't think it's a good idea in general to add extensions to the upstream APIs unless we know it's for use cases that are stable, and unfortunately, right now sessions aren't yet stable in the spec, so I wouldn't like to get in a situation in which we're stuck at having to maintain public APIs that add deprecated attributes, in case these change before becoming stable.

core/src/main/java/io/opentelemetry/android/OpenTelemetryRumImpl.kt
```diff
@@ -25,13 +26,24 @@ internal class OpenTelemetryRumImpl(
...
     override fun emitEvent(
         eventName: String,
         body: String,
         attributes: Attributes,
     ) {
-        val logRecordBuilder = logger.logRecordBuilder()
-        logRecordBuilder
+        logger
+            .logRecordBuilder()
+            .setSessionIdentifiersWith(sessionProvider)
```
LikeTheSalad, line 46: This is related to [my previous comment](https://github.com/open-telemetry/opentelemetry-android/pull/1470/changes#r2741551408). We don't need to manually add attributes when creating a single signal.

core/src/main/java/io/opentelemetry/android/internal/processors/SessionIdLogRecordAppender.kt
```diff
@@ -5,19 +5,25 @@
...
+ */
 internal class SessionIdLogRecordAppender(
     private val sessionProvider: SessionProvider,
 ) : LogRecordProcessor {
     override fun onEmit(
         context: Context,
         logRecord: ReadWriteLogRecord,
     ) {
-        logRecord.setAttribute(SESSION_ID, sessionProvider.getSessionId())
+        logRecord.setSessionIdentifiersWith(sessionProvider)
```
LikeTheSalad, line 27: I don't think the difference from using an extension function here is worth creating one, especially since this use case should be done only here for all logs.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
```diff
@@ -94,6 +94,7 @@ object OpenTelemetryRumInitializer {
             else -> "none"
         }
 
+    @OptIn(Incubating::class)
```
LikeTheSalad, line 97: Is there a need to annotate this private function as Incubating?

android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionConfig.kt
```diff
@@ -9,11 +9,87 @@ import kotlin.time.Duration
...
+ *   ends and a new one will be created when the app becomes active again.
+ *
+ * @param maxLifetime maximum duration a session can remain active regardless of app activity.
+ *   Default is 4 hours, meaning even if the app stays in foreground continuously, sessions will
+ *   rotate every 4 hours to prevent unbounded session growth and ensure fresh session context.
+ *
+ * @see io.opentelemetry.android.agent.session.SessionManager
+ * @see io.opentelemetry.android.session.SessionProvider
+ */
+class SessionConfig(
```
LikeTheSalad, line 52: I'm a bit confused about why we need to expose this class?

android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/factory/SessionManagerFactory.kt
LikeTheSalad, line 32: What's the intended use case for this factory? `SessionManager` is supposed to be our opinionated implementation of `SessionProvider`, so users shouldn't ever need to manually create `SessionManager`.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/factory/SessionProviderFactory.kt
```diff
@@ -0,0 +1,29 @@
...
+ * Factory for creating [SessionProvider] instances.
+ *
+ * This interface follows the Factory design pattern to enable flexible dependency injection
+ * and testing of session management components. Implementations can provide custom session
+ * providers with different behaviors while maintaining a consistent creation interface.
+ *
+ * @see SessionManagerFactory
+ * @see SessionProvider
+ */
+interface SessionProviderFactory {
```
LikeTheSalad, line 21: I'm also not sure what could be the use case for this factory. Users should be able to create their own implementation of `SessionProvider` and [pass it to OpenTelemetryRumBuilder](https://github.com/open-telemetry/opentelemetry-android/blob/495ef25dc7b4798c0684d35b1be34e489170adcc/core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.kt#L260) if they want a custom session management.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/factory/SessionManagerFactory.kt
LikeTheSalad: Btw, OpenTelemetryRumBuilder [already allows to set](https://github.com/open-telemetry/opentelemetry-android/blob/495ef25dc7b4798c0684d35b1be34e489170adcc/core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.kt#L260) an instance of `SessionProvider`. If we wanted to allow users of `android-agent` to provide their own session logic, we could allow them to set their own instance of `SessionProvider` in the initializer function, but this is something that would require a deeper talk, as the agent is supposedly our opinionated way of configuring OpenTelemetryRumBuilder.

LikeTheSalad (changes requested):
Thank you for taking the time to work on these changes, @gregory-at-cvs. I've left a couple of comments and some questions regarding the approaches used here.

PR #1476, opened 2025-12-12
Add ability to override resource from OpenTelemetryRumInitializer

### LINKED ISSUE #1471: Allow otel Resource customization through agent/initializer.
See https://github.com/open-telemetry/opentelemetry-android/issues/1257#issuecomment-3617125034

The ask is to be able to customize and/or override parts of the Resource. This is possible through the `OpenTelemetryRumBuilder` class in `core`, but we want users to use the agent and discourage direct usage of the OTRB where possible.

fractalwrench:
## Goal

Makes it possible to override resource attributes from `OpenTelemetryRumInitializer`. This adds a new API to the DSL that allows actions to be performed on a `ResourceBuilder` instance when the `OpenTelemetryRumBuilder` is instantiated.

I did consider creating a more restrictive API than exposing `ResourceBuilder` as it does couple the project's API to opentelemetry-java's API. However, I figured the benefits of being able to merge resources and folks potentially being familiar with the existing `Resource` API surface outweighed the cons.

Fixes #1471

## Testing

Added unit tests to confirm the builder is altered.

### DIFF
--- a/android-agent/api/android-agent.api
+++ b/android-agent/api/android-agent.api
@@ -45,6 +45,7 @@ public final class io/opentelemetry/android/agent/dsl/OpenTelemetryConfiguration
 	public final fun globalAttributes (Lkotlin/jvm/functions/Function0;)V
 	public final fun httpExport (Lkotlin/jvm/functions/Function1;)V
 	public final fun instrumentations (Lkotlin/jvm/functions/Function1;)V
+	public final fun resource (Lkotlin/jvm/functions/Function1;)V
 	public final fun session (Lkotlin/jvm/functions/Function1;)V
 }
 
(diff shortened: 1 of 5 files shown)


fractalwrench:
@cleverchuk it's kind of tricky to test on `OpenTelemetryRumInitializer` right now, as there isn't anything that publicly exposes the created resource object. I'm hoping to create some smoke tests in an unrelated PR that create `OpenTelemetryRum` and assert that a payload is received via OkHttp's MockWebServer, so if I get that over the line it would include coverage for this.

breedx-splk:
> How about a test for `OpenTelemetryRumInitializer.kt` to exercise the change?

I agree that this is important to have test for, but it can be done with a follow-up effort. Let's get this in now. Thanks all!

PR #1485, opened 2025-12-18
Add smoke tests for OpenTelemetryRum

fractalwrench:
## Goal

Begins to address #405 by creating an integration test that asserts `OpenTelemetryRum` is capable of exporting OTLP. This is achieved by [using `MockWebServer`](https://github.com/square/okhttp/tree/master/mockwebserver) and deserializing requests into OpenTelemetry's protobuf models.

For now, I've just asserted that _something_ is exported for both logs and traces, as that gives reassurance that the agent isn't completely broken. In future PRs we have the option of asserting against the Protobuf model class or asserting against a golden file of known output. I'd also envision writing more test cases to exercise different parts of the API.

Proof of the usefulness of these tests is demonstrated by the fact it found a bug with the disk buffering configuration. If `diskBuffering.enabled` was set to `false` in the DSL, this was ignored due to a logic bug, meaning disk buffering effectively couldn't be disabled via `OpenTelemetryRumInitializer`. I've fixed this in the changeset. In future we should probably also add some test cases that confirm data is delivered when using the disk buffering exporter too.

### DIFF
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OpenTelemetryConfiguration.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OpenTelemetryConfiguration.kt
@@ -8,7 +8,6 @@ package io.opentelemetry.android.agent.dsl
 import io.opentelemetry.android.Incubating
 import io.opentelemetry.android.agent.dsl.instrumentation.InstrumentationConfiguration
 import io.opentelemetry.android.config.OtelRumConfig
-import io.opentelemetry.android.features.diskbuffering.DiskBufferingConfig
 import io.opentelemetry.api.common.Attributes
 import io.opentelemetry.sdk.resources.ResourceBuilder
 
@@ -19,16 +18,12 @@ import io.opentelemetry.sdk.resources.ResourceBuilder
 @OpenTelemetryDslMarker
 class OpenTelemetryConfiguration internal constructor(
     internal val rumConfig: OtelRumConfig = OtelRumConfig(),
+    internal val diskBufferingConfig: DiskBufferingConfigurationSpec = DiskBufferingConfigurationSpec(rumConfig),
 ) {
     internal val exportConfig = HttpExportConfiguration()
     internal val sessionConfig = SessionConfiguration()
-    internal val diskBufferingConfig = DiskBufferingConfigurationSpec()
     internal val instrumentations = InstrumentationConfiguration(rumConfig)
     internal var resourceAction: ResourceBuilder.() -> Unit = {}
-
-    init {
-        diskBuffering {}
-    }
 
     /**
      * Configures how OpenTelemetry should export telemetry over HTTP.
@@ -63,7 +58,6 @@ class OpenTelemetryConfiguration internal constructor(
      */
     fun diskBuffering(action: DiskBufferingConfigurationSpec.() -> Unit) {
         diskBufferingConfig.action()
-        rumConfig.setDiskBufferingConfig(DiskBufferingConfig.create(enabled = diskBufferingConfig.enabled))
     }
 
     /**
--- a/android-agent/src/test/kotlin/io/opentelemetry/android/agent/smoke/FakeOpenTelemetryServer.kt
+++ b/android-agent/src/test/kotlin/io/opentelemetry/android/agent/smoke/FakeOpenTelemetryServer.kt
@@ -1,0 +1,106 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.android.agent.smoke
+
+import io.opentelemetry.proto.collector.logs.v1.ExportLogsServiceRequest
+import io.opentelemetry.proto.collector.trace.v1.ExportTraceServiceRequest
+import mockwebserver3.Dispatcher
+import mockwebserver3.MockResponse
+import mockwebserver3.MockWebServer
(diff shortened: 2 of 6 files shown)


android-agent/src/test/kotlin/io/opentelemetry/android/agent/smoke/OpenTelemetryRumSmokeTest.kt
```diff
@@ -0,0 +1,86 @@
...
+                diskBufferingConfig.enabled(false)
+            },
+            action = {
+                val logger = openTelemetry.logsBridge.get("logger")
+                logger.logRecordBuilder().setBody("Hello world").emit()
+            },
+        )
+
+        val logRequest = server.awaitLogRequest()
+        checkNotNull(logRequest)
```
breedx-splk: Can we try and make this a bit more test-like?

```suggestion
        assertThat(
            logRequest
                .getResourceLogs(0)
                .scopeLogsList
                .first { x ->
                    x.scope.name.equals("logger")
                }.getLogRecords(0)
                .body.stringValue,
        ).isEqualTo("Hello world")
```

android-agent/src/test/kotlin/io/opentelemetry/android/agent/smoke/FakeOpenTelemetryServer.kt
```diff
@@ -0,0 +1,64 @@
...
+            ExportTraceServiceRequest.parseFrom(it)
+        }
+
+    /**
+     * Waits for a log request or throws after a timeout.
+     */
+    fun awaitLogRequest(): ExportLogsServiceRequest =
+        awaitRequest {
+            ExportLogsServiceRequest.parseFrom(it)
+        }
```
breedx-splk: These will explode (and fail tests I assume) if the agent happens to export one (unexpected) signal type before another, right? That is, if we're awaiting a trace and a log export happens first, we'll get an exception?

If that's the case, then I'm very minimally concerned that this could cause flappy tests, because I think there's nothing or at least very little that coordinates the order of those exports. I hate to over-engineer things, but I wonder if we'd be better off queueing up requests and then peeking at them to determine which deserializer to use....or just keep trying from the queue/buffer until one works? 

Imagine a future world where a (more complicated) smoke tests wants to test both logs and spans, and the order isn't predictable. We can kick that can down the road, but it's food for thought.

android-agent/src/test/kotlin/io/opentelemetry/android/agent/smoke/FakeOpenTelemetryServer.kt
```diff
@@ -0,0 +1,64 @@
...
+import java.io.InputStream
+import java.util.concurrent.TimeUnit
+import java.util.zip.GZIPInputStream
+
+/**
+ * Provides a fake server that can be used to test OTLP export. It launches a mock HTTP server
+ * and provides functions that can wait for requests to be received and then deserializes them
+ * into Protobuf models that can be asserted against.
+ */
+internal class FakeOpenTelemetryServer {
```
breedx-splk, line 26: We have a fake collector in the upstream testing utilities, but I like how simple this is.

android-agent/src/test/kotlin/io/opentelemetry/android/agent/smoke/OpenTelemetryRumSmokeTest.kt
fractalwrench: I've enhanced the test assertion here for both the log + trace test cases.

android-agent/src/test/kotlin/io/opentelemetry/android/agent/smoke/FakeOpenTelemetryServer.kt
fractalwrench: Yep, that's a problem. I've altered the code so that it uses a `Dispatcher` instead that places requests into a queue. The test case then searches for any log/trace that matches a predicate & waits if it hasn't arrived yet. I think this should work better for more complex test cases.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OpenTelemetryConfiguration.kt
```diff
@@ -63,7 +58,6 @@ class OpenTelemetryConfiguration internal constructor(
      */
     fun diskBuffering(action: DiskBufferingConfigurationSpec.() -> Unit) {
         diskBufferingConfig.action()
-        rumConfig.setDiskBufferingConfig(DiskBufferingConfig.create(enabled = diskBufferingConfig.enabled))
```
breedx-splk, line 66: A little surprised to see these unrelated changes in this PR but whatever, it's fine that this moves into the config spec class.

android-agent/src/test/kotlin/io/opentelemetry/android/agent/smoke/FakeOpenTelemetryServer.kt
```diff
@@ -0,0 +1,106 @@
...
+    private val logRequests: MutableCollection<ExportLogsServiceRequest> = ConcurrentLinkedQueue()
+    private val traceRequests: MutableCollection<ExportTraceServiceRequest> =
+        ConcurrentLinkedQueue()
+
+    private val requestHandler: Dispatcher =
+        object : Dispatcher() {
+            override fun dispatch(request: RecordedRequest): MockResponse {
+                when (request.target) {
+                    "/v1/traces" -> {
```
breedx-splk, line 35: yeah having separate endpoints is great.

PR #1486, opened 2025-12-19
Add ability to override clock

fractalwrench:
## Goal

Adds the ability to override the `Clock` used by specifying an instance either on `OpenTelemetryRumInitializer` or `OpenTelemetryRumBuilder`. This partially addresses #1364 as it allows overriding from the default behavior, and will also aid testing since it'll be possible to provide a fake implementation that is locked to a specific time.

Most of the changeset is just wiring through the `Clock` instance to various classes & test code. In a subsequent PR I'll look at implementing a better default clock for Android.

## Testing

Added unit tests for the configuration DSL.

### DIFF
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
@@ -42,12 +42,6 @@ object OpenTelemetryRumInitializer {
         val cfg = OpenTelemetryConfiguration()
         configuration(cfg)
 
-        val sessionConfig =
-            SessionConfig(
-                cfg.sessionConfig.backgroundInactivityTimeout,
-                cfg.sessionConfig.maxLifetime,
-            )
-
         // ensure we're using the Application Context to prevent potential leaks.
         // if context.applicationContext is null (e.g. called from within attachBaseContext),
         // fallback to the supplied context.
@@ -67,8 +61,9 @@ object OpenTelemetryRumInitializer {
 
         return RumBuilder
             .builder(ctx, cfg.rumConfig)
-            .setSessionProvider(createSessionProvider(ctx, sessionConfig))
+            .setSessionProvider(createSessionProvider(ctx, cfg))
             .setResource(resource)
+            .setClock(cfg.clock)
             .addSpanExporterCustomizer {
                 OtlpHttpSpanExporter
                     .builder()
@@ -101,10 +96,16 @@ object OpenTelemetryRumInitializer {
 
     private fun createSessionProvider(
         context: Context,
-        sessionConfig: SessionConfig,
+        cfg: OpenTelemetryConfiguration,
     ): SessionProvider {
-        val timeoutHandler = SessionIdTimeoutHandler(sessionConfig)
+        val sessionConfig =
+            SessionConfig(
+                cfg.sessionConfig.backgroundInactivityTimeout,
+                cfg.sessionConfig.maxLifetime,
+            )
+        val clock = cfg.clock
+        val timeoutHandler = SessionIdTimeoutHandler(sessionConfig, clock)
         Services.get(context).appLifecycle.registerListener(timeoutHandler)
-        return SessionManager.create(timeoutHandler, sessionConfig)
+        return SessionManager.create(timeoutHandler, sessionConfig, clock)
     }
 }
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OpenTelemetryConfiguration.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OpenTelemetryConfiguration.kt
@@ -10,6 +10,7 @@ import io.opentelemetry.android.agent.dsl.instrumentation.InstrumentationConfigu
 import io.opentelemetry.android.config.OtelRumConfig
 import io.opentelemetry.android.features.diskbuffering.DiskBufferingConfig
 import io.opentelemetry.api.common.Attributes
+import io.opentelemetry.sdk.common.Clock
 import io.opentelemetry.sdk.resources.ResourceBuilder
 
 /**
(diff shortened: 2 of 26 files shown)


instrumentation/activity/src/main/java/io/opentelemetry/android/instrumentation/activity/startup/AppStartupTimer.java
```diff
@@ -26,8 +25,8 @@ public class AppStartupTimer {
     private static final long MAX_TIME_TO_UI_INIT = TimeUnit.MINUTES.toNanos(1);
 
     // exposed so it can be used for the rest of the startup sequence timing.
-    private final AnchoredClock startupClock = AnchoredClock.create(Clock.getDefault());
-    private final long firstPossibleTimestamp = startupClock.now();
+    @Nullable private AnchoredClock startupClock = null;
+    private long firstPossibleTimestamp;
```
fractalwrench, line 29: In practice `AppStartupTimer` is only initialized at the same time as when `start` is invoked.

instrumentation/activity/src/main/java/io/opentelemetry/android/instrumentation/activity/startup/AppStartupTimer.java
```diff
@@ -53,11 +54,6 @@ public Span start(Tracer tracer) {
         return appStart;
     }
 
-    /** Returns the epoch timestamp in nanos calculated by the startupClock. */
-    public long clockNow() {
```
fractalwrench, line 57: This and several other bits were unused so I removed them.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OpenTelemetryConfiguration.kt
```diff
@@ -28,6 +29,11 @@ class OpenTelemetryConfiguration internal constructor(
         diskBuffering {}
     }
 
+    /**
+     * Configures the [Clock] used for capturing telemetry.
```
breedx-splk:
```suggestion
     * Configures the [Clock] used for capturing telemetry. Defaults to Clock.getDefault().
```

android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
```diff
@@ -61,7 +62,8 @@ object OpenTelemetryRumInitializer {
         val metricsEndpoint = cfg.exportConfig.metricsEndpoint()
         return RumBuilder
             .builder(ctx, cfg.rumConfig)
-            .setSessionProvider(createSessionProvider(ctx, sessionConfig))
+            .setSessionProvider(createSessionProvider(ctx, sessionConfig, cfg.clock))
```
breedx-splk: I think it'll be slightly cleaner now to move the creation of the `sessionConfig` variable into the `createSessionProvider()` method, and just pass in the `cfg`, reducing the argument count to 2 here and shrinking the initialize method length.

android-agent/src/test/kotlin/io/opentelemetry/android/agent/dsl/ClockConfigTest.kt
```diff
@@ -0,0 +1,33 @@
...
+            object : Clock {
+                override fun now(): Long = 0
+
+                override fun nanoTime(): Long = 0
+            }
+        val otelConfig =
+            OpenTelemetryConfiguration().apply {
+                clock = fakeClock
+            }
+        assertSame(fakeClock, otelConfig.clock)
```
breedx-splk: nit: we have been using assertJ elsewhere -- so it becomes something like `assertThat(otelConfig.clock).isSameAs(fakeClock)`. Would be nice to stay consistent.

instrumentation/activity/src/main/java/io/opentelemetry/android/instrumentation/activity/startup/AppStartupTimer.java
```diff
@@ -116,35 +112,4 @@ private void clear() {
...
-
-        @Override
-        public void run() {
-            // check whether an activity has been created
-            if (!startupTimer.uiInitStarted) {
-                Log.d(RumConstants.OTEL_RUM_LOG_TAG, "Detected background app start");
-                startupTimer.isStartedFromBackground = true;
-            }
-        }
-    }
```
breedx-splk, line 149: oh yeah I think way back when all of this somehow helped to account for app starts that were triggered by notifications or somehow other background activity. I guess that hasn't been wired up for quite some time?

breedx-splk (commented):
So I like this approach -- I think the implementation is pretty straightforward and easy to follow, so not really many comments about that.

I respect that the description says that it partially address some issue(s), so there's likely still more to be done. What makes me pause is that exposing this new api on the agent kinda suggests to users that they should start providing some clock of their own design in certain cases (eg. when they want to handle sleeps or NTP or whatever)...which I think is NOT what we want most users to be doing. 

Ideally, we will provide our own internal, robust Clock implementation that satisfies the existing issues (again, time slips during sleep, drift accounted by a time server, etc). User provided clock might be a rare exception then, which may be justified on the OpenTelemetryRumBuilder but maybe not all the way up on the agent.

Thoughts?

breedx-splk:
Oh, I also see the follow-up PR now, sorry for the thrash. It's probably still worth discussing whether or not we want this all the way up in the agent, or if just the OpenTelemetryRumBuilder is sufficient.

android-agent/src/test/kotlin/io/opentelemetry/android/agent/dsl/ClockConfigTest.kt
fractalwrench: Got it, will try to remember for future. My muscle memory uses the default JUnit assertions but it's definitely worth staying consistent with whatever convention we're using already.

fractalwrench:
@breedx-splk I agree I wouldn't expect most folks to be supplying their own clock implementation, the default implementation should take care of these sorts of corner cases. I do think it's helpful for our testing to override the global clock and it's also useful as an escape hatch for folks who do want to use a different implementation.

Longer-term I'm of the opinion that `OpenTelemetryRumInitializer` should supersede `OpenTelemetryRumBuilder` although that is quite a long-term goal and obviously requires discussion with other folks. I'd be fine with some solution like an internal interface/only putting the clock on `OpenTelemetryRumBuilder` for now, as long as it doesn't impact the ability to fake the current time in tests.

PR #1488, opened 2025-12-19
Implement `app.installation.id` resource attribute 

### LINKED ISSUE #922: Implement `app.installation.id` resource attribute
The experimental semconv [`app.installation.id`](https://opentelemetry.io/docs/specs/semconv/attributes-registry/app/) got accepted recently (https://github.com/open-telemetry/semantic-conventions/pull/1897), I think it would be a useful addition to the existing attributes provided by the SDK.

What makes its implementation a bit tricky, is that there are multiple ways you can implement this resource attribute on Android. In the specs we have listed these four options:

> - [Firebase Installation ID](https://firebase.google.com/docs/projects/manage-installations).
> - A globally unique UUID which is persisted across sessions in your application.
> - [App set ID](https://developer.android.com/identity/app-set-id).
> - [`Settings.getString(Settings.Secure.ANDROID_ID)`](https://developer.android.com/reference/android/provider/Settings.Secure#ANDROID_ID).

As @bidetofevil [pointed out](https://github.com/open-telemetry/semantic-conventions/pull/1897/files#r2002116621) the most practical choice for this SDK is probably number two (UUID), however, I could imaging exposing a setter for each of the options, as [they all have pros and cons](https://github.com/open-telemetry/semantic-conventions/issues/1874#issuecomment-2638963799).

```kotlin
OpenTelemetryRum.builder(this, config)
  .appInstallationIdProvider(UuidAppInstallationIdProvider()) // default behavior, if nothing is specified
  // or
  .appInstallationIdProvider(FirebaseAppInstallationIdProvider())
  // or
  .appInstallationIdProvider(AppSetIdAppInstallationIdProvider())
  // or
  .appInstallationIdProvider(AndroidIdAppInstallationIdProvider())
  // or
  .appInstallationIdProvider(object : AppInstallationIdProvider() { ... })
  // ...
```

fractalwrench:
## Goal

Closes #922 by implementing the `app.installation.id` resource attribute. This has been implemented as a random UUID that is persisted in `SharedPreferences` and is read at launch.

I've also made `AndroidResource` internal as it seems like an implementation detail, and reworked the test assertions slightly.

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/AndroidResource.kt
+++ b/core/src/main/java/io/opentelemetry/android/AndroidResource.kt
@@ -12,6 +12,7 @@ import io.opentelemetry.semconv.ServiceAttributes.SERVICE_NAME
 import io.opentelemetry.semconv.ServiceAttributes.SERVICE_VERSION
 import io.opentelemetry.semconv.TelemetryAttributes.TELEMETRY_SDK_VERSION
 import io.opentelemetry.semconv.incubating.AndroidIncubatingAttributes.ANDROID_OS_API_LEVEL
+import io.opentelemetry.semconv.incubating.AppIncubatingAttributes.APP_INSTALLATION_ID
 import io.opentelemetry.semconv.incubating.DeviceIncubatingAttributes.DEVICE_MANUFACTURER
 import io.opentelemetry.semconv.incubating.DeviceIncubatingAttributes.DEVICE_MODEL_IDENTIFIER
 import io.opentelemetry.semconv.incubating.DeviceIncubatingAttributes.DEVICE_MODEL_NAME
@@ -19,7 +20,9 @@ import io.opentelemetry.semconv.incubating.OsIncubatingAttributes.OS_DESCRIPTION
 import io.opentelemetry.semconv.incubating.OsIncubatingAttributes.OS_NAME
 import io.opentelemetry.semconv.incubating.OsIncubatingAttributes.OS_TYPE
 import io.opentelemetry.semconv.incubating.OsIncubatingAttributes.OS_VERSION
+import java.util.UUID
 
+private const val SHARED_PREF_FILE = "opentelemetry-android"
 private const val DEFAULT_APP_NAME = "unknown_service:android"
 
 object AndroidResource {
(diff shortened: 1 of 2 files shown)


MustafaHaddara (approved):
seems reasonable to me

core/src/main/java/io/opentelemetry/android/AndroidResource.kt
```diff
@@ -41,9 +43,23 @@ object AndroidResource {
...
+        val prefs = context.getSharedPreferences(SHARED_PREF_FILE, 0)
+        val installId = prefs.getString(APP_INSTALLATION_ID.key, null)
+
+        if (installId == null) {
+            val id = UUID.randomUUID().toString()
+            prefs.edit().putString(APP_INSTALLATION_ID.key, id).apply()
+            return id
+        } else {
+            return installId
+        }
```
breedx-splk: remove redundant else.

```suggestion
        return installId
```

core/src/main/java/io/opentelemetry/android/AndroidResource.kt
```diff
@@ -12,17 +12,19 @@ import io.opentelemetry.semconv.ServiceAttributes.SERVICE_NAME
...
 import io.opentelemetry.semconv.incubating.OsIncubatingAttributes.OS_DESCRIPTION
 import io.opentelemetry.semconv.incubating.OsIncubatingAttributes.OS_NAME
 import io.opentelemetry.semconv.incubating.OsIncubatingAttributes.OS_TYPE
 import io.opentelemetry.semconv.incubating.OsIncubatingAttributes.OS_VERSION
+import java.util.UUID
 
 private const val DEFAULT_APP_NAME = "unknown_service:android"
 
-object AndroidResource {
+internal object AndroidResource {
```
breedx-splk: 👍🏻 good call.

core/src/main/java/io/opentelemetry/android/AndroidResource.kt
```diff
@@ -79,4 +95,6 @@ object AndroidResource {
                 .append(")")
                 .toString()
         }
+
+    private const val SHARED_PREF_FILE = "opentelemetry-android"
```
breedx-splk: let's move this up to the top next to the other constant, to keep them easier to find.

core/src/main/java/io/opentelemetry/android/AndroidResource.kt
```diff
@@ -41,9 +43,23 @@ object AndroidResource {
             .put(OS_TYPE, "linux")
             .put(OS_VERSION, Build.VERSION.RELEASE)
             .put(OS_DESCRIPTION, oSDescription)
+            .put(APP_INSTALLATION_ID, readInstallId(context))
             .build()
     }
 
+    private fun readInstallId(context: Context): String {
+        val prefs = context.getSharedPreferences(SHARED_PREF_FILE, 0)
+        val installId = prefs.getString(APP_INSTALLATION_ID.key, null)
```
breedx-splk, line 54: Initial thought is that it might help to have a comment that the pref key is the same as the resource attribute key.....but maybe that's unnecessary. They don't NEED to be the same, but I think it's convenient that they are.

core/src/main/java/io/opentelemetry/android/AndroidResource.kt
fractalwrench: I ended up needing to keep this `public` for now as `AndroidResource` is currently used from another module (specifically `OpenTelemetryRumInitializer`)

PR #1462, opened 2025-12-05
Update docs for strategy on supporting minimum versions

fractalwrench:
## Goal

Updates the documentation around our strategy for what minimum versions this project supports. This may need a bit of discussion in the SIG. If discussion ends up being a blocker I'd be happy to split the changeset into controversial/non-controversial points, so we can at least get something documented.

Fixes #466

PR #1489, opened 2025-12-19
Bump minimum supported Kotlin to 2.0

fractalwrench:
## Goal

Bumps the minimum supported Kotlin version to 2.0, which allows updating to build with 2.3.0 (#1481 failed as additional steps are required). 2.0 is the [lowest supported version](https://kotlinlang.org/docs/compatibility-guide-23.html#drop-support-in-language-version-for-1-8-and-1-9) by 2.3.0 and emits a warning when targeting the `apiVersion`, so we'll need to bump again when Kotlin 2.4 is released.


instrumentation/compose/click/src/main/kotlin/io/opentelemetry/instrumentation/compose/click/ComposeClickEventGenerator.kt
```diff
@@ -41,7 +41,8 @@ internal class ComposeClickEventGenerator(
                     .setAttribute(APP_SCREEN_COORDINATE_X, motionEvent.x.toLong())
                     .emit()
 
-                composeTapTargetDetector.findTapTarget(window.decorView, motionEvent.x, motionEvent.y)?.let { layoutNode ->
+                val node: LayoutNode? = composeTapTargetDetector.findTapTarget(window.decorView, motionEvent.x, motionEvent.y)
```
fractalwrench, line 44: The compose code + tests failed in 2.0 without specifying an explicit type for this function.

common/api/common.api
```diff
@@ -81,6 +81,7 @@ public final class io/opentelemetry/android/common/internal/features/networkattr
 	public static final field TRANSPORT_VPN Lio/opentelemetry/android/common/internal/features/networkattributes/data/NetworkState;
 	public static final field TRANSPORT_WIFI Lio/opentelemetry/android/common/internal/features/networkattributes/data/NetworkState;
 	public static final field TRANSPORT_WIRED Lio/opentelemetry/android/common/internal/features/networkattributes/data/NetworkState;
+	public static fun getEntries ()Lkotlin/enums/EnumEntries;
```
fractalwrench, line 84: `entries` is generated by the compiler from Kotlin 1.9 onwards

fractalwrench:
CodeQL doesn't support Kotlin 2.3 yet, but is working on it here: https://github.com/github/codeql/issues/20661

breedx-splk:
Since I went ahead and merged #1462, can we update the note in `VERSIONING.md` as part of this PR please, @fractalwrench ?

fractalwrench:
@breedx-splk done. This does drop support for older versions of Kotlin so I would consider this a breaking change for when we do merge this.

marandaneto (approved):

fractalwrench:
@breedx-splk @LikeTheSalad this now passes CI as CodeQL has added support for Kotlin 2.3. We should decide when to merge this as it does drop support for older Kotlin versions.

LikeTheSalad:
> @breedx-splk @LikeTheSalad this now passes CI as CodeQL has added support for Kotlin 2.3. We should decide when to merge this as it does drop support for older Kotlin versions.

You're right. I think this is a good topic for the next SIG meeting.

common/api/common.api
breedx-splk: ...today I learned....

gradle/gradle-daemon-jvm.properties
```diff
@@ -0,0 +1,13 @@
...
+toolchainUrl.LINUX.AARCH64=https\://api.foojay.io/disco/v3.0/ids/29ee363f71d060405f729a8f1b7f7aef/redirect
+toolchainUrl.LINUX.X86_64=https\://api.foojay.io/disco/v3.0/ids/ecd23fd7707c683afbcd6052998cb6a9/redirect
+toolchainUrl.MAC_OS.AARCH64=https\://api.foojay.io/disco/v3.0/ids/10fc3bf1ee0001078a473afe6e43cfdb/redirect
+toolchainUrl.MAC_OS.X86_64=https\://api.foojay.io/disco/v3.0/ids/9c55677aff3966382f3d853c0959bfb2/redirect
+toolchainUrl.UNIX.AARCH64=https\://api.foojay.io/disco/v3.0/ids/29ee363f71d060405f729a8f1b7f7aef/redirect
+toolchainUrl.UNIX.X86_64=https\://api.foojay.io/disco/v3.0/ids/ecd23fd7707c683afbcd6052998cb6a9/redirect
+toolchainUrl.WINDOWS.AARCH64=https\://api.foojay.io/disco/v3.0/ids/39846e8427e64a3824c13e399d7d813c/redirect
+toolchainUrl.WINDOWS.X86_64=https\://api.foojay.io/disco/v3.0/ids/ac151d55def6b6a9a159dc4cb4642851/redirect
+toolchainVendor=JETBRAINS
+toolchainVersion=21
```
breedx-splk: do we know if renovate supports these?

instrumentation/compose/click/src/main/kotlin/io/opentelemetry/instrumentation/compose/click/ComposeClickEventGenerator.kt
breedx-splk: thanks for catching it

breedx-splk (approved):
Thanks for taking this on...seems like it's time! 🚀

breedx-splk:
I think kotlin 1.8 had its last release around the time this project was started in march 2023. 😅

fractalwrench:
Updated docs + changelog with a warning as discussed in the SIG today.

breedx-splk:
Needs a rebase @fractalwrench

PR #1493, opened 2026-01-03
Retry exporting unsuccessfully exported batches

bencehornak:
### Current behavior

Since #1326 when the SDK exports batches from the disk it essentially ignores any exporting failures and deletes the unsuccessfully exported batches from the disk. Consequently, valuable telemetry might be dropped silently, if there is any exception taking place in the OTel exporter (e.g. IOException due to the device being offline or the backend experiencing a temporary outage).

### Proposed behavior

The export scheduler stops exporting on the first failure to prevent the deletion of unconsumed signals. On the next periodic run, the exporting will be retried from the point where the previous attempt left off.

### Note

I think the root cause of the problem is the wrong example in the [disk buffering module's README](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/disk-buffering/README.md#reading-data), see https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2539.

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/features/diskbuffering/SignalFromDiskExporter.kt
+++ b/core/src/main/java/io/opentelemetry/android/features/diskbuffering/SignalFromDiskExporter.kt
@@ -79,7 +79,12 @@ class SignalFromDiskExporter
          * @return `true`, if all exporting completed successfully, `false` otherwise.`
          */
         @WorkerThread
-        fun exportAllSignalsFromDisk(): Boolean = exportSpansFromDisk() && exportMetricsFromDisk() && exportLogsFromDisk()
+        fun exportAllSignalsFromDisk(): Boolean {
+            val successSpans = exportSpansFromDisk()
+            val successMetrics = exportMetricsFromDisk()
+            val successLogs = exportLogsFromDisk()
+            return successSpans && successMetrics && successLogs
+        }
 
         companion object {
             private var instance: SignalFromDiskExporter? = null
--- a/core/src/test/java/io/opentelemetry/android/features/diskbuffering/scheduler/SignalFromDiskExporterTest.kt
+++ b/core/src/test/java/io/opentelemetry/android/features/diskbuffering/scheduler/SignalFromDiskExporterTest.kt
@@ -120,8 +120,12 @@ class SignalFromDiskExporterTest {
     @Test
     fun `Return false when spans export fails`() {
         every { spanStorage.iterator() }.returns(makeIterator(size = 1))
+        every { metricStorage.iterator() }.returns(makeIterator(size = 1))
+        every { logStorage.iterator() }.returns(makeIterator(size = 1))
 
         every { spanExporter.export(any()) }.returns(CompletableResultCode.ofFailure())
+        every { metricExporter.export(any()) }.returns(CompletableResultCode.ofSuccess())
+        every { logExporter.export(any()) }.returns(CompletableResultCode.ofSuccess())
 
         assertThat(instance.exportAllSignalsFromDisk()).isFalse()
         verify { spanExporter.export(any()) }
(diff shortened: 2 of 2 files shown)


LikeTheSalad (approved):
Thank you! 🙏

core/src/main/java/io/opentelemetry/android/features/diskbuffering/SignalFromDiskExporter.kt
```diff
@@ -29,75 +28,58 @@ class SignalFromDiskExporter
...
-            var atLeastOneWorked = exportBatchOfSpans()
-            if (exportBatchOfMetrics()) {
-                atLeastOneWorked = true
-            }
-            if (exportBatchOfLogs()) {
-                atLeastOneWorked = true
-            }
-            return atLeastOneWorked
-        }
+        fun exportAllSignalsFromDisk(): Boolean = exportSpansFromDisk() && exportMetricsFromDisk() && exportLogsFromDisk()
```
breedx-splk: One downside to this approach is that a failure in one telemetry receiver can impact the ability of the others to even attempt export. That is, if a client is configured to use separate endpoints for metrics, traces, and logs, an outage in the spans receiver endpoint will cause the metrics and logs to not even be attempted!

It's probably a minor point, because I think that most users will have the same endpoint for all telemetry, but this is not necessarily the case. Because the signals are stored separately, we _could_ attempt to export the other types even when one type fails. I'm not sure it's worth it, though...and doing so might cause some overengineering.

core/src/main/java/io/opentelemetry/android/features/diskbuffering/SignalFromDiskExporter.kt
```diff
@@ -29,75 +28,58 @@ class SignalFromDiskExporter
...
+         * @return `true`, if the exporting completely succeeded, `false` otherwise.
          */
         @WorkerThread
-        @Throws(IOException::class)
-        fun exportBatchOfLogs(): Boolean = export(logStorage, logExporter::export)
+        fun exportLogsFromDisk(): Boolean = export(logStorage, logExporter::export)
 
+        /**
+         * Exports all persisted spans/metrics/logs from the disk and deletes the successfully
+         * consumed batches from the disk. It aborts the exporting on the first failure.
```
breedx-splk, line 59: I don't love that the javadocs on these addresses the deletion -- as pointed out in the other repo/issue/PR, the deletion is a side effect of iteration. I know we're going to be addressing this, but because this class doesn't actually do the deletion, I don't think it should be describing behavior performed by a collaborator...unless that is explicitly called out.

core/src/main/java/io/opentelemetry/android/features/diskbuffering/SignalFromDiskExporter.kt
bencehornak: Good point, I oversaw this possibility. It doesn't cost much to implement the right way, I will make sure that all 3 methods are called regardless of failures.

core/src/main/java/io/opentelemetry/android/features/diskbuffering/SignalFromDiskExporter.kt
bencehornak: The deletion is implicitly there due to the current tricky API (`next()` with side-effect). Since you can't see the deleting/retrying spelled out in the current version, I'd argue that it's even more important to describe when data is removed from the persistent storage and whether a failure is retried. The android-sdk makes an opinionated choice here (there are at least two ways of consuming the disk-buffering lib, the old way by ignoring failures and the proposed way by retrying failures), so that choice is what I aimed to document here.

I can try to improve the documentation of the opinionated part, e.g. I could put an emphasis on the retry mechanism, like

```suggestion
         * Exports all persisted spans/metrics/logs from the disk.
         * If an export fails, the method aborts and retries exporting the
         * affected signals on the next invocation.
```

would that work for you @breedx-splk?

breedx-splk: I mean, I guess that's fine...but I still think it's incorrect for a javadoc to describe the internal workings of one of its collaborators. If (when!) the collaborator changes, we are left with incorrect javadoc. 

I think we can just roll with this for now and address that in a follow-up PR effort.

PR #1497, opened 2026-01-07
Add a one-off release workflow for the v1.0.0 release.

breedx-splk:
I have concluded that it's better just to do this one-off special release for v1.0.0 rather than attempt to make all of the release pipeline steps support the `-rc.1` suffix. It's too different from the other patch release workflows, and there are definitely too many moving pieces to be confident in getting this correct.

So...

The process will be this:

* merge this PR
* manually create an merge a prepare-release branch PR into `release/v1.0.x` (#1498)
* run this workflow to release v1.0.0
* manually create the GH release
* manually update `main` branch with PR to update changelog and readmes to 1.1.0
* delete this one-off workflow with another PR

after that we can put the `-rc.1` complications behind us....until the next time. 🙈

### DIFF
--- a/.github/workflows/release-1.0.0.yml
+++ b/.github/workflows/release-1.0.0.yml
@@ -1,0 +1,56 @@
+---
+name: "Special one-off 1.0.0 release"
+
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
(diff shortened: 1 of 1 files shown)


bidetofevil (approved):
LGTM

PR #1498, opened 2026-01-07
Prepare v1.0.0

breedx-splk:
Done manually instead of using the workflow, due to `-rc.1` suffix complicating things in a nonstandard way.


.github/workflows/release-1.0.0.yml
```diff
@@ -0,0 +1,56 @@
+---
+name: "Special one-off 1.0.0 release"
```
breedx-splk, line 2: Because this will only be run once manually, I think it will be sourced from the main branch, but including here just in case.

PR #1510, opened 2026-01-15
Tighten test assertion for OpenTelemetryRumSmokeTest

fractalwrench:
## Goal

Tightens a test assertion for `OpenTelemetryRumSmokeTest`. I believe this might be the cause of a [flaky test on CI](https://github.com/open-telemetry/opentelemetry-android/actions/runs/21014913071/job/60417939533?pr=1509#step:6:1526) where `testLogExported` fails to find the correct log.

The way the assertion was written before just grabbed the first log, whereas now it specifically waits until a log matching the body is received by the fake server. This should handle the case where the agent emits multiple logs and the test case just grabs whatever is sent first.

I've also tweaked the github workflows so that [`--stacktrace` is passed as an argument ](https://docs.gradle.org/current/userguide/logging.html#stacktraces) and that exceptions during tests have their full stacktrace logged out which should aid debugging .

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -24,7 +24,7 @@ jobs:
       - name: touch local props
         run: touch demo-app/local.properties
       - name: run gradle check
-        run: ./gradlew check -x detekt detektDebug detektDebugUnitTest detektDebugAndroidTest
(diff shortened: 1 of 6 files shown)


breedx-splk (approved):
Thanks!

PR #1525, opened 2026-01-21
Upgrade to AGP 9.0.0

priettt:
## Goal

Upgrade Android Gradle Plugin from 8.13.2 to 9.0.0, following the instructions here: https://developer.android.com/build/releases/agp-9-0-0-release-notes

## Changes

- Bump `androidPlugin` version to 9.0.0
- Remove explicit `org.jetbrains.kotlin.android` plugin declarations (now applied automatically by AGP 9)
- Enable `resValues` build feature explicitly in demo-app (required in AGP 9)
- Remove unused `rum.version` resource value from core module (legacy from Splunk RUM donation, the code now uses `BuildConfig.OTEL_ANDROID_VERSION` instead)
- Fix spacing in log warning messages in `SdkPreconfiguredRumBuilder` (new lint rule from AGP 9)
- Replaced detektDebug, detektDebugUnitTest and detektDebugAndroidTest (not working in AGP 9) with just detekt, adding test source sets to the detekt configuration
- Set explicit `aarMetadata.minCompileSdk` to 34, since AGP 9 defaults it to `compileSdk` (36) which would force consumers to compile with SDK 36. Setting it to 34 aligns with Google Play's minimum target SDK requirement.
- Allow cleartext HTTP traffic in test modules (httpurlconnection, okhttp3, okhttp3-websocket). AGP 9 changed the default `targetSdk` from `minSdk` to `compileSdk`, which means `targetSdk` is now 36 and cleartext HTTP is blocked by default, breaking test requests.

### DIFF
--- a/instrumentation/httpurlconnection/testing/src/main/AndroidManifest.xml
+++ b/instrumentation/httpurlconnection/testing/src/main/AndroidManifest.xml
@@ -1,4 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android" >
     <uses-permission android:name="android.permission.INTERNET" />
+    <application android:usesCleartextTraffic="true" />
 </manifest>
--- a/instrumentation/okhttp3-websocket/testing/src/main/AndroidManifest.xml
+++ b/instrumentation/okhttp3-websocket/testing/src/main/AndroidManifest.xml
@@ -1,4 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
     <uses-permission android:name="android.permission.INTERNET" />
+    <application android:usesCleartextTraffic="true" />
 </manifest>
--- a/instrumentation/okhttp3/testing/src/main/AndroidManifest.xml
+++ b/instrumentation/okhttp3/testing/src/main/AndroidManifest.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
-
     <uses-permission android:name="android.permission.INTERNET" />
+    <application android:usesCleartextTraffic="true" />
 </manifest>


.github/workflows/build.yaml
```diff
@@ -24,7 +24,7 @@ jobs:
       - name: touch local props
         run: touch demo-app/local.properties
       - name: run gradle check
-        run: ./gradlew check -x detekt detektDebug detektDebugUnitTest detektDebugAndroidTest --stacktrace
+        run: ./gradlew check --stacktrace
```
LikeTheSalad, line 27: > Replaced detektDebug, detektDebugUnitTest and detektDebugAndroidTest (not working in AGP 9) with just detekt

If I understood correctly, we should at least leave `detekt` in this line?

priettt: `detekt` hooks into the `check` gradle task, so when you run `check`, it will also run `detekt`.

You can verify it by running this command, that will not return anything as there are no differences in the tasks that will be executed:

```(diff <(./gradlew check --dry-run 2>&1 | grep "SKIPPED" | sort) <(./gradlew check detekt --dry-run 2>&1 | grep "SKIPPED" | sort))```

Comment (LikeTheSalad) [reply .github/workflows/build.yaml:27]: Awesome, thanks for the clarification!

File: core/build.gradle.kts [hunk 2]
```diff
@@ -50,6 +50,7 @@ android {
 
     buildFeatures {
         buildConfig = true
+        resValues = true
```
fractalwrench: What part of opentelemetry-android requires resource value generation?

priettt: we injected `rum.version` but it seems it isn't used anywhere. I'll remove it

LikeTheSalad:
@priettt Thank you for the time you've taken to address this issue. It's almost ready to go, but [some Android tests](https://github.com/open-telemetry/opentelemetry-android/actions/runs/21364107699/job/61490961868?pr=1525) are failing after these changes. Can you have a look?

fractalwrench:
`aarMetadata.minCompileSdk` should also be explicitly set to `34` IMO. There's a behavior change in AGP 9 that automatically sets it to the same as the `compileSdk` value, which I first [spotted here](https://github.com/open-telemetry/opentelemetry-kotlin/pull/214)

instrumentation/httpurlconnection/testing/src/main/AndroidManifest.xml
```diff
@@ -1,4 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android" >
     <uses-permission android:name="android.permission.INTERNET" />
+    <application android:usesCleartextTraffic="true" />
```
breedx-splk, line 4: What's this about?

breedx-splk: Oh right sorry, you addressed this in the description: it's to allow test requests, which fail under agp 9. 👍🏻

breedx-splk (approved):
Thanks for taking this on! Much appreciated.

PR #1530, opened 2026-01-23
Update spotless to 8.2.0

fractalwrench:
Updates Spotless to 8.2.0 and fixes some of the existing style violations that were breaking the renovate PR in #1527. I assume that Spotless has added some new rules or tweaked the defaults which is why these weren't picked up previously.


breedx-splk:
Supersedes #1527

Issue #1475, opened 2025-12-12
How to add client TLS in the OpenTelemetryRumInitializer?


brsgina:
Hi,

I am using the `OpenTelemetryRumInitializer.initialize` function in my Android project to create the `OpenTelemetryRum` instance and I would like to set a client TLS. With io.opentelemetry.android.instrumentation (+ io.opentelemetry.exporter.otlp) it was possible to build a `OtlpHttpLogRecordExporter` and set the client TLS to it with the `addLoggerProviderCustomizer` function:

`val logExporter = OtlpHttpLogRecordExporter.builder().setEndpoint(BASE_URL).setClientTls(keyBytes, crtBytes) .build()`

But if I understand correctly, there is currently no solution for this in the initializer configuration. Is there currently a way to set up a TLS certificate, or will it be possible to specify something in the configuration later?

Thank you in advance for the response!

breedx-splk:
Hi @brsgina thanks for the inquiry. Just to be clear, we are talking about client-certs, and not just enabling TLS (any old `https` endpoint url will work just fine for regular TLS). It looks like we don't currently expose a way to configure client TLS in the `OpenTelemetryRumInitializer`! 

This can currently be accomplished with the `OpenTelemetryRumBuilder` from the `core` module, but it is a bit more work to configure. The `Otlp*ExporterBuilder` classes from the java core repos have a `setClientTls()` method on them.

Given that Android apps run around in the wild and the endpoints need to talk directly with the devices, I could absolutely see the desire to have client-side certs for ingest servers to be able to validate that data is coming from a trusted source. I think this is something we should add at the level of the initializer.

brsgina:
Hi @breedx-splk , thank you for the quick response. Yes we are talking abut client-certs and yes, this would be a useful new feature :) Can you tell me approximately when or how soon I can expect a released version that already includes this configuration option?

breedx-splk:
> Hi [@breedx-splk](https://github.com/breedx-splk) , thank you for the quick response. Yes we are talking abut client-certs and yes, this would be a useful new feature :) 

😎 Cool thanks for the clarification.

> Can you tell me approximately when or how soon I can expect a released version that already includes this configuration option?

Well, that's not generally how this works. It's up to the community to agree on and build the feature, so timelines are difficult to predict. We're in the holiday season now, and you can expect work to slow down considerably for a while. I have a partial implementation started, but if I'm honest it probably won't get completed until January sometime. Thanks again!

[...]

PR #1537, opened 2026-01-26
Expose ssl context for dsl config

DamienL-F:
Our team needed to pass an SSL context and `OpenTelemetryConfiguration` was not providing a way for it.

### DIFF
--- a/android-agent/api/android-agent.api
+++ b/android-agent/api/android-agent.api
@@ -14,6 +14,12 @@ public final class io/opentelemetry/android/agent/connectivity/Compression : jav
 	public static fun values ()[Lio/opentelemetry/android/agent/connectivity/Compression;
 }
 
+public final class io/opentelemetry/android/agent/connectivity/SSLContextConnectivity {
+	public fun <init> (Ljavax/net/ssl/SSLContext;Ljavax/net/ssl/X509TrustManager;)V
+	public final fun getSslContext ()Ljavax/net/ssl/SSLContext;
+	public final fun getSslX509TrustManager ()Ljavax/net/ssl/X509TrustManager;
+}
+
 public final class io/opentelemetry/android/agent/dsl/DiskBufferingConfigurationSpec : io/opentelemetry/android/agent/dsl/instrumentation/CanBeEnabledAndDisabled {
 	public fun enabled (Z)V
 }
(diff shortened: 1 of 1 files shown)


breedx-splk:
~~Resolves #1475~~ maybe that is something similar but different enough to leave open.

breedx-splk:
Please run `./gradlew spotlessApply` to run the auto-formatter.

DamienL-F:
> Resolves #1475

I'm not sure about that. #1475 is about TLS, not SSL.
But #1475 would probably need a similar development.

DamienL-F:
> Thanks! This looks good to me, but I do wish we had some additional test coverage around this. How hard would it be to add?

I added unit tests on http export config.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/connectivity/SSLContextConnectivity.kt
```diff
@@ -0,0 +1,14 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.android.agent.connectivity
+
+import javax.net.ssl.SSLContext
+import javax.net.ssl.X509TrustManager
+
+data class SSLContextConnectivity(
```
fractalwrench: Data classes are discouraged in public APIs in Kotlin: https://kotlinlang.org/docs/api-guidelines-backward-compatibility.html#avoid-using-data-classes-in-your-api

I'd suggest using a regular class instead and also provide it [using a type-safe builder](https://kotlinlang.org/docs/type-safe-builders.html) which would be more in keeping with how the rest of the DSL is configured.

```kotlin
fun sslContext(action: SslConfiguration.() -> Unit) {

}

class SslConfiguration(
    var sslContext: SSLContext? = null,
    var sslX509TrustManager: X509TrustManager? = null,
)
```

DamienL-F: I think i got the point.
That being said, what disturbs me about this is `OtlpHttpSpanExporterBuilder.setSslContext` need for both `SSLContext/X509TrustManager` not being null.
So it wouldn't make sense doing `sslContext { sslContext =  }` without `sslX509TrustManager =`.

fractalwrench: My main concern here is exposing a data class as part of the public API. I'm less fussed about the type-safe builder.

DamienL-F: Ok, i changed the data class to class.

breedx-splk:
Hi @DamienL-F -- I think we're good with getting this in, but it looks like the merge never happened due to the build being broken.

...and it's only "broken" because we strictly track changes to the api, and this introduces a new one. As such, you will need to regenerate the api baseline with `./gradlew apiDump` and commit+push the results. It's easy to overlook this step, which we do have [documented at the bottom of CONTRIBUTING.md](https://github.com/open-telemetry/opentelemetry-android/blob/main/CONTRIBUTING.md#api-compatibility). 

Thanks again!

DamienL-F:
> Hi @DamienL-F -- I think we're good with getting this in, but it looks like the merge never happened due to the build being broken.
> 
> ...and it's only "broken" because we strictly track changes to the api, and this introduces a new one. As such, you will need to regenerate the api baseline with `./gradlew apiDump` and commit+push the results. It's easy to overlook this step, which we do have [documented at the bottom of CONTRIBUTING.md](https://github.com/open-telemetry/opentelemetry-android/blob/main/CONTRIBUTING.md#api-compatibility).
> 
> Thanks again!

Thanks. I missed it. This is done.

PR #1539, opened 2026-01-26
Instrumentation api change - option 1

LikeTheSalad:
DO NOT MERGE

This is a proposal for this issue: https://github.com/open-telemetry/opentelemetry-android/issues/1541 - It's meant to highlight the key changes, without addressing potential compilation errors, if any. If chosen as the approach we'd like to go with, then I'll add the rest of the changes needed for the build to pass.

### DIFF
--- a/agent-api/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
+++ b/agent-api/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
@@ -18,6 +18,10 @@ interface OpenTelemetryRum {
      */
     val openTelemetry: OpenTelemetry
 
+    val clock: Clock
+
+    val sessionProvider: SessionProvider
+
     /**
      * Get the client session ID associated with this instance of the RUM instrumentation library.
      * Note: this value will change throughout the lifetime of an application instance, so it is
--- a/agent-api/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
+++ b/agent-api/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
@@ -7,6 +7,7 @@ package io.opentelemetry.android
 
 import io.opentelemetry.api.OpenTelemetry
 import io.opentelemetry.api.common.Attributes
+import io.opentelemetry.sdk.common.CompletableResultCode
 
 /**
  * Entrypoint for the OpenTelemetry Real User Monitoring library for Android.
@@ -43,6 +44,8 @@ interface OpenTelemetryRum {
         body: String = "",
         attributes: Attributes = Attributes.empty(),
     )
+
+    fun flushLogRecords(): CompletableResultCode
 
     /**
      * Initiates orderly shutdown of this OpenTelemetryRum instance. After this method completes,
--- a/instrumentation/android-instrumentation/src/main/java/io/opentelemetry/android/instrumentation/InstallationContext.kt
+++ b/instrumentation/android-instrumentation/src/main/java/io/opentelemetry/android/instrumentation/InstallationContext.kt
@@ -8,14 +8,10 @@ package io.opentelemetry.android.instrumentation
 import android.app.Application
 import android.content.Context
 import io.opentelemetry.android.OpenTelemetryRum
-import io.opentelemetry.android.session.SessionProvider
-import io.opentelemetry.sdk.common.Clock
 
 class InstallationContext(
     val context: Context,
     val openTelemetry: OpenTelemetryRum,
-    val sessionProvider: SessionProvider,
-    val clock: Clock,
 ) {
     val application: Application? = context as? Application
 }
--- a/instrumentation/android-instrumentation/src/main/java/io/opentelemetry/android/instrumentation/InstallationContext.kt
+++ b/instrumentation/android-instrumentation/src/main/java/io/opentelemetry/android/instrumentation/InstallationContext.kt
@@ -7,13 +7,13 @@ package io.opentelemetry.android.instrumentation
 
 import android.app.Application
 import android.content.Context
+import io.opentelemetry.android.OpenTelemetryRum
 import io.opentelemetry.android.session.SessionProvider
-import io.opentelemetry.api.OpenTelemetry
 import io.opentelemetry.sdk.common.Clock
 
 class InstallationContext(
     val context: Context,
-    val openTelemetry: OpenTelemetry,
+    val openTelemetry: OpenTelemetryRum,
     val sessionProvider: SessionProvider,
     val clock: Clock,
 ) {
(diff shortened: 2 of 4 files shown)


instrumentation/crash/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashReporterInstrumentation.kt
```diff
@@ -27,7 +26,7 @@ class CrashReporterInstrumentation : AndroidInstrumentation {
         val crashReporter = CrashReporter(additionalExtractors)
 
         // TODO avoid using OpenTelemetrySdk methods, only use the ones from OpenTelemetry api.
```
breedx-splk, line 28: todo can go away if we do this

instrumentation/android-instrumentation/src/main/java/io/opentelemetry/android/instrumentation/InstallationContext.kt
```diff
@@ -7,15 +7,11 @@ package io.opentelemetry.android.instrumentation
...
-import io.opentelemetry.api.OpenTelemetry
-import io.opentelemetry.sdk.common.Clock
+import io.opentelemetry.android.OpenTelemetryRum
 
 class InstallationContext(
     val context: Context,
-    val openTelemetry: OpenTelemetry,
-    val sessionProvider: SessionProvider,
-    val clock: Clock,
+    val openTelemetry: OpenTelemetryRum,
```
breedx-splk, line 14: nitpick: I don't love this name now -- because it suggests something from the upstream api and not a rum class. I might call it `rum` or `openTelemetryRum`.

agent-api/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
```diff
@@ -40,6 +45,8 @@ interface OpenTelemetryRum {
         attributes: Attributes = Attributes.empty(),
     )
 
+    fun flushLogRecords(): CompletableResultCode
```
fractalwrench, line 48: IMO either `shutdown()` should achieve this functionality or we can expose `forceFlush()`. It feels like `flushLogRecords()` is probably exposed to make one particular feature work, but IMO it'd be preferable just to have one API that flushes all telemetry, _if_ we even want to expose that at all..

agent-api/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
```diff
@@ -18,6 +19,10 @@ interface OpenTelemetryRum {
      */
     val openTelemetry: OpenTelemetry
 
+    val clock: Clock
+
+    val sessionProvider: SessionProvider
```
fractalwrench, line 24: Happy with this but might be worth thinking about whether `SessionProvider.noop()` should still be exposed. Will end-users actually invoke that function?

instrumentation/android-instrumentation/src/main/java/io/opentelemetry/android/instrumentation/InstallationContext.kt
```diff
@@ -7,15 +7,11 @@ package io.opentelemetry.android.instrumentation
 
 import android.app.Application
 import android.content.Context
-import io.opentelemetry.android.session.SessionProvider
-import io.opentelemetry.api.OpenTelemetry
-import io.opentelemetry.sdk.common.Clock
+import io.opentelemetry.android.OpenTelemetryRum
 
 class InstallationContext(
```
fractalwrench, line 12: We could perhaps consider whether `InstallationContext` is required anymore given that `Context` and `OpenTelemetryRum` could plausibly be passed into the `install/uninstall` functions instead.

If we do keep it, perhaps it's worth renaming to something like `InstallationParams` given that `Context` has a special meaning on Android already, _and_ that OTel already has a separate concept of context.

instrumentation/android-instrumentation/src/main/java/io/opentelemetry/android/instrumentation/InstallationContext.kt
```diff
@@ -7,15 +7,11 @@ package io.opentelemetry.android.instrumentation
...
     val context: Context,
```
fractalwrench, line 13: I definitely like the fact that there are only 2 params left over here now.

instrumentation/android-instrumentation/src/main/java/io/opentelemetry/android/instrumentation/InstallationContext.kt
LikeTheSalad: Makes sense. I'm up for renaming it to `rum`.

agent-api/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
LikeTheSalad: > It feels like flushLogRecords() is probably exposed to make one particular feature work

That's true, it'd only be useful for crash reporting, at least for now.

> shutdown() should achieve this functionality

I'm open to that option, though I would have some questions. For example, is it ok to shut down RUM from an instrumentation? Or maybe it won't be needed to do so because `shutdown()` _should_ be called on a crash automatically anyway (via our shutdownHook)? - If we decide on the latter, can we ensure that the app will have enough time to get to the SDK shutdown function before the OS kills the app?

> or we can expose forceFlush()

I'm open to going with a "one method to flush all signals" approach. I'd still be a bit wary about the time it'd need to take to finish executing it all before the OS kills the app on a crash, though. But it sounds better than relying on the shutdown function, because `forceFlush` would be called directly from the instrumentation.

instrumentation/android-instrumentation/src/main/java/io/opentelemetry/android/instrumentation/InstallationContext.kt
LikeTheSalad: I don't have a strong opinion on this, so I'm ok with either removing it or renaming it 👍

agent-api/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
LikeTheSalad: Good point. I guess it might be needed in case they wanted to create their own implementation of `OpenTelemetryRum` without sessions? It sounds like an edge case, though, so maybe it's still fine to remove it.

agent-api/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
fractalwrench: After thinking about this some more on the other PR, I would lean towards just not exposing this as a public API for now (and perhaps forever). It could be an implementation detail of the SDK: https://github.com/open-telemetry/opentelemetry-android/pull/1540#pullrequestreview-3710139559

PR #1540, opened 2026-01-26
Instrumentation api change - option 2

LikeTheSalad:
DO NOT MERGE

This is a proposal for this issue: https://github.com/open-telemetry/opentelemetry-android/issues/1541 - It's meant to highlight the key changes, without addressing potential compilation errors, if any. If chosen as the approach we'd like to go with, then I'll add the rest of the changes needed for the build to pass.

### DIFF
--- a/agent-api/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
+++ b/agent-api/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
@@ -18,6 +18,10 @@ interface OpenTelemetryRum {
      */
     val openTelemetry: OpenTelemetry
 
+    val clock: Clock
+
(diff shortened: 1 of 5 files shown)


fractalwrench (commented):
This comment on #1539 applies to this PR too: https://github.com/open-telemetry/opentelemetry-android/pull/1539/changes#r2730782298

Basically: does it make sense to expose this functionality for only the log record signals, or consider exposing a way to orderly shutdown the SDK as a whole?

I'd also suggest that we don't necessarily need to expose this as part of the public API, with a few internal changes. `OpenTelemetryRum` could set its own `UncaughtExceptionHandler` separate from the crash instrumentation and shutdown signal collection in an orderly manner, without needing to provide a public API.

PR #1543, opened 2026-01-27
Session timeout lifecycle registration on main looper

DamienL-F:
`OpenTelemetryRumInitializer.initialize` can be a bit long.
In our team, with our device, it was in average around 500ms.

We tried not to do it on the main thread but we faced an issue with the session timeout lifecycle registration that must be done on the main.

So here is a fix to ensure it's done on the main looper.

It worked well for our team. Now we're able to call `OpenTelemetryRumInitializer.initialize` not on the main thread.

### DIFF
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
@@ -7,6 +7,8 @@ package io.opentelemetry.android.agent
 
 import android.app.Application
 import android.content.Context
+import android.os.Handler
+import android.os.Looper
 import io.opentelemetry.android.AndroidResource
 import io.opentelemetry.android.Incubating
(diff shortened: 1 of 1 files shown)


fractalwrench:
I'd be interested to see _why_ it was taking 500ms to initialize and whether there's any bottlenecks that can be optimized in addition to allowing initialization off the main thread. What devices did you use for benchmarking?

I'm generally not keen to suggest initializing SDKs off the main thread as it makes it more likely that telemetry capture gets missed early in the application lifecycle. However, the flip side of that is that opentelemetry-android does need to be quick to initialize.

DamienL-F:
Our team is using homemade device.
But the problem may be present as well on mainstream smartphone.

I think it's still a good idea to make the SDK init possible out of the main thread.
Team will choose what they want :
- The main thread to be sure no capture get missed
- An other thread but they have to manage a pile/an await (or not depending on what they want)
For example, our team is using a `awaitInit` method before sending log/metrics.

Issue #1439, opened 2025-12-02
Bump minSdk to 23


breedx-splk:
Our current supported minSdk is 21, which is essentially Android 5 Lollipop, released in 2014.
For the 2.x release, we will want to bump our minSdk to 23, which is Android 6 Marshmallow from 2015. This is old enough for most apps, and is part of the long tail, with < 1% of the install base ([ref](https://composables.com/android-distribution-chart)).

Furthermore, minSdk >= 23 is required for contemporary androidx/jetpack libraries ([ref](https://apilevels.com/)).

When we do this work, we should also:

* update docs to indicate minSdk is now 23
* update `renovate.json5` to enable updates for androidx libs.

fractalwrench:
In addition to this I also think we should update the VERSIONING.md doc to state that we'll aim to support the last 4 minor Kotlin versions (just like Jetbrains does)

breedx-splk:
Oops, this was resolved in #1650.

PR #1545, opened 2026-01-27
[DNM] Instrumentation API change suggestions

fractalwrench:
## Goal

Displays some changes that I think would be beneficial to make to the Instrumentation API before going stable, in addition to Cesar's outstanding PRs #1439 and #1540.

This will not compile right now as I wanted to demonstrate the intent of the changes before investing the time in making all the changes. I also haven't touched `InstallationContext` much because Cesar's PRs already cover that.


android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/instrumentation/AndroidInstrumentationLoaderImpl.kt
fractalwrench, line 11: This class should be made internal or moved to a non-stable module as it's an implementation detail

instrumentation/android-instrumentation/api/android-instrumentation.api
```diff
@@ -1,37 +1,19 @@
 public abstract interface class io/opentelemetry/android/instrumentation/AndroidInstrumentation {
```
fractalwrench, line 1: The general idea here is to reduce the amount of public symbols in the API surface.

instrumentation/android-instrumentation/src/main/java/io/opentelemetry/android/instrumentation/InstrumentationParams.kt
```diff
@@ -0,0 +1,15 @@
+package io.opentelemetry.android.instrumentation
+
+import android.app.Application
+import android.content.Context
+import io.opentelemetry.android.session.SessionProvider
+import io.opentelemetry.api.OpenTelemetry
+import io.opentelemetry.sdk.common.Clock
+
+interface InstrumentationParams {
```
fractalwrench, line 9: Previously called `InstallationContext`. I've also made this an interface so that the implementation can be swapped out easily.

I've not altered the properties that are accessible - that can be discussed further on Cesar's PRs.

instrumentation/android-instrumentation/api/android-instrumentation.api
```diff
@@ -1,37 +1,19 @@
...
 	public abstract fun getName ()Ljava/lang/String;
-	public abstract fun install (Lio/opentelemetry/android/instrumentation/InstallationContext;)V
-	public fun uninstall (Lio/opentelemetry/android/instrumentation/InstallationContext;)V
+	public abstract fun install (Lio/opentelemetry/android/instrumentation/InstrumentationParams;)V
+	public abstract fun uninstall (Lio/opentelemetry/android/instrumentation/InstrumentationParams;)V
 }
 
 public abstract interface class io/opentelemetry/android/instrumentation/AndroidInstrumentationLoader {
```
fractalwrench, line 7: Ideally I think we'd hide this altogether. Is it really required as part of our public API?

instrumentation/android-instrumentation/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentation.kt
```diff
@@ -26,19 +26,17 @@ interface AndroidInstrumentation {
...
      * This method can be called to uninstall the instrumentation. Implementations should remove all
      * used resources and shut down cleanly.
      *
-     * @param ctx The InstallationContext under which the instrumentation had been removed.
+     * @param ctx The InstrumentationParams under which the instrumentation had been removed.
      */
-    fun uninstall(ctx: InstallationContext) {
-        // NOP default implementation
-    }
+    fun uninstall(ctx: InstrumentationParams)
```
fractalwrench, line 39: Made this mandatory rather than optional. IMO several instrumentations could implement `uninstall` and currently don't because they've forgotten this function exists.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/instrumentation/AndroidInstrumentationLoaderImpl.kt
LikeTheSalad: I agree.

instrumentation/android-instrumentation/api/android-instrumentation.api
LikeTheSalad: > Is it really required as part of our public API?

I couldn't find a reason why it might be needed as a public api right now, so I agree is better to hide it. We can always expose it later if needed.

instrumentation/android-instrumentation/src/main/java/io/opentelemetry/android/instrumentation/InstrumentationParams.kt
```diff
@@ -0,0 +1,15 @@
...
+    val context: Context
+    val openTelemetry: OpenTelemetry
+    val sessionProvider: SessionProvider
+    val clock: Clock
+    val application: Application?
```
LikeTheSalad, line 14: There's something about adding this val to an interface that doesn't sound too ideal to me. I know it's there already for `InstallationContext`, but it's not part of the constructor, so it's more of a convenience getter to avoid casting `Context` manually. Maybe here it could be a function with a default body that does the context cast?

fractalwrench: I'd be happy with that approach

fractalwrench:
Superseded by other PRs.

PR #1553, opened 2026-01-29
Convert fragment instrumentation to kotlin

fractalwrench:
## Goal

Converts the remaining fragment instrumentation code to Kotlin. I've made the classes internal as part of this change.

## Testing

Relied on existing unit tests for all classes.


instrumentation/fragment/src/test/java/io/opentelemetry/android/instrumentation/fragment/FragmentTracerTest.kt
```diff
@@ -54,11 +55,12 @@ internal class FragmentTracerTest {
...
-            FragmentTracer
-                .builder(mockk<Fragment>())
-                .setTracer(tracer)
-                .setActiveSpan(activeSpan)
-                .build()
+            FragmentTracer(
+                mockk<Fragment>(),
+                tracer,
+                "",
+                activeSpan,
```
breedx-splk: Because the constructor has 4 args, it might be nice (for readability) to use the parameter names here at the invocation site (and elsewhere in this file).

fractalwrench: Good call. I've updated the applicable call sites to use named arguments.

PR #1554, opened 2026-01-29
Convert ActiveSpan to Kotlin

fractalwrench:
## Goal

Converts `ActiveSpan` to Kotlin

## Testing

Added a unit test. There is additional indirect test coverage in the activity/fragment instrumentation tests.

### DIFF
--- a/instrumentation/common-api/src/test/java/io/opentelemetry/android/instrumentation/common/ActiveSpanTest.kt
+++ b/instrumentation/common-api/src/test/java/io/opentelemetry/android/instrumentation/common/ActiveSpanTest.kt
@@ -5,9 +5,13 @@
 
 package io.opentelemetry.android.instrumentation.common
 
+import io.mockk.every
 import io.mockk.mockk
 import io.mockk.verify
 import io.opentelemetry.api.trace.Span
+import io.opentelemetry.api.trace.SpanContext
+import io.opentelemetry.api.trace.TraceFlags
+import io.opentelemetry.api.trace.TraceState
 import io.opentelemetry.context.Context
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertFalse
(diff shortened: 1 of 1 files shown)


instrumentation/common-api/src/main/java/io/opentelemetry/android/instrumentation/common/ActiveSpan.kt
```diff
@@ -0,0 +1,53 @@
...
+    fun spanInProgress(): Boolean = span != null
+
+    // it's fine to not close the scope here, will be closed in endActiveSpan()
+    fun startSpan(spanCreator: () -> Span) {
+        // don't start one if there's already one in progress
+        if (span != null) {
+            return
+        }
+        span = spanCreator()
+        span?.makeCurrent()
```
breedx-splk: Ugh! I think the original code overspecified `this.span` which might have borked the auto-conversion. In any case, this is missing a scope assignment now, which could be quite bad....

```suggestion
        span = spanCreator()
        scope = span?.makeCurrent()
```

instrumentation/common-api/src/test/java/io/opentelemetry/android/instrumentation/common/ActiveSpanTest.kt
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    @Test
+    fun testActiveSpanCreate() {
+        val activeSpan = ActiveSpan { "prev_screen" }
+        assertFalse(activeSpan.spanInProgress())
+
+        val span = mockk<Span>(relaxed = true)
+        activeSpan.startSpan { span }
+        assertTrue(activeSpan.spanInProgress())
```
breedx-splk, line 56: Thanks for adding these tests....It would be great if we could also make some assertions about the scope too...even with an internal method or something.

fractalwrench: Added a couple of checks using mockk's `verify` function.

breedx-splk: It's true, but none of them really address the `scope` as far as I can tell...

fractalwrench: Had a think about this and altered the assertions to check the value of `Context.current()`, which should give more confidence that the scope is being closed correctly.

PR #1556, opened 2026-01-30
Update errorprone

LikeTheSalad:
Supersedes https://github.com/open-telemetry/opentelemetry-android/pull/1548

### DIFF
--- a/buildSrc/src/main/kotlin/otel.errorprone-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.errorprone-conventions.gradle.kts
@@ -36,11 +36,11 @@ tasks {
         options.errorprone {
             if (name.lowercase(Locale.getDefault()).contains("test")) {
                 // just disable all error prone checks for test
-                enabled = false
+                enabled.set(false)
                 compilingTestOnlyCode.set(true)
             } else {
                 if (isAndroidProject) {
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/kotlin/otel.errorprone-conventions.gradle.kts
```diff
@@ -24,7 +24,6 @@ val libs = extensions.getByType<VersionCatalogsExtension>().named("libs")
 dependencies {
     errorprone(libs.findLibrary("nullaway").get())
     errorprone(libs.findLibrary("errorprone-core").get())
-    errorproneJavac(libs.findLibrary("errorprone-javac").get())
```
fractalwrench, line 27: Was this library unused?

fractalwrench (approved):
LGTM assuming that the `errorprone-javac` library isn't used

buildSrc/src/main/kotlin/otel.errorprone-conventions.gradle.kts
LikeTheSalad: It was unnecessary since version 3.0.0, according to their release notes, and the configuration `errorproneJavac` was removed in this version, so it wouldn't compile if I left it there.

PR #1557, opened 2026-01-30
Convert Websocket instrumentation to Kotlin

fractalwrench:
## Goal

Converts the Websocket instrumentation to Kotlin. As part of this I consolidated a couple of classes into `WebsocketListenerWrapper` but otherwise left the `WebsocketListenerWrapper` symbol alone as it appears to be referenced by ByteBuddy code.

## Testing

Added unit tests.


LikeTheSalad (approved):
Thank you! 🙏

PR #1558, opened 2026-01-30
Convert HttpUrlConnection instrumentation to Kotlin

fractalwrench:
## Goal

Converts some classes in the `HttpUrlConnection` instrumentation to Kotlin.

## Testing

Relied on existing test cases.


instrumentation/httpurlconnection/library/src/main/java/io/opentelemetry/instrumentation/library/httpurlconnection/internal/HttpUrlHttpAttributesGetter.kt
```diff
@@ -0,0 +1,64 @@
...
+        val httpURLConnection = connection as HttpURLConnection
+        return httpURLConnection.requestMethod
+    }
+
+    override fun getUrlFull(connection: URLConnection): String = connection.getURL().toExternalForm()
+
+    override fun getHttpRequestHeader(
+        connection: URLConnection,
+        name: String,
+    ): MutableList<String?> =
```
breedx-splk, line 23: Bummer we have this java type baggage. 😿

Issue #1541, opened 2026-01-26
Enhance instrumentation API


LikeTheSalad:
## Introduction

After going stable with our `andrid-agent` module, we've discussed plans to stabilize other modules as well, in a recent Android SIG meeting.

One of the modules that could be ideal as a next candidate to become stable is [android-instrumentation](https://github.com/open-telemetry/opentelemetry-android/tree/main/instrumentation/android-instrumentation), where the base abstraction for all of OTel Android instrumentations is defined ([AndroidInstrumentation](https://github.com/open-telemetry/opentelemetry-android/blob/e725d67e26a136a6ed80501bfc7871bb29cec4e5/instrumentation/android-instrumentation/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentation.kt#L23)), which would open the possibilities of marking specific instrumentations as stable right after.

## The problem

However, before we stabilize the instrumentation API, I think we should make sure that it's flexible enough so that it can be extended in the future without having to introduce breaking changes or having to resort to workarounds, as we currently do with our CrashReporterInstrumentation, where we cast an `OpenTelemetry` instance to `OpenTelemetrySdk` for it to work properly:

https://github.com/open-telemetry/opentelemetry-android/blob/e725d67e26a136a6ed80501bfc7871bb29cec4e5/instrumentation/crash/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashReporterInstrumentation.kt#L29-L30

## Proposals

I've put together a couple of proposals to make the instrumentation API more flexible, hoping it can be extended as needed in the future:

* [Option 1](https://github.com/open-telemetry/opentelemetry-android/pull/1539) - Ensures that our `OpenTelemetryRum` abstraction can support common use cases, such as flushing log records (needed for our crash instrumentation).
* [Option 2](https://github.com/open-telemetry/opentelemetry-android/pull/1540) - Creates separate/optional APIs that implementations of `OpenTelemetryRum` would have to opt-in to enable some use cases, such as flushing log records (needed for our crash instrumentation).

[...]

PR #1568, opened 2026-02-05
Move `AndroidInstrumentationLoader` out of `android-instrumentation` module

fractalwrench:
## Goal

Moves `AndroidInstrumentationLoader` out of the `android-instrumentation` module as discussed in #1545 and #1541. This avoids exposing the implementation detail as part of the public API which will be required before marking that module as stable.

I've additionally made `uninstall()` on the `AndroidInstrumentation` interface mandatory for implementations, as IMO this should always be specified (even if it's a no-op) and several instrumentations seem to have forgotten about this functionality.

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentationLoader.kt
+++ b/core/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentationLoader.kt
@@ -15,7 +15,7 @@ interface AndroidInstrumentationLoader {
      * @param type The type of the instrumentation to retrieve.
      * @return The instrumentation instance if available, null otherwise.
      */
-    fun <T : AndroidInstrumentation> getByType(type: Class<out T>): T
+    fun <T : AndroidInstrumentation> getByType(type: Class<out T>): T?
 
     /**
      * Provides all registered instrumentations.
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/instrumentation/ActivityLifecycleConfiguration.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/instrumentation/ActivityLifecycleConfiguration.kt
@@ -20,23 +20,23 @@ class ActivityLifecycleConfiguration internal constructor(
     private val config: OtelRumConfig,
 ) : ScreenLifecycleConfigurable,
     CanBeEnabledAndDisabled {
-    private val activityLifecycleInstrumentation: ActivityLifecycleInstrumentation by lazy {
+    private val activityLifecycleInstrumentation: ActivityLifecycleInstrumentation? by lazy {
         AndroidInstrumentationLoader.get().getByType(ActivityLifecycleInstrumentation::class.java)
     }
 
     override fun tracerCustomizer(value: (Tracer) -> Tracer) {
-        activityLifecycleInstrumentation.setTracerCustomizer(value)
+        activityLifecycleInstrumentation?.setTracerCustomizer(value)
     }
 
     override fun screenNameExtractor(value: ScreenNameExtractor) {
-        activityLifecycleInstrumentation.setScreenNameExtractor(value)
+        activityLifecycleInstrumentation?.setScreenNameExtractor(value)
     }
 
     override fun enabled(enabled: Boolean) {
         if (enabled) {
-            config.allowInstrumentation(activityLifecycleInstrumentation.name)
+            activityLifecycleInstrumentation?.name?.let { config.allowInstrumentation(it) }
         } else {
-            config.suppressInstrumentation(activityLifecycleInstrumentation.name)
+            activityLifecycleInstrumentation?.name?.let { config.suppressInstrumentation(it) }
         }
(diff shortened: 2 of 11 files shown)


instrumentation/android-instrumentation/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentation.kt
```diff
@@ -36,9 +36,7 @@ interface AndroidInstrumentation {
      *
      * @param ctx The InstallationContext under which the instrumentation had been removed.
      */
-    fun uninstall(ctx: InstallationContext) {
-        // NOP default implementation
-    }
+    fun uninstall(ctx: InstallationContext)
```
breedx-splk: This was intentional. There was a default implementation created as part of #1109 which was inspired by #1071. There was an umbrella tracking issue here in #1128 that was intended to track this work over time, but it was (accidentally?) auto-closed when #1369 was merged. 

Anyway, I've reopened #1128.

I guess we could just continue with this, since we have the tracking issue for each instrumentation...but I kinda hate to see so many empy/no-op instrumentations.

Is the default interface implementation problematic for some other reason?

fractalwrench: @breedx-splk I've reverted the change for the purposes of getting the rest of this PR merged, but think we should still do this.

IMO it's slightly problematic in that a default implementation makes it less obvious that instrumentation should implement an uninstall function that cleans up. For example, the compose/session instrumentation both set listeners that aren't unregistered. Perhaps that is an academic point though given that `uninstall()` isn't generally called in a production app?

LikeTheSalad: > Is the default interface implementation problematic for some other reason?

My 2 cents: I see the ergonomic benefit of having a default empty body for `uninstall()`, since not all the instrumentations need it. But the flip side is that it's easy to overlook in cases where it is needed, leaving us with 2 tradeoffs:

* Annoying creators of instrumentations with having to have an empty `uninstall()` method in cases it's not needed.
* Risking memory leaks or the like for the cases where it's needed and creators overlooked it.

Between those, I'd go with the first one of making an `uninstall()` impl mandatory (and annoying for the cases it's not needed), because it seems to be the least problematic by comparison.

breedx-splk: Thanks for the discussion. I think @LikeTheSalad does a good job of summarizing the tradeoffs, but we still differ on preference. It's fine tho, I'm not losing sleep over it.

> a default implementation makes it less obvious that instrumentation should implement an uninstall function that cleans up

I see where you're coming from @fractalwrench but I take this to mean "instrumentations MAY choose to implement `uninstall()` and it will be called if individual instrumentations need to do cleanup. Those that don't need it can omit it."

Not a huge deal either way, but thanks for talking it out.

core/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentationLoader.kt
```diff
@@ -17,7 +15,7 @@ interface AndroidInstrumentationLoader {
      * @param type The type of the instrumentation to retrieve.
      * @return The instrumentation instance if available, null otherwise.
      */
-    fun <T : AndroidInstrumentation> getByType(type: Class<out T>): T?
+    fun <T : AndroidInstrumentation> getByType(type: Class<out T>): T
```
breedx-splk: The removal of the null possibility makes the javadoc incorrect now.

As we work toward stabilizing the instrumentation api, we should give this some extra thought. I suspect there are cases where the classpath doesn't contain the desired type, in which case we now `error()`...which is fine, but minimally we should document it for the caller to know about. Overall, to me, as much as I hate nulls, the contract feels slightly weaker. Maybe we should declare an explicit exception type? I dunno. 

I don't wanna hold the PR up for this, but we should continue discussing I think.

fractalwrench: This doesn't change behavior as the code was calling `error()` before anyway if an instrumentation was null. IMO the signature just makes that existing approach explicit: https://github.com/open-telemetry/opentelemetry-android/pull/1568/changes#diff-d40eb4026322ddf37c2a20c0ce7c84206607fd52d13be3483abbcfc168859a7fL42

I think my preference would be for this function to return `T?` and then print a warning if instrumentation failed to load, as that feels a better approach than erroring. I'll leave some time for other folks to add their opinions and then amend this PR if there's consensus

breedx-splk: Yeah I like that. The error seems too extreme.

fractalwrench:
I've tweaked the function signature for `getByType` so that it has a nullable return type as suggested in the review comments.

PR #1570, opened 2026-02-06
Convert activity instrumentation code to kotlin

fractalwrench:
## Goal

Converts `ActivityTracer` and `ActivityTracerCache` to Kotlin

## Testing

Relied on existing unit tests.


instrumentation/activity/src/main/java/io/opentelemetry/android/instrumentation/activity/ActivityTracerCache.kt
```diff
@@ -0,0 +1,64 @@
...
+            startupTimer: AppStartupTimer,
+            screenNameExtractor: ScreenNameExtractor,
+        ) : this({ activity: Activity ->
+            ActivityTracer(
+                activity,
+                ActiveSpan(visibleScreenTracker::previouslyVisibleScreen),
+                tracer,
+                startupTimer,
+                screenNameExtractor.extract(activity),
+                null,
```
breedx-splk: I think this defaults now, so can be omitted? Or maybe it would be better to use named params for readability, since the constructor is pretty large.

fractalwrench: Removed this param and added named parameters for the rest.

PR #1571, opened 2026-02-06
Convert network services to Kotlin

fractalwrench:
## Goal

Converts some more network code in the `services` module to Kotlin.

## Testing

Relied on existing test cases.


services/src/main/java/io/opentelemetry/android/internal/services/network/CarrierFinder.kt
```diff
@@ -0,0 +1,114 @@
...
+        if (isValidString(simOperator) && simOperator.length >= 5) {
+            mcc = simOperator.take(3)
+            mnc = simOperator.substring(3)
+        }
+        val isoCountryCode = telephonyManager.simCountryIso
+        if (isValidString(isoCountryCode)) {
+            iso = isoCountryCode
+        }
+        listOf(mcc, mnc, iso)
+    }
```
breedx-splk: Do we have some guarantee that the sim information cannot change!? I think that this is a functional change that I'm not entirely convinced is a good idea.

In the java implementation, the method always went back to the `telephonyManager` in order to fetch the sim info. I'd be worried about this reporting old values now...

breedx-splk (changes requested):
Blocking to discuss behavior change.

services/src/main/java/io/opentelemetry/android/internal/services/network/CarrierFinder.kt
fractalwrench: I'd assumed that the SIM would never change but [the documentation](https://developer.android.com/reference/android/telephony/TelephonyManager#getSimOperator()) does indicate that this isn't true (particularly with e-sims I assume). I've converted the lazy property into a function so `CarrierFinder` will check each time.

Issue #1564, opened 2026-02-04
app.jank log sent with bad timestamp


fleficher:
The `app.jank` event log is being sent with a bad timestamp, resulting in an error when the request is being made:

```json
{
        "1": {
          "1": "app.jank"
        },
        "2": {
          "6": [
            {
              "1": "app.jank.frame_count",
              "2": {
                "3": 1
              }
            },
            {
              "1": "app.jank.period",
              "2": "!\u0000\u0000\u0000\u0000\u0000\u0000�?"
            },
            {
              "1": "app.jank.threshold",
              "2": "!����Mb�?"
            },
            {
              "1": "network.connection.type",
              "2": {
                "1": "wifi"
              }
            },
            {
              "1": "screen.name",
              "2": {
                "1": "HomeActivity"
              }
            },
            {
              "1": "session.id",
              "2": ""
            }
          ],
          "11": {
            "Type": "Buffer (8 bytes)",
            "As string": "\flæ(œ\u0001\u0000\u0000",
            "As hex": "0C 6C E6 28 9C 01 00 00"
          },
          "12": "app.jank"
        }
      }
```
Result in an HTTP 400 error:
```json
{
  "2": "entry for stream '{deployment_environment=\"dev\"}' has timestamp too old: 1970-01-01T00:29:30Z, oldest acceptable timestamp is: 2026-01-28T13:46:58Z"
}
```

fractalwrench:
Thanks for the report @fleficher, does this happen consistently for all app.jank logs or is this something that you see X% of the time?

breedx-splk:
@fleficher thanks for this report, but I have a couple additional questions:

* what is this json format and how are you creating it? I don't see a timestamp anywhere in the data you have provided. 
* Are you using disk buffering? If so, can you try without?
* what is providing that error message?

fleficher:
Hey 👋  @breedx-splk @fractalwrench 

> what is this json format and how are you creating it? I don't see a timestamp anywhere in the data you have provided.

I don't create it myself, it's the sdk doing it. I'm just capturing the HTTP requests json body sent by the SDK.

> Are you using disk buffering? If so, can you try without?

Yes, I tried to disabled it, but same issue.

> what is providing that error message?

Calls to `v1/logs` API, we're using Grafana to report the logs.

> Does this happen consistently for all app.jank logs or is this something that you see X% of the time?

It's happenning 100% of the time. I've also seen other events having the same issue:
[...]

PR #1574, opened 2026-02-06
Make clock baseline nanos

breedx-splk:
It looks like the new clock implementation from #1487 expects the baseline to be an offset number of nanos. This is obtained when the clock is created, but does subtraction between millis and nanos, without conversion. I think this creates an incorrect baseline, the result of which ends up being timestamps very close to the epoch.

Related somewhat to #1564.


LikeTheSalad (approved):
LGTM, cheers!

PR #1580, opened 2026-02-10
DNM - instrumentation API change sample

bidetofevil:


bidetofevil:
Not meant to be merged. This is only to illustrate my point in the discussions in #1541

breedx-splk:
@bidetofevil We think this has already migrated in a direction so I'm closing this. Please feel free to reopen if you want to continue down this path.

Issue #1577, opened 2026-02-09
We need a way to discover span and logs customiser coming from different instrumentation


atulgpt:
Having this help to create a instrumentation which can add different attributes to spans(, or/and) logs.

One example use case is to create [location instrumentation](https://github.com/open-telemetry/opentelemetry-android/issues/1491), which is a opt-in instrumentation but when added wants to add different [location related attributes](https://opentelemetry.io/docs/specs/semconv/registry/attributes/geo/) to each spans and logs

We can explore something like Service Provider similar to we did for discovering different instrumentation

breedx-splk:
The purpose of instrumentation is not to augment/change telemetry data created by other instrumentation. Instead, instrumentation should generate new telemetry data (spans, events, logs, metrics). 

Altering existing telemetry can be accomplished with standard OpenTelemetry sdk components -- like a `SpanProcessor` or `LogRecordProcessor`. We have examples of these in the codebase already. If you create these custom components,  you will need to use the `OpenTelemetryRumBuilder` to wire these into the creation of the OpenTelemetry sdk, as the initializer provided by the agent does not currently offer this. For example, you can use `addTracerProviderCustomizer()` on `OpenTelemetryRumBuilder` to call `addSpanProcessor()` on the given `SdkTracerProviderBuilder`.

Please clarify if you are looking for something else that isn't already offered. I think that instrumentation is not a great approach for this.

atulgpt:
> The purpose of instrumentation is not to augment/change telemetry data created by other instrumentation. Instead, instrumentation should generate new telemetry data (spans, events, logs, metrics).

If we go by this then my suggestion doesn't make sense, as I wanted to create a instrumentation which can add attributes to spans and logs created by the SDK.
Do we know why we have restricted the scope of instrumentation to only `The purpose of instrumentation is not to augment/change telemetry data created by other instrumentation. Instead, instrumentation should generate new telemetry data (spans, events, logs, metrics).`?

I am aware of `SpanProcessor`, `LogRecordProcessor` and methods like `addTracerProviderCustomizer` to add these processors. But to use these we have to add the dependency in the SDK itself

My use-case it to create a Geo-location instrumentation, which can add Geo attributes for each spans, and logs created by the SDK but should be optional so SDK by default don't even add the location related dependency unless client wants this functionality

fractalwrench:
[...]

PR #1586, opened 2026-02-12
Add ability to customize some aspects of the otel sdk creation

breedx-splk:
Related to #1577

This adds some additional customization capabilities to the initialization DSL. We've had users ask for similar things quite a few times now...and rather than always having to resort to using the `OpenTelemetryRumBuilder` I figured we could try and find some middle ground -- something that allows users to get the benefits of the initializer while still being able to do niche/bespoke customizations when needed.

This does change the api by adding functionality, but is backwards compatible and thus non-breaking.

### DIFF
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OpenTelemetryConfiguration.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OpenTelemetryConfiguration.kt
@@ -25,6 +25,7 @@ class OpenTelemetryConfiguration internal constructor(
      * Configures the [Clock] used for capturing telemetry.
      */
     var clock: Clock = OtelAndroidClock(),
+    internal val sdkCustomizations: OtelSdkCustomizationsSpec = OtelSdkCustomizationsSpec(),
 ) {
     internal val exportConfig = HttpExportConfiguration()
     internal val sessionConfig = SessionConfiguration()
@@ -57,6 +58,14 @@ class OpenTelemetryConfiguration internal constructor(
      */
     fun globalAttributes(action: () -> Attributes) {
         rumConfig.setGlobalAttributes(action())
+    }
+
+    /**
+     * You can use this to customize additional nonstandard aspects of the OpenTelemetry SDK.
+     * Most users should not need to provide any customizations here.
+     */
+    fun otelSdkCustomizations(action: OtelSdkCustomizationsSpec.() -> Unit) {
+        sdkCustomizations.action()
     }
 
     /**
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OtelSdkCustomizationsSpec.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OtelSdkCustomizationsSpec.kt
@@ -9,6 +9,10 @@ import io.opentelemetry.sdk.logs.SdkLoggerProviderBuilder
 import io.opentelemetry.sdk.metrics.SdkMeterProviderBuilder
 import io.opentelemetry.sdk.trace.SdkTracerProviderBuilder
 
+typealias TracerProviderCustomizer = Function1<SdkTracerProviderBuilder, SdkTracerProviderBuilder>
+typealias LoggerProviderCustomizer = Function1<SdkLoggerProviderBuilder, SdkLoggerProviderBuilder>
+typealias MeterProviderCustomizer = Function1<SdkMeterProviderBuilder, SdkMeterProviderBuilder>
+
 @OpenTelemetryDslMarker
 class OtelSdkCustomizationsSpec internal constructor() {
     internal val tracerProviderCustomizers: MutableList<TracerProviderCustomizer> = mutableListOf()
@@ -50,16 +54,4 @@ class OtelSdkCustomizationsSpec internal constructor() {
     fun customizeMeterProvider(customizer: MeterProviderCustomizer) {
         meterProviderCustomizers.add(customizer)
     }
-}
-
-fun interface TracerProviderCustomizer {
-    fun customize(builder: SdkTracerProviderBuilder): SdkTracerProviderBuilder
-}
-
-fun interface LoggerProviderCustomizer {
-    fun customize(builder: SdkLoggerProviderBuilder): SdkLoggerProviderBuilder
-}
-
-fun interface MeterProviderCustomizer {
-    fun customize(builder: SdkMeterProviderBuilder): SdkMeterProviderBuilder
 }
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OtelSdkCustomizationsSpec.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OtelSdkCustomizationsSpec.kt
@@ -40,7 +40,7 @@ class OtelSdkCustomizationsSpec internal constructor() {
     }
 
     /**
-     * Modify the creation of the OpenTelemetry LoggerProvider by providing
+     * Modify the creation of the OpenTelemetry MeterProvider by providing
      * your own customizer here. The customizer is a function
      * that receives an instance of [SdkMeterProviderBuilder] and returns an
      * instance of [SdkMeterProviderBuilder], which should almost always be
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OtelSdkCustomizationsSpec.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OtelSdkCustomizationsSpec.kt
@@ -1,0 +1,65 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.android.agent.dsl
+
+import io.opentelemetry.sdk.logs.SdkLoggerProviderBuilder
+import io.opentelemetry.sdk.metrics.SdkMeterProviderBuilder
+import io.opentelemetry.sdk.trace.SdkTracerProviderBuilder
+
+@OpenTelemetryDslMarker
+class OtelSdkCustomizationsSpec internal constructor() {
+    internal val tracerProviderCustomizers: MutableList<TracerProviderCustomizer> = mutableListOf()
+    internal val loggerProviderCustomizers: MutableList<LoggerProviderCustomizer> = mutableListOf()
+    internal val meterProviderCustomizers: MutableList<MeterProviderCustomizer> = mutableListOf()
+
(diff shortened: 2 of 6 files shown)


android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OtelSdkCustomizationsSpec.kt
```diff
@@ -0,0 +1,65 @@
...
+    fun customize(builder: SdkTracerProviderBuilder): SdkTracerProviderBuilder
+}
+
+fun interface LoggerProviderCustomizer {
+    fun customize(builder: SdkLoggerProviderBuilder): SdkLoggerProviderBuilder
+}
+
+fun interface MeterProviderCustomizer {
+    fun customize(builder: SdkMeterProviderBuilder): SdkMeterProviderBuilder
+}
```
breedx-splk: So these types aren't completely necessary -- but I thought that it read a bit cleaner than using the underlying functional methods from the SDK. If we're worried about the increase in surface area, then we could probably just remove them and refer to the functions...but it's a bit verbose.

What do you think?

LikeTheSalad: What's the use case you have in mind for these customizers in the initializer? They seem to be providing a huge hammer to me. If I'm not mistaken, they would allow users to override our processors and our exporters, so they might get into trouble by doing so and then seeing that sessions, disk buffering, and export parameters don't work as configured via the DSL options.

fractalwrench: `typealias` might be an option here if you want a degree of separation but without creating an extra interface

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OtelSdkCustomizationsSpec.kt
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+package io.opentelemetry.android.agent.dsl
+
+import io.opentelemetry.sdk.logs.SdkLoggerProviderBuilder
+import io.opentelemetry.sdk.metrics.SdkMeterProviderBuilder
+import io.opentelemetry.sdk.trace.SdkTracerProviderBuilder
+
+@OpenTelemetryDslMarker
+class OtelSdkCustomizationsSpec internal constructor() {
```
fractalwrench, line 17: Is it worth considering marking these new APIs as 'experimental' or 'incubating'? I suppose this is a more general question on what stability guarantees we want to provide for new APIs in modules that are otherwise stable?

My personal opinion is that marking a new API as experimental for a couple of versions is fairly normal on Android. Would be curious to hear others thoughts.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OtelSdkCustomizationsSpec.kt
```diff
@@ -0,0 +1,65 @@
...
+     * the same instance. If a new instance is returned, the operation can be
+     * destructive.
+     */
+    fun customizeMeterProvider(customizer: MeterProviderCustomizer) {
+        meterProviderCustomizers.add(customizer)
+    }
+}
+
+fun interface TracerProviderCustomizer {
+    fun customize(builder: SdkTracerProviderBuilder): SdkTracerProviderBuilder
```
fractalwrench: One potential downside is that this does tie our API further to symbols declared in opentelemetry-java's SDK module. If we ever do decide to use opentelemetry-kotlin then this takes us further from that goal - but perhaps that's ok for now?

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OpenTelemetryConfiguration.kt
```diff
@@ -59,6 +60,14 @@ class OpenTelemetryConfiguration internal constructor(
         rumConfig.setGlobalAttributes(action())
     }
 
+    /**
+     * You can use this to customize additional nonstandard aspects of the OpenTelemetry SDK.
+     * Most users should not need to provide any customizations here.
+     */
+    fun otelSdkCustomizations(action: OtelSdkCustomizationsSpec.() -> Unit) {
```
fractalwrench, line 67: I think we should add some sort of guarantee about configuration order here. For example, this could be the final action after all the other configuration (such as OTLP exporters) has been setup on the builder object.

fractalwrench:
I agree with the goal of allowing users to configure the underlying OpenTelemetry instance more. I think this changeset opens up some interesting questions about the tradeoffs of how best to achieve that.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OtelSdkCustomizationsSpec.kt
breedx-splk: > `typealias` might be an option here if you want a degree of separation but without creating an extra interface

Yes! Thank you. I think the `typealias` is considerably cleaner here.

breedx-splk: > What's the use case you have in mind for these customizers in the initializer? 

Good question. The ones that I've heard, and what was discussed in #1577, is the ability to add custom `SpanProcessor`s into the sdk. Without some help from the initializer, the developer would need to use the `OpenTelemetryRumBuilder` directly, missing out on the benefits/convenience of the Initializer.  

Another use case that has been brought up more than once is the ability to expire/swap HTTP exporter headers/tokens...so basically customizing the exporters. That's not covered here, but I think we will need to address it before too long.

> They seem to be providing a huge hammer to me. If I'm not mistaken, they would allow users to override our processors and our exporters, so they might get into trouble by doing so and then seeing that sessions, disk buffering, and export parameters don't work as configured via the DSL options.

You're completely correct -- this is a big hammer! You're also right, that I think this allows users to break things that the agent tries to help with. I think this is already true of the upstream autoconfigure as well -- it's possible to ab/use the autoconfigure callbacks to undo some of the defaults (like breaking the BSP which is responsible for exporting, for example).

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OtelSdkCustomizationsSpec.kt
breedx-splk: Seems like a reasonably good idea to me. Open to ideas on how to actually do this in practice.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OtelSdkCustomizationsSpec.kt
breedx-splk: Right, this did occur to me as I was hacking on this, and I respect that you're especially sensitive to this (with KSP compat being an end goal). I'm not sure how to do things like this without some amount of dependency on java core for now....

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OpenTelemetryConfiguration.kt
breedx-splk: I agree this is important to consider. I think right now that the implementation does these last, but I'm not sure how to bake that into this contract.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OtelSdkCustomizationsSpec.kt
fractalwrench: The Incubating annotation in `android-agent` is one way of achieving this. It's probably worth chatting about this at the SIG as it's a broader discussion of how we approach adding new APIs to otherwise stable modules

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OtelSdkCustomizationsSpec.kt
fractalwrench: I think the benefits outweigh the downside here tbh.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OtelSdkCustomizationsSpec.kt
LikeTheSalad: Got it, thanks for the details.

> The ones that I've heard, and what was discussed in https://github.com/open-telemetry/opentelemetry-android/issues/1577, is the ability to add custom SpanProcessors into the sdk.

This is fair enough. I believe we should be more specific with the api surface additions to make its evolution more iterative, though. I see the benefits of adding a big hammer like this one in terms of making it future-proof, but I'm afraid it can cause strange and difficult to address isssues down the line. Based on that, why don't we provide an API to add processors specifically instead?

> Another use case that has been brought up more than once is the ability to expire/swap HTTP exporter headers/tokens...so basically customizing the exporters.

This should probably be a separate API that would allow to customize the exporters that we create in the agent. Not only to avoid providing a big hammer, but also because of how the Java SDK works. By adding a [Signal]ProviderCustomizer as proposed in this PR, I don't think that would help users to customize our exporters, as they can't access those via the Java SDK builders. So a specific exporter customizer seems to be the only feasible approach.

> You're completely correct -- this is a big hammer! You're also right, that I think this allows users to break things that the agent tries to help with. I think this is already true of the upstream autoconfigure as well -- it's possible to ab/use the autoconfigure callbacks to undo some of the defaults (like breaking the BSP which is responsible for exporting, for example).

Don't get me wrong, I do not to like the idea of having to be some sort of police for our users and to tell them what they should and shouldn't do with their code. The thing is that experienced users can already do whatever they want to by using the rum builder, while being conscious that they might break things, which is why I believe that whoever decides to use the agent is expecting not only simplicity but also reliability, and it will become difficult to guarantee the reliability if we make it too flexible. Because of that, and to being able to provide better support for agent users when they run into trouble, I think it's best to increase the API surface to support concrete use cases in mind, such as allowing for custom processors in this case, and thus prevent other accidental changes.

breedx-splk:
Thanks for a good discussion, everyone. After the SIG meeting today, I think we concluded that the specific use case of adding geo attributes would be better handled by an existing API which we overlooked previously -- namely the `globalAttributes(supplier)` which leverages the internal `GlobalAttributesSpanAppender`.

With that, I'm going to convert o draft for now, in case we want to continue discussing the tradeoffs...but I definitely agree that the approach provided in this PR is pretty broad and opens up some specific problems (big footgun, sdk in api footprint).

PR #1632, opened 2026-03-04
Enhancing instrumentation manual install

LikeTheSalad:
Related to: https://github.com/open-telemetry/opentelemetry-android/issues/1541

These changes are intended to address "_challenge 1_", as explained in [this comment](https://github.com/open-telemetry/opentelemetry-android/issues/1541#issuecomment-3812235676). By making the `OpenTelemetryRum` implementation take care of the creation of an `InstallationContext` object, which contains the proper contextual dependencies that were used to create the `OpenTelemetry` instance.

### DIFF
--- a/agent-api/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
+++ b/agent-api/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
@@ -5,6 +5,7 @@
 
 package io.opentelemetry.android
 
+import io.opentelemetry.android.instrumentation.AndroidInstrumentation
 import io.opentelemetry.api.OpenTelemetry
 import io.opentelemetry.api.common.Attributes
 
@@ -39,6 +40,15 @@ interface OpenTelemetryRum {
         body: String = "",
         attributes: Attributes = Attributes.empty(),
     )
+
+    /**
+     * Installs an [AndroidInstrumentation] using the internal context of this [OpenTelemetryRum]
+     * instance. This creates the appropriate [io.opentelemetry.android.instrumentation.InstallationContext]
+     * with the correct context objects that were used to build this instance.
+     *
+     * @param instrumentation The instrumentation to install.
+     */
+    fun installInstrumentation(instrumentation: AndroidInstrumentation)
 
     /**
      * Initiates orderly shutdown of this OpenTelemetryRum instance. After this method completes,
--- a/agent-api/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
+++ b/agent-api/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
@@ -48,7 +48,7 @@ interface OpenTelemetryRum {
      *
      * @param instrumentation The instrumentation to install.
      */
-    fun installInstrumentation(instrumentation: AndroidInstrumentation)
+    fun install(instrumentation: AndroidInstrumentation)
 
     /**
      * Initiates orderly shutdown of this OpenTelemetryRum instance. After this method completes,
--- a/core/src/main/java/io/opentelemetry/android/OpenTelemetryRumImpl.kt
+++ b/core/src/main/java/io/opentelemetry/android/OpenTelemetryRumImpl.kt
@@ -5,21 +5,29 @@
 
 package io.opentelemetry.android
 
+import android.content.Context
+import io.opentelemetry.android.instrumentation.AndroidInstrumentation
+import io.opentelemetry.android.instrumentation.InstallationContext
 import io.opentelemetry.android.session.SessionProvider
 import io.opentelemetry.api.OpenTelemetry
 import io.opentelemetry.api.common.Attributes
 import io.opentelemetry.api.logs.Logger
 import io.opentelemetry.sdk.OpenTelemetrySdk
+import io.opentelemetry.sdk.common.Clock
 
 internal class OpenTelemetryRumImpl(
     private val openTelemetrySdk: OpenTelemetrySdk,
     private val sessionProvider: SessionProvider,
+    private val context: Context,
+    private val clock: Clock,
     private val onShutdown: Runnable,
 ) : OpenTelemetryRum {
     private val logger: Logger =
         openTelemetrySdk.logsBridge
             .loggerBuilder("io.opentelemetry.rum.events")
             .build()
+
+    private val manuallyInstalledInstrumentations = mutableListOf<AndroidInstrumentation>()
 
     override val openTelemetry: OpenTelemetry = openTelemetrySdk
 
@@ -38,7 +46,17 @@ internal class OpenTelemetryRumImpl(
             .emit()
     }
 
+    override fun installInstrumentation(instrumentation: AndroidInstrumentation) {
+        val ctx = InstallationContext(context, openTelemetrySdk, sessionProvider, clock)
+        instrumentation.install(ctx)
+        manuallyInstalledInstrumentations.add(instrumentation)
+    }
+
     override fun shutdown() {
+        val ctx = InstallationContext(context, openTelemetrySdk, sessionProvider, clock)
+        for (instrumentation in manuallyInstalledInstrumentations) {
+            instrumentation.uninstall(ctx)
+        }
         onShutdown.run()
     }
 }
--- a/core/src/main/java/io/opentelemetry/android/OpenTelemetryRumImpl.kt
+++ b/core/src/main/java/io/opentelemetry/android/OpenTelemetryRumImpl.kt
@@ -46,7 +46,7 @@ internal class OpenTelemetryRumImpl(
             .emit()
     }
 
-    override fun installInstrumentation(instrumentation: AndroidInstrumentation) {
+    override fun install(instrumentation: AndroidInstrumentation) {
         val ctx = InstallationContext(context, openTelemetrySdk, sessionProvider, clock)
         instrumentation.install(ctx)
         manuallyInstalledInstrumentations.add(instrumentation)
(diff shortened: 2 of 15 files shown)


agent-api/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
```diff
@@ -40,6 +41,15 @@ interface OpenTelemetryRum {
...
     )
 
+    /**
+     * Installs an [AndroidInstrumentation] using the internal context of this [OpenTelemetryRum]
+     * instance. This creates the appropriate [io.opentelemetry.android.instrumentation.InstallationContext]
+     * with the correct context objects that were used to build this instance.
+     *
+     * @param instrumentation The instrumentation to install.
+     */
+    fun install(instrumentation: AndroidInstrumentation)
```
fractalwrench, line 51: Did we want to consider marking this as an incubating/beta API in some way?

core/src/main/java/io/opentelemetry/android/OpenTelemetryRumImpl.kt
```diff
@@ -38,7 +46,17 @@ internal class OpenTelemetryRumImpl(
             .emit()
     }
 
+    override fun install(instrumentation: AndroidInstrumentation) {
+        val ctx = InstallationContext(context, openTelemetrySdk, sessionProvider, clock)
+        instrumentation.install(ctx)
+        manuallyInstalledInstrumentations.add(instrumentation)
```
fractalwrench, line 52: Is it worth guarding against misuse by checking whether the reference is already in the collection?

core/src/main/java/io/opentelemetry/android/OpenTelemetryRumImpl.kt
```diff
@@ -38,7 +46,17 @@ internal class OpenTelemetryRumImpl(
             .emit()
     }
 
+    override fun install(instrumentation: AndroidInstrumentation) {
+        val ctx = InstallationContext(context, openTelemetrySdk, sessionProvider, clock)
+        instrumentation.install(ctx)
```
fractalwrench, line 51: Worth a try-catch for the `install/uninstall` functions, given that it's user-submitted code that probably shouldn't terminate the entire process?

fractalwrench (commented):
Thanks for looking at this. I left a few comments inline but would be broadly happy with this sort of approach.

breedx-splk (changes requested):
I'm failing to understand how adding the `install(instrumentation)` method on our public/stable `OpenTelemetryRum` interface is making things better or what problem we're solving. A couple of problems I see with this approach:

1. I think this change makes things somewhat MORE confusing for users -- who now have to decide if their instrumentation should be added via `OpenTelemetryRumBuilder.addInstrumentation(instrumentation)` or if they should wait until after they build the `OpenTelemetryRum` instance in order to call `install(instrumentation)` on that. 
2. You've introduced mutable state to the `OpenTelemetryRumImpl` instance. Before this change, basically all the instrumentation state was configured and established at "build time". Once the rum instance was created, things didn't change...and there are numerous benefits to this. Along these same lines, I think it's confusing that a user can now just choose at any old random time to install an instrumentation. I think it's easier to reason about the code when all setup/installation is done at creation time.

I've read through the related issues and comments a couple times, but I still can't make sense of what this helps. 

There's some suggestion that the `InstallationContext` is more accurate or something with this approach, but I think it's just passing along the same instances as before.

LikeTheSalad:
Thanks for the detailed feedback, @breedx-splk 

It's been a while since I explained the issues in [this comment](https://github.com/open-telemetry/opentelemetry-android/issues/1541#issuecomment-3812235676), and I wasn't aware of your opinion on it until now, which is why I proceeded to write the changes. Still, I'm glad that we can discuss it further, because it'd certainly be a big change if this gets through, so it's better to align on the details beforehand.

## The problem

The current instrumentation API looks like this:

Screenshot 2026-03-06 at 10 16 31

It's all public, which means **nothing** stops any user from using it. For example, they can implement their own instrumentation and manually call `install()` at any time.

However, if they choose to do so, they'd need to construct an `InstallationContext` object, which requires all the fields shown in the diagram above. So the main roadblock is, **how can a user construct an `InstallationContext` object?**

Here's what we provide our users when initializing the agent:

Screenshot 2026-03-06 at 10 20 31

Which means that, out of all the dependencies from `InstallationContext`, we only help with **1** of them (the `OpenTelemetry` object). So here's what happens with the rest:

* For `Context`, they can get it themselves from the Android SDK, although there's a potential issue here, which is that we need a specific type of Context, which is the "Application Context". This is important because, if an instrumentation impl stores the context for whatever reason, and the user provides a screen-related context, then that can cause a memory leak.
* For `SessionProvider`, users would have to create their own object, which would naturally not provide the same `session.id` values as the one we configure internally when initializing `OpenTelemetryRum`. This can cause session inconsistency issues across the app.
* For `Clock`, users would either have to provide their own implementation, or use the default one, none of which is the implementation that we use when initializing the agent, which can cause time-related inconsistencies.
* An additional issue, is that we intend for `InstallationContext` to be "extendable" to adapt to future instrumentation use cases in the future, in case some extra dependencies are needed by then. But once the API is stable, we won't be able to enforce those new dependencies in the constructor, as that would cause a breaking change. So, our only option would be to adding them with default args, which would most likely suffer from similar inconsistencies as the previously mentioned dependencies.

## Potential solutions

You can find more details in [this issue](https://github.com/open-telemetry/opentelemetry-android/issues/1541). But essentially the options discussed there were 2:

- Adding getters for all of the `InstallationContext` dependencies to `OpenTelemetryRum`, and that way getting rid of the need to instantiate a third object to install instrumentations, as we'd only need to pass `OpenTelemetryRum` to install an instrumentation, where it could find everything it needs and it's extendable without introducing breaking changes. This option was considered dangerous because we would provide instrumentations the ability to shut the agent down, given that the `OpenTelemetryRum` api allows it.
- Removing the need for users to construct the `InstallationContext` themselves, which would not only make it easier for them to manually install instrumentations, but also ensure that the instrumentations are consistent with the whole context used in `OpenTelemetryRum`. This option sounded like the least problematic in the thread, hence I created this PR with it.

Based on your comments, it sounds like you're only taking into account the ability to install instrumentations at the time of initializing `OpenTelemetryRum`. This option is also valid, in the sense that it would prevent the issues mentioned above, though for it to work, we should explain in the docs that there's a "recommended way" of installing instrumentations, and explain that we wouldn't support issues that may arise from attempting to install them via other methods. Also, we should probably make the API more restricted if possible, to reduce the chances of users trying to manually install it by calling `install()` by themselves. It'd probably be a strict requirement, given that it's technically possible to still install instrumentations after the agent is initialized, but at least it would make the maintenance easier, so I'm not opposed to going with this option. But we need to be aligned, so let me know what you think.

breedx-splk:
Thanks @LikeTheSalad for the reply and discussion.

> It's all public, which means nothing stops any user from using it. For example, they can implement their own instrumentation and manually call install() at any time.

Sure, but I actually don't want users to call `install()` on instrumentation. This should be a responsibility of the agent (and/or OpenTelemetryRum startup/creation).

>However, if they choose to do so, they'd need to construct an InstallationContext object, which requires all the fields shown in the diagram above. 

Right, and if you recall, we used to not have an `InstallationContext` class at all -- it used to be something like `install(app, otel, etc)`...and we intentionally added `InstallationContext` class so that we could modify what gets passed to instrumentations without breaking the contract. It's much easier to change the internals off the `InstallationContext` implementation instead of changing the signature of the `install()` method (or that was the thinking at the time anyway).

>So the main roadblock is, how can a user construct an InstallationContext object?

So that's the thing: I don't want users to ever have to create instances of this! It's not a user's responsibility. I wish I could hide it or make it internal or something. If they have some custom `AndroidInstrumentation` that they've built, then they should wire that up to the agent (or builder) and away we go. The agent is then also responsible for managing the lifecycle and ensuring that installed bits are uninstalled, etc...without the user needing to manage this.

If we want to relax this and allow users to install `AndroidInstrumentation` implementations willy-nillly whenever they choose, then this is a pretty big change. If we do that, then maybe it's less bad to let the `OpenTelemetryRumImpl` contain the instance of the `InstallationContext` (it's effectively a singleton), and that could be accessed by users doing self-service install or whatever. One thing that I don't like about this approach immediately is that it then provides multiple ways to get at the same things (eg. `otelRum.getRumSessionId()` and `otelRum.getInstrumentationInstallationContext().getSessionProvider().getSessionId()`) and it probably exposes more public interface than we want.

Another idea: I think if we put `getClock()` on `OpenTelemetryRum` we could delete the `InstallationContext` class and just go back to `install(Context, OpenTelemetryRum)`.

LikeTheSalad:
> So that's the thing: I don't want users to ever have to create instances of this! It's not a user's responsibility. I wish I could hide it or make it internal or something.

I think we all agree on this. The changes proposed in this PR are an option to prevent users from having to manually instantiate `InstallationContext`, while still keeping this type around.

> If we want to relax this and allow users to install AndroidInstrumentation implementations willy-nillly whenever they choose, then this is a pretty big change.

I think it's the other way around. Users have always had the option to install instrumentations willy-nilly whenever they choose. We could instead do what I mentioned earlier, of at least documenting that there's only a "supported way" of doing so, and that we wouldn't provide support for issues that may arise from installing instrumentations in a "non-supported" way. Though it would probably cause issues, as it would be a virtual constraint.

> Another idea: I think if we put getClock() on OpenTelemetryRum we could delete the InstallationContext class and just go back to install(Context, OpenTelemetryRum).

This sounds like the first option I mentioned earlier, where we'd get rid of `InstallationContext` and instead provide its dependencies via OpenTelemetryRum. I've always liked that option, though it was considered dangerous because we'd expose the `shutdown()` method to instrumentations. But I like that it reduces the api surface and I guess the shutdown concern can be verified for all the instrumentation impl that we define in OTel Android, so it should be manageable.

LikeTheSalad:
Not needed after https://github.com/open-telemetry/opentelemetry-android/pull/1645

PR #1645, opened 2026-03-11
Instrumentation api ehnancement

LikeTheSalad:
These changes affect the `AndroidInstrumentation` API. The API changes themselves aren't substantial, but there are many existing instrumentations that use this API, which is why a lot of files have been changed. Most of those changes are mechanical adaptations to the new signatures.

### Providing an `OpenTelemetryRum` instance to instrumentations

```kotlin
// BEFORE
interface AndroidInstrumentation {
    fun install(ctx: InstallationContext)
    fun uninstall(ctx: InstallationContext)
}
```

```kotlin
// AFTER
interface AndroidInstrumentation {
    fun install(context: Context, openTelemetryRum: OpenTelemetryRum)
    fun uninstall(context: Context, openTelemetryRum: OpenTelemetryRum)
}
```

This removes the need for a separate `InstallationContext` wrapper. Instrumentations now receive the `OpenTelemetryRum` instance created by the agent directly, alongside the Android `Context`.

### Extending and cleaning up `OpenTelemetryRum`

`OpenTelemetryRum` now exposes `SessionProvider` and `Clock`, which are the instances initialized by the agent. This gives instrumentations (and end users) access to the same objects backing the agent's telemetry, ensuring consistency.

`getRumSessionId()` has been removed since it was equivalent to `sessionProvider.getSessionId()`. Exposing the `SessionProvider` directly is more flexible, as callers also gain access to future additions to `SessionProvider` from a single getter.

### Module dependency cleanup

`InstallationContext` has been deleted.

The `session` module's dependency on `agent-api` was unused and has been removed. The `android-instrumentation` module's dependencies on `session`, `services`, and `opentelemetry-sdk` existed only because of `InstallationContext` and have been removed as well.


instrumentation/android-instrumentation/api/android-instrumentation.api
```diff
@@ -1,15 +1,6 @@
...
-
-public final class io/opentelemetry/android/instrumentation/InstallationContext {
-	public fun <init> (Landroid/content/Context;Lio/opentelemetry/api/OpenTelemetry;Lio/opentelemetry/android/session/SessionProvider;Lio/opentelemetry/sdk/common/Clock;)V
-	public final fun getApplication ()Landroid/app/Application;
-	public final fun getClock ()Lio/opentelemetry/sdk/common/Clock;
-	public final fun getContext ()Landroid/content/Context;
-	public final fun getOpenTelemetry ()Lio/opentelemetry/api/OpenTelemetry;
-	public final fun getSessionProvider ()Lio/opentelemetry/android/session/SessionProvider;
+	public abstract fun install (Landroid/content/Context;Lio/opentelemetry/android/OpenTelemetryRum;)V
+	public fun uninstall (Landroid/content/Context;Lio/opentelemetry/android/OpenTelemetryRum;)V
```
fractalwrench, line 4: I like that this reduces the API surface substantially and reduces the number of concepts exposed to library consumers.

fractalwrench (approved):
I'm happy with this approach. I think it's better than #1632 as it reduces the API surface and the number of concepts that folks need to know about.

marandaneto (approved):

agent-api/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
```diff
@@ -19,11 +21,14 @@ interface OpenTelemetryRum {
     val openTelemetry: OpenTelemetry
 
     /**
-     * Get the client session ID associated with this instance of the RUM instrumentation library.
-     * Note: this value will change throughout the lifetime of an application instance, so it is
-     * recommended that you do not cache this value, but always retrieve it from here when needed.
+     * Get the [SessionProvider] associated with this instance.
      */
-    fun getRumSessionId(): String
+    val sessionProvider: SessionProvider
```
breedx-splk, line 26: I like this PR and this approach in general, however...

⚠️  This change exposes a non-stable interface `SessionProvider` from a stable interface `OpenTelemetryRum`, and I think this is a real problem.

So I can see two options, maybe there are others:

* don't expose the `SessionProvider` and keep the `getSessionId()` method. Every usage of the `SessionProvider` just calls `getSessionId()` anyway, so this helps with some law of Demeter violations anyway.
* stabilize the session module. This is also a bit risky until we have more support from the spec/semconf on this.

Thoughts?

LikeTheSalad: Option 1 won't work for an existing instrumentation that needs to observe session changes, so we'd need to find a workaround just for that instrumentation, which I wouldn't feel comfortable with tbh.

If we must stabilize the session module, I'm ok with doing so in a way that only the `SessionProvider` interface becomes stable. It only provides a session ID anyway, which is quite unlikely to go away by the time the session's semconv becomes stable, so I'm not worried that there would be an issue with that. And even if there is, to me it's still worth making it stable now and deprecating its session ID getter later in case the "session.id" attribute name changes or the like, which I really think is quite unlikely at this point.

breedx-splk: > Option 1 won't work for an existing instrumentation that needs to observe session changes, so we'd need to find a workaround just for that instrumentation, which I wouldn't feel comfortable with tbh.

Oh you might be thinking of the `SessionManager` (which isA `SessionProvider`), which is also a `SessionPublisher` and exposes `addObserver()`. In this instrumentation api tho, we only have the `SessionProvider`, so I don't see any way for existing instrumentations to be registering for changes. Am I missing something?

core/src/test/java/io/opentelemetry/android/instrumentation/TestAndroidInstrumentation.kt
```diff
@@ -5,17 +5,26 @@
...
 class TestAndroidInstrumentation : AndroidInstrumentation {
     override val name: String = "test"
 
     var installed = false
         private set
 
-    override fun install(ctx: InstallationContext) {
+    override fun install(
+        context: Context,
+        openTelemetryRum: OpenTelemetryRum,
```
Sabaev, line 19: This change forces users to depend on OpenTelemetryRum and initialize it, even if they already have their own OpenTelemetry configuration and only want to reuse specific instrumentations (e.g., OkHttpInstrumentation, fragments instrumentation, etc.).

Sabaev: As a consequence, users cannot simply reuse instrumentations. They are also unable to share instrumentations built on plain OpenTelemetry or on a future open-telemetry-kotlin implementation.

agent-api/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
LikeTheSalad: > I don't see any way for existing instrumentations to be registering for changes.

We're currently doing so [here](https://github.com/open-telemetry/opentelemetry-android/blob/c30129731d36dcd03d921497aee501c917c1fa91/instrumentation/sessions/src/main/kotlin/io/opentelemetry/android/instrumentation/sessions/SessionInstrumentation.kt#L23), in the session event instrumentation.

We don't need to make them all stable; we could make `SessionProvider` stable, then the other APIs would be incubating or something like that, which would be fine for our session instrumentation, as we'd handle any breaking changes there, if any, around the session observer tools, without it affecting users.

core/src/test/java/io/opentelemetry/android/instrumentation/TestAndroidInstrumentation.kt
LikeTheSalad: From a usage point of view, it is not as different as having to create an instance of `InstallationContext` to provide your vanilla OpenTelemetry instance, as you would have to provide other dependencies there that you might not need for your use case. We've discussed the details over [here in this other PR](https://github.com/open-telemetry/opentelemetry-android/pull/1632) if you're interested. With this approach, if you'd like to use the instrumentations as standalone ones without the agent, note that we won't provide support for that use case regardless of this change getting merged or not, as discussed in this week's SIG meeting, though it would still be possible to do so if you really wanted to, by providing your vanilla OpenTelemetry instance in your own `OpenTelemetryRum.openTelemetry` getter implementation, as OpenTelemetryRum is just an interface.

agent-api/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
breedx-splk: > We're currently doing so [here](https://github.com/open-telemetry/opentelemetry-android/blob/c30129731d36dcd03d921497aee501c917c1fa91/instrumentation/sessions/src/main/kotlin/io/opentelemetry/android/instrumentation/sessions/SessionInstrumentation.kt#L23), in the session event instrumentation.

Oh damn that is such a hack! :) I didn't realize we were doing that....

I also concede that instrumentations _should_ (in theory) be able to monitor for session changes, and I think it should be ok to stabilize `SessionProvider` and `SessionPublisher` interfaces, because they're pretty small...but I do think we'd want to do that first before making this change here.

fractalwrench: Another option: could we avoid `SessionProvider/getSessionId()` on `OpenTelemetryRum` for now, and use some internal API to access the session ID? The session ID is only required by the session instrumentation right now, so it feels like we could skip for now & add it onto the stable API later.

surbhiia (approved):

PR #1660, opened 2026-03-19
Feature/TLS cert configuration

### LINKED ISSUE #1475: How to add client TLS in the OpenTelemetryRumInitializer?

brsgina:
Resolves #1475 

This modification allows us to set a client TLS certificate for OtlpHttpExporters in the configuration parameter via the initialize() function.


marandaneto:
This is a serious security issue. APKs are trivially decompilable (e.g., apktool, jadx), so anyone can extract the
 private key and impersonate your client

where are you reading this info from? hardcoded or read at runtime from keystore?

brsgina:
@marandaneto We are aware of the risks of embedding secrets in the APK. The endpoint is protected with mutual TLS (mTLS), so client authentication is based on a certificate, not just a static key.

The private key is stored in the Android Keystore and is not directly extractable. The app never exposes the raw key material; it is only used via the secure hardware-backed keystore APIs.

Additionally, even if someone decompiles the APK, they cannot simply impersonate the client without access to the keystore-protected private key and a valid device context.
Of course, we understand that mobile environments cannot be considered fully secure, so mTLS is combined with additional server-side checks and monitoring.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/connectivity/ClientTlsConnectivity.kt
```diff
@@ -0,0 +1,12 @@
+package io.opentelemetry.android.agent.connectivity
+
+/**
+ * The client key and the certificate chain to use for verifying client when TLS is enabled.
+ *
+ * @param privateKeyPem client key PKCS8 in PEM format.
+ * @param certificatePem client certificate chain in PEM format.
+ */
+class ClientTlsConnectivity(
```
breedx-splk, line 13: I would like to mark this class with `@Incubating` so that we can send a really clear message about this not being solidified yet.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/HttpExportConfiguration.kt
```diff
@@ -34,6 +35,12 @@ class HttpExportConfiguration internal constructor() {
      */
     var sslContext: SSLContextConnectivity? = null
 
+    /**
+     * Sets ths client key and the certificate chain to use for verifying client
+     * for all requests when TLS is enabled.
+     */
+    var clientTls: ClientTlsConnectivity? = null
```
breedx-splk, line 44: Can we mark this as incubating as well?

android-agent/src/main/kotlin/io/opentelemetry/android/agent/connectivity/HttpEndpointConnectivity.kt
```diff
@@ -10,28 +10,50 @@ internal class HttpEndpointConnectivity private constructor(
     private val headers: Map<String, String>,
     private val compression: Compression,
     private val sslContext: SSLContextConnectivity?,
+    private val clientTls: ClientTlsConnectivity?
```
breedx-splk: I think having a default null value helps to simplify a few places where the `clientTls` isn't passed.

```suggestion
    private val clientTls: ClientTlsConnectivity? = null
```

android-agent/src/test/kotlin/io/opentelemetry/android/agent/dsl/HttpExportConfigurationTest.kt
```diff
@@ -191,20 +229,101 @@ internal class HttpExportConfigurationTest {
...
-        config.metricsEndpoint().assertEndpointConfig("${url}v1/metrics", headers, Compression.GZIP, expectedSslContext)
+        config.spansEndpoint()
+            .assertEndpointConfig(
+                "${url}v1/traces",
+                headers,
+                Compression.GZIP,
+                expectedSslContext,
+                null
+
+            )
```
breedx-splk: this formatting is weird...but also, if this was defaulted to null in the param list, we wouldn't need to change these lines.

android-agent/src/test/kotlin/io/opentelemetry/android/agent/dsl/HttpExportConfigurationTest.kt
```diff
@@ -191,20 +229,101 @@ internal class HttpExportConfigurationTest {
...
+                expectedClientTls
+            )
     }
 
     private fun HttpEndpointConnectivity.assertEndpointConfig(
         expectedUrl: String,
         expectedHeaders: Map<String, String>,
         expectedCompression: Compression,
         expectedSslContext: SSLContextConnectivity?,
+        expectedClientTls: ClientTlsConnectivity?
```
breedx-splk:
```suggestion
        expectedClientTls: ClientTlsConnectivity? = null
```

breedx-splk (commented):
I like this. I had a few small suggestions, and I do feel like we should definitely try to flag this as incubating. Thanks for the contribution!

breedx-splk:
> This is a serious security issue. APKs are trivially decompilable (e.g., apktool, jadx), so anyone can extract the private key and impersonate your client
> 
> where are you reading this info from? hardcoded or read at runtime from keystore?

I think this will virtually always be the case....but this makes it significantly more difficult.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/connectivity/ClientTlsConnectivity.kt
brsgina: I added it.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/HttpExportConfiguration.kt
brsgina: Yes, I added it.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/connectivity/HttpEndpointConnectivity.kt
brsgina: Yes, I added this default value.

android-agent/src/test/kotlin/io/opentelemetry/android/agent/dsl/HttpExportConfigurationTest.kt
brsgina: I added this default value.

android-agent/src/test/kotlin/io/opentelemetry/android/agent/dsl/HttpExportConfigurationTest.kt
brsgina: Fixed.

breedx-splk (approved):
LGTM thanks!

android-agent/src/main/kotlin/io/opentelemetry/android/agent/connectivity/ClientTlsConnectivity.kt
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * The client key and the certificate chain to use for verifying client when TLS is enabled.
+ *
+ * @param privateKeyPem client key PKCS8 in PEM format.
+ * @param certificatePem client certificate chain in PEM format.
+ */
+
+@Incubating
+class ClientTlsConnectivity(
```
fractalwrench, line 13: My general preference would be for interfaces rather than classes in public APIs, as it avoids exposing `<init>` functions publicly and also tends to allow for more flexibility if we ever want to switch out a specific implementation for another. That feedback isn't blocking however.

fractalwrench (approved):
LGTM

Issue #1576, opened 2026-02-09
Implement Instrumentation for Scroll and Gesture Events


abhishek-dream11:
I would like to propose the addition of automated (or semi-automated) instrumentation for Scroll and Gesture events within the OpenTelemetry Android SDK. These events provide critical insights into user interactions and application responsiveness that are currently missing from the standard span collection.

fractalwrench:
@abhishek-dream11 thanks for opening this. What specific scroll and gesture events do you think should be captured, and what information would be useful to capture as telemetry?

abhishek-dream11:
@fractalwrench we can capture all scroll and gesture events (Swipe Gesture, Pinch Gesture), we can capture -
Scroll Events: Measure how far users scroll (read depth). Infinite Scroll Performance: Detect scroll lag,
Scroll Direction, Scroll Distance, Scroll Velocity.

DavidGrath:
Good Day,
I find this issue is interesting to work on, so I would like to voice my thoughts and get some feedback
First of all, I imagine that the use case for these gestures varies significantly by application, and so it's unlikely that most apps would want to collect all gestures by default:
* Scroll - drag and fling
* Tap
* Double-Tap
* Scale
* Long Press
* Drag

 Examples I can easily think of include:
* A finance app: lots of tapping to enter credentials, scrolling mainly for a "transactions" screen, barely any zooming
* A reading app: much more scrolling than the finance app, a significant amount of zooming, too for the reader to adjust scales and fonts and so on
* A social app: still scroll heavy, though zooming is restricted to image posts. Double-taps may also be significant here

And so with that in mind, I imagine that subscribing to only a subset of these gestures is necessary e.g
```kotlin
rum.registerGestures(GestureType.SCALE, GestureType.DOUBLE_TAP, GestureType.SCROLL)
```
Mainly to reduce telemetry noise. I also imagine that the possibility of custom-defined gestures exists, although I won't talk about that.

The next issue becomes "scope":
* Every screen of the finance app?
* The "PostDetails" screen of the social app?
* The "DocumentContent" custom view of the reading app?

[...]

PR #1681, opened 2026-04-09
Added basic double-tap, based off view click by cleverchuk. Addresses #1642 and #1576

DavidGrath:
The core difference between this and the original View Click is the use of a `GestureDetector` instead of relying on `MotionEvent` directly. As brought up in my comments on the issue, I'd like to know if it's desirable to add configurations to selectively track the double-tap event per-view via some form of allowlist. Then there's the issue of interplay between this and the original `view.click`


DavidGrath:
Okay, the steps I took before opening this PR include running `spotlessApply` and `testDebugUnitTest` successfully. After fixing the missing proguard file and `detekt` errors, please I'd like to know if there's anything else I would need to do to ensure a successful set of checks

DavidGrath:
I'll make the relevant fixes and revert within 48 hours

instrumentation/double-tap/src/main/kotlin/internal/ViewUtils.kt
```diff
@@ -0,0 +1,4 @@
+package internal
+
+internal const val APP_SCREEN_DOUBLE_TAP_EVENT_NAME = "app.screen.doubletap"
+internal const val VIEW_DOUBLE_TAP_EVENT_NAME = "app.widget.doubletap"
```
thompson-tomo: Could we not use a generic click/tap event name and use attribute to ie distinguish between the type ie single/double click

DavidGrath: Thanks for the review.

The idea is to eventually expand to other gestures including scroll and pinch, that's the reasoning behind me leaving single tap distinct from double tap

thompson-tomo: I have no problem with scroll & pinch having dedicated name as they would also have different attributes. However in the case of clicks (single vs double) they would have the same attributes and the number of definitions triples when you consider that it could also be a left vs right click when using a mouse.

DavidGrath: Okay, I see it working out like this:

* Add the first source and class from `InputDevice.getSources()` source as 2 attributes
* Add `MotionEvent.getToolType()` as an attribute
* Add `MotionEvent.getButtonState()` as an attribute if it's greater than 0
* Merge what I've done so far into the original view click and delete the new double tap folder

I'd be translating the returned integers into strings

What do you think?

DavidGrath: Oh, and also add `click.type` for either single or double

thompson-tomo: I like that as otherwise we would have had so many different event names.

The question is attribute names. I am thinking:

- `hw.pointer.type` for the tool type as it is describing the hardware that was used
- `hw.pointer.button` for the button state with stylus_primary shortened to primary etc.
- `hw.pointer.clicks` for how many clicks ie 1 or 2

DavidGrath: Okay, I assume we're only focusing on `SOURCE_CLASS_POINTER` since that's where gestures typically come from. I think it's good, although I'd prefer hardware in full instead of shortened.

DavidGrath: I'll try and get back to this on or before the 18th

thompson-tomo: No worries, note `hw.*` is a defined namespace in semconv. 😉

DavidGrath: Ah. Thanks for pointing that out. Noted

DavidGrath:
Main points of discussion:
* If the button state isn't primary, then should the event still be sent or actively suppressed?
* I've changed the original code to indirectly rely on `ViewConfiguration.getDoubleTapTimeout()` through the use of `onSingleTapConfirmed`. Is this an okay change or should I rely on `onSingleTapUp` instead?

DavidGrath:
@thompson-tomo , please how do the changes look?

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
```diff
@@ -26,28 +32,90 @@ internal class ViewClickEventGenerator(
...
+                val toolTypeHasButtons = toolTypeInt == MotionEvent.TOOL_TYPE_MOUSE || toolTypeInt == MotionEvent.TOOL_TYPE_STYLUS
+                val isButtonPrimary = buttonStateInt == MotionEvent.BUTTON_PRIMARY || buttonStateInt == MotionEvent.BUTTON_STYLUS_PRIMARY
+                if(toolTypeHasButtons && !isButtonPrimary) {
+                    return false
+                }
+                val buttonState = buttonStateToString(buttonStateInt)
+
+                if(buttonState != null) {
+                    appEvent.setAttribute(HARDWARE_POINTER_BUTTON, buttonState)
+                }
```
thompson-tomo: Not sure if we need to restrict it to primary given we already have an attribute to indicate which button.

```suggestion
                val toolTypeHasButtons = toolTypeInt == MotionEvent.TOOL_TYPE_MOUSE || toolTypeInt == MotionEvent.TOOL_TYPE_STYLUS
                val buttonState = null
                if(toolTypeHasButtons) {
                    val buttonStateInt = motionEvent.buttonState
                    buttonState = buttonStateToString(buttonStateInt)
                }
                if(buttonState != null) {
                    appEvent.setAttribute(HARDWARE_POINTER_BUTTON, buttonState)
                }
```

DavidGrath: Okay, I'll need to delete the corresponding test I wrote for this as well

thompson-tomo: Sure, note the change should be replicated to single click.

DavidGrath: Done

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
thompson-tomo, line 140: Should we also be adding the `hw.pointer.*` attributes on her as well?

DavidGrath: I think it might not be necessary, since `app.screen.click` is always sent

DavidGrath: And it already has the same attributes

cleverchuk (approved):
May need to standardize the new attributes in semconv otherwise LGTM!

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
```diff
@@ -26,28 +32,94 @@ internal class ViewClickEventGenerator(
...
+    private val viewCoordinates = IntArray(2)
+
+    fun startTracking(window: Window) {
+        windowRef = WeakReference(window)
+        val currentCallback = window.callback
+        window.callback = WindowCallbackWrapper(currentCallback, this)
+    }
+
+    fun generateClick(motionEvent: MotionEvent?) {
+        if(motionEvent != null) {
```
fractalwrench:
```suggestion
        if (motionEvent != null) {
```

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
```diff
@@ -26,28 +32,94 @@ internal class ViewClickEventGenerator(
...
+
+
+                val toolTypeHasButtons = toolTypeInt == MotionEvent.TOOL_TYPE_MOUSE || toolTypeInt == MotionEvent.TOOL_TYPE_STYLUS
+                var buttonState: String? = null
+                if(toolTypeHasButtons) {
+                    val buttonStateInt = motionEvent.buttonState
+                    buttonState = buttonStateToString(buttonStateInt)
+                }
+
+                if(buttonState != null) {
```
fractalwrench:
```suggestion
                if (buttonState != null) {
```

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
```diff
@@ -26,28 +32,94 @@ internal class ViewClickEventGenerator(
...
+                val appEvent = createEvent(APP_SCREEN_CLICK_EVENT_NAME)
+                    .setAttribute(APP_SCREEN_COORDINATE_Y, motionEvent.y.toLong())
+                    .setAttribute(APP_SCREEN_COORDINATE_X, motionEvent.x.toLong())
+                    .setAttribute(HARDWARE_POINTER_CLICKS, 1)
+                    .setAttribute(HARDWARE_POINTER_TYPE, toolType)
+
+
+                val toolTypeHasButtons = toolTypeInt == MotionEvent.TOOL_TYPE_MOUSE || toolTypeInt == MotionEvent.TOOL_TYPE_STYLUS
+                var buttonState: String? = null
+                if(toolTypeHasButtons) {
```
fractalwrench:
```suggestion
                if (toolTypeHasButtons) {
```

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
```diff
@@ -26,28 +32,94 @@ internal class ViewClickEventGenerator(
...
+                val appEvent = createEvent(APP_SCREEN_CLICK_EVENT_NAME)
                     .setAttribute(APP_SCREEN_COORDINATE_Y, motionEvent.y.toLong())
                     .setAttribute(APP_SCREEN_COORDINATE_X, motionEvent.x.toLong())
-                    .emit()
+                    .setAttribute(HARDWARE_POINTER_CLICKS, 2)
+                    .setAttribute(HARDWARE_POINTER_TYPE, toolType)
+
+                val toolTypeHasButtons = toolTypeInt == MotionEvent.TOOL_TYPE_MOUSE || toolTypeInt == MotionEvent.TOOL_TYPE_STYLUS
+                var buttonState: String? = null
+                if(toolTypeHasButtons) {
```
fractalwrench:
```suggestion
                if (toolTypeHasButtons) {
```

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
```diff
@@ -26,28 +32,94 @@ internal class ViewClickEventGenerator(
...
+                    val buttonStateInt = motionEvent.buttonState
+                    buttonState = buttonStateToString(buttonStateInt)
+                }
+
+                if(buttonState != null) {
```
fractalwrench:
```suggestion
                if (buttonState != null) {
```

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
```diff
@@ -26,28 +32,94 @@ internal class ViewClickEventGenerator(
...
+                val toolType = toolTypeToString(toolTypeInt)
+
+                val appEvent = createEvent(APP_SCREEN_CLICK_EVENT_NAME)
                     .setAttribute(APP_SCREEN_COORDINATE_Y, motionEvent.y.toLong())
                     .setAttribute(APP_SCREEN_COORDINATE_X, motionEvent.x.toLong())
-                    .emit()
+                    .setAttribute(HARDWARE_POINTER_CLICKS, 2)
+                    .setAttribute(HARDWARE_POINTER_TYPE, toolType)
+
+                val toolTypeHasButtons = toolTypeInt == MotionEvent.TOOL_TYPE_MOUSE || toolTypeInt == MotionEvent.TOOL_TYPE_STYLUS
```
fractalwrench, line 51: I would consider driving this and other information via an enum in `ViewUtils` as that will reduce the amount of logic in this class. Here's a non-exhaustive example of what I mean:

```
enum class TapEvent(
    val motionEvent: Int,
    val toolType: Int,
) {


    val toolTypeDescription: String // derive from MotionEvent.TOOL_TYPE_*
    val hasButtons: Boolean = toolTypeInt == MotionEvent.TOOL_TYPE_MOUSE || toolTypeInt == MotionEvent.TOOL_TYPE_STYLUS
}
```

DavidGrath: Thank you, I've applied the other suggestions. I'll look into this and give my feedback by Wednesday or Thursday

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
thompson-tomo: That is true but I was thinking of the use case of where the collector is filtering events and might filter out app.click hence the view click event should be self sufficient.

instrumentation/view-click/src/test/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickInstrumentationTest.kt
```diff
@@ -293,37 +323,411 @@ class ViewClickInstrumentationTest {
...
+        viewClickActivityCallback.onActivityResumed(activity)
+        verify {
+            window.callback = capture(wrapperCapturingSlot)
+        }
+
+        for(motionEvent in doubleTapSequence) {
+            wrapperCapturingSlot.captured.dispatchTouchEvent(motionEvent)
+        }
+        val events = openTelemetryRule.logRecords
+        Assertions.assertThat(events).hasSize(2)
```
breedx-splk: Please use the static import for these.
```suggestion
        assertThat(events).hasSize(2)
```

instrumentation/view-click/src/test/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickInstrumentationTest.kt
```diff
@@ -293,37 +323,411 @@ class ViewClickInstrumentationTest {
...
+        }
+
+        wrapperCapturingSlot.captured.dispatchTouchEvent(doubleTapSequence[0])
+        wrapperCapturingSlot.captured.dispatchTouchEvent(doubleTapSequence[1])
+        fastForwardDoubleTapTimeout()
+        wrapperCapturingSlot.captured.dispatchTouchEvent(doubleTapSequence[2])
+        wrapperCapturingSlot.captured.dispatchTouchEvent(doubleTapSequence[3])
+
+        val events = openTelemetryRule.logRecords
+        Assertions.assertThat(events).hasSize(2)
```
breedx-splk:
```suggestion
       assertThat(events).hasSize(2)
```

DavidGrath: Okay, I've checked and seen that I used `Assertions` like this 4 more times. I'll fix them all for consistency

DavidGrath:
Ah, the stabilization changes. Merging with main and reworking my tests

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
DavidGrath: @fractalwrench, I've used a separate class, although I didn't know how to use an enum specifically like you asked, so I used a regular one. Is `TapEvent` alright? I'll work on fixing my broken tests when I can later today

DavidGrath:
> May need to standardize the new attributes in semconv otherwise LGTM! 

Thanks @cleverchuk .

@thompson-tomo, I'll open an issue in semconv over the weekend for the attributes. Anything I need to note?
@sandy2008, do you think this helps your proposal?

thompson-tomo:
@DavidGrath the only thing I can think of would be linking to this PR.

Might be worthwhile to share it with the swift sig for their feedback.

DavidGrath:
> @DavidGrath the only thing I can think of would be linking to this PR.
> 
> Might be worthwhile to share it with the swift sig for their feedback.

@thompson-tomo I didn't make the issue on time as planned, I'll still try and look into it this week

breedx-splk (approved):
Thanks again for the contribution @DavidGrath!

I'm going to create an issue to add this instrumentation to the demo app so that users can see it in a real use case. It would be amazing if you could pick that up as well, but no pressure.

breedx-splk:
Ok, I've created #1707 and #1708 as follow-up tasks. 🚀

PR #1683, opened 2026-04-13
docs: align HTTP client READMEs with stable semantic conventions

### LINKED ISSUE #1393: For android network instrumentation spans we should use `http.response.status_code` for status code
Currently we are using `http.status_code` as per doc [here](https://github.com/open-telemetry/opentelemetry-android/tree/main/instrumentation/okhttp3) same is true for other network lib instrumentation

`http.response.status_code` is defined [here](https://opentelemetry.io/docs/specs/semconv/registry/attributes/http/#http-response-status-code)

ashishvz:
## Summary

We debugged and tested client HTTP spans from a sample app and confirmed the instrumentation stack **already emits the latest stable HTTP semantic conventions** (for example `http.response.status_code`, `http.request.method`, and `url.full`). **No code changes** were required.

This PR only **updates READMEs** so the documented attribute names match what we saw at runtime:

- `instrumentation/okhttp3/README.md`
- `instrumentation/httpurlconnection/README.md`

Fixes #1393

## Verification

See screenshot below: client span attributes from a real request (e.g. HTTPS GET) showing stable keys including **`http.response.status_code`** and related HTTP / network fields.

Screenshot 2026-04-13 at 4 34 49 PM

### DIFF
--- a/instrumentation/httpurlconnection/README.md
+++ b/instrumentation/httpurlconnection/README.md
@@ -34,14 +34,15 @@ the following occurs: the response stream is fully read, the stream is closed,
 - Name: Determined by the HTTP span name extractor (typically `HTTP {method}` or derived from the URL)
 - Description: Client-side HTTP request
 - Attributes (following OpenTelemetry [HTTP semantic conventions](https://opentelemetry.io/docs/specs/semconv/http/http-spans/)):
-  - `http.method` — request method (GET, POST, etc.)
-  - `http.url` — full URL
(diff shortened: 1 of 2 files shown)


fractalwrench (approved):
LGTM - thanks for investigating this.

LikeTheSalad (approved):

ashishvz:
Thank you @fractalwrench and @LikeTheSalad for approving and merging the PR.

PR #1688, opened 2026-04-16
feat(demo-app): OkHttp build-time instrumentation and demo screen

### LINKED ISSUE #419: [demo-app] Set up build-time OkHttp instrumentation
The demo app wants to show off instrumentation features, so we should set up the build-time autoinstrumentation ok OkHttp. This also means using an OkHttp client somewhere in the app. I recommend setting up a button, probably somewhere on the main screen, that when pressed, does an HTTP call. Maybe a few buttons, one that generates a known good HTTP 200, another for 404, and another that's a completely connection failure (similar to the splunk demo app).

ashishvz:
## Summary
Implements [issue #419](https://github.com/open-telemetry/opentelemetry-android/issues/419): the demo app now uses **build-time OkHttp autoinstrumentation** (ByteBuddy + okhttp3 agent/library) and includes a dedicated screen to fire sample HTTP traffic for manual verification in a collector.

### Gradle / composite build (`demo-app/`)
- Apply **`net.bytebuddy.byte-buddy-gradle-plugin`** (version aligned with repo catalog in `demo-app/gradle/libs.versions.toml`).
- Add **`implementation`** `io.opentelemetry.android.instrumentation:okhttp3-library` and **`byteBuddy`** `okhttp3-agent`, plus **`okhttp`** for app code.
- Extend **`demo-app/settings.gradle.kts`** `dependencySubstitution` for `okhttp3-library` and `okhttp3-agent` when using the composite `includeBuild("..")`.

### App behavior
- **`OkHttpDemoActivity`** + **`OkHttpDemoScreen`**: Material3 **Scaffold** / **TopAppBar**, back navigation, **56dp**-tall **OutlinedButton** actions:
  - GET `httpbin` **200** and **404**
  - GET **connection failure** (`127.0.0.1:1`)
  - **POST** JSON to `https://httpbin.org/post`
  - **Slow GET** `https://httpbin.org/delay/3`
- **`MainActivity`**: **`enableEdgeToEdge()`**, root **`safeDrawingPadding()`**, scroll + **`spacedBy`** layout; **“OkHttp instrumentation”** opens the new activity; title uses **`onBackground`** for dark theme.
- **`DemoAppTheme`**: **transparent** status and navigation bar colors; **`isAppearanceLightStatusBars`** / **`isAppearanceLightNavigationBars`** from **background luminance** so system bar icons stay readable in light/dark/dynamic color.

## Issue
Fixes https://github.com/open-telemetry/opentelemetry-android/issues/419

### DIFF
--- a/demo-app/src/main/java/io/opentelemetry/android/demo/OkHttpDemoScreen.kt
+++ b/demo-app/src/main/java/io/opentelemetry/android/demo/OkHttpDemoScreen.kt
@@ -85,7 +85,7 @@ fun OkHttpDemoScreen(onBack: () -> Unit) {
                 color = MaterialTheme.colorScheme.onSurfaceVariant,
             )
             Spacer(modifier = Modifier.height(8.dp))
-            OutlinedButton(
+            OkHttpDemoActionButton(
                 onClick = {
                     scope.launch {
                         val msg =
@@ -102,17 +102,9 @@ fun OkHttpDemoScreen(onBack: () -> Unit) {
                         }
                     }
                 },
-                modifier = Modifier.fillMaxWidth().height(56.dp),
-                border = BorderStroke(1.dp, DemoAccentBlue),
-                colors =
-                    ButtonDefaults.outlinedButtonColors(
-                        containerColor = Color.Transparent,
-                        contentColor = DemoAccentBlue,
-                    ),
-            ) {
-                Text("GET known 200")
-            }
-            OutlinedButton(
+                text = "GET known 200",
+            )
+            OkHttpDemoActionButton(
                 onClick = {
                     scope.launch {
                         val msg =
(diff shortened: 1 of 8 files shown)


ashishvz:
@fractalwrench PR is ready for review!

demo-app/src/main/java/io/opentelemetry/android/demo/OkHttpDemoScreen.kt
```diff
@@ -0,0 +1,228 @@
...
+                        }
+                    }
+                },
+                modifier = Modifier.fillMaxWidth().height(56.dp),
+                border = BorderStroke(1.dp, DemoAccentBlue),
+                colors =
+                    ButtonDefaults.outlinedButtonColors(
+                        containerColor = Color.Transparent,
+                        contentColor = DemoAccentBlue,
+                    ),
```
fractalwrench: These attributes are shared between several other `OutlinedButton` calls. Could you try extracting the creation of the button to a function so that the style only needs to be specified in one place?

fractalwrench (approved):
LGTM with one minor nit about UI code organisation. Thanks for taking this on!

demo-app/src/main/java/io/opentelemetry/android/demo/OkHttpDemoScreen.kt
ashishvz: Sure, will implement it.

ashishvz:
@fractalwrench Thanks for approving and merging the code!

fractalwrench:
Thanks very much for the contribution @ashishvz!

PR #1714, opened 2026-04-30
Add agent dsl to disable instrumentations

### LINKED ISSUE #411: Programmatically disabling instrumentation on classpath
This is a tracking issue that follows up on [this comment ](https://github.com/open-telemetry/opentelemetry-android/pull/408#discussion_r1625869332) in #408.

Now that we've moved to the fancy new instrumentation API, we have a case where all instrumentation that is included at build-time should show up on the classpath at runtime in order to be disabled. There may be cases, however, where users want to programmatically disable some instrumentation at runtime based on configuration or something else (a probabilistic model, whatever the case may be).

It's perfectly reasonable to defer this until we have an actual ask from an end-user, but I wanted to make sure it had been captured.

breedx-splk:
Resolves #411.

By default, the instrumentation loader will find all instrumentation on the classpath and load it. This new DSL method allows suppressing some of those instrumentations if they might be unwanted.

I like the simplicity of this approach, because it just leverages the existing mechanism built into the `OtelRumConfig` from core and the existing `InstrumentationConfiguration` DSL. One shortcoming, though, is that users might want this to be as lazy as possible -- evaluated just before install time (so that they can make a calculation or whatever). That approach would be considerably more involved.


fractalwrench (commented):
This feels like a reasonable approach. I'm not convinced on the name, perhaps something like `disabledInstrumentations` or `suppressedInstrumentation` is more descriptive?

While reviewing I also noticed that `OtelRumConfig` has `shouldDiscoverInstrumentations()`. I'd be inclined to keep the scope of this PR tight, but whether we need something similar should be part of related discussion IMO.

breedx-splk:
> This feels like a reasonable approach. I'm not convinced on the name, perhaps something like `disabledInstrumentations` or `suppressedInstrumentation` is more descriptive?

I didn't include the word 'instrumentation' in the name because it's already scoped to the 'instrumentations' block....which would make it redundant. 

```
instrumentations {
  suppressing (
    "unwanted.instrumentation.name",
    "something.unwanted"
  )
  ...
}
```

I think I see where you're coming from, though. Nothing else (yet?) in the DSL uses the infinitive verb form (tho it's not super uncommon in other DSLs I think). Maybe you would like `suppress` or `suppressed` better? I'm open to changing it.
 
> While reviewing I also noticed that `OtelRumConfig` has `shouldDiscoverInstrumentations()`. I'd be inclined to keep the scope of this PR tight, but whether we need something similar should be part of related discussion IMO.

Oh that definitely occurred to me as well, and we should add that....agree not in this PR.

breedx-splk:
Created #1716 to follow up on this idea.

demo-app/src/main/java/io/opentelemetry/android/demo/OtelDemoApplication.kt
```diff
@@ -40,7 +40,16 @@ class OtelDemoApplication : Application() {
...
+                    instrumentations {
+                        suppressing (
+                            "unwanted.instrumentation.name",
+                            "something.unwanted"
+                        )
+                        screenOrientation {
+                            enabled(false)
+                        }
+                    }
+                },
```
LikeTheSalad, line 52: Considering that there's already a way to do this for built-in instrumentations with `enabled(false)`, I wonder if we can use a name that intuitively explains that these 2 apis do the same, only that the new one is targeted towards non-built-in instrumentations? We know this is the case, but for a regular user, it might be confusing to say (I'll disable X instrumentation and then suppress Y) and think that the outcome is the same for both.

Maybe `disableByName` could do?

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/instrumentation/InstrumentationConfiguration.kt
```diff
@@ -84,4 +84,10 @@ class InstrumentationConfiguration internal constructor(
     fun screenOrientation(configure: ScreenOrientationConfiguration.() -> Unit) {
         screenOrientation.configure()
     }
+
+    fun suppressing(vararg instrumentationsToExclude: String) {
```
breedx-splk, line 94: Add javadoc that explains what this actually does.

demo-app/src/main/java/io/opentelemetry/android/demo/OtelDemoApplication.kt
breedx-splk: We did a pretty good job at bikeshedding names in the SIG meeting and concluded that we should be ok just leaving it was `suppressing`. 🤘🏻

LikeTheSalad (approved):
Cheers!

fractalwrench (approved):
LGTM - thanks

PR #1718, opened 2026-05-01
Add agent DSL to add session observers

### LINKED ISSUE #328: Session Id Refresh Callback
When session expires we need a callback or  hook to identify that the session has ended based on that we can do clean up of cached session id. Or trigger any custom logic as when required if callback implemented.

breedx-splk:
Resolves #328.

As discussed in SIG earlier this week, we think there is room now to allow users to provide session observers via the DSL.


LikeTheSalad (approved):
Cheers!

PR #1719, opened 2026-05-02
Added view click to demo app and android agent

### LINKED ISSUE #1707: Add double tap instrumentation to demo app
#1681 added new instrumentation that is capable of reporting double tap events on a View.

I'm not too opinionated about it, but ideally this would be application functionality that would show how the data looks under various possible situations.

If the app already works well enough to demonstrate these events, then this issue can just be a validation step to ensure that the telemetry generated matches what's in the readme.

DavidGrath:
Closes https://github.com/open-telemetry/opentelemetry-android/issues/1707


DavidGrath:
I notice none of the other `instrumentation` package classes have any tests. Should I bother with coverage here?

android-agent/build.gradle.kts
```diff
@@ -29,6 +29,7 @@ dependencies {
     implementation(project(":instrumentation:startup"))
     implementation(project(":instrumentation:sessions"))
     implementation(project(":instrumentation:screen-orientation"))
+    implementation(project(":instrumentation:view-click"))
```
LikeTheSalad: I'm not sure we've discussed adding this as a default instrumentation for the agent? If the intent of the changes is to only add a demo app example, there's no need to update the agent; the dependency can be added directly into the demo app's Gradle file.

LikeTheSalad (changes requested):
> I notice none of the other instrumentation package classes have any tests. Should I bother with coverage here?

For demo app changes, there's no need. The demo app changes look good, though I'm blocking to confirm we want the agent changes as well.

android-agent/build.gradle.kts
DavidGrath: I remember having issues when trying to add it via the substitutions in the settings.gradle file. I'll try again and revert later today with more feedback

DavidGrath: Thank you

DavidGrath: Okay, so adding `view-click` alone caused a complaint from the IDE that the super interface is required to be present, and that's why I added `android-instrumentation` directly. Feels like I could do better, though

DavidGrath: Line removed

demo-app/src/main/java/io/opentelemetry/android/demo/about/Feature.kt
```diff
@@ -40,6 +40,14 @@ fun getFeatureList(): List<Feature> {
             """
             - Provides access to the OpenTelemetry APIs for manual instrumentation, allowing developers to create custom spans and events as needed.
             """.trimIndent()
+        ),
+        Feature(
+            "View Click Detection",
+            """
+            - Modelled as a log event.
```
breedx-splk: Simply event. The "log" part is implied and an implementation detail.

```suggestion
            - Modelled as an event.
```

breedx-splk (approved):
Looks good to me. One small suggestion. Thanks for the contribution @DavidGrath !

demo-app/src/main/java/io/opentelemetry/android/demo/about/Feature.kt
DavidGrath: Done. Thank you

LikeTheSalad (approved):
Thank you! 🙏

PR #1721, opened 2026-05-05
Revert "docs: require signed commits (#1717)"

breedx-splk:
This reverts commit 3c20dc1395d54de72107a3e3009b8990dd42d163.

Apparently I completely hallucinated the idea that we required signed commits. We definitely do not.

(sorry, I really just confused this with the Splunk repos, which do require it).


fractalwrench (approved):
LGTM. _Do_ we want to consider signed commits in future?

PR #1722, opened 2026-05-05
Clarify release cadence schedule

breedx-splk:
Tweaks the release cadence wording. Because we follow upstream, and upstream roughly publishes on the Wednesday after the second Monday, we can say that we aim for the Tuesday after the third Monday. 

This helps with machine calculation/estimate.


fractalwrench (approved):
LGTM

LikeTheSalad (approved):

PR #1715, opened 2026-05-01
feat: support fullUrl in EndpointConfiguration to bypass default signal path appending

Cloud-Architect-Emma:
Fixes #1712

## Summary

When using `HttpExportConfiguration`, the signal-specific paths (`/v1/logs`, `/v1/traces`, `/v1/metrics`) are always appended to the base URL. This makes it impossible to use a custom endpoint path like `/v2/logs`.

This PR adds a `fullUrl` property to `EndpointConfiguration` that, when set, bypasses the automatic path appending and uses the provided URL as-is.

## Usage

```kotlin
httpExport {
    baseUrl = "https://example.com"
    logs {
        fullUrl = "https://example.com/v2/logs"  // used as-is, no path appended
    }
    // traces still uses: https://example.com/v1/traces
    // metrics still uses: https://example.com/v1/metrics
}
```

## Changes

- Added `fullUrl: String?` property to `EndpointConfiguration`
- Updated `HttpExportConfiguration` to use `fullUrl` when set, falling back to existing `url` + path behaviour
- Updated `HttpEndpointConnectivity.forTraces/forLogs/forMetrics` to accept a `fullUrl` flag
- Added tests covering the new behaviour

PR #1723, opened 2026-05-05
feat: support fullUrl in EndpointConfiguration to bypass default signal path appending

### LINKED ISSUE #1712: How can I specify a url for my logs in HttpExportConfiguration with different endpoint than v1/logs?
Hi all,

In the OpenTelemetryRumInitializer.initialize function I configure the httpExport base url like this:
`httpExport {
        baseUrl = BASE_URL }`

 But these endpoints are hardcoded and they are appended to the base url:
- for traces: baseUrl.trimEnd('/') + "/v1/traces"
- for logs: baseUrl.trimEnd('/') + "/v1/logs"
- for metrics: baseUrl.trimEnd('/') + "/v1/metrics"

Even if inside the http export configuration if I add a different url to the logs, these endpoints are still appended.
`httpExport {
                        baseUrl = BASE_URL
                        logs {
                            url = "$BASE_URL/v2/logs"
                        }
 }
`

I would like to use a different path, for example not v1/logs but v2/logs. How can I do that? 
(With io.opentelemetry.android.instrumentation (+ io.opentelemetry.exporter.otlp) it was possible to configure my own endpoint for logs, so if its not possible here, what is the reason for it?)

Cloud-Architect-Emma:
Fixes #1712

## Summary
When using `HttpExportConfiguration`, the signal-specific paths (`/v1/logs`, 
`/v1/traces`, `/v1/metrics`) are always appended to the base URL. This makes 
it impossible to use a custom endpoint path like `/v2/logs`.

This PR adds a `fullUrl` property to `EndpointConfiguration` that, when set, 
bypasses the automatic path appending and uses the provided URL as-is.

## Usage
```kotlin
httpExport {
    baseUrl = "https://example.com"
    logs {
        fullUrl = "https://example.com/v2/logs"  // used as-is, no path appended
    }
    // traces still uses: https://example.com/v1/traces
    // metrics still uses: https://example.com/v1/metrics
}
```

## Changes
- Added `fullUrl: String?` property to `EndpointConfiguration`
- Updated `HttpExportConfiguration` to use `fullUrl` when set, falling back 
  to existing `url` + path behaviour
- Updated `HttpEndpointConnectivity.forTraces/forLogs/forMetrics` to accept 
  a `fullUrl` flag
- Fixed detekt violations (trailing commas, line length)
- Added tests covering the new behaviour

## Notes
This supersedes PR #1715 which had CLA identity issues.

### DIFF
--- a/android-agent/api/android-agent.api
+++ b/android-agent/api/android-agent.api
@@ -33,9 +33,11 @@ public final class io/opentelemetry/android/agent/dsl/DiskBufferingConfiguration
 
 public final class io/opentelemetry/android/agent/dsl/EndpointConfiguration {
 	public final fun getCompression ()Lio/opentelemetry/android/agent/connectivity/Compression;
+	public final fun getFullUrl ()Ljava/lang/String;
 	public final fun getHeaders ()Ljava/util/Map;
 	public final fun getUrl ()Ljava/lang/String;
 	public final fun setCompression (Lio/opentelemetry/android/agent/connectivity/Compression;)V
+	public final fun setFullUrl (Ljava/lang/String;)V
 	public final fun setHeaders (Ljava/util/Map;)V
 	public final fun setUrl (Ljava/lang/String;)V
 }
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/EndpointConfiguration.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/EndpointConfiguration.kt
@@ -2,7 +2,6 @@
  * Copyright The OpenTelemetry Authors
  * SPDX-License-Identifier: Apache-2.0
  */
-
 package io.opentelemetry.android.agent.dsl
 
 import io.opentelemetry.android.agent.connectivity.Compression
@@ -13,9 +12,17 @@ import io.opentelemetry.android.agent.connectivity.Compression
 @OpenTelemetryDslMarker
 class EndpointConfiguration internal constructor(
     /**
-     * URL for HTTP export requests.
+     * Base URL for HTTP export requests. The signal-specific path (e.g. /v1/logs) will be
+     * appended automatically.
      */
     var url: String,
+    /**
+     * Full URL for HTTP export requests. When set, this URL is used as-is without appending
+     * any signal-specific path. Use this to specify a completely custom endpoint path,
+     * for example "https://example.com/v2/logs".
+     * Note: when set, this overrides the baseUrl from the surrounding httpExport block entirely.
+     */
+    var fullUrl: String? = null,
     /**
      * Headers that should be attached to HTTP export requests.
      */
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/connectivity/HttpEndpointConnectivity.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/connectivity/HttpEndpointConnectivity.kt
@@ -2,7 +2,6 @@
  * Copyright The OpenTelemetry Authors
  * SPDX-License-Identifier: Apache-2.0
  */
-
 package io.opentelemetry.android.agent.connectivity
 
 internal class HttpEndpointConnectivity private constructor(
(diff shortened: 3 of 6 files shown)


Cloud-Architect-Emma:
Hi @breedx-splk, quick update,  the line endings and detekt issues are 
now fixed! EasyCLA is passing.

The only remaining blocker is a conflict in android-agent/api/android-agent.api. 
This file is auto-generated by Gradle and I'm unable to run it locally 
without the Android SDK set up.

Could you resolve it by running:

  ./gradlew :android-agent:apiDump

and pushing to my branch? That should be the very last thing needed 
before this PR is ready for review.

I'm really sorry for all the trouble and deeply appreciate your patience. 
Thank you so much!

android-agent/api/android-agent.api
```diff
@@ -121,5 +122,4 @@ public final class io/opentelemetry/android/agent/dsl/instrumentation/SlowRender
 	public final fun detectionPollInterval-LRDsOJo (J)V
 	public final fun enableVerboseDebugLogging ()V
 	public fun enabled (Z)V
-}
-
+}
```
breedx-splk: It's weird. Apparently there are newline issues, and this is what the [failure is about](https://scans.gradle.com/s/sg6fymf6p6him).
```suggestion
}


```

.gitattributes
```diff
@@ -0,0 +1,2 @@
+﻿* text=auto eol=lf
+*.kt text eol=lf
```
breedx-splk: Let's not add this file in this PR please. Thanks.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/EndpointConfiguration.kt
```diff
@@ -13,9 +12,16 @@ import io.opentelemetry.android.agent.connectivity.Compression
...
     /**
-     * URL for HTTP export requests.
+     * Base URL for HTTP export requests. The signal-specific path (e.g. /v1/logs) will be
+     * appended automatically.
      */
     var url: String,
+    /**
+     * Full URL for HTTP export requests. When set, this URL is used as-is without appending
+     * any signal-specific path. Use this to specify a completely custom endpoint path,
+     * for example "https://example.com/v2/logs".
```
breedx-splk, line 22: I think there should also be a note here that explains this this overrides the `baseUrl` from the surrounding block -- so that if you provide this `fullUrl`, then the `baseUrl` won't be used at all.

breedx-splk (commented):
This seems like a reasonable approach to me...thanks for the contribution @Cloud-Architect-Emma!

There are two `gradlew` files that should not be part of this PR, as well as the `.gitattributes`, so lets get this removed and then move it forward. Thanks again for your patience!

breedx-splk:
Oh there are some compilation failures too in the `HttpEndpointConnectivityTest`. I'd recommend setting up the Android SDK so that you can build and test this project on your machine. 😉

Cloud-Architect-Emma:
Hi @breedx-splk, all checks are now passing! I've addressed all your feedback:
- Removed .gitattributes, gradlew.bat, and demo-app/gradlew.bat
- Fixed compile error in HttpEndpointConnectivityTest.kt  
- Updated KDoc to note that fullUrl overrides baseUrl entirely
- Fixed API file trailing newlines

Ready for your approval whenever you get a chance. Thank you so much for your patience throughout this whole process!

Cloud-Architect-Emma:
@breedx-splk, PR passed all checks. Can you please review at your convenience? Thnaks

demo-app/gradlew.bat
breedx-splk, line 1: Oops, we don't want this file to be deleted! We just don't expect it to be changed in this PR. Please revert.

gradlew.bat
breedx-splk, line 1: Oops, we don't want this file to be deleted! We just don't expect it to be changed in this PR. Please revert.

Cloud-Architect-Emma: file have been revert

demo-app/gradlew.bat
Cloud-Architect-Emma: Files is revised

gradlew.bat
Cloud-Architect-Emma: File is reverted

Cloud-Architect-Emma: file have been reverted

breedx-splk (approved):
Thanks for the contribution @Cloud-Architect-Emma! This is looking good to me now, and it's going to be useful for users to be able to customize the specific url for a given signal. 

There's one result from this which is potentially odd looking, but I can't think of a better way of doing it that remains backwards compatible. 

```kotlin
configuration = {
    httpExport {
        baseUrl = "http://10.0.2.2:4318"
        logs {
            url = "http://something.else.com"
            fullUrl = "https://something.other.net/v1/whatever"
        }
    }
...
}
```

I think that would be the worst case....and the solution IMO is pretty simple -- users shouldn't do that. :) Documentation is probably enough. 👍🏻 

I would like the other @open-telemetry/android-maintainers to chime in as well before we merge. Thanks again for seeing this through.

fractalwrench (approved):
LGTM and thanks for the contribution. I agree that documentation should be sufficient to mitigate the comment @breedx-splk made.

Cloud-Architect-Emma:
Thank you @breedx-splk and @fractalwrench for the approvals! 
Happy to address any remaining feedback from other maintainers 
if needed, otherwise ready to merge whenever convenient.

Cloud-Architect-Emma:
Hi @breedx-splk, the detekt check is now failing with 1 weighted 
issue after I added test coverage for the fullUrl=true paths. 

Could you help by running:
./gradlew :android-agent:apiDump :android-agent:detekt

and pushing the results to my branch? I'm unable to run the Android 
Gradle build locally without the Android SDK.

Thank you so much for your patience, really appreciate it!

Cloud-Architect-Emma:
Hi @breedx-splk, I reverted the additional test I added (which was 
causing a detekt failure I couldn't identify locally without the 
Android SDK). The PR is back to the state you approved.

Could you let me know which detekt rule the new test was violating? 
I'd like to add proper test coverage for the fullUrl=true paths but 
want to make sure it passes detekt. 

Thank you!

Cloud-Architect-Emma:
Hi @breedx-splk and @fractalwrench, the PR is back to the approved 
state with all required checks passing. The only non-required check 
failing is Codecov coverage which reflects the new code paths added. 
Ready to merge whenever convenient!

PR #1727, opened 2026-05-07
Fix stale PR auto-close workflow

breedx-splk:
You can see in #1467 (for example) that the automation was removing and re-adding the stale tag, which prevented it from being closed. This was an unexpected/clumsy interaction between the needs-author-feedback workflow and the stale workflow.


fractalwrench (approved):
LGTM

PR #1729, opened 2026-05-08
Remove zero duration span from slow rendering instrumentation.

breedx-splk:
We deprecated the old zero-duration span compatibility 8+ months ago, so it's long overdue for removal. Events are the way.


fractalwrench (approved):
LGTM

LikeTheSalad (approved):

PR #1730, opened 2026-05-08
[demo-app] Remove zero duration span

breedx-splk:
Let's not encourage users to generate zero-duration spans for things that are better represented as events.


demo-app/src/main/java/io/opentelemetry/android/demo/DemoViewModel.kt
```diff
@@ -27,12 +26,4 @@ class DemoViewModel : ViewModel() {
     private fun updateSession() {
         // TODO
     }
-
-    private fun sendTrace(
```
breedx-splk, line 31: this wasn't being used anywhere anymore.

LikeTheSalad (approved):

PR #1731, opened 2026-05-08
Remove deprecated sdk ready listeners.

breedx-splk:
Users should migrate to the otel ready listeners instead, and use the api instead of sdk.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### ⚠️⚠️ Breaking changes
+
+- Remove deprecated sdk ready listeners. Users should use the otel ready listeners instead.
+  ([#1731](https://github.com/open-telemetry/opentelemetry-android/pull/1731)) 
+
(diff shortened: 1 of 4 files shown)


fractalwrench (approved):
LGTM. Do we have a policy for when deprecated symbols are removed from the SDK? Does it need to be in a minor or major version bump?

LikeTheSalad (approved):

breedx-splk:
> LGTM. Do we have a policy for when deprecated symbols are removed from the SDK? Does it need to be in a minor or major version bump?

I don't think we have a stated policy for this repo. In general, otel will try and not introduce breaking changes, and we try and be "stable by default". The `core` APIs aren't yet stable, so hopefully users aren't too surprised, and it's why we did at least one release of deprecation -- to send a strong signal that a change will happen in the near future.

I think that granular tracking of every non-stable api change would be too much of a maintenance burden. 

You can read lots of words about stability from the otel perspective [here](https://opentelemetry.io/docs/specs/otel/versioning-and-stability/). :)

PR #1736, opened 2026-05-12
Retool the scheduled work service and from-disk exporter

breedx-splk:
So I started looking into #947 but got distracted by the uncomfortable structure of the periodic work service and the various handler classes. This seems like a big change, but it's mostly some naming changes. I'll try and summarize what matters:

* To help make this transition cleaner, I worked in a new package called `periodic` that replaces the `periodicwork` package.
* The `PeriodicWork` service api is now `PeriodicTaskScheduler`.
* The `PeriodicTaskScheduler` starts a `PeriodicRunnable` that is runnable, stoppable, and also provides its period as a `Duration`.
* The implementation of the service now uses coroutines instead of the fairly complicated thread management that existed previously.
* The `ExportScheduleHandler` and the specific implementation were renamed, hopefully for clarity.
 
I tried to remove use of the suffix `Handler` because I think it's too vague and doesn't provide a ton of value. 

I think this improves the groundwork to start addressing #947, which, as it turns out, is quite a bit more complicated than it seems -- due to the fact that disk buffering has some internal lifecycle durations for files, in addition to the periodic disk scanning in this PR.


services/src/main/java/io/opentelemetry/android/internal/services/periodic/PeriodicTaskSchedulerImpl.kt
```diff
@@ -0,0 +1,25 @@
...
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+
+class PeriodicTaskSchedulerImpl : PeriodicTaskScheduler {
+    private val job = SupervisorJob()
+    private val scope = CoroutineScope(job + Dispatchers.Default)
+
+    override fun start(runnable: PeriodicRunnable) {
+        scope.launch {
```
LikeTheSalad, line 18: I think we should store the result Job from this call to being able to cancel it before this line, should `start` is called twice, otherwise we'll get 2 coroutines running.

services/src/main/java/io/opentelemetry/android/internal/services/periodic/PeriodicTaskSchedulerImpl.kt
```diff
@@ -0,0 +1,25 @@
...
+            while (!runnable.shouldStop()) {
+                runnable.run()
```
LikeTheSalad: The SupervisorJob is a good call in that it will prevent the scope from getting killed should this line throws an error, but it won't prevent the job running this loop from getting terminated. So we should wrap it in a try/catch block to keep the loop.

services/src/main/java/io/opentelemetry/android/internal/services/periodic/PeriodicTaskSchedulerImpl.kt
```diff
@@ -0,0 +1,25 @@
...
+
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+
+class PeriodicTaskSchedulerImpl : PeriodicTaskScheduler {
+    private val job = SupervisorJob()
+    private val scope = CoroutineScope(job + Dispatchers.Default)
```
LikeTheSalad: It's usually a good practice to provide the dispatcher via a constructor parameter and avoid hardcoding one, this makes it easier to write tests.

LikeTheSalad (commented):
Looks good, I just added some comments on the coroutine usage.

Also, regarding:

> is quite a bit more complicated than it seems -- due to the fact that disk buffering has some internal lifecycle durations for files

The internal lifecycle durations are configurable, and the OTel Android polling times are fully handled on the OTel Android side, so it's probably just a matter of making the value from [this line](https://github.com/open-telemetry/opentelemetry-android/pull/1736/changes#diff-1f855dc11e1623a0814ef0bbc2954107b71848c917442e3e9e19de3a371c3061R13) configurable, to being able to address the issue you referenced in the description.

core/src/main/java/io/opentelemetry/android/features/diskbuffering/scheduler/DiskBufferingEnablement.kt
```diff
@@ -0,0 +1,34 @@
...
+import io.opentelemetry.android.internal.services.periodic.PeriodicTaskScheduler
+import java.util.concurrent.atomic.AtomicBoolean
+
+class DiskBufferingEnablement(
+    private val signalFromDiskExporter: SignalFromDiskExporter,
+    private val taskScheduler: PeriodicTaskScheduler,
+) : ScheduleEnablement {
+    private val enabled = AtomicBoolean(false)
+    @Volatile
+    private var periodicExporter: PeriodicExporter? = null
```
fractalwrench: nit: is there potential to use `AtomicReference<PeriodicExporter?>` instead of two separate properties?

fractalwrench (commented):
The renames look good from my POV. If those are the non-controversial parts of this changeset, would it be worth extracting them to a separate PR so that they can get merged down quicker & also reduce the diff?

core/src/main/java/io/opentelemetry/android/features/diskbuffering/scheduler/DiskBufferingEnablement.kt
breedx-splk: Yeah I like that and think it's marginally cleaner. Will make that change.

services/src/main/java/io/opentelemetry/android/internal/services/periodic/PeriodicTaskSchedulerImpl.kt
breedx-splk: Thanks yeah good idea. Did that, and leveraged some coroutine library to not do realtime/sleepy testing...which I despise anyway.

services/src/main/java/io/opentelemetry/android/internal/services/periodic/PeriodicTaskSchedulerImpl.kt
breedx-splk: Yeah, this is a really good point. I've made that change. 👍🏻

services/src/main/java/io/opentelemetry/android/internal/services/periodic/PeriodicTaskSchedulerImpl.kt
breedx-splk: I think that's ok, and it's actually part of the current design. Do you think allowing multiple jobs to run within the same coroutine scope is a problem? I figured that's something we might not use right now but might want to be able to leverage it, so I didn't ever consider guarding _against_ it...

breedx-splk:
> The renames look good from my POV. If those are the non-controversial parts of this changeset, would it be worth extracting them to a separate PR so that they can get merged down quicker & also reduce the diff?

I hear ya, but unfortunately in this particular case the renames also come with some restructuring, so it's not something that can just be done cleanly as a first step. Hopefully this isn't too bad and not a deal breaker.

services/src/main/java/io/opentelemetry/android/internal/services/periodic/PeriodicTaskSchedulerImpl.kt
LikeTheSalad: > Do you think allowing multiple jobs to run within the same coroutine scope is a problem?

I think it's more about running several coroutines within the same scope, which will run in parallel given that we use the default dispatcher, and all I know is that concurrency tends to make things fun :) so I'd recommend setting it up properly when/if we need it. It's not a blocker, though, because iirc we don't expose this type to users, so if there's only one place we call it, it should be safe.

LikeTheSalad (approved):
Cheers!

fractalwrench (approved):
LGTM

PR #1741, opened 2026-05-13
Added long press to click detection

DavidGrath:
References #1576 

@thompson-tomo , are these new attributes fine by you?

### DIFF
--- a/instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
+++ b/instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
@@ -11,11 +11,13 @@ import android.view.View
 import android.view.ViewGroup
 import android.view.Window
 import io.opentelemetry.android.instrumentation.view.click.internal.APP_SCREEN_CLICK_EVENT_NAME
+import io.opentelemetry.android.instrumentation.view.click.internal.APP_SCREEN_LONG_PRESS_EVENT_NAME
 import io.opentelemetry.android.instrumentation.view.click.internal.HARDWARE_POINTER_BUTTON
 import io.opentelemetry.android.instrumentation.view.click.internal.HARDWARE_POINTER_CLICKS
 import io.opentelemetry.android.instrumentation.view.click.internal.HARDWARE_POINTER_TYPE
 import io.opentelemetry.android.instrumentation.view.click.internal.TapEvent
 import io.opentelemetry.android.instrumentation.view.click.internal.VIEW_CLICK_EVENT_NAME
+import io.opentelemetry.android.instrumentation.view.click.internal.VIEW_LONG_PRESS_EVENT_NAME
 import io.opentelemetry.api.common.Attributes
 import io.opentelemetry.api.logs.LogRecordBuilder
 import io.opentelemetry.api.logs.Logger
@@ -74,6 +76,24 @@ internal class ViewClickEventGenerator(
             }
             return false
         }
+
+        override fun onLongPress(e: MotionEvent) {
+            windowRef?.get()?.let { window ->
+
+                val tapEvent = TapEvent(e)
+
+                createEvent(APP_SCREEN_LONG_PRESS_EVENT_NAME, tapEvent)
+                    .setAttribute(APP_SCREEN_COORDINATE_Y, e.y.toLong())
+                    .setAttribute(APP_SCREEN_COORDINATE_X, e.x.toLong())
+                    .emit()
+
+                findTargetForTap(window.decorView, e.x, e.y)?.let { view ->
+                    createEvent(VIEW_LONG_PRESS_EVENT_NAME, tapEvent)
+                        .setAllAttributes(createViewAttributes(view))
+                        .emit()
+                }
+            }
+        }
     }
 
     val gestureDetector = GestureDetector(null, gestureListener)
@@ -101,13 +121,24 @@ internal class ViewClickEventGenerator(
         windowRef = null
     }
 
-    private fun createEvent(name: String, tapEvent: TapEvent, clicks: Int): LogRecordBuilder =
-        eventLogger
+    /**
+     * @param clicks 0 for other event, 1 for single tap, 2 for double tap
+     */
+    private fun createEvent(name: String, tapEvent: TapEvent, clicks: Int = 0): LogRecordBuilder {
+
+        val logger = eventLogger
             .logRecordBuilder()
             .setEventName(name)
-            .setAttribute(HARDWARE_POINTER_CLICKS, clicks)
             .setAttribute(HARDWARE_POINTER_TYPE, tapEvent.toolTypeDescription)
             .setAttribute(HARDWARE_POINTER_BUTTON, tapEvent.buttonStateDescription)
+
+        val isTapEvent = clicks > 0
+        if(isTapEvent) {
+            logger.setAttribute(HARDWARE_POINTER_CLICKS, clicks)
+        }
+        return logger
+    }
+
 
     private fun createViewAttributes(view: View): Attributes {
         val builder = Attributes.builder()
--- a/instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
+++ b/instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
@@ -12,10 +12,7 @@ import android.view.ViewGroup
 import android.view.Window
 import io.opentelemetry.android.instrumentation.view.click.internal.APP_SCREEN_CLICK_EVENT_NAME
 import io.opentelemetry.android.instrumentation.view.click.internal.APP_SCREEN_LONG_PRESS_EVENT_NAME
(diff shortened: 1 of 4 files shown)


instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
```diff
@@ -101,14 +121,25 @@ internal class ViewClickEventGenerator(
...
+
+        val logger = eventLogger
             .logRecordBuilder()
             .setEventName(name)
-            .setAttribute(HARDWARE_POINTER_CLICKS, clicks)
             .setAttribute(HARDWARE_POINTER_TYPE, tapEvent.toolTypeDescription)
             .setAttribute(HARDWARE_POINTER_BUTTON, tapEvent.buttonStateDescription)
 
+        val isTapEvent = clicks > 0
+        if(isTapEvent) {
```
fractalwrench:
```suggestion
        if (isTapEvent) {
```

instrumentation/view-click/src/test/kotlin/TestUtils.kt
```diff
@@ -161,6 +161,50 @@ fun getSingleTapSequence(x: Float, y: Float, toolType: Int = MotionEvent.TOOL_TY
...
 }
 
+fun getLongPressSequence(x: Float, y: Float, toolType: Int = MotionEvent.TOOL_TYPE_FINGER, buttonState: Int = 0,
+                         stayDownLongEnough: Boolean = true)
+        : Array<MotionEvent> {
+    require(toolType in allowedToolTypes) {
+        "Invalid tool type"
+    }
+
+    if(buttonState != 0) {
```
fractalwrench:
```suggestion
    if (buttonState != 0) {
```

instrumentation/view-click/src/test/kotlin/TestUtils.kt
```diff
@@ -161,6 +161,50 @@ fun getSingleTapSequence(x: Float, y: Float, toolType: Int = MotionEvent.TOOL_TY
...
+
+    val initialTime = SystemClock.uptimeMillis()
+
+    val pointerProperties = MotionEvent.PointerProperties()
+    pointerProperties.id = 0
+    pointerProperties.toolType = toolType
+
+    val pointerCoords = PointerCoordsBuilder.newBuilder().setCoords(x, y).build()
+    val delay = if(stayDownLongEnough) {
+//        val allowanceTime = 100L
```
fractalwrench:
```suggestion
```

instrumentation/view-click/src/test/kotlin/TestUtils.kt
```diff
@@ -161,6 +161,50 @@ fun getSingleTapSequence(x: Float, y: Float, toolType: Int = MotionEvent.TOOL_TY
...
+    }
+
+    val initialTime = SystemClock.uptimeMillis()
+
+    val pointerProperties = MotionEvent.PointerProperties()
+    pointerProperties.id = 0
+    pointerProperties.toolType = toolType
+
+    val pointerCoords = PointerCoordsBuilder.newBuilder().setCoords(x, y).build()
+    val delay = if(stayDownLongEnough) {
```
fractalwrench:
```suggestion
    val delay = if (stayDownLongEnough) {
```

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
```diff
@@ -101,14 +121,25 @@ internal class ViewClickEventGenerator(
         windowRef = null
     }
 
-    private fun createEvent(name: String, tapEvent: TapEvent, clicks: Int): LogRecordBuilder =
-        eventLogger
+    /**
+     * @param clicks 0 for other event, 1 for single tap, 2 for double tap
+     */
+    private fun createEvent(name: String, tapEvent: TapEvent, clicks: Int = 0): LogRecordBuilder {
```
fractalwrench: This might be overkill, but I wonder whether it's worth creating an enum that holds `TapEvent` and `clicks` and returns the attributes that should be used for each event. IMO it could avoid logic such as `if(isTapEvent)` growing if we continue to add more touch events to the instrumentation.

DavidGrath: Would it be something like this?

```kotlin
//TapEvent renamed to TapEventMetadata
sealed class Gesture(
    val tapEventMetadata: TapEventMetadata
) {
    class LongPress(val t: TapEvent): Gesture(t)
    class Click(val t: TapEvent, clicks: Int): Gesture(t)
    class Scale(val t: TapEvent, val scaleFactor: Double): Gesture(t)
    class Scroll(val t: TapEvent, val velocityX: Double, val velocityY: Double): Gesture(t)
}
```

fractalwrench: Yep, something along those lines. `Gesture` could even define `val attributes: Map<String, Any>` and generate the attributes that are required for each event.

DavidGrath: Noted. I'll apply these changes over the weekend

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/internal/ViewUtils.kt
```diff
@@ -58,3 +61,21 @@ internal class TapEvent(
     }
 
 }
+
+internal sealed class Gesture(val m: MotionEvent) {
+    private val t = TapEventMetadata(m)
+    val attributeBuilder = Attributes.builder()
```
DavidGrath: @fractalwrench , so is this alright or should I stick with a plain `Map` instead of `Attributes`? I feel the dangling builder object is something I might need to walk back on

fractalwrench: I would lean slightly to just returning `Attributes` rather than a builder but I don't think that needs to block.

fractalwrench (approved):
LGTM

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/internal/ViewUtils.kt
breedx-splk: I don't think there's much value in holding a mutable object in an immutable `val`, versus just having a mutable private `var`. So, my recommendation is to just have 

`var attributes = Attributes.empty()`

...and then in the 3 places where you need to update it, you can do 

`attributes = attributes.toBuilder().put(xxx, yyy).build()`.

instrumentation/view-click/src/test/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewLongPressInstrumentationTest.kt
```diff
@@ -0,0 +1,277 @@
...
+            }
+
+        every { window.decorView } returns mockViewGroup
+        viewClickActivityCallback.onActivityResumed(activity)
+
+        verify {
+            window.callback = capture(wrapperCapturingSlot)
+        }
+
+
```
breedx-splk:
```suggestion

```

breedx-splk (approved):
Had one small feedback about the attributes builder, but otherwise looking good. Thanks for the contribution!

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/internal/ViewUtils.kt
DavidGrath: Okay, done. Thanks for the review

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/internal/ViewUtils.kt
```diff
@@ -58,3 +61,21 @@ internal class TapEvent(
     }
 
 }
+
+internal sealed class Gesture(val m: MotionEvent) {
```
marandaneto, line 65: sharing as context
eg https://github.com/PostHog/posthog-android/blob/d2cc7f7a9bb715a2902b454e33bc4912cba585b9/posthog-android/src/main/java/com/posthog/android/replay/PostHogReplayIntegration.kt#L303-L323
i've seen errors where acessing MotionEvent may trigger native errors, so its better to call `MotionEvent.obtain` and later `recycle`
this was the case if accessing or processing in a background thread, not sure if its the case here

DavidGrath: If the change is to simply use an `obtain` to duplicate it, then I think I can do that. I'd need to take my time to look at your own code so I can reproduce a test case for it, too, so it'll be a bit

LikeTheSalad (approved):
Thank you! 🙏

PR #1749, opened 2026-05-18
Treat empty fullUrl as null/missing

breedx-splk:
Minor improvement after #1723, which slightly changed how an empty configuration string would be handled.

Prior to #1723 we would check for empty before falling back. This same approach needs to also happen with the fullUrl, so that if a user configures an empty string as a fullUrl, we still just use the locally scoped url.


fractalwrench (approved):
LGTM

LikeTheSalad (approved):

PR #1763, opened 2026-05-20
Move constants from `RumConstants`

fractalwrench:
## Goal

Moves several constants out of `RumConstants` into the instrumentation where they are used. This does leave a couple of constants behind in `RumConstants` that are used across several modules. I figured we could address those separately from the easy ones that were only referenced from one module.


instrumentation/startup/api/startup.api
```diff
@@ -1,4 +1,11 @@
 public final class io/opentelemetry/android/instrumentation/startup/SdkInitializationEvents : io/opentelemetry/android/internal/initialization/InitializationEvents {
+	public static final field INIT_EVENT_ANR_MONITOR Ljava/lang/String;
```
LikeTheSalad: Do we need to make these public?

fractalwrench: @LikeTheSalad I can make `SdkInitializationEvents` non-public if you'd prefer? Right now the fields are technically public because `SdkInitializationEvents` is. I don't see that symbol being used anywhere by code within the SDK..

LikeTheSalad: Got it. It's just that I wouldn't like to increase the api surface unless we know it's needed. Regarding making `SdkInitializationEvents` non-public, it probably is not ideal given that it's marked as an auto service. What if we add these to a new, internal type within the startup module?

fractalwrench: Moved these constants to a new internal object.

LikeTheSalad (approved):
Thank you!

breedx-splk (approved):
Thanks for helping reduce the public surface area of the `common` module.  I'm in favor of proceeding with this, but I do want to call out one downside: This spreads around (distributes) these bespoke semantic conventions in a way that makes it harder to find the list of all custom names/constants that we're using.

Ideally, all of these will one day live in semantic conventions, and so these will go away entirely. Until then, though, this change makes it slightly more challenging to identify/locate the things we haven't yet brought into semconv. I think it's manageable tho, so let's do this.

PR #1767, opened 2026-05-21
Refactor ServicesFactory

fractalwrench:
## Goal

Refactors `ServicesFactory` so that `CacheStorage` and `PeriodicTaskScheduler` are moved to `core` and made an internal implementation detail, as this is the only module where they were invoked apart from in test code. I've additionally tried to move away from using the static `Services.get()` as that complicates injection of fakes/mocks. The remainder of properties left on `ServicesFactory` are harder to change without altering the instrumentation API.


LikeTheSalad (approved):
This is great, cheers!

core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.kt
```diff
@@ -305,8 +331,15 @@ class OpenTelemetryRumBuilder internal constructor(
      */
     fun build(): OpenTelemetryRum {
         val services = Services.get(context)
+        val cacheStorage: Lazy<CacheStorage> = lazy { cacheStorageOverride ?: CacheStorageImpl(context) }
+        val periodicTaskScheduler: Lazy<PeriodicTaskScheduler> =
+            lazy { periodicTaskSchedulerOverride ?: PeriodicTaskSchedulerImpl() }
```
breedx-splk, line 336: I think there's a little room for improvement here -- it's a tad confusing to have two fields and two local variables, and then we pass the local vars into other methods. Why not just remove the `override` suffix from the fields, ensure they're created if not assigned near the start of `build()` (as it exists here), and then just reference the fields from the class instance and don't pass them around.

breedx-splk: ...also this comment shouldn't be blocking. We can pick that up in a follow-up enhancement PR.

breedx-splk (approved):
Cool stuff, thanks! I think we can proceed with this, and I'll follow up the merge with a PR to address my comment. Cheers.

PR #1768, opened 2026-05-22
Extract CurrentNetworkProvider from ServicesFactory

fractalwrench:
## Goal

Extracts `CurrentNetworkProvider` from `ServicesFactory`. This follows on from #1767 by attempting to extract one of the services that instrumentation depends on via static calls and isn't easy to inject due to SPI. I've therefore created `NetworkProviderHolder` which does that job, and injected `NetworkProvider` as a parameter where possible.


breedx-splk:
Closing due to #1781. Thanks again @fractalwrench !

PR #1771, opened 2026-05-26
OpenTelemetryRumBuilder - Use lateinit vars instead of nullable/lazy.

breedx-splk:
This is a follow-up to [this comment](https://github.com/open-telemetry/opentelemetry-android/pull/1767#discussion_r3305537466), which was part of #1767.

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.kt
+++ b/core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.kt
@@ -122,8 +122,8 @@ class OpenTelemetryRumBuilder internal constructor(
     private var resource: Resource = createDefault(context)
     private var exportScheduleEnablement: ScheduleEnablement? = null
     private var sessionProvider: SessionProvider = SessionProvider.getNoop()
-    private var cacheStorageOverride: CacheStorage? = null
-    private var periodicTaskSchedulerOverride: PeriodicTaskScheduler? = null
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.kt
```diff
@@ -331,9 +331,8 @@ class OpenTelemetryRumBuilder internal constructor(
      */
     fun build(): OpenTelemetryRum {
         val services = Services.get(context)
-        val cacheStorage: Lazy<CacheStorage> = lazy { cacheStorageOverride ?: CacheStorageImpl(context) }
-        val periodicTaskScheduler: Lazy<PeriodicTaskScheduler> =
-            lazy { periodicTaskSchedulerOverride ?: PeriodicTaskSchedulerImpl() }
+        cacheStorage = if (::cacheStorage.isInitialized) cacheStorage else CacheStorageImpl(context)
```
LikeTheSalad, line 334: In my opinion, the usage of `::lateinitvar.isInitialized` is a clear hint that said var must be nullable. Unless we're certain that those lateinit vars shall be set before they're referenced anywhere in the code, we should keep them nullable; otherwise, we'll be introducing a risk of getting a runtime exception, which defeats the purpose of lateinit var. Nullable vars are safer to use for these kinds of scenarios.

breedx-splk:
not that important, just feeling it out.

PR #1773, opened 2026-05-26
Use kotlin published semconv instead of java

breedx-splk:
I'm not 100% sure we're ready for this, but I wanted to see what it would look like to start using the semconv that are published from the [kotlin project](https://github.com/open-telemetry/opentelemetry-kotlin). Sorry that this (obviously) touches a lot of files, but most of them are just import changes.

There are two things for the reviewer to note:

1. Kotlin doesn't currently publish separate artifacts nor use packages for incubating attributes. Instead, it uses an `@IncubatingApi` annotation.
2. Kotlin attribute keys are not typed like Java's `AttributeKey<T>` classes, they are simply strings. This means that several places require the use of `stringKey()` or `longKey()` to get the type info back into place. This may change eventually when this android project uses the kotlin api instead of the java api.

### DIFF
--- a/android-agent/build.gradle.kts
+++ b/android-agent/build.gradle.kts
@@ -30,8 +30,7 @@ dependencies {
     implementation(project(":instrumentation:sessions"))
     implementation(project(":instrumentation:screen-orientation"))
 
-    testImplementation(libs.opentelemetry.semconv)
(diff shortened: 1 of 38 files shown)


bidetofevil (approved):
Hell yeah lets do it

LikeTheSalad (approved):

PR #1780, opened 2026-06-02
Added scroll and fling to view-click. References #1576

DavidGrath:
Adds scroll and fling. Adds suggestions from https://github.com/open-telemetry/opentelemetry-android/pull/1741#discussion_r3273710893 by @marandaneto 

At this point, I think the instrumentation has gone beyond the name of "click". I'd like to ask how to address the nomenclature:
* Move the non-click code to a new "gesture" instrumentation?
* Rename the existing module instead?

### DIFF
--- a/instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
+++ b/instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
@@ -11,10 +11,14 @@ import android.view.View
 import android.view.ViewGroup
 import android.view.Window
 import io.opentelemetry.android.instrumentation.view.click.internal.APP_SCREEN_CLICK_EVENT_NAME
+import io.opentelemetry.android.instrumentation.view.click.internal.APP_SCREEN_FLING_EVENT_NAME
 import io.opentelemetry.android.instrumentation.view.click.internal.APP_SCREEN_LONG_PRESS_EVENT_NAME
+import io.opentelemetry.android.instrumentation.view.click.internal.APP_SCREEN_SCROLL_EVENT_NAME
 import io.opentelemetry.android.instrumentation.view.click.internal.Gesture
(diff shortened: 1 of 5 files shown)


DavidGrath:
I'm not sure exactly why the x and y distances are negated. I assumed it would just be the vertical axis because of how the coordinate system works compared to a regular Cartesian plane with the y axis flipped. But yeah, I'll leave it like that

instrumentation/view-click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
```diff
@@ -38,59 +42,110 @@ internal class ViewClickEventGenerator(
         override fun onDoubleTap(motionEvent: MotionEvent): Boolean {
             windowRef?.get()?.let { window ->
 
-
+                val safeEvent = MotionEvent.obtain(motionEvent)
```
breedx-splk, line 45: It's not immediately clear to me how/why this is "safer" than just using the non-nullable parameter `motionEvent` as before. Is it mutable or something or ?

breedx-splk: Oh nevermind, I think I answered my own question: The event can be `.recycle()`d while in use, so we make a copy that we then recycle below. Makes sense. A comment could have helped a little for those of us that don't know the platform well enough. 😅

breedx-splk (approved):
This is great! Thanks @DavidGrath .

breedx-splk:
@DavidGrath did you want to repackage this instrumentation before merge, or would you prefer to do that in a follow-up effort/PR?

DavidGrath:
@breedx-splk , in a follow up, so that I can add the scaling gesture as well

breedx-splk:
> @breedx-splk , in a follow up, so that I can add the scaling gesture as well

That works for me. Please open a tracking issue in the meantime. 👍🏻

PR #1783, opened 2026-06-02
Introduce InstrumentationConfigurator

breedx-splk:
Currently, we have a limitation in that `AutoService` instrumentations that are automatically found and loaded on the classpath cannot really be customized. The service loader just loads them and then immediately calls install. This limits users ability to tweak specific instrumentations to their needs.

So this PR adds a new interface called `InstrumentationConfigurator`. Implementations of this are found on the classpath, loaded by the service loader, and then they are invoked with the instrumentation before `install()` is called. The instances have a name field, and this name field allows the configurator to be linked/matched to the correct instrumentation.

We already have the DSL, but we cannot create type-specific DSL wrappers for 3rd party or other user-specific instrumentations. We should expect to see 1st party instrumentation libraries, and we want there to be a nice, uniform way of configuring these. 

If we like this approach, then later on we could consider adding support for `InstrumentationConfigurator` in the DSL as well.

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/instrumentation/InstrumentationConfigurators.kt
+++ b/core/src/main/java/io/opentelemetry/android/instrumentation/InstrumentationConfigurators.kt
@@ -2,7 +2,7 @@ package io.opentelemetry.android.instrumentation
 
 import java.util.ServiceLoader
 
-private typealias Konfigurator = InstrumentationConfigurator<AndroidInstrumentation>
+private typealias Konfigurator = InstrumentationConfigurator<*>
 
 internal class InstrumentationConfigurators
     private constructor(private val configurators: Map<String, List<Konfigurator>>) {
@@ -13,22 +13,23 @@ internal class InstrumentationConfigurators
         }
 
         // Exists for testing
-        @Suppress("UNCHECKED_CAST")
-        fun create(loader: Loader<Konfigurator>): InstrumentationConfigurators {
+        fun create(loader: Loader): InstrumentationConfigurators {
             val configurators = loader.load(Konfigurator::class.java)
                 .groupBy { it.instrumentationName }
             return InstrumentationConfigurators(configurators)
         }
     }
 
-    fun configure(instrumentation: AndroidInstrumentation){
-        configurators[instrumentation.name]?.forEach { it.configure(instrumentation) }
+    @Suppress("UNCHECKED_CAST")
+    fun configure(instrumentation: AndroidInstrumentation) {
+        configurators[instrumentation.name]
+            ?.forEach { (it as InstrumentationConfigurator<AndroidInstrumentation>).configure(instrumentation) }
     }
 }
 
 /**
  * Exists for testing
  */
-internal fun interface Loader<S: Konfigurator> {
-    fun load(service: Class<S>): Iterable<S>
+internal fun interface Loader {
+    fun load(service: Class<Konfigurator>): Iterable<Konfigurator>
 }
--- a/core/src/main/java/io/opentelemetry/android/instrumentation/InstrumentationConfigurators.kt
+++ b/core/src/main/java/io/opentelemetry/android/instrumentation/InstrumentationConfigurators.kt
@@ -1,0 +1,23 @@
+package io.opentelemetry.android.instrumentation
+
+import java.util.ServiceLoader
+
+internal class InstrumentationConfigurators
+    private constructor(
+        private val configurators: Map<String, List<InstrumentationConfigurator<AndroidInstrumentation>>>,
+    ) {
+
+    companion object {
+        @Suppress("UNCHECKED_CAST")
+        fun create(): InstrumentationConfigurators {
+            val configurators = ServiceLoader
+                .load(InstrumentationConfigurator::class.java)
+                .groupBy { it.instrumentationName }
+            return InstrumentationConfigurators(configurators as Map<String, List<InstrumentationConfigurator<AndroidInstrumentation>>>)
+        }
+    }
+
+    fun configure(instrumentation: AndroidInstrumentation){
+        configurators[instrumentation.name]?.forEach { it.configure(instrumentation) }
+    }
+}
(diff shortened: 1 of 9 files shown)


DavidGrath (commented):
Looks great. It also seems like a potential first step towards declarative, if I'm not mistaken. I'll wait till the SIG meeting for further discussion

instrumentation/android-instrumentation/src/main/java/io/opentelemetry/android/instrumentation/InstrumentationConfigurator.kt
```diff
@@ -0,0 +1,20 @@
...
+    /**
+     * The [AndroidInstrumentation.name] value of the instrumentation this configurator applies to.
+     * This name must exactly match the name returned by the AndroidInstrumentation.name.
+     */
+    val instrumentationName: String
+
+    /**
+     * Called to configure an [instrumentation] before [AndroidInstrumentation.install] is called.
+     */
+    fun configure(instrumentation: T)
```
fractalwrench, line 22: Are there limitations that prevent us from putting `configure()` on `AndroidInstrumentation`? That would avoid requiring two separate interfaces that are coupled by `name/instrumentationName`.

breedx-splk: > Are there limitations that prevent us from putting configure() on AndroidInstrumentation?

No, not really, but I think it doesn't solve this specific use case -- which is being able to configure/customize instrumentation modules before `install()` is called. Even if `configure()` existed on each module, who could call it? The idea is to be able to configure instrumentations from non-agent code (in user code).

We still need some form of SPI (`@AutoService`) to allow developers to customize instrumentations at runtime, before they are installed.

breedx-splk: I think I understand the hesitation around coupling through the name -- but really, names don't change that often. We have 5+ years of prior art with java instrumentation modules, and configuration SPIs that use their names...and I can't think of this having been a problem really. 🤷🏻

fractalwrench: Ah, I think I understand now. I'd been assuming that developers would always be writing their own `AndroidInstrumentation` implementations and could therefore control how they were configured, but that's not necessarily true.

fractalwrench (approved):
LGTM

core/src/main/java/io/opentelemetry/android/instrumentation/InstrumentationConfigurators.kt
```diff
@@ -0,0 +1,35 @@
...
+        fun create(): InstrumentationConfigurators {
+            return create { s -> ServiceLoader.load(s) }
+        }
+
+        // Exists for testing
+        fun create(loader: Loader): InstrumentationConfigurators {
+            val configurators = loader.load(Konfigurator::class.java)
+                .groupBy { it.instrumentationName }
+            return InstrumentationConfigurators(configurators)
+        }
```
lemnik: Given that the second `create` is for testing (where the `service` `Class` in the `Loader` goes unused), would making it `inline` possibly be a bit cleaner?

```suggestion
        fun create(): InstrumentationConfigurators {
            return create { ServiceLoader.load(Konfigurator::class.java) }
        }

        // Exists for testing
        inline fun create(loader: () -> Iterable<Konfigurator>): InstrumentationConfigurators {
            val configurators = loader().groupBy { it.instrumentationName }
            return InstrumentationConfigurators(configurators)
        }
```

docs/WRITING_INSTRUMENTATION.md
```diff
@@ -131,6 +131,34 @@ Configuration must happen before `OpenTelemetryRum` is built. The loader discove
...
+invoked prior to installation. A configurator is matched to an instrumentation by `instrumentationName`, 
+which must equal the target instrumentation's `AndroidInstrumentation.name`. There can be multiple
+configurators for any `AndroidInstrumentation`.
+
+Configurators must be annotated with `@AutoService`:
+
+```kotlin
+@AutoService(InstrumentationConfigurator::class)
+class MyFeatureConfigurator : InstrumentationConfigurator<SomeInstrumentation> {
+    override val instrumentationName: String = "some-instrumentation"
```
LikeTheSalad, line 150: This is not exactly a deal-breaker for me, but since we're already providing the instrumentation type when implementing `InstrumentationConfigurator`, can we use that type to map configuration to instrumentation instead of using the name?

breedx-splk: It would be nice, I agree...but due to type erasure of generics, I don't know of a great way of doing that. It's probably made worse by the ServiceLoader not really understanding/supplying generics.

Please let me know if you find a way.

core/src/main/java/io/opentelemetry/android/instrumentation/InstrumentationConfigurators.kt
breedx-splk: Yeah, thanks for the idea. I like that it gets rid of the `Loader` interface then too, so it's a little bit cleaner.

docs/WRITING_INSTRUMENTATION.md
LikeTheSalad: I see what you mean. The one alternative that comes to my mind would be to do something like this:

```kotlin
interface InstrumentationConfigurator<T : AndroidInstrumentation> {
    val instrumentationType: Class<T>
    // ...
}
```

Which still requires overriding something, just like with the name approach, but I think it's less error-prone in general. Still, the current state doesn't cause a huge concern (it can mostly be affected by typos), so I'll approve the PR in case the alternative I propose doesn't work for you.

LikeTheSalad (approved):
Cheers! I provided an alternative for the main interface [here](https://github.com/open-telemetry/opentelemetry-android/pull/1783#discussion_r3551569201) in case it helps.

docs/WRITING_INSTRUMENTATION.md
breedx-splk: Ok cool let's try this then....I think it's less fragile than the string/name. I wonder about inheritance, but I don't think it's any worse than the string situation. We can always change it.

Issue #1776, opened 2026-05-28
Removing unwanted exporters


brsgina:
Hi, 

If I would like to export only logs, can I turn off the span and metric exporters?
In the initialize function the span and metric exporters are always added, but I don't have an endpoint for that on my backend because I don't want to collect them. So right now I've got a 404 error, because exporting is continuously happening in the background at certain intervals. Can I turn off that somehow, or is it possible to make adding these exporters optional?

LikeTheSalad:
Hi @brsgina 

> can I turn off the span and metric exporters?

It is not possible at the moment; however, it should be possible to add support for this use case into the DSL, and from my perspective, it'd be fine to do so.

breedx-splk:
I agree that this is a nice feature to add.

@brsgina I'm not aware of anything today that is generating metrics on Android, so I'm a little surprised to hear that you were seeing 404s for these. Can you help us understand what component was generating these metrics? Maybe you can temporarily set up a logging collector to help troubleshoot? Also, for the unwanted Spans, can you simply disable those instrumentations that generate them? If not, which specific instrumentations are causing you these problems?

Regardless, some noop implementations may be possible with some code changes, but we should be sensitive to the subtle differences between "turn off the exporter" and "turn off the signal". Having a signal that does instrumentation, generates telemetry, and then just throws it away is wasteful. It would be better to not generate the telemetry in the first place.

There is no way to wire this up today, but I should point out that the spec does mandate the existence of the [NoOp metrics api](https://opentelemetry.io/docs/specs/otel/metrics/noop/). We should probably favor wiring that up over an exporter that drops generated metric data.

bidetofevil:
Create [this PR](https://github.com/open-telemetry/opentelemetry-android/pull/1785) to address this in the project without upstream changes. It requires wrapping the `OpenTelemetry` implementation created in the SDK and should be a decent solution until this is better supported in the SDK.

breedx-splk:
@brsgina I'm going to close this because I think #1785 will be a good fit for what you're asking for. This will allow you to disable signals entirely, without the need to disable exporters. 

Feel free to reopen if this does not suit your needs or if you want to continue brainstorming/discussing. Thanks for raising this!

brsgina:
Sorry for my late reply, yes this change was a perfect fit for that, thank you!

PR #1785, opened 2026-06-03
Allow each signal API to be thoroughly disabled

bidetofevil:
Add the ability to disable each signal API via the config DSL that provides a no-op provider instead of just not configuring any processors. This eliminates the overhead of instrumentation that are still configured but whose telemetry doesn't ever get recorded and sent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ private fun initOTel(context: Context): OpenTelemetryRum? =
                 globalAttributes {
                     Attributes.of(stringKey("demo-version"), "test")
                 }
-                disableMetrics(true)
+                disableMetrics()
             }
         )
     }.onFailure {
--- a/README.md
+++ b/README.md
@@ -87,6 +87,7 @@ private fun initOTel(context: Context): OpenTelemetryRum? =
                 globalAttributes {
                     Attributes.of(stringKey("demo-version"), "test")
                 }
+                disableMetrics(true)
             }
         )
     }.onFailure {
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
@@ -55,26 +55,36 @@ object OpenTelemetryRumInitializer {
             }
 
         val rumConfig = OtelRumConfig()
-        val builder = RumBuilder.builder(ctx, rumConfig)
-        val cfg = OpenTelemetryConfiguration(rumConfig, instrumentationLoader = builder.instrumentationLoader)
-        configuration(cfg)
+        return RumBuilder.builder(ctx, rumConfig).apply {
+            val cfg = OpenTelemetryConfiguration(
+                rumConfig = rumConfig,
+                instrumentationLoader = instrumentationLoader
+            ).also(configuration)
 
-        val spansEndpoint = cfg.exportConfig.spansEndpoint()
-        val logsEndpoints = cfg.exportConfig.logsEndpoint()
-        val metricsEndpoint = cfg.exportConfig.metricsEndpoint()
+            setSessionProvider(createSessionProvider(Services.get(ctx).appLifecycle, cfg))
+            setResource(
+                AndroidResource.createDefault(ctx).toBuilder().apply {
+                    cfg.resourceAction(this)
+                }.build()
+            )
+            setClock(cfg.clock)
 
-        val resourceBuilder = AndroidResource.createDefault(ctx).toBuilder()
-        cfg.resourceAction(resourceBuilder)
-        val resource = resourceBuilder.build()
-
-        return builder
-            .setSessionProvider(createSessionProvider(Services.get(ctx).appLifecycle, cfg))
-            .setResource(resource)
-            .setClock(cfg.clock)
-            .addSpanExporterCustomizer { createSpanExporter(spansEndpoint) }
-            .addLogRecordExporterCustomizer { createLogExporter(logsEndpoints) }
-            .addMetricExporterCustomizer { createMetricExporter(metricsEndpoint) }
-            .build()
+            if (rumConfig.tracingEnabled) {
+                addSpanExporterCustomizer {
+                    createSpanExporter(cfg.exportConfig.spansEndpoint())
+                }
+            }
+            if (rumConfig.loggingEnabled) {
+                addLogRecordExporterCustomizer {
+                    createLogExporter(cfg.exportConfig.logsEndpoint())
+                }
+            }
+            if (rumConfig.metricsEnabled) {
+                addMetricExporterCustomizer {
+                    createMetricExporter(cfg.exportConfig.metricsEndpoint())
+                }
+            }
+        }.build()
     }
 
     private fun createSpanExporter(endpoint: HttpEndpointConnectivity): SpanExporter =
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OpenTelemetryConfiguration.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OpenTelemetryConfiguration.kt
@@ -36,22 +36,22 @@ class OpenTelemetryConfiguration internal constructor(
     /**
      * Disable tracing in the SDK by providing no-op implementations that don't incur overhead even if instrumentation creates spans
      */
-    fun disableTracing(disable: Boolean) {
-        rumConfig.setTracingEnabled(!disable)
+    fun disableTracing() {
+        rumConfig.setTracingDisabled(false)
     }
 
     /**
-     * Disable logging in the SDK by providing no-op implementations that don't incur overhead even if instrumentation emits logs
+     * Disable logging and events in the SDK by providing no-op implementations that don't incur overhead even if instrumentation emits logs
      */
-    fun disableLogging(disable: Boolean) {
-        rumConfig.setLoggingEnabled(!disable)
+    fun disableLogging() {
+        rumConfig.setLoggingDisabled(false)
     }
 
     /**
      * Disable metrics in the SDK by providing no-op implementations that don't incur overhead even if instrumentation records metrics
      */
-    fun disableMetrics(disable: Boolean) {
-        rumConfig.setMetricsEnabled(!disable)
+    fun disableMetrics() {
+        rumConfig.setMetricsEnabled(false)
     }
 
     /**
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OpenTelemetryConfiguration.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OpenTelemetryConfiguration.kt
@@ -37,21 +37,21 @@ class OpenTelemetryConfiguration internal constructor(
      * Disable tracing in the SDK by providing no-op implementations that don't incur overhead even if instrumentation creates spans
      */
     fun disableTracing() {
-        rumConfig.setTracingDisabled(true)
+        rumConfig.setTracingDisabled()
     }
 
     /**
      * Disable logging and events in the SDK by providing no-op implementations that don't incur overhead even if instrumentation emits logs
      */
     fun disableLogging() {
-        rumConfig.setLoggingDisabled(true)
+        rumConfig.setLoggingDisabled()
     }
 
     /**
      * Disable metrics in the SDK by providing no-op implementations that don't incur overhead even if instrumentation records metrics
      */
     fun disableMetrics() {
-        rumConfig.setMetricsDisabled(true)
+        rumConfig.setMetricsDisabled()
     }
 
     /**
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OpenTelemetryConfiguration.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OpenTelemetryConfiguration.kt
@@ -37,21 +37,21 @@ class OpenTelemetryConfiguration internal constructor(
      * Disable tracing in the SDK by providing no-op implementations that don't incur overhead even if instrumentation creates spans
      */
     fun disableTracing() {
-        rumConfig.setTracingDisabled(false)
+        rumConfig.setTracingDisabled(true)
     }
 
     /**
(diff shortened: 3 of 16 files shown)


core/src/main/java/io/opentelemetry/android/config/OtelRumConfig.kt
```diff
@@ -24,6 +24,27 @@ class OtelRumConfig {
     private val suppressedInstrumentations: MutableList<String> = mutableListOf()
     private var diskBufferingConfigImpl: DiskBufferingConfig = create()
 
+    /**
+     * Whether the tracing API is enabled. Toggle with using [setTracingEnabled]
+     */
+    @Volatile
```
LikeTheSalad: Is there a need to make these volatile?

core/src/main/java/io/opentelemetry/android/DisableableOpenTelemetry.kt
```diff
@@ -0,0 +1,45 @@
...
+import io.opentelemetry.api.OpenTelemetry
+import io.opentelemetry.api.logs.LoggerProvider
+import io.opentelemetry.api.metrics.MeterProvider
+import io.opentelemetry.api.trace.TracerProvider
+import io.opentelemetry.context.propagation.ContextPropagators
+
+/**
+ * A wrapper around [delegate] that provides no-op implementations for the providers of each API that are disabled.
+ */
+internal class DisableableOpenTelemetry(
```
LikeTheSalad, line 17: Do we need this new type? Or is it possible to set the noop instances within the rum builder?

breedx-splk: It's a good question. I kinda like having this in one place, but we might be able to use *ProviderCustomizers to do the same thing...

core/src/main/java/io/opentelemetry/android/config/OtelRumConfig.kt
bidetofevil: I suppose its creation and consumption are all done on the main thread, and this object is thrown away after the SDK is initialized, so I can remove this if you feel it makes things slightly more convoluted.

core/src/main/java/io/opentelemetry/android/DisableableOpenTelemetry.kt
bidetofevil: I also want to do it in the creation of the original SDK instance in `OpenTelemetryRumBuilder`, but I need an `Sdk*Provider` there, while the noops are just `*Provider`. 

But are you saying we can use the customizer to wire the noops in even?

LikeTheSalad: I meant that, for example, if you wanted to provide a noop tracer provider, one place we could do so is here:

https://github.com/open-telemetry/opentelemetry-android/blob/71b575702ddbc9b588eb9725cd8a1b90e91dd9f2/core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.kt#L346

So the config check might be done there to decide whether to build the actual tracer provider or to just use `TracerProvider.noop()`.

core/src/main/java/io/opentelemetry/android/config/OtelRumConfig.kt
LikeTheSalad: Yeah. I don't see a reason why making it thread-safe, at least not now. Plus, if it were needed, then I think we should apply the same safety guards to the rest of the config options. But for now I think it's better to keep it simple.

core/src/main/java/io/opentelemetry/android/DisableableOpenTelemetry.kt
bidetofevil: So `TraceProvider.noop()` returns `TraceProvider` but `setTracerProvider()` expects `SdkTracerProvider` -- that's the crux of the problem. And not calling it creates a default tracer which isn't a no-op, but simply doesn't export because it has no processor.

LikeTheSalad: > `setTracerProvider()` expects `SdkTracerProvider`

Wow. That's odd. I didn't get it at first, but I understand what you meant now. Thanks for the clarification. Let's leave it as is then.

bidetofevil: I started this thinking that was the only place I'd be changing too, but didn't realize what I needed to do until I got in there 😅

breedx-splk:
Related to #1776

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OpenTelemetryConfiguration.kt
```diff
@@ -33,6 +33,27 @@ class OpenTelemetryConfiguration internal constructor(
     internal val instrumentations = InstrumentationConfiguration(rumConfig, instrumentationLoader)
     internal var resourceAction: ResourceBuilder.() -> Unit = {}
 
+    /**
+     * Disable tracing in the SDK by providing no-op implementations that don't incur overhead even if instrumentation creates spans
+     */
+    fun disableTracing(disable: Boolean) {
+        rumConfig.setTracingEnabled(!disable)
+    }
```
breedx-splk: I think these would benefit from not allowing the inverse logic, especially when they are enabled by default. 

```suggestion
    fun disableTracing() {
        rumConfig.setTracingEnabled(false)
    }
```

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OpenTelemetryConfiguration.kt
```diff
@@ -33,6 +33,27 @@ class OpenTelemetryConfiguration internal constructor(
...
+
+    /**
+     * Disable logging in the SDK by providing no-op implementations that don't incur overhead even if instrumentation emits logs
+     */
+    fun disableLogging(disable: Boolean) {
+        rumConfig.setLoggingEnabled(!disable)
+    }
```
breedx-splk:
```suggestion
    fun disableLogging() {
        rumConfig.setLoggingEnabled(false)
    }
```

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OpenTelemetryConfiguration.kt
```diff
@@ -33,6 +33,27 @@ class OpenTelemetryConfiguration internal constructor(
...
+
+    /**
+     * Disable metrics in the SDK by providing no-op implementations that don't incur overhead even if instrumentation records metrics
+     */
+    fun disableMetrics(disable: Boolean) {
+        rumConfig.setMetricsEnabled(!disable)
+    }
```
breedx-splk:
```suggestion
    fun disableMetrics() {
        rumConfig.setMetricsEnabled(false)
    }
```

core/src/main/java/io/opentelemetry/android/DisableableOpenTelemetry.kt
breedx-splk: Thanks for clarifying. I was also surprised by this.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/OpenTelemetryConfiguration.kt
```diff
@@ -33,6 +33,27 @@ class OpenTelemetryConfiguration internal constructor(
...
 
+    /**
+     * Disable tracing in the SDK by providing no-op implementations that don't incur overhead even if instrumentation creates spans
+     */
+    fun disableTracing(disable: Boolean) {
+        rumConfig.setTracingEnabled(!disable)
+    }
+
+    /**
+     * Disable logging in the SDK by providing no-op implementations that don't incur overhead even if instrumentation emits logs
```
breedx-splk: Maybe we also mention that this silences events too?

android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
```diff
@@ -55,26 +55,36 @@ object OpenTelemetryRumInitializer {
...
+            if (rumConfig.loggingEnabled) {
+                addLogRecordExporterCustomizer {
+                    createLogExporter(cfg.exportConfig.logsEndpoint())
+                }
+            }
+            if (rumConfig.metricsEnabled) {
+                addMetricExporterCustomizer {
+                    createMetricExporter(cfg.exportConfig.metricsEndpoint())
+                }
+            }
```
breedx-splk, line 86: One interesting edge case of this is that if (for some reason) telemetry data was left on disk, and, after an upgrade, one of the signals was no longer turned on, we wouldn't have an exporter to handle that data on disk. Presumably we would read it and export it to a noop, effectively losing it.

I think I'm fine with that, but just wanted to make a note of it in case it hadn't been considered.

android-agent/src/test/kotlin/io/opentelemetry/android/agent/smoke/OpenTelemetryRumSmokeTest.kt
```diff
@@ -40,71 +44,109 @@ class OpenTelemetryRumSmokeTest {
...
+                disableLogging(false)
+            },
+            action = {
+                recordSpan()
+                recordLog()
+            },
+        )
+
+        server.awaitLogRequest { findLog(it) }
+        assertEquals(0, server.traceRequestCount())
```
breedx-splk: Prefer fluent assertj assertions.
```suggestion
        assertThat(server.traceRequestCount()).isZero()
```

android-agent/src/test/kotlin/io/opentelemetry/android/agent/smoke/OpenTelemetryRumSmokeTest.kt
```diff
@@ -40,71 +44,109 @@ class OpenTelemetryRumSmokeTest {
...
-            server.awaitTraceRequest {
-                it
-                    .getResourceSpans(0)
-                    .getScopeSpans(0)
-                    .getSpans(0)
-                    .name == spanName
-            }
-        assertTraceRequestReceived(traceRequest)
+        server.awaitTraceRequest { findSpan(it) }
+        assertEquals(0, server.logRequestCount())
```
breedx-splk:
```suggestion
        assertThat(server.logRequestCount()).isZero()
```

core/src/main/java/io/opentelemetry/android/config/OtelRumConfig.kt
```diff
@@ -122,6 +140,33 @@ class OtelRumConfig {
...
+    }
+
+    /**
+     * Enables or disables the metrics API for the SDK. When disabled, the SDK will provide no-op implementations to instrumentation
+     * that use this API.
+     */
+    fun setMetricsEnabled(enabled: Boolean): OtelRumConfig {
+        metricsEnabled = enabled
+        return this
+    }
```
breedx-splk: I think my preference would be to have these as `disable` methods with no arg, since the default is enabled.

core/src/main/java/io/opentelemetry/android/OpenTelemetryRumImpl.kt
```diff
@@ -9,22 +9,19 @@ import io.opentelemetry.android.session.SessionProvider
 import io.opentelemetry.api.OpenTelemetry
 import io.opentelemetry.api.common.Attributes
 import io.opentelemetry.api.logs.Logger
-import io.opentelemetry.sdk.OpenTelemetrySdk
 import io.opentelemetry.sdk.common.Clock
 
 internal class OpenTelemetryRumImpl(
-    private val openTelemetrySdk: OpenTelemetrySdk,
+    override val openTelemetry: OpenTelemetry,
```
breedx-splk, line 15: 👍🏻

breedx-splk (commented):
Left some comments, but it's looking mostly good to me! Thanks for picking this up.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt
bidetofevil: Ideally, any persisted telemetry will be properly exported regardless of whether the API is enabled. I think if it's OK for now that it isn't, though in my mind, they should be decoupled enough in theory for that to be possible.

At any rate, I don't think I'll worry about that for now?

core/src/main/java/io/opentelemetry/android/config/OtelRumConfig.kt
bidetofevil: Gonna make the method changes but I'm also going to keep the internal variables as `*enabled`

README.md
```diff
@@ -87,6 +87,7 @@ private fun initOTel(context: Context): OpenTelemetryRum? =
                 globalAttributes {
                     Attributes.of(stringKey("demo-version"), "test")
                 }
+                disableMetrics(true)
```
breedx-splk: Update to match changes.
```suggestion
                disableMetrics()
```

core/src/test/java/io/opentelemetry/android/config/OtelRumConfigTest.kt
```diff
@@ -70,4 +70,23 @@ class OtelRumConfigTest {
...
+        val config = OtelRumConfig()
+        assertThat(config.tracingEnabled).isTrue()
+        assertThat(config.loggingEnabled).isTrue()
+        assertThat(config.metricsEnabled).isTrue()
+    }
+
+    @Test
+    fun `signal enablement can be toggled`() {
+        val config = OtelRumConfig()
+        assertThat(config.setTracingDisabled(false).tracingEnabled).isFalse()
```
breedx-splk: This is so confusing 🤣

bidetofevil: Yeah I'll rename all of these thorughly

bidetofevil:
Made all the disablements one-way in both the DSL and config object. All the tests should match that verbiage too

breedx-splk (approved):
Yeah that cleaned up quite nicely, thanks for sticking with it. I think this is improved over the original idea to disable exporters. Appreciate it.

PR #1801, opened 2026-06-09
Add thermal status instrumentation (#1799)

### LINKED ISSUE #1799: Thermal status instrumentation
Following up on [this comment](https://github.com/open-telemetry/opentelemetry-android/issues/1286#issuecomment-3370537803) in #1286, let's see about adding some new instrumentation for [changes in thermal status](https://developer.android.com/reference/android/os/PowerManager.html#addThermalStatusListener(java.util.concurrent.Executor,%20android.os.PowerManager.OnThermalStatusChangedListener)).

This should be a new instrumentation module that should start as opt-in.

More information about the thermal APIs, including the status code values, is here: https://source.android.com/docs/core/power/thermal-mitigation#thermal-api

hitkall:
## What

Adds a new, **opt-in** instrumentation module (`instrumentation/thermal`) that listens for device thermal-status changes and emits one log event per change.

Fixes #1799

## Approach

- **Event-driven, no polling.** `ThermalInstrumentation.install()` registers a `PowerManager.OnThermalStatusChangedListener` via `PowerManager.addThermalStatusListener(executor, listener)` on a single-thread executor. `uninstall()` calls `removeThermalStatusListener` and shuts the executor down.
- **API 29+ gated.** The thermal APIs (`addThermalStatusListener`, `THERMAL_STATUS_*`) are only available on API level 29 (Android Q). Since the project `minSdk` is 23, `install()` gates on `Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q` and no-ops on older devices. The API-29 call sites are annotated `@RequiresApi(29)` so Animal Sniffer is satisfied.
- **Opt-in.** This module is **not** added to `android-agent`, so it is not bundled by default — consumers add the dependency and install it explicitly. The README documents this.
- `ThermalDetector` maps the integer status codes to readable strings (`none`, `light`, `moderate`, `severe`, `critical`, `emergency`, `shutdown`, with an `unknown` fallback) and emits via the logs bridge.

This module closely mirrors the existing `instrumentation/screen-orientation` module in structure, style, and conventions.

## Telemetry

| | |
|---|---|
| Scope | `io.opentelemetry.thermal` |
| Event name | `device.thermal_state` |
| Attribute | `android.thermal.state` (string) |

> **Note on naming:** the event name (`device.thermal_state`) and attribute (`android.thermal.state`) are **proposals**. Happy to align them with whatever the maintainers prefer for semantic conventions.

## Tests

`ThermalDetectorTest` (JUnit 4 + MockK + AssertJ + `OpenTelemetryRule`) covers:
- a single status change emits the right event name + attribute,
- the full code→name mapping for every known status,
- the `unknown` fallback for an unrecognized code.

## Verification

Run locally against JDK 17 + Android SDK 36:
- `./gradlew spotlessApply`
- `./gradlew :instrumentation:thermal:test` (3 tests pass)
- `./gradlew apiDump` (generated `api/thermal.api`)
- `./gradlew apiCheck`
- `./gradlew :instrumentation:thermal:assemble`
- (also `detekt` + `lintDebug` for the module)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/instrumentation/thermal/README.md
+++ b/instrumentation/thermal/README.md
@@ -24,9 +24,9 @@ scope of `io.opentelemetry.thermal`.
 * Name: `device.thermal_status.change`
 * Description: An event representing a change in the device's thermal throttling status.
 * Attributes:
-    * `android.thermal.state` - A string value that contains the new thermal state.
-    Values are `none`, `light`, `moderate`, `severe`, `critical`, `emergency`,
-    `shutdown`, or `unknown`.
+    * `android.thermal.throttling_status` - A string value that contains the new thermal
+    throttling status. Values are `none`, `light`, `moderate`, `severe`, `critical`,
+    `emergency`, `shutdown`, or `unknown`.
 
 ## Installation
 
--- a/instrumentation/thermal/README.md
+++ b/instrumentation/thermal/README.md
@@ -1,0 +1,40 @@
+# Thermal Status Instrumentation
+
+Status: development
+
+The OpenTelemetry thermal status instrumentation for Android detects when the device's
+thermal throttling status changes (for example, when the device starts to overheat).
+
+It listens for changes via [`PowerManager.addThermalStatusListener`](https://developer.android.com/reference/android/os/PowerManager#addThermalStatusListener(java.util.concurrent.Executor,%20android.os.PowerManager.OnThermalStatusChangedListener))
+(no polling) and emits one event per status change.
+
+The thermal status APIs are only available on API level 29 (Android Q) and higher, so this
+instrumentation no-ops on older devices.
+
+This instrumentation is not currently enabled by default.
+
+## Telemetry
+
+This instrumentation produces the following telemetry, with an instrumentation
+scope of `io.opentelemetry.thermal`.
+
+### Thermal State
+
+* Type: Event
+* Name: `device.thermal_state`
+* Description: An event representing a change in the device's thermal throttling status.
+* Attributes:
+    * `android.thermal.state` - A string value that contains the new thermal state.
+    Values are `none`, `light`, `moderate`, `severe`, `critical`, `emergency`,
+    `shutdown`, or `unknown`.
+
+## Installation
+
+This instrumentation does not come with the [android agent](../../android-agent) out of the
+box, so you need to manually install it as described below.
+
+### Adding dependencies
+
+```kotlin
+implementation("io.opentelemetry.android.instrumentation:thermal:LATEST_VERSION")
+```
--- a/instrumentation/thermal/README.md
+++ b/instrumentation/thermal/README.md
@@ -21,7 +21,7 @@ scope of `io.opentelemetry.thermal`.
 ### Thermal State
 
 * Type: Event
-* Name: `device.thermal_state`
+* Name: `device.thermal_status.change`
 * Description: An event representing a change in the device's thermal throttling status.
 * Attributes:
     * `android.thermal.state` - A string value that contains the new thermal state.
--- a/instrumentation/thermal/src/main/AndroidManifest.xml
+++ b/instrumentation/thermal/src/main/AndroidManifest.xml
@@ -1,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+
+</manifest>
--- a/instrumentation/thermal/src/main/AndroidManifest.xml
+++ b/instrumentation/thermal/src/main/AndroidManifest.xml
@@ -1,4 +1,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<manifest xmlns:android="http://schemas.android.com/apk/res/android">
-
-</manifest>
--- a/instrumentation/thermal/src/main/java/io/opentelemetry/android/instrumentation/thermal/ThermalDetector.kt
+++ b/instrumentation/thermal/src/main/java/io/opentelemetry/android/instrumentation/thermal/ThermalDetector.kt
@@ -26,14 +26,14 @@ internal class ThermalDetector(
 ) : PowerManager.OnThermalStatusChangedListener {
     internal companion object {
         const val EVENT_NAME = "device.thermal_status.change"
-        const val THERMAL_STATE = "android.thermal.state"
+        const val THERMAL_THROTTLING_STATUS = "android.thermal.throttling_status"
     }
 
     override fun onThermalStatusChanged(status: Int) {
         logger
             .logRecordBuilder()
             .setEventName(EVENT_NAME)
-            .setAttribute(THERMAL_STATE, status.name)
+            .setAttribute(THERMAL_THROTTLING_STATUS, status.name)
             .emit()
     }
 
--- a/instrumentation/thermal/src/main/java/io/opentelemetry/android/instrumentation/thermal/ThermalDetector.kt
+++ b/instrumentation/thermal/src/main/java/io/opentelemetry/android/instrumentation/thermal/ThermalDetector.kt
@@ -25,7 +25,7 @@ internal class ThermalDetector(
     private val logger: Logger,
 ) : PowerManager.OnThermalStatusChangedListener {
     internal companion object {
-        const val EVENT_NAME = "device.thermal_state"
+        const val EVENT_NAME = "device.thermal_status.change"
         const val THERMAL_STATE = "android.thermal.state"
     }
 
--- a/instrumentation/thermal/src/main/java/io/opentelemetry/android/instrumentation/thermal/ThermalDetector.kt
+++ b/instrumentation/thermal/src/main/java/io/opentelemetry/android/instrumentation/thermal/ThermalDetector.kt
@@ -29,11 +29,11 @@ internal class ThermalDetector(
         const val THERMAL_STATE = "android.thermal.state"
     }
 
-    private fun emitLog(state: String) {
+    override fun onThermalStatusChanged(status: Int) {
         logger
             .logRecordBuilder()
             .setEventName(EVENT_NAME)
-            .setAttribute(THERMAL_STATE, state)
+            .setAttribute(THERMAL_STATE, status.name)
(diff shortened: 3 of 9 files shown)


instrumentation/thermal/src/main/java/io/opentelemetry/android/instrumentation/thermal/ThermalInstrumentation.kt
```diff
@@ -0,0 +1,93 @@
...
+            applicationContext.getSystemService(Context.POWER_SERVICE) as? PowerManager
+                ?: return
+
+        val logger =
+            openTelemetryRum.openTelemetry
+                .logsBridge
+                .loggerBuilder("io.opentelemetry.$name")
+                .build()
+
+        val executor = Executors.newSingleThreadExecutor()
```
breedx-splk, line 50: A bit of a drag to burn an entire thread on something that is probably rarely used. I'm not entirely sure what best practice around this is. We have a few other instrumentations that create their own, so it's not a show-stopper. We might want to consider a future mechanism for sharing these so that it's a little more scalable.

instrumentation/thermal/src/main/java/io/opentelemetry/android/instrumentation/thermal/ThermalDetector.kt
```diff
@@ -0,0 +1,57 @@
...
+ * The thermal status APIs are only available on API level 29 (Android Q) and higher.
+ *
+ * @param logger The [Logger] instance used to record thermal status change events.
+ */
+@RequiresApi(29)
+internal class ThermalDetector(
+    private val logger: Logger,
+) : PowerManager.OnThermalStatusChangedListener {
+    internal companion object {
+        const val EVENT_NAME = "device.thermal_state"
```
breedx-splk: If the `state` is the thing we're communicating via the attribute, then I think the name of the event itself should be the "what happened", for which I'll propose this change:

```suggestion
        const val EVENT_NAME = "device.thermal_status.change"
```

We can think about semconv more thoroughly later, but having it try and be slightly more conformant now may help....I think?

instrumentation/thermal/src/main/java/io/opentelemetry/android/instrumentation/thermal/ThermalDetector.kt
```diff
@@ -0,0 +1,57 @@
...
+                PowerManager.THERMAL_STATUS_MODERATE -> "moderate"
+                PowerManager.THERMAL_STATUS_SEVERE -> "severe"
+                PowerManager.THERMAL_STATUS_CRITICAL -> "critical"
+                PowerManager.THERMAL_STATUS_EMERGENCY -> "emergency"
+                PowerManager.THERMAL_STATUS_SHUTDOWN -> "shutdown"
+                else -> "unknown"
+            }
+        }
+
+    override fun onThermalStatusChanged(status: Int) {
```
breedx-splk: nitpick: prefer to have the interface method above the private methods that it calls into.

instrumentation/thermal/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+
+</manifest>
```
breedx-splk: I think this can be removed,  unless you have a reason for needing it? I see that the screen orientation also has an empty one, but maybe that can be removed as well (separately from this PR tho).

instrumentation/thermal/src/test/java/io/opentelemetry/android/instrumentation/thermal/ThermalDetectorTest.kt
```diff
@@ -0,0 +1,87 @@
...
+import io.opentelemetry.android.instrumentation.thermal.ThermalDetector.Companion.THERMAL_STATE
+import io.opentelemetry.api.common.AttributeKey
+import io.opentelemetry.sdk.testing.junit4.OpenTelemetryRule
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+
+class ThermalDetectorTest {
+    private lateinit var sut: ThermalDetector
```
breedx-splk: I'm not sure what `sut` is a convention for or shortening of ("[s]omething [u]nder [t]est"?), but it's not something we really use. Maybe just call it `detector` instead?

instrumentation/thermal/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Telemetry
+
+This instrumentation produces the following telemetry, with an instrumentation
+scope of `io.opentelemetry.thermal`.
+
+### Thermal State
+
+* Type: Event
+* Name: `device.thermal_state`
```
breedx-splk: reminder to change readme to match the name change suggested above (or whatever it lands on).

instrumentation/thermal/src/main/java/io/opentelemetry/android/instrumentation/thermal/ThermalDetector.kt
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @param logger The [Logger] instance used to record thermal status change events.
+ */
+@RequiresApi(29)
+internal class ThermalDetector(
+    private val logger: Logger,
+) : PowerManager.OnThermalStatusChangedListener {
+    internal companion object {
+        const val EVENT_NAME = "device.thermal_state"
+        const val THERMAL_STATE = "android.thermal.state"
```
breedx-splk: I wonder if this attribute should be more specifically about the throttling and impact and not the thermal state. It's a subtle distinction, but the [docs](https://source.android.com/docs/core/power/thermal-mitigation#thermal-api) talk about throttling more than the temperature. Anyway, I don't want to overthink this in the first pass, so non-blocking comment. Just thinking out loud.

breedx-splk (commented):
Thanks @hitkall this is great. I offered a few remarks, but this is definitely on track. 

I also think you might have missed adding this module to the `settings.gradle.kts` in the root of the project. I don't think it'll be included and compiled until you do that.

hitkall:
Thanks for the thorough review, really appreciate it!
Addressing everything:
settings.gradle.kts — I think this is actually handled already. The includeFromDir("instrumentation") function in settings.gradle.kts walks the directory tree looking for build.gradle.kts files up to depth 3, so instrumentation/thermal gets auto-discovered without a manual entry. That said, I want to double-check that api/thermal.api was committed properly — its absence might be what made the module look unregistered in CI. Will verify and push a fix if it's missing.
Executor — totally fair point. A dedicated thread for something that fires maybe a handful of times a day is wasteful. Happy to follow whatever pattern the project lands on for sharing executors across instrumentations — flagging it as a known issue in the code for now sounds reasonable if there's no existing pattern to hook into yet.
device.thermal_status.change — accepted, makes more sense as "what happened". Will update EVENT_NAME, the README, and the test references.
Method ordering — will fix, onThermalStatusChanged above the private extension.
AndroidManifest.xml — will delete it.
sut → detector — will rename throughout the test.
android.thermal.state attribute — I like the direction of framing it around throttling impact rather than raw thermal state. Open to android.thermal.throttling_status or similar if you have a preference — otherwise I'll leave it as-is for now and we can revisit when semconv gets more formal here.
Will push an update addressing all of the above shortly.

@breedx-splk

instrumentation/thermal/src/main/java/io/opentelemetry/android/instrumentation/thermal/ThermalDetector.kt
hitkall: @breedx-splk 

After reading the thermal-mitigation docs more carefully, I think you're right — the API reports throttling severity (0 = no throttling, 6 = shutdown), not temperature state. Happy to rename the attribute to android.thermal.throttling_status to better reflect that. Let me know if you'd prefer a different name or want to leave it for a follow-up semconv discussion.

instrumentation/thermal/src/main/java/io/opentelemetry/android/instrumentation/thermal/ThermalDetector.kt
hitkall: Accepted — `device.thermal_status.change` is cleaner. Will update in the next push along with the other feedback.

breedx-splk (approved):
Looking good! Thanks for the help and for the quick responses.

hitkall:
Thanks for the review @breedx-splk! All feedback is addressed in the latest commits:

- Renamed the attribute to `android.thermal.throttling_status` — agreed the API reports throttling severity (0 = none … 6 = shutdown) rather than temperature, so this is more accurate. README updated to match.
- Renamed the event to `device.thermal_status.change`.
- Moved the interface method (`onThermalStatusChanged`) above the private helpers.
- Removed the empty `AndroidManifest.xml`.
- Renamed `sut` → `detector` in the detector test.
- Added `ThermalInstrumentationTest` covering the API-29 gate, listener registration/removal, and the uninstall-before-install no-op.

Re: `settings.gradle.kts` — the module is picked up automatically by `includeFromDir("instrumentation")`, and CI is compiling/testing `:instrumentation:thermal`, so no manual `include` is needed.

The event/attribute names are still open to whatever the semantic-conventions discussion lands on. Let me know if you'd like anything else.

hitkall:
@breedx-splk are we good to merge?

breedx-splk:
> @breedx-splk are we good to merge?

I think so, yeah! I haven't merged already because I'm giving our other maintainers and approvers a chance to review and chime in. Just FYI, we're flexible with this stuff, but this is what we've previously described as "[tier 3](https://github.com/open-telemetry/opentelemetry-android/blob/main/CONTRIBUTING.md#tier-3--new-features-and-significant-internal-changes)" PR, so a second approval and a little simmer time is appreciated.

Thanks again!

DavidGrath (approved):
It's okay by me. Thanks

PR #1802, opened 2026-06-10
Use jank semconv constants

breedx-splk:
Allows us to remove our hard-coded constants in favor of semconv.


LikeTheSalad (approved):
Cheers!

PR #1691, opened 2026-04-20
Renaming crash reporter event

LikeTheSalad:
Following this semantic convention change: https://github.com/open-telemetry/semantic-conventions/pull/3448

PR #1804, opened 2026-06-10
crash event name changed, update doc to match

breedx-splk:
This changed in #1691 but it looks like the doc didn't match.


LikeTheSalad (approved):
Cheers!

PR #1811, opened 2026-06-12
Add power save mode instrumentation

### LINKED ISSUE #1800: Power Save mode instrumentation
Following up on [this comment ](https://github.com/open-telemetry/opentelemetry-android/issues/1286#issuecomment-3370537803)in #1286, let's see about adding some new instrumentation to report [changes in the power saving mode](https://developer.android.com/reference/android/os/PowerManager#ACTION_POWER_SAVE_MODE_CHANGED). This should be a new instrumentation module that should be opt-in for starters.

GabrielBBaldez:
Closes #1800.

Adds an opt-in instrumentation module that reports changes to the device's power save (battery saver) mode, following the pattern of the existing `thermal` instrumentation.

## What it does

- Registers a `BroadcastReceiver` for `PowerManager.ACTION_POWER_SAVE_MODE_CHANGED` (no polling); on each change it reads `PowerManager.isPowerSaveMode` and emits one event.
- Event `device.power_save_mode.change` with a boolean attribute `android.power_save_mode.enabled`, under instrumentation scope `io.opentelemetry.power_save_mode`.
- Opt-in: not wired into the android-agent by default, consistent with `thermal`.
- The power save mode APIs (`isPowerSaveMode` / `ACTION_POWER_SAVE_MODE_CHANGED`) are available since API 21, which is below `minSdk`, so no version gating is required.

## Notes

- Structure mirrors the `thermal` module (instrumentation + detector + unit tests + API dump + README).
- The event name and attribute are a first proposal — happy to adjust to match any preferred semantic conventions.

## Verification

`./gradlew :instrumentation:power-save-mode:build` passes locally: compile, unit tests (Robolectric), lint, API check, detekt, and assemble.

### DIFF
--- a/instrumentation/power-save-mode/src/main/java/io/opentelemetry/android/instrumentation/powersavemode/PowerSaveModeInstrumentation.kt
+++ b/instrumentation/power-save-mode/src/main/java/io/opentelemetry/android/instrumentation/powersavemode/PowerSaveModeInstrumentation.kt
@@ -1,0 +1,59 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.android.instrumentation.powersavemode
+
+import android.content.Context
+import android.content.IntentFilter
+import android.os.PowerManager
+import com.google.auto.service.AutoService
+import io.opentelemetry.android.OpenTelemetryRum
+import io.opentelemetry.android.instrumentation.AndroidInstrumentation
+
+/**
+ * An Android instrumentation module that installs and manages [PowerSaveModeDetector].
+ *
+ * The power save mode APIs ([PowerManager.isPowerSaveMode] and
+ * [PowerManager.ACTION_POWER_SAVE_MODE_CHANGED]) are available since API level 21, which is below
+ * this library's minSdk, so no version gating is required.
+ */
+@AutoService(AndroidInstrumentation::class)
+class PowerSaveModeInstrumentation : AndroidInstrumentation {
+    private var detector: PowerSaveModeDetector? = null
+
+    override fun install(
+        context: Context,
+        openTelemetryRum: OpenTelemetryRum,
+    ) {
+        val applicationContext = context.applicationContext
+        val powerManager =
+            applicationContext.getSystemService(Context.POWER_SERVICE) as? PowerManager
+                ?: return
+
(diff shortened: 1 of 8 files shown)


instrumentation/power-save-mode/src/test/java/io/opentelemetry/android/instrumentation/powersavemode/PowerSaveModeInstrumentationTest.kt
DavidGrath, line 1: I'm a bit iffy about this class - to me it borders on "testing the implementation". Other than that, though, this PR looks fine to me

GabrielBBaldez: Good call — reworked it to assert behavior instead of the registration calls: it now installs the instrumentation, broadcasts `ACTION_POWER_SAVE_MODE_CHANGED`, and checks the event is emitted (and that it stops after uninstall), reusing the `OpenTelemetryRule` + Robolectric setup from the detector test. Pushed in 2012457.

DavidGrath: Thanks. Looks good

DavidGrath (approved):

instrumentation/power-save-mode/src/main/java/io/opentelemetry/android/instrumentation/powersavemode/PowerSaveModeInstrumentation.kt
```diff
@@ -0,0 +1,59 @@
...
+ */
+@AutoService(AndroidInstrumentation::class)
+class PowerSaveModeInstrumentation : AndroidInstrumentation {
+    private var detector: PowerSaveModeDetector? = null
+
+    override fun install(
+        context: Context,
+        openTelemetryRum: OpenTelemetryRum,
+    ) {
+        val applicationContext = context.applicationContext
```
breedx-splk, line 30: I'd move this closer to where it's used, or just inline it.

breedx-splk (approved):
Looks good. I had one very minor nit, but I'm going to merge anyway. Thanks for the contribution!

Issue #1814, opened 2026-06-15
Federated semantic conventions :: Phased Approach


breedx-splk:
For historical reasons, this repository contains many bespoke, nonstandard semantic conventions scattered throughout. These exist outside of machine-readable, spec-compliant semconv yaml, and they live primarily as hard-coded string constants for attributes, resource attributes, and event names.

This is not desirable for many reasons. Firstly:

1. Conventions are scattered throughout the code, so there isn't a single place to find them.
2. Without yaml and a schema, semantic conventions could drift and become inconsistent in names, types, or requirement or stability levels.
3. Code generation isn't possible.
4. Documentation generation isn't possible. To demonstrate, the telemetry documented in instrumentation READMEs must be manually generated and maintained (and has been wrong several times).
5. Auditing and other tooling isn't possible.

With the relatively recent acceptance of the [semconv v2 otep](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/4815-semantic-conventions-schema-v2.md), we have a nice path forward with federated semantic conventions. I propose that we create and begin using local, federated semantic conventions in this android repo.

Why not just add them to the main [semantic-conventions](https://github.com/open-telemetry/semantic-conventions)?

* the main reasons are described in the [motivations](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/4815-semantic-conventions-schema-v2.md#motivation) section of the OTEP, which says `[...] language-specific instrumentations should be able to publish their own conventions that are only applicable within their ecosystem. They should also have a means to communicate the schema they use via [Schema URL](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.54.0/specification/schemas/README.md#schema-url).`
* too many conventions already here are android-specific
* federating here allows us to move faster, without the need to have broad (huge) community consensus
* having conventions in this repo will provide a solid foundation for later merging/federating with other client semconv

These semconv changes have the possibility of touching MANY many files and inflating PR size, resulting in difficult reviews. To mitigate this, I am proposing a phased approach, with each of these phases being a single PR:

1. Create a new module that houses the `registry.yaml` and `events.yaml` semconv, along with the `manifest.yaml`. These will contain duplicate names, types, and descriptions etc. for every semantic convention in the repo.
2. Add a build step to generate the attribute semantic conventions. When run, this will create kotlin source files that contain the attribute name constant values.
3. Replace all usages of attributes with the generated constant values.
[...]

PR #1815, opened 2026-06-15
Federated semantic conventions phase 1

breedx-splk:
Part of #1814. AI assisted.

This establishes a new module to hold our android-specific semconv. This PR pulls in all of the semantic conventions that I could find and creates attribute and event entries for these. 

There is not yet any weaver or code generation. As described in #1814, this will come in the following phase(s).


LikeTheSalad (approved):
Cheers!

PR #1816, opened 2026-06-16
Truncated stack traces to 1,000 lines max by default

### LINKED ISSUE #1195: Truncate large stacktraces captured by crash instrumentation
Stacktraces captured by the crash instrumentation do not current impose limits on the number of stackframes that are captured. This is likely to lead to poor outcomes where the stacktrace is very large such as in the case of a `StackOverflowError`. We should impose some reasonable upper limit of stackframes that should be captured in a log, e.g. 1000.

Originally spotted in https://github.com/open-telemetry/opentelemetry-android/pull/1172

DavidGrath:
Resolves #1195

### DIFF
--- a/instrumentation/crash/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashReporter.kt
+++ b/instrumentation/crash/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashReporter.kt
@@ -59,7 +59,7 @@ internal class CrashReporter(
 
     private fun stackTraceToTruncatedString(throwable: Throwable, truncateThreshold: Int = 1000): String {
         val stringWriter = StringWriter()
-        val printWriter = MaxLinesPrintWriter(sw, truncateThreshold)
+        val printWriter = MaxLinesPrintWriter(stringWriter, truncateThreshold)
         throwable.printStackTrace(printWriter)
         printWriter.flush()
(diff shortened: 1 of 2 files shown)


instrumentation/crash/src/main/java/io/opentelemetry/android/instrumentation/crash/CrashReporter.kt
```diff
@@ -34,6 +57,14 @@ internal class CrashReporter(
         Thread.setDefaultUncaughtExceptionHandler(handler)
     }
 
+    private fun stackTraceToTruncatedString(throwable: Throwable, truncateThreshold: Int = 1000): String {
+        val sw = StringWriter()
+        val cw = MaxLinesPrintWriter(sw, truncateThreshold)
+        throwable.printStackTrace(cw)
+        cw.flush()
+        return sw.toString()
```
breedx-splk: We should use more conventional variable names, and not shortened abbreviations.

```suggestion
        val stringWriter = StringWriter()
        val printWriter = MaxLinesPrintWriter(sw, truncateThreshold)
        throwable.printStackTrace(printWriter)
        printWriter.flush()
        return stringWriter.toString()
```

breedx-splk (approved):
Thanks!

PR #1818, opened 2026-06-16
Remove empty manifest files

breedx-splk:


DavidGrath:
Hmmm, so it seems they were actually needed according to the logs. To be closed? Maybe we could instead add XML comments to explain their importance

LikeTheSalad:
> Hmmm, so it seems they were actually needed according to the logs. To be closed? Maybe we could instead add XML comments to explain their importance

Yeah it seems like the AGP needs to find this file, even if it's empty.

breedx-splk:
Yup! But now we know. Thanks all.

PR #1819, opened 2026-06-16
Use app.screen.name and allow opt-in for older semconv

breedx-splk:
This adds a new configuration item that can be leveraged through the DSL:

```kotlin
semanticConventions {
  useLatestExperimental = false
}
```

This allows users to continue using existing semantic conventions and opt into NOT using the latest/changed/new conventions. This helps with migrations. When we reach 2.0, we can remove the existing mappings and just stick with the current/latest. Users who do nothing will, by default, get the new names. I think this is reasonable in the short term, but only because nearly none of the semconv are anywhere near stable yet.

Right now there are only two mappings:
* `app.screen.name` -> `screen.name`
* `app.crash` -> `device.crash`

Ideally this would be even fancier -- and maybe schema aware or let the user choose which version they want, but I'm doing the stupid/simpler thing for now until we have our local semconv in place.


breedx-splk:
relates to #1814

LikeTheSalad (approved):
Cheers!

PR #1828, opened 2026-06-18
Federated semantic conventions phase 2

breedx-splk:
Part of #1814.

This adds a weaver template and gradle script to run weaver and generate these constants files.

In the next phase, we will replace production usages of these constants with references to the new semconv constant classes.


breedx-splk:
I followed the convention from the `opentelemetry-kotlin` repo, which expects a local `weaver` executable to be available. If we hate that and would rather use a docker image (more like the `semantic-conventions-java` repo), we can change it....I'm flexible.

semconv/src/main/kotlin/io/opentelemetry/android/semconv/LastAttributes.kt
```diff
@@ -0,0 +1,19 @@
...
+
+package io.opentelemetry.android.semconv
+
+/**
+ * Autogenerated OpenTelemetry Android semantic conventions for 'last'.
+ *
+ * DO NOT EDIT - AUTOGENERATED FILE
+ */
+@Suppress("unused")
+object LastAttributes {
```
breedx-splk: LOL this also makes it very clear that we will want to change this.

semconv/templates/registry/kotlin/kotlin_semconv_template.kt.j2
```diff
@@ -0,0 +1,20 @@
...
+object {{ my_class_name }} {
+{% for attribute in ctx.attributes | rejectattr("name", "in", ctx.excluded_attributes) %}
+
+    {{ [attribute.brief, concat_if("\n\nNotes:\n\n", attribute.note)] | comment | indent(4) -}}
+{% if attribute is deprecated %}
+    @Deprecated("{{ attribute.deprecated.note | replace('\n', ' ') | trim  }}")
+{% endif %}
+    const val {{ attribute.name | screaming_snake_case }}: String = "{{ attribute.name }}"
+{% endfor %}
+}
```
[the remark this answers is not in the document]
breedx-splk: addressed.

semconv/src/main/kotlin/io/opentelemetry/android/semconv/ActivityAttributes.kt
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.android.semconv
+
+/**
+ * Autogenerated OpenTelemetry Android semantic conventions for 'activity'.
+ *
+ * DO NOT EDIT - AUTOGENERATED FILE
```
LikeTheSalad: Do we need to track these auto-generated source files? To me, it sounds similar to tracking protobuf-generated ones, for example. Usually these kinds of sources get into the "build" dir and then are plugged into the android sources and compilation process to be available for reference and automatically up-to-date whenever the template changes.

bidetofevil (approved):
LGTM

LikeTheSalad (approved):
Thank you!

semconv/src/main/kotlin/io/opentelemetry/android/semconv/ActivityAttributes.kt
LikeTheSalad: Btw this is not a blocker, so we can take a look at it later.

breedx-splk: Ok, thanks for the idea. I went ahead and implemented it (with help from codex) in this PR. I really like not having the generated sources committed for constants, and probably for the event files too. This PR shrunk in half!

Please have another look. I'll plan on merging soon. Appreciate it!

breedx-splk: Oh, interesting. That last commit that doesn't add the generated sources highlights a challenge with this approach -- basically "the build" now requires weaver to always be available. I hadn't considered that. I'm going to see how ugly it gets to add weaver somewhat more broadly/generically.

breedx-splk: I'm talking myself out of this for now. I don't want to burden new contributors with a requirement of needing weaver available locally. I think it probably also makes the build a bit slower.

Issue #1821, opened 2026-06-17
DSL vs imperative configuration and disk buffering DSL config


bencehornak:
I've recently faced the dilemma of using the new DSL configuration API (i.e. `OpenTelemetryRumInitializer`): I wanted to make use of the "reasonable defaults" that it comes with, however, there was just one thing I wanted to override: the disk buffering configuration by increasing the exporting interval. Since the related disk buffering properties (`maxFileAgeForWriteMillis`, `minFileAgeForReadMillis`, `maxFileAgeForReadMillis`) are not exposed (you can only enable or disable it currently, see below), this meant that I had no choice, but to fully ditch the DSL API and include all the boilerplate in my project that the `android-agent` module is supposed to hide.

https://github.com/open-telemetry/opentelemetry-android/blob/ad76f8937c49527134926b2e43d37cfcfdb1f253/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/DiskBufferingConfigurationSpec.kt#L13-L32

So I have some questions related to my recent experiences:

1. Are you planning to make the DSL configuration be on par with the imperative configuration API (i.e. `RumBuilder`)?
2. If not:
   1. Would it make sense to document both APIs as viable alternatives with pros and cons in the `README` (currently the usage of `RumBuilder` requires a lot of reverse-engineering due to the absence of docs)?
   2. The `OpenTelemetryRumBuilder` API must be officially part of the stable API (as opposed to the [disclaimer](https://github.com/open-telemetry/opentelemetry-android/blob/1922ff471a0e309236a623d1543169f66ce22cd1/core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.kt#L84-L85), that it is internal), right?
   3. What do you think about exposing the `RumBuilder` object of the `OpenTelemetryRumInitializer` DSL entry point class in some way to allow further customizations before the `OpenTelemetryRum` object is built?

Thanks for your feedback in advance!

LikeTheSalad:
Hi @bencehornak 

It's not something we've written anywhere AFAIK, but my understanding is that we'd like to go with option 1, to the point where `RumBuilder` would only be needed for very niche scenarios. Ideally, we should have everything needed in the DSL, is just that we don't know what's fully needed, and I don't think it's wise to add too much stuff upfront due to maintenance concerns, so feedback like this one is the missing piece before we decide to add more config options. So thanks for raising this issue!

[...]

PR #1834, opened 2026-06-22
Make the disk buffering export period configurable

### LINKED ISSUE #1823: Configurable `exportPeriod` for disk buffering related `PeriodicExporter`
### Problem

Currently, the `PeriodicExporter` is statically configured to wake up every 10 seconds and to check if there is anything to export from the disk:

https://github.com/open-telemetry/opentelemetry-android/blob/e6062e8073a39b6180396e3ec4e710851aacf93f/core/src/main/java/io/opentelemetry/android/features/diskbuffering/scheduler/PeriodicExporter.kt#L11-L14

I have overridden the defaults for the disk buffering to write the files for at least 15 minutes to reduce the number of calls on my backend (I have tens of thousands of devices producing logs around the clock, so the number of requests would incur extreme amount of costs with the default disk buffering config). However, I'm unable to reduce the frequency of the `PeriodicExporter` to check the buffers less frequently, so it is checking the buffers 90 times more frequently than necessary.

### Proposal
Use the `DiskBufferingConfig.minFileAgeForReadMillis` property for `PeriodicExporter.exportPeriod`? (I'm open for better suggestions)

thc1006:
## Summary

The disk-buffering `PeriodicExporter` polled the disk for buffered signals on a fixed
10-second interval that could not be changed. Apps that deliberately buffer for longer (for
example by raising `maxFileAgeForWriteMillis` to reduce backend traffic) still paid for a
10-second poll, checking the disk far more often than data could become ready.

This adds a dedicated, configurable export period.

## Changes

- `DiskBufferingConfig` gains `exportPeriodMillis` (default 10 seconds), alongside the existing
  disk-buffering fields. The field is appended last so the positional `create()` overloads stay
  source- and binary-compatible.
- `create()` falls back to the default when given a non-positive value, mirroring how the
  existing min/max fields are fixed up.
- The value is threaded through `DiskBufferingEnablement` into the `PeriodicExporter`. Default
  behavior is unchanged.

## Notes for reviewers

- This is the `:core` half discussed in #1823. The friendlier high-level DSL
  (`exportingPeriod` / `discardAfter`) is tracked separately in #1821 and is intentionally not
  included here to keep this change focused.
- The new option deliberately does not reuse `minFileAgeForReadMillis`, in line with the
  guidance on the issue and the direction of open-telemetry/opentelemetry-java-contrib#2897.
- Public API change in `:core`; `core.api` is updated accordingly. `:android-agent` is unchanged.

## Testing

- `DiskBufferingConfigTest`: default, custom value, and non-positive fallback.
- `DiskBufferingEnablementTest`: the configured period reaches `PeriodicExporter.period()`.

Resolves #1823

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfig.kt
+++ b/core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfig.kt
@@ -15,6 +15,7 @@ const val MAX_CACHE_FILE_SIZE: Int = 1024 * 1024
 const val DEFAULT_MAX_FILE_AGE_FOR_WRITE_MS = 30L
 const val DEFAULT_MIN_FILE_AGE_FOR_READ_MS = 33L
 const val DEFAULT_MAX_FILE_AGE_FOR_READ_MS = 18L
+const val DEFAULT_EXPORT_PERIOD_MS = 10L
 
 data class DiskBufferingConfig
     @JvmOverloads
@@ -31,6 +32,10 @@ data class DiskBufferingConfig
          * `null`, a default directory inside the application's cache directory will be used.
          */
         val signalsBufferDir: File? = null,
+        /**
+         * How often the exporter wakes up to read buffered signals from disk and send them.
+         */
+        val exportPeriodMillis: Long = TimeUnit.SECONDS.toMillis(DEFAULT_EXPORT_PERIOD_MS),
     ) {
         companion object {
             /**
@@ -49,12 +54,19 @@ data class DiskBufferingConfig
                 maxCacheFileSize: Int = MAX_CACHE_FILE_SIZE,
                 debugEnabled: Boolean = false,
                 signalsBufferDir: File? = null,
+                exportPeriodMillis: Long = TimeUnit.SECONDS.toMillis(10),
             ): DiskBufferingConfig {
                 var minRead = minFileAgeForReadMillis
                 if (minFileAgeForReadMillis <= maxFileAgeForWriteMillis) {
                     minRead = maxFileAgeForWriteMillis + 5
                     Log.w(OTEL_RUM_LOG_TAG, "minFileAgeForReadMillis must be greater than maxFileAgeForWriteMillis")
                     Log.w(OTEL_RUM_LOG_TAG, "overriding minFileAgeForReadMillis from $minFileAgeForReadMillis to $minRead")
+                }
+                var exportPeriod = exportPeriodMillis
+                if (exportPeriod <= 0) {
+                    exportPeriod = TimeUnit.SECONDS.toMillis(10)
+                    Log.w(OTEL_RUM_LOG_TAG, "exportPeriodMillis must be greater than 0")
+                    Log.w(OTEL_RUM_LOG_TAG, "overriding exportPeriodMillis from $exportPeriodMillis to $exportPeriod")
                 }
                 return DiskBufferingConfig(
                     enabled = enabled,
@@ -65,6 +77,7 @@ data class DiskBufferingConfig
                     maxCacheFileSize = maxCacheFileSize,
                     debugEnabled = debugEnabled,
                     signalsBufferDir = signalsBufferDir,
+                    exportPeriodMillis = exportPeriod,
                 )
             }
         }
--- a/core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfig.kt
+++ b/core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfig.kt
@@ -54,7 +54,7 @@ data class DiskBufferingConfig
                 maxCacheFileSize: Int = MAX_CACHE_FILE_SIZE,
                 debugEnabled: Boolean = false,
                 signalsBufferDir: File? = null,
-                exportPeriodMillis: Long = TimeUnit.SECONDS.toMillis(10),
+                exportPeriodMillis: Long = TimeUnit.SECONDS.toMillis(DEFAULT_EXPORT_PERIOD_MS),
             ): DiskBufferingConfig {
                 var minRead = minFileAgeForReadMillis
                 if (minFileAgeForReadMillis <= maxFileAgeForWriteMillis) {
@@ -64,7 +64,7 @@ data class DiskBufferingConfig
                 }
                 var exportPeriod = exportPeriodMillis
                 if (exportPeriod <= 0) {
-                    exportPeriod = TimeUnit.SECONDS.toMillis(10)
+                    exportPeriod = TimeUnit.SECONDS.toMillis(DEFAULT_EXPORT_PERIOD_MS)
                     Log.w(OTEL_RUM_LOG_TAG, "exportPeriodMillis must be greater than 0")
                     Log.w(OTEL_RUM_LOG_TAG, "overriding exportPeriodMillis from $exportPeriodMillis to $exportPeriod")
                 }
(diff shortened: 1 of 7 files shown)


thc1006:
Thanks for the really helpful discussion on #1823, and thanks @bencehornak for raising it with such clear context. Since it's labeled `help wanted`, I hope it's okay that I went ahead and took a first pass at it.

This implements the `:core` side that was discussed: a dedicated `exportPeriodMillis` on `DiskBufferingConfig` rather than reusing `minFileAgeForReadMillis`, following @LikeTheSalad's suggestion. It defaults to the current 10s, is appended last so the existing positional `create()` overloads stay binary-compatible, and `create()` falls back to the default for non-positive values, the same way the min/max fields are already fixed up.

Two points I'd genuinely value your steer on:

- I kept it as a `Long` millis field on `DiskBufferingConfig` (to match the neighboring fields) instead of a method on `OtelRumConfig`. Happy to change if you'd prefer the latter.
- I left the higher-level `:android-agent` DSL (`exportingPeriod` / `discardAfter`) out so this PR stays focused, since that part is still being shaped in #1821. Glad to fold it in here instead if you'd rather.

Since this touches the public API, there's no rush at all on my side. Whenever you or the other approvers have a moment, any feedback would be very welcome, and I'm happy to reshape it however works best.

core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfig.kt
```diff
@@ -49,13 +54,20 @@ data class DiskBufferingConfig
                 maxCacheFileSize: Int = MAX_CACHE_FILE_SIZE,
                 debugEnabled: Boolean = false,
                 signalsBufferDir: File? = null,
+                exportPeriodMillis: Long = TimeUnit.SECONDS.toMillis(10),
```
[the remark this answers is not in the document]
thc1006: Good catch, thank you. Switched the create() default to DEFAULT_EXPORT_PERIOD_MS in 624436db so it no longer hard-codes the value.

core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfig.kt
```diff
@@ -49,13 +54,20 @@ data class DiskBufferingConfig
...
             ): DiskBufferingConfig {
                 var minRead = minFileAgeForReadMillis
                 if (minFileAgeForReadMillis <= maxFileAgeForWriteMillis) {
                     minRead = maxFileAgeForWriteMillis + 5
                     Log.w(OTEL_RUM_LOG_TAG, "minFileAgeForReadMillis must be greater than maxFileAgeForWriteMillis")
                     Log.w(OTEL_RUM_LOG_TAG, "overriding minFileAgeForReadMillis from $minFileAgeForReadMillis to $minRead")
                 }
+                var exportPeriod = exportPeriodMillis
+                if (exportPeriod <= 0) {
+                    exportPeriod = TimeUnit.SECONDS.toMillis(10)
```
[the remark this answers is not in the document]
thc1006: Done in 624436db, the non-positive fallback now uses DEFAULT_EXPORT_PERIOD_MS as well, so all three defaults stay in sync.

LikeTheSalad (approved):
LGTM. Thank you! 🙏

core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfig.kt
```diff
@@ -15,6 +15,7 @@ const val MAX_CACHE_FILE_SIZE: Int = 1024 * 1024
 const val DEFAULT_MAX_FILE_AGE_FOR_WRITE_MS = 30L
 const val DEFAULT_MIN_FILE_AGE_FOR_READ_MS = 33L
 const val DEFAULT_MAX_FILE_AGE_FOR_READ_MS = 18L
+const val DEFAULT_EXPORT_PERIOD_MS = 10L
```
breedx-splk: This constant is the number of seconds, so the `_MS` suffix here is confusing. Please rename it to `DEFAULT_EXPORT_PERIOD_S`. Thanks!

core/src/main/java/io/opentelemetry/android/features/diskbuffering/scheduler/DiskBufferingEnablement.kt
```diff
@@ -8,15 +8,18 @@ package io.opentelemetry.android.features.diskbuffering.scheduler
 import io.opentelemetry.android.features.diskbuffering.SignalFromDiskExporter
 import io.opentelemetry.android.internal.services.periodic.PeriodicTaskScheduler
 import java.util.concurrent.atomic.AtomicReference
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.seconds
 
 internal class DiskBufferingEnablement(
     private val signalFromDiskExporter: SignalFromDiskExporter,
     private val taskScheduler: PeriodicTaskScheduler,
+    private val exportPeriod: Duration = 10.seconds,
```
breedx-splk: This value is already defined in `DiskBufferingConfig`....so can we reuse it here instead of hard coding the literal?

breedx-splk:
Thanks for the contribution @thc1006, it's appreciated. I had two small requests around the default constants, but otherwise this looks good and I think we can get this in soon.

I think you were correct to follow the established conventions here, but in general, if we were to use Duration objects instead of longs holding milliseconds, much of these issues would go away. Doesn't need to be in this PR tho, could be a follow-up item. Thanks again! 🥂

thc1006:
> Thanks for the contribution @thc1006, it's appreciated. I had two small requests around the default constants, but otherwise this looks good and I think we can get this in soon.
> 
> I think you were correct to follow the established conventions here, but in general, if we were to use Duration objects instead of longs holding milliseconds, much of these issues would go away. Doesn't need to be in this PR tho, could be a follow-up item. Thanks again! 🥂

OK I will follow-up this issue.

core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfig.kt
DavidGrath: @breedx-splk  Oh, yeah, about that. I've recently been going through the codebase myself and came across this file.  
The other constants also feel a bit confusing since they're being converted from `TimeUnit.SECONDS` and `TimeUnit.HOURS` . Maybe this is a chance to rename the existing 3 accordingly? Or we leave them for now in anticipation of `Duration`?

thc1006: Thanks, both. Renamed it to `DEFAULT_EXPORT_PERIOD_S` (and updated `core.api`) in 98780156.

@DavidGrath fair point on the existing three. For now I've left `DEFAULT_MAX_FILE_AGE_FOR_WRITE_MS`, `DEFAULT_MIN_FILE_AGE_FOR_READ_MS`, and `DEFAULT_MAX_FILE_AGE_FOR_READ_MS` as they are, to keep this PR to its single purpose, and since they may well be reworked when disk buffering moves to `Duration` (#1821). Happy to rename them here too if you'd both prefer, or to open a small follow-up, whatever you think works best.

core/src/main/java/io/opentelemetry/android/features/diskbuffering/scheduler/DiskBufferingEnablement.kt
thc1006: Good point, thanks. It now reuses `DEFAULT_EXPORT_PERIOD_S` instead of the hard-coded literal, in 83fd7bc6.

core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfig.kt
DavidGrath: @thc1006, noted. Jason and Cesar might not be available this weekend, so for now you might need to hold on

thc1006: Thanks for the heads-up, @DavidGrath, no rush at all on my side. Everything raised so far is addressed and CI is green, so whenever Jason and Cesar have time is completely fine. Appreciate you taking the time to review.

breedx-splk: Yeah we can do all the Duration stuff in a follow-up effort. Feel free to create an issue if you'd like.

breedx-splk (approved):

PR #1841, opened 2026-06-25
Reorganise gesture instrumentation into separate modules

DavidGrath:
References #1576

To solve the issue of shared code being exposed, I created a new annotation similar to `Incubating`.
Also left 2 TODO's to be discussed.

I tried working on the Scale gesture as well, but it became so involved that I decided it should just be a follow-up: I added another module for the sake of working with Robolectric more effectively, amongst other things


instrumentation/view-common/src/main/kotlin/io/opentelemetry/android/instrumentation/view/common/ViewUtils.kt
LikeTheSalad, line 1: If I understand correctly, this is the only file that needs to be shared across the 2 `view-*` modules? If that's the case, I think the `common-api` might be too much of a generic place for it. This is the first time we want to split a couple of instrumentations that share the same "scope", so please bear with me as we go through the process of designing how these use cases should look like.

Ideally, I think we should have a "common-view-scoped" shared module, where we put this `ViewUtils` type. Regarding the folder structure, what do you think about the following:

```
instrumentation/
├─ view/
│  ├─ click/
│  ├─ scroll/
│  ├─ common/
```

instrumentation/common-api/src/main/java/io/opentelemetry/android/instrumentation/internal/InternalViewApi.kt
LikeTheSalad, line 1: I think this is a cool idea, but I'm not sure it's needed to go this far, given that the package for `ViewUtils` already contains `internal`; it should mean to any user that they shouldn't use that type because it can change at any time (we can add an explanatory comment like [this one](https://github.com/open-telemetry/opentelemetry-android/blob/2b861fbbe755aae1a814a51564eab4610d1bbd4e/core/src/main/java/io/opentelemetry/android/internal/initialization/InitializationEvents.kt#L12), for example).

LikeTheSalad (commented):
Thank you! 🙏 I like the idea in general, but I wanted to discuss more about the folder structure and where to put scoped common apis.

instrumentation/common-api/src/main/java/io/opentelemetry/android/instrumentation/internal/InternalViewApi.kt
DavidGrath: Thanks, I actually like this idea
My initial impression was that the module would be broken down into top-level instrumentations, but putting them under a common module as suggested above actually works better because now I can use `internal` exclusively. I'll work with that

breedx-splk:
Sorry about the merge conflicts @DavidGrath .... this got caught up in the re-linting of the entire kotlin codebase.

breedx-splk:
@DavidGrath there are a number of open/unanswered copilot review comments. Please have a look at those, respond or close. That can help move this forward (we don't want a bunch of dangling threads).

To me, it also looks like both `view-click` and `view-scroll` depend on `view-common`, but `view-common` doesn't include the `otel.publish-conventions`, so how is that supposed to work once those hit maven?

instrumentation/view-common/build.gradle.kts
```diff
@@ -0,0 +1,22 @@
+plugins {
+    id("otel.android-library-conventions")
```
breedx-splk, line 2: This probably also needs the publish conventions, so that the dependent modules can consume it from maven/sonatype after they are published.

DavidGrath:
So the main issue with the demo app was the fact that view click project clashed with the compose click project, and also, common clashed with the root common package. Updating the project names in the root settings file was how I got around it

DavidGrath:
This is the stacktrace that I used to decide to change the name of `:instrumentation:view:common` to `view-common`. It most likely clashed with the root `:common` project

```
FAILURE: Build failed with an exception.

* What went wrong:
Circular dependency between the following tasks:
:opentelemetry-android:instrumentation:view:common:generateReleaseLintModel
\--- :opentelemetry-android:instrumentation:view:common:generateReleaseLintModel (*)

(*) - details omitted (listed previously)
...
```

instrumentation/view/common/build.gradle.kts
```diff
@@ -0,0 +1,22 @@
+plugins {
+    id("otel.android-library-conventions")
```
DavidGrath: Done

demo-app/build.gradle.kts
```diff
@@ -73,7 +73,7 @@ dependencies {
     implementation("io.opentelemetry.android.instrumentation:sessions")
     implementation("io.opentelemetry.android.instrumentation:okhttp3-library")
     byteBuddy("io.opentelemetry.android.instrumentation:okhttp3-agent")
-    implementation("io.opentelemetry.android.instrumentation:view-click")
+    implementation("io.opentelemetry.android.instrumentation:view:click")
```
benjoseph-grafana, line 76: I believe this shouldn't change since the artifact id is still `view-click`

demo-app/settings.gradle.kts
```diff
@@ -36,7 +36,7 @@ includeBuild("..") {
             .using(project(":instrumentation:okhttp3:agent"))
         substitute(module("io.opentelemetry.android.instrumentation:android-instrumentation"))
             .using((project(":instrumentation:android-instrumentation")))
-        substitute(module("io.opentelemetry.android.instrumentation:view-click"))
-            .using(project(":instrumentation:view-click"))
+        substitute(module("io.opentelemetry.android.instrumentation:view:click"))
```
benjoseph-grafana, line 39: revert if updating the `build.gradle.kts` file

instrumentation/view/click/build.gradle.kts
```diff
@@ -19,6 +19,7 @@ dependencies {
     implementation(project(":agent-api"))
     implementation(project(":instrumentation:android-instrumentation"))
     implementation(project(":semconv"))
+    implementation(project(":instrumentation:view:view-common"))
```
benjoseph-grafana, line 22:
```suggestion
    implementation(project(":instrumentation:view:common"))
```

instrumentation/view/common/src/main/kotlin/io/opentelemetry/android/instrumentation/view/common/ViewClickActivityCallback.kt
```diff
@@ -3,21 +3,28 @@
...
-package io.opentelemetry.android.instrumentation.view.click
+package io.opentelemetry.android.instrumentation.view.common
 
 import android.app.Activity
+import android.view.Window
 import io.opentelemetry.android.internal.services.visiblescreen.activities.DefaultingActivityLifecycleCallbacks
 
-internal class ViewClickActivityCallback(
-    private val viewClickEventGenerator: ViewClickEventGenerator,
+class ViewClickActivityCallback(
```
benjoseph-grafana, line 12: nit: rename now that scroll uses it.

instrumentation/view/click/README.md
benjoseph-grafana, line 1: Still documents scroll/fling that click no longer emits

instrumentation/view/common/build.gradle.kts
```diff
@@ -0,0 +1,23 @@
...
+
+android {
+    namespace = "io.opentelemetry.android.instrumentation.view.common"
+
+    defaultConfig {
+        consumerProguardFiles("consumer-rules.pro")
+    }
+}
+
+dependencies {
```
benjoseph-grafana, line 16: missing `api(platform(libs.opentelemetry.platform.alpha))`?

instrumentation/view/click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
```diff
@@ -0,0 +1,176 @@
...
+        builder.put(APP_WIDGET_ID, view.id.toString())
+
+        builder.put(APP_SCREEN_COORDINATE_X, view.x.toLong())
+        builder.put(APP_SCREEN_COORDINATE_Y, view.y.toLong())
+        return builder.build()
+    }
+}
+
+internal const val APP_SCREEN_CLICK_EVENT_NAME = "app.screen.click"
+internal const val VIEW_CLICK_EVENT_NAME = "app.widget.click"
```
benjoseph-grafana, line 176: can we import these from semconv?

instrumentation/view/common/src/main/kotlin/io/opentelemetry/android/instrumentation/view/common/ViewUtils.kt
```diff
@@ -54,7 +57,7 @@ internal class TapEventMetadata(
     }
 }
 
-internal sealed class Gesture(
+sealed class Gesture(
```
benjoseph-grafana, line 60: nit: move these to an internal package and add docs "not for public use"

instrumentation/view/click/src/main/kotlin/io/opentelemetry/android/instrumentation/view/click/ViewClickEventGenerator.kt
```diff
@@ -0,0 +1,176 @@
...
+
+    override fun consumeEvent(event: MotionEvent?) {
+        if (event != null) {
+            gestureDetector.onTouchEvent(event)
+        }
+    }
+
+    override fun stopTracking() {
+        windowRef?.get()?.run {
+            if (callback is WindowCallbackWrapper) {
```
benjoseph-grafana, line 117: click and scroll both share one `WindowCallbackWrapper`, so one instrumentation can peel off the other.

benjoseph-grafana (commented):
hey @DavidGrath, some minor comments to get this merge ready

DavidGrath:
Thanks for the review, @benjoseph-grafana , sorry I missed these, I must have skipped the email somehow. I'll address them sometime within the following week

Issue #1842, opened 2026-06-26
Network instrumentation exposed from core


LikeTheSalad:
In the latest version `1.5.0` we accidentally added the [network change auto instrumentation](https://github.com/open-telemetry/opentelemetry-android/tree/main/instrumentation/network) as a [dependency of core](https://github.com/open-telemetry/opentelemetry-android/blob/56c2d4907159fea8c39ab93499d30d8810ed7b39/core/build.gradle.kts#L64), causing this instrumentation to become enabled transitively by just adding the core dependency into a consumer project.

We need to rework `core` to avoid exposing such instrumentation by default.

DavidGrath:
@LikeTheSalad closing this

PR #1847, opened 2026-06-30
Remove dependency from core to network instrumentation

breedx-splk:
Part of #1842. 

Note: this targets the 1.5.x branch so that we can start a patch release. We will cherry pick to `main` later.

In order for `core` to not depend on the network instrumentation, some movement was required.

* log and span appenders moved out of the network instrumentation and into `core`.
* non-instrumentation network code moved back into services, where it remains decoupled from the `ServiceFactory`.

No functional changes -- just moving things around and regenerating the api spec.


breedx-splk:
Bah, looks like I started from `main` instead of the release branch. Lemme fix that.

breedx-splk:
Oh nevermind, it's just that we broke CodeQL a while back I guess and it's just still failing.

core/src/main/java/io/opentelemetry/android/internal/processors/NetworkAttributesLogRecordAppender.kt
```diff
@@ -1,11 +1,7 @@
-/*
- * Copyright The OpenTelemetry Authors
- * SPDX-License-Identifier: Apache-2.0
- */
-
-package io.opentelemetry.android.instrumentation.network.internal
+package io.opentelemetry.android.internal.processors
```
DavidGrath, line 6: License header omission?

DavidGrath (commented):
LGTM otherwise

core/src/main/java/io/opentelemetry/android/internal/processors/NetworkAttributesLogRecordAppender.kt
breedx-splk: Super weird, yeah. There were 3 that were missing it! I've added it back. Thanks for catching it. 

Looks like the PR check is missing that lint step somehow. I'll look at fixing that too, so that this is automated.

DavidGrath: Just seeing this. I didn't get notified of this or your subsequent mention in #1853 either, weird

DavidGrath (approved):

core/src/main/java/io/opentelemetry/android/internal/processors/NetworkAttributesLogRecordAppender.kt
DavidGrath: Oh, okay, I found the email. My bad

LikeTheSalad (approved):
Cheers!

PR #1850, opened 2026-07-02
Check semconv generated sources in PR builds

breedx-splk:
Part of #1814. Basically phase 2.5. 😁 

This adds a PR build check to ensure that the source files generated from weaver are up to date when the yaml changes. This is one of the downsides of having the source checked in -- we run the risk of those files being out of sync with the yaml definitions. This build step helps guard against that.

This step only runs when there are changes in the semconv module, and it runs weaver and checks to see if there is any diff. If there is, then it will fail the pr check and output an error message showing how to regenerate the source.

### DIFF
--- a/.github/workflows/pr-check.yaml
+++ b/.github/workflows/pr-check.yaml
@@ -6,6 +6,68 @@ permissions:
   contents: read
 
 jobs:
+  semconv-generation:
+    name: "semconv-generation"
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


LikeTheSalad:
@breedx-splk - I think the changes from this other PR should prevent us from having to add these extra steps in the CI process: https://github.com/open-telemetry/opentelemetry-android/pull/1876 - please take a look 🙏

breedx-splk:
Going draft in favor of #1876. When that lands we can close this one entirely.

PR #1851, opened 2026-07-02
disable codeql

breedx-splk:
Every PR build now shows an ❌ in the favicon and in the status checks, because we broke CodeQL when we merged #1786. CodeQL does not yet support kotlin 2.4.0, so until it does, all of our builds will be "broken". CodeQL isn't in the critical path, but it's obnoxious to have all these false failures everywhere.

So let's disable it for now by renaming the workflow, so that we can easily rename it back.

Upstream tracking issue is here: https://github.com/github/codeql/issues/21938


bidetofevil (approved):
LGTM

PR #1856, opened 2026-07-02
Enable Kotlin Spotless checks

### LINKED ISSUE #1853: Fix spotless kotlin check/format
In [this comment](https://github.com/open-telemetry/opentelemetry-android/pull/1847#discussion_r3511042974) @DavidGrath caught a real problem with the spotless linting process -- and it appears that the spotless linting checks aren't running correctly for kotlin files. 🤦🏻 

So fixing that in CI/CD is the right thing to do, but it will immediately cause many many build failures because the source has drifted.

To work around that, I'll do this as a series of PRs that just lint each module in isolation. Once all modules pass the linter, I'll wire it back up to the CI/CD process.

breedx-splk:
resolves #1853

enables spotlessKotlin for Android modules. Once all the linting is done, we can rebase this and it should pass....but we can't merge this before that.

### DIFF
--- a/buildSrc/src/main/kotlin/otel.spotless-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.spotless-conventions.gradle.kts
@@ -4,23 +4,13 @@ plugins {
     id("com.diffplug.spotless")
 }
 
-var kotlinConfigured = false
-
 spotless {
     java {
         googleJavaFormat().aosp()
         licenseHeaderFile(rootProject.file("gradle/spotless.license.java"), "(package|import|public)")
         target("src/**/*.java")
     }
-    plugins.withId("org.jetbrains.kotlin.jvm") {
-        configureKotlinOnce(this@spotless)
-    }
-    plugins.withId("org.jetbrains.kotlin.android") {
-        configureKotlinOnce(this@spotless)
-    }
-    plugins.withId("kotlin-android") {
-        configureKotlinOnce(this@spotless)
-    }
+    configureKotlin(this@spotless)
     kotlinGradle {
         ktlint()
     }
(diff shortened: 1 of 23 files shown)


breedx-splk:
@open-telemetry/android-approvers ready for a look

buildSrc/src/main/kotlin/otel.spotless-conventions.gradle.kts
```diff
@@ -57,6 +62,15 @@ if (project == rootProject) {
     }
 }
 
+fun configureKotlinOnce(
+    spotlessExtension: SpotlessExtension,
+) {
+    if (!kotlinConfigured) {
```
LikeTheSalad: It seems like this is the one meaningful change, but I'm not sure I see the connection between adding this flag and actually applying the kotlin spotless checks :(

buildSrc/src/main/kotlin/otel.spotless-conventions.gradle.kts
```diff
@@ -4,17 +4,22 @@ plugins {
...
     }
     plugins.withId("org.jetbrains.kotlin.jvm") {
-        configureKotlin(this@spotless)
+        configureKotlinOnce(this@spotless)
     }
     plugins.withId("org.jetbrains.kotlin.android") {
-        configureKotlin(this@spotless)
+        configureKotlinOnce(this@spotless)
+    }
+    plugins.withId("kotlin-android") {
```
LikeTheSalad: After seeing this line, I suspect this might be the reason why the Spotless check stopped working for Kotlin sources. When we bumped AGP to version 9.x, we had to remove the explicit kotlin android plugin declaration, so this block never executed. Maybe what we could do is to always call `configureKotlin` unconditionally and remove both `plugins.withId` blocks that check for the existence of a Kotlin android plugin (lines 18 and 21)?

breedx-splk: oh yeah great catch! I forgot about that history, and this works well. I'll make that fix. Appreciate it!

LikeTheSalad (approved):
Cheers! It looks like there are still some files left that need to get reformatted, but other than that this looks good to me.

fractalwrench (approved):
LGTM

PR #1859, opened 2026-07-02
Re-lint the instrumentation modules

breedx-splk:
Applies spotless to the instrumentation modules, as part of #1853.

Sorry that this one got a little big -- I figured it would be kinda stupid to do all the individual instrumentation modules separately.


instrumentation/view-click/src/test/kotlin/TestUtils.kt
```diff
@@ -75,66 +92,156 @@ fun getDoubleTapSequence(x: Float, y: Float, toolType: Int = MotionEvent.TOOL_TY
...
+                arrayOf(pointerCoords),
+                0,
+                buttonState,
+                1f,
+                1f,
+                0,
+                0,
+                0,
+                0,
+            ),
```
DavidGrath:
```
The function getDoubleTapSequence is too long (150). The maximum length is 60. [LongMethod]
```
Oh, my. Do we add these as exceptions to `detekt` or perhaps I create a wrapper function over `MotionEvent` to reduce the line proliferation?

LikeTheSalad (approved):

instrumentation/view-click/src/test/kotlin/TestUtils.kt
breedx-splk: yeah I just went ahead and did refactoring to that test helper to reduce method length.

DavidGrath: Alright. Thanks

PR #1860, opened 2026-07-02
Re-lint the remaining modules

breedx-splk:
applies spotless to the remaining modules, as part of #1853.

this picks up:

* agent-api
* common
* test-common
* semconv (egad)

Should be the last of them.


fractalwrench (approved):
LGTM. Ignoring the CodeQL job as that doesn't support Kotlin 2.4 right now.

PR #1861, opened 2026-07-02
Add stacktrace logging to snapshot publish

breedx-splk:
The build is broken when trying to publish snapshots for :instrumentation:okhttp3-websocket:library. This adds --stacktrace and --info so that we can do additional troubleshooting.


bidetofevil (approved):
LGTM

PR #1865, opened 2026-07-05
Documentation fixes

DavidGrath:
Various changes made relating to markdown and KDoc as I went through the code recently.
For the `@see` part, I noticed it wasn't making links properly when I hovered over it, and I found it's because the tag requires a class or method identifier


demo-app/README.md
```diff
@@ -66,9 +66,6 @@ The OpenTelemetry Android Demo App currently supports the following features:
 ### Known Gaps
 As of now, there are a few areas where the instrumentation might not be comprehensive:
 
-* HTTP Client Instrumentation  
-OpenTelemetry Android supports automatic instrumentation for HTTP client libraries. This feature captures spans for HTTP requests with details. However, the demo app does not currently demonstrate this feature as it doesn't make any network requests.
```
breedx-splk, line 70: Why remove this?

DavidGrath: @breedx-splk, mainly because of the introduction of `OkHttpDemoActivity` and the use of httpbin

DavidGrath: That is, from #1688

breedx-splk: Ah! Thanks.

.lychee.toml
```diff
@@ -4,7 +4,7 @@ max_retries = 6
 max_concurrency = 4
 
 # Check link anchors
-include_fragments = true
+include_fragments = "full"
```
breedx-splk, line 7: I was a little surprised to see this change among docs changes.....but the docs say that this causes it to check both anchors and full urls, so `true` seemed like a bunk value anyway.

breedx-splk (approved):
This looks good. The copilot suggestion about indenting is valid tho eh?

.lychee.toml
DavidGrath: Oh, for this, I downloaded the most recent lychee binary and ran it locally and the true caused it to fail, that was why. I could have also tried with the original version from the docker file as well, though

DavidGrath:
> This looks good. The copilot suggestion about indenting is valid tho eh?

Yeah, true. Done now

PR #1868, opened 2026-07-06
Remove dependency from core to network instrumentation (#1847)

breedx-splk:
This is just a cherry pick of the commits in #1847

* remove dependency from core to network instrumentation
* add missing headers


fractalwrench (approved):
LGTM once merge conflicts are resolved

PR #1872, opened 2026-07-06
update lychee version

breedx-splk:
Renovate had not been updating lychee because of how we had the version expressed!

When a recent PR changed `= true` to `= full`', it broke the main build a yet another way. So this update should:

* get us on the latest
* fix the build
* allow renovate to update us with new versions.

<phew>


breedx-splk:
related to #1865

fractalwrench (approved):
LGTM

LikeTheSalad (approved):

PR #1873, opened 2026-07-06
Increase metaspace and stop daemon before publish

breedx-splk:
Well after much flailing, it looks like the next thing to try and fix the main build is to beef up the metaspace and to stop the daemon between the build and publish phases.


fractalwrench (approved):
LGTM

LikeTheSalad (approved):

PR #1876, opened 2026-07-07
Semconv build time

### LINKED ISSUE #1867: Avoid tracking generated semconv sources in git
We should generate those at build time, similarly to how protobuf sources are generated. Plus, once that's done, we shouldn't need this workaround in the CI actions: https://github.com/open-telemetry/opentelemetry-android/pull/1850

LikeTheSalad:
Closes #1867 

These changes do the following:

- Enable semconv's compilation process to download the weaver binary into the project's `build` dir to use it from there and to avoid asking contributors to install it on their machines (these changes **DO NOT** install weaver as an OS-wide app, only put weaver's binary into the OTel Android's "semconv/build" dir, so it's only usable from within OTel Android's semconv tasks).
- Attaches the `generateSemanticConventions` task to "semconv"'s build process, so that the generated Kotlin files are created automatically without having to manually run `generateSemanticConventions`.
- Configures Renovate to update weaver's version automatically.

Both tasks (downloading weaver and generating Kotlin sources) are cacheable, so they will only run when some changes are available. For example, if weaver is downloaded on a first run, the second run won't do it again (unless weaver's version is manually changed). Same for generating Kotlin sources with "generateSemanticConventions", it will only run when there are semconv changes available.


gradle/libs.versions.toml
```diff
@@ -21,6 +21,8 @@ fragment = "1.8.9"
 koverGradlePlugin = "0.9.8"
 openTelemetryProto = "1.10.0-alpha"
 protobufKotlin = "4.35.1"
+weaver = "0.24.2"
+tukaaniXz = "1.10"
```
fractalwrench: Is this necessary? opentelemetry-kotlin does something similar without an extra dependency, so it'd be good to know one way another whether the kotlin repository is missing some advantage of this approach: https://github.com/open-telemetry/opentelemetry-kotlin/blob/main/semconv/build.gradle.kts#L74

fractalwrench (commented):
I had a question about how the weaver archive is opened but think this is the right approach to take

breedx-splk:
Sorry about the huge set of linting changes....this needs a rebase.

.github/renovate.json5
```diff
@@ -99,5 +99,25 @@
...
     }
-  ]
+  ],
+  customManagers: [
+    {
+      // weaver's version in the catalog isn't tied to a Maven coordinate (it's a CLI binary
+      // downloaded from GitHub Releases by semconv/build.gradle.kts), so the default gradle
+      // manager can't see it. Tell Renovate to treat it as a github-releases dependency instead.
+      // Release tags are "v"-prefixed (e.g. v0.24.2) but the catalog value isn't, matching how
+      // the download URL is built (`"v$version"`), so strip the prefix when comparing/writing.
```
breedx-splk, line 109: oh yeah this is cool. I've done similar before with setting up maven/gradle to source from gh releases, but not renovate. ...stoked that this can work!

gradle/libs.versions.toml
breedx-splk: This additional/new 3rd party library is probably a deal-breaker for me. I like the approach here overall, so I created an example PR against your PR (https://github.com/LikeTheSalad/opentelemetry-android/pull/2) showing how this might be avoided. There are likely to be other approaches as well.

LikeTheSalad: > Is this necessary? opentelemetry-kotlin does something similar without an extra dependency, so it'd be good to know one way another whether the kotlin repository is missing some advantage of this approach

It seems like the Kotlin repository is downloading a `.zip` file, which is natively supported by Gradle, so for that use case it's fine. I think Gradle also supports the `tar.gz` format, but unfortunately it doesn't support `tar.xz`, which for whatever reason is the one they decided to use for weaver's non-Windows packages 🤷‍♂️

> This additional/new 3rd party library is probably a deal-breaker for me. I like the approach here overall, so I created an example PR against your PR (https://github.com/LikeTheSalad/opentelemetry-android/pull/2) showing how this might be avoided

Thank you, @breedx-splk 🙏 this is a very clever solution. I knew that this third-party lib wasn't going to be popular, so I tried multiple options to avoid using it without luck, but I didn't think of using an OS CLI tool. I guess the only concern is that it has to be present in every non-Windows OS, but I think that's already the case, so it should be fine. I just updated this PR to use your approach, cheers!

.github/renovate.json5
LikeTheSalad: It is pretty cool! I learned about it thanks to this PR from trask: https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2193 - I'm not sure it's properly configured, though, but I think there's no way to test it locally, so let's see what happens 🤞

fractalwrench (approved):
LGTM

PR #1879, opened 2026-07-07
Remove anchor from google link, fix lychee

breedx-splk:
Builds are failing the link checker due to this dynamic anchor or whatever. I don't think it's that important, and I think it's newer behavior that started failing after our lychee upgrade yesterday.


DavidGrath (approved):
Aw, man. Minor bummer that it needs to be broken apart now

breedx-splk:
> Aw, man. Minor bummer that it needs to be broken apart now

Yeah, well I figure it was the lazy way of getting it working again. :)

Issue #197, opened 2023-12-20
Use real time from an NTP server


LikeTheSalad:
Android devices' clocks can be inaccurate, which can cause issues when analyzing distributed traces. One option that helps and doesn't rely on vendor-specific capabilities would be to get the real-time from an NTP server and set it in a custom Clock implementation when initializing the Android RUM agent.

breedx-splk:
Are you aware of or does Elastic have prior work around this feature? I think this seems like a good idea, perhaps with a couple small downsides:

* increased network overhead (tho ntp is pretty small indeed)
* slightly more complicated clock behavior when ntp is unavailable or when telemetry is created _during_ ntp time sync.
* time-based correlation between otel data and in-app data sourced from system clock becomes difficult.

LikeTheSalad:
We have it in Elastic atm, I'd say the most concerning downside is when some telemetry is created before the ntp clock is available, although the TrueTime lib caches the time in the disk so the issue should only happen the first time the app is open (unless there's a network connectivity issue or something that prevents the request from succeeding). I'm also not sure if we should do some periodic updates to it even if it's already cached, so I think there's quite a lot to discuss about it in general, however, it seems to be the only solution that only depends on the mobile device, there might be other alternatives to this issue but would require making the servers aware of it afaik. I think @marandaneto has some insights on this.

marandaneto:
Might be needed for v1, assuming a device has the wrong date and time, and after upgrading the SDK with an NTP server, the telemetry timestamps may seen off (before and after the upgrade).

LikeTheSalad:
Update: It seems like Android might have come up with a [solution](https://developer.android.com/reference/android/os/SystemClock#currentNetworkTimeClock()) (for API 33+ though 😅).

marandaneto:
another alternative just out of the oven https://android-developers.googleblog.com/2025/02/trustedtime-api-introducing-reliable-approach-to-time-keeping-for-apps.html?m=1

marandaneto:
I am using https://developer.android.com/reference/android/os/SystemClock#currentNetworkTimeClock() for >= TIRAMISU and its working well so no need for a 3rd party

fractalwrench:
I'd suggest we discuss this when #1487 goes through review or close this issue

DavidGrath:
Re-opening

PR #1880, opened 2026-07-08
Anchor OtelAndroidClock to GNSS or Network time when available

lemnik:
Addresses #197 by attempting to anchor the `OtelAndroidClock` against either the GNSS (navigation satellite) provided time, or the network time (NTP) when they are available on the device.

The priority order is:

- GNSS time
- Network time
- Device wall time

The PR doesn't attempt to identify which clock was used as the anchor point since the previous "wall time only" behaviour already caused drift problems, since its not clear what (if anything) could be done with that information and this should remain a generic implementation.

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/OtelAndroidClock.kt
+++ b/core/src/main/java/io/opentelemetry/android/OtelAndroidClock.kt
@@ -5,22 +5,53 @@
 
 package io.opentelemetry.android
 
+import android.os.Build
 import android.os.SystemClock
+import androidx.annotation.RequiresApi
 import io.opentelemetry.sdk.common.Clock
 
 /**
  * An implementation of [Clock] that takes a baseline by subtracting
- * [SystemClock.elapsedRealtimeNanos] from [System.currentTimeMillis] and then
+ * [SystemClock.elapsedRealtimeNanos] from the best available wall time and then
  * adds that baseline to [SystemClock.elapsedRealtimeNanos] to provide a monotonic wall-clock time.
+ *
+ * When selecting the anchor clock we try (in order):
+ * - [SystemClock.currentGnssTimeClock]
+ * - [SystemClock.currentNetworkTimeClock]
+ * - [System.currentTimeMillis]
  *
  * This avoids the downside of [System.currentTimeMillis] not being monotonic on Android (i.e.
  * the clock doesn't consistently tick when the process is in the background or cached). It also
  * avoids the problem of [SystemClock.elapsedRealtimeNanos] not providing wall-clock time.
  */
 class OtelAndroidClock : Clock {
-    private val baselineNanos = System.currentTimeMillis() * 1_000_000 - nanoTime()
+    private val baselineNanos = currentTimeMillis() * 1_000_000 - nanoTime()
 
     override fun now(): Long = baselineNanos + nanoTime()
 
     override fun nanoTime(): Long = SystemClock.elapsedRealtimeNanos()
+
+    private fun currentTimeMillis(): Long {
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/opentelemetry/android/OtelAndroidClock.kt
```diff
@@ -5,22 +5,53 @@
...
+    private val baselineNanos = currentTimeMillis() * 1_000_000 - nanoTime()
 
     override fun now(): Long = baselineNanos + nanoTime()
 
     override fun nanoTime(): Long = SystemClock.elapsedRealtimeNanos()
+
+    private fun currentTimeMillis(): Long {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+            try {
+                return currentGnssTimeMillis()
```
LikeTheSalad, line 50: I would like to understand what this does a bit better since I'm not familiar with it. It looks like it uses the device's GPS, if so, does that mean that it requires some some of manifest permissions?

core/src/main/java/io/opentelemetry/android/OtelAndroidClock.kt
```diff
@@ -5,22 +5,53 @@
...
  */
 class OtelAndroidClock : Clock {
-    private val baselineNanos = System.currentTimeMillis() * 1_000_000 - nanoTime()
+    private val baselineNanos = currentTimeMillis() * 1_000_000 - nanoTime()
 
     override fun now(): Long = baselineNanos + nanoTime()
 
     override fun nanoTime(): Long = SystemClock.elapsedRealtimeNanos()
+
+    private fun currentTimeMillis(): Long {
```
LikeTheSalad, line 47: Based on the SIG meeting yesterday, my understanding is that we'd like these NTP/GNSS time sources to be opt-in, at least for now. Maybe we can add some config in the DSL for it?

core/src/main/java/io/opentelemetry/android/OtelAndroidClock.kt
lemnik: It does use the GPS/Navigation satellite based time, but indirectly so it doesn't require any special permissions. It can throw exceptions when there haven't been any location requests. In theory any location fix by any app sets or corrects this clock, but like the `currentNetworkTimeMillis` clock once they're set they continue to run.

core/src/main/java/io/opentelemetry/android/OtelAndroidClock.kt
lemnik: Given that we already have `OpenTelemetryConfiguration.clock` maybe the opt-in is some boolean flags on the `OtelAndroidClock` constructor (defaulting to `false`)?

core/src/main/java/io/opentelemetry/android/OtelAndroidClock.kt
LikeTheSalad: Got it. Thank you for the details.

core/src/main/java/io/opentelemetry/android/OtelAndroidClock.kt
LikeTheSalad: > some boolean flags on the OtelAndroidClock constructor (defaulting to false)

I like this idea. It doesn't clearly show that this option is available, as it's not directly part of the DSL config, but given there were a lot of concerns in the SIG meeting about what would happen if the time source changes often or the like, I think it's good to do this until we're ready to advertise it more clearly when we are more confident about its stability.

LikeTheSalad (approved):
Thank you! 🙏

Issue #764, opened 2025-01-24
Support for Native Crash Reporting in OpenTelemetry Android


MdNaushad:
OpenTelemetry-Android currently does not handle native crashes (e.g., crashes caused by C/C++ code within Android applications)

Native crashes are common in applications that use JNI (Java Native Interface) for interacting with native libraries or in apps that include high-performance, low-level code written in C or C++. These crashes can result in a poor user experience, and without proper reporting, developers lack visibility into the root causes.

JuozasIzi1:
second that

breedx-splk:
One challenge to this that we have discussed on SIG meetings is the fact that currently the android agent doesn't persist its session information. When a native crash happens, the app likely won't have the ability to respond to it until next launch -- at which point the session information is lots. 

Some vendors periodically write state to the device, which allows them to place the native crash information in the correct session on relaunch. We should consider something similar.

fractalwrench:
An alternative approach is [ApplicationExitInfo](https://developer.android.com/reference/android/app/ApplicationExitInfo) which works for native crashes from API 33+, which doesn't require any signal handlers _or_ for opentelemetry-android to persist data. It has downsides (the API level, and requires symbolicating the file to get useful info)

ashishvz:
Hi @fractalwrench and @breedx-splk,

I would like to research and explore the possibilities of native crash detection and reporting. Embrace android SDK has a dedicated module for this signal. 

Refer to -> https://github.com/embrace-io/embrace-android-sdk/tree/main/embrace-android-instrumentation-crash-ndk

Let me know your thought's!

yduartep:
What about extending the existing CrashReporter instrumentation to replay **REASON_CRASH_NATIVE** via **ApplicationExitInfo** on the next launch (API 30+)?

That’s how we handle it in **Grafana Faro React Native SDK**: no signal handlers, no persisting session state at crash time. On startup we read `getHistoricalProcessExitReasons()`, open `getTraceInputStream()` when present (protobuf tombstone on API 31+), normalize to a compact text backtrace, and emit a **device.crash** log with the tombstone in a **trace** attribute. Symbolication stays server-side in our case.

Session attribution is imperfect in **our case** — the crash is reported under the **new session** after relaunch, not the one active when the process died. We attach **ApplicationExitInfo metadata** (`timestamp`, `pid`, `processName`, `signal`) for correlation instead.

ashishvz:
[...]

Issue #1403, opened 2025-11-19
Incorporate `versionCode` in `service.version` and start adding `app.build_id`


atulgpt:
Currently in `AndroidResource` we use [`[packageInfo.versionName](https://github.com/open-telemetry/opentelemetry-android/blob/0cfb9b555d47f0df26553676754c61a933ece2c4/core/src/main/java/io/opentelemetry/android/AndroidResource.kt#L62)`](https://github.com/open-telemetry/opentelemetry-android/blob/0cfb9b555d47f0df26553676754c61a933ece2c4/core/src/main/java/io/opentelemetry/android/AndroidResource.kt#L62) as the `service.version` value here:
[https://github.com/open-telemetry/opentelemetry-android/blob/150491e66d5401dfdb8acb01df56d61ed60ddb38/core/src/main/java/io/opentelemetry/android/AndroidResource.kt#L31](https://github.com/open-telemetry/opentelemetry-android/blob/150491e66d5401dfdb8acb01df56d61ed60ddb38/core/src/main/java/io/opentelemetry/android/AndroidResource.kt#L31)

Should we update this to combine both `versionName` and `versionCode`? Since Android apps in the Play Store can share the same `versionName`, it likely makes sense to use `versionName_versionCode` instead.

Otel also defines [`[app.build_id](https://opentelemetry.io/docs/specs/semconv/registry/attributes/app/#app-build-id)`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/app/#app-build-id) as:

> A unique identifier representing the installation of an application on a specific device [1]

We should also start populating that as currently we are not doing that?

We should therefore start populating `app.build_id`, and if we do this, we may also need two additional attributes to report `versionName` and `versionCode` separately so those values aren’t lost.

breedx-splk:
Related to #922.

Thanks @atulgpt, but I feel that this issue is a bit misleading. Yes, we do read the `packageInfo.versionName` in the `AndroidResource`, but we only use it to populate the `service.version` resource attribute. Are we populating `app.build_id` anywhere yet? I don't think so.

So maybe you can rework this to indicate that you want us to start populating the `app.build_id` resource attribute, and the suggestion here is what you think it could contain. Make sense?

atulgpt:
Yes you are correct, I had some local change and got confused. 
> Are we populating app.build_id anywhere yet?

Yes we don't populate that info

[...]

PR #1887, opened 2026-07-10
feat(native-crash): add persisted crash replay

aranhave:
## Summary

- add a standalone native-crash instrumentation module
- replay a persisted native crash marker as `app.crash` on the next app startup
- persist the crashed process session, app version/build, and OS context
- ignore missing or malformed markers and replay each valid marker only once

Because the crash is reported by a later app process, the persisted context is attached to the event instead of relying on the new process resource values.

## Scope

This is the first part of #764 and only adds the replay path. A follow-up PR will add the JNI/C signal handler that writes the marker when a native crash occurs.

Native stack unwinding, symbol upload, symbolication, and backend changes are intentionally out of scope.

## Testing

- focused replay tests covering valid, missing, malformed, and consumed markers
- session context update test
- module formatting, checks, API validation, and debug/release assembly

### DIFF
--- a/instrumentation/native-crash/README.md
+++ b/instrumentation/native-crash/README.md
@@ -1,0 +1,40 @@
+# Native Crash Instrumentation
+
+Status: development
+
+The native crash instrumentation replays a persisted native crash as an `app.crash` event when
+the application next starts.
+
+This first increment provides the module, persisted marker/context format, and replay path. Native
+signal capture is intentionally left for a follow-up change.
+
+## Telemetry
+
+The replayed event uses the original crash timestamp and includes:
+
+* `exception.type`
+* `exception.message`
+* `session.id`, when available
+* `app.build_id`, when available
+* `service.version`, when available
+* `os.name`
+* `os.version`
+
+The app and OS fields are persisted before a crash so the replayed event describes the process that
+crashed rather than the process that reports it.
+
+## Installation
+
+Add the instrumentation dependency:
+
+```kotlin
+implementation("io.opentelemetry.android.instrumentation:native-crash:1.5.0-alpha")
+```
+
+The module is discovered and installed automatically when it is present on the runtime classpath.
+
+## Limitations
+
+Native signal handling, native stack unwinding, symbol upload, and symbolication are not included
+in this increment. Until signal handling is added, this module only provides the replay side of the
+native crash reporting flow.
--- a/instrumentation/native-crash/README.md
+++ b/instrumentation/native-crash/README.md
@@ -5,8 +5,9 @@ Status: development
 The native crash instrumentation replays a persisted native crash as an `app.crash` event when
 the application next starts.
 
-This first increment provides the module, persisted marker/context format, and replay path. Native
-signal capture is intentionally left for a follow-up change.
+This first increment provides the module, persisted marker/context format, and replay path. Each
+crash uses a separate marker so unreadable records can be retried without blocking other crashes.
+Native signal capture is intentionally left for a follow-up change.
 
 ## Telemetry
 
@@ -37,8 +38,9 @@ The module is discovered and installed automatically when it is present on the r
 
 Native signal handling, native stack unwinding, symbol upload, and symbolication are not included
 in this increment. Until signal handling is added, this module only provides the replay side of the
-native crash reporting flow.
+native crash reporting flow. It does not emit native stack frames, so there is no native stack to
+symbolicate in this change.
 
-The persisted crash marker is deleted immediately after the event is emitted. Replay is therefore
-at most once: if the application exits before the telemetry is exported, the crash event may be
-lost.
+Each persisted crash marker is deleted immediately after its event is emitted. Replay is therefore
+at most once per marker: if the application exits before the telemetry is exported, that crash
+event may be lost.
--- a/instrumentation/native-crash/README.md
+++ b/instrumentation/native-crash/README.md
@@ -16,7 +16,6 @@ The replayed event uses the original crash timestamp and includes:
 * `exception.type`
 * `exception.message`
 * `session.id`, when available
-* `app.build_id`, when available
 * `service.version`, when available
 * `os.name`
 * `os.version`
--- a/instrumentation/native-crash/README.md
+++ b/instrumentation/native-crash/README.md
@@ -6,8 +6,9 @@ The native crash instrumentation replays a persisted native crash as an `app.cra
 the application next starts.
 
 This first increment provides the module, persisted marker/context format, and replay path. Each
-crash uses a separate marker so unreadable records can be retried without blocking other crashes.
-Native signal capture is intentionally left for a follow-up change.
+crash uses a separate marker containing its signal metadata and crash-time context. Unreadable
+records can be retried without blocking other crashes. Native signal capture is intentionally left
+for a follow-up change.
 
 ## Telemetry
 
@@ -20,8 +21,8 @@ The replayed event uses the original crash timestamp and includes:
 * `os.name`
 * `os.version`
 
-The app and OS fields are persisted before a crash so the replayed event describes the process that
-crashed rather than the process that reports it.
+The app and OS fields are stored with each crash marker so the replayed event describes the process
+that crashed rather than the process that reports it.
 
 ## Installation
 
@@ -35,10 +36,10 @@ The module is discovered and installed automatically when it is present on the r
 
 ## Limitations
 
-Native signal handling, native stack unwinding, symbol upload, and symbolication are not included
-in this increment. Until signal handling is added, this module only provides the replay side of the
-native crash reporting flow. It does not emit native stack frames, so there is no native stack to
-symbolicate in this change.
+Native signal handling and native stack capture are not included in this increment. Until signal
+handling is added, this module only provides the replay side of the native crash reporting flow.
+It does not create or attach a binary crash dump. Symbol upload and symbolication are downstream
+concerns and require a separate design once native stack frames are available.
 
 Each persisted crash marker is deleted immediately after its event is emitted. Replay is therefore
 at most once per marker: if the application exits before the telemetry is exported, that crash
--- a/instrumentation/native-crash/README.md
+++ b/instrumentation/native-crash/README.md
@@ -38,3 +38,7 @@ The module is discovered and installed automatically when it is present on the r
 Native signal handling, native stack unwinding, symbol upload, and symbolication are not included
 in this increment. Until signal handling is added, this module only provides the replay side of the
 native crash reporting flow.
+
+The persisted crash marker is deleted immediately after the event is emitted. Replay is therefore
+at most once: if the application exits before the telemetry is exported, the crash event may be
+lost.
--- a/instrumentation/native-crash/README.md
+++ b/instrumentation/native-crash/README.md
@@ -5,10 +5,9 @@ Status: development
 The native crash instrumentation replays a persisted native crash as an `app.crash` event when
 the application next starts.
 
-This first increment provides the module, persisted marker/context format, and replay path. Each
-crash uses a separate marker containing its signal metadata and crash-time context. Unreadable
-records can be retried without blocking other crashes. Native signal capture is intentionally left
-for a follow-up change.
+This first increment provides the module, persisted marker/context format, and replay path. It uses
+one marker for the most recent crash and a separate context snapshot maintained while the app is
+running. Native signal capture is intentionally left for a follow-up change.
 
 ## Telemetry
 
@@ -21,8 +20,8 @@ The replayed event uses the original crash timestamp and includes:
 * `os.name`
 * `os.version`
 
-The app and OS fields are stored with each crash marker so the replayed event describes the process
-that crashed rather than the process that reports it.
+The app and OS fields are read from the persisted crash-time context before it is replaced with the
+new process context, so the replayed event describes the process that crashed.
 
 ## Installation
 
@@ -41,6 +40,7 @@ handling is added, this module only provides the replay side of the native crash
 It does not create or attach a binary crash dump. Symbol upload and symbolication are downstream
 concerns and require a separate design once native stack frames are available.
 
-Each persisted crash marker is deleted immediately after its event is emitted. Replay is therefore
-at most once per marker: if the application exits before the telemetry is exported, that crash
-event may be lost.
+The persisted crash marker is deleted immediately after its event is emitted. Replay is therefore
+at most once: if the application exits before the telemetry is exported, that crash event may be
+lost. A later change may add support for preserving multiple consecutive startup crashes.
+Unreadable or malformed markers are discarded rather than retried.
--- a/instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
+++ b/instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
@@ -31,6 +31,7 @@ import io.opentelemetry.kotlin.semconv.SessionAttributes.SESSION_ID
 import java.io.File
 import java.io.FileInputStream
 import java.io.FileOutputStream
+import java.io.IOException
 import java.time.Instant
 import java.util.Properties
 
@@ -160,6 +161,7 @@ internal class FileNativeCrashStore(
         )
     }
 
+    @Synchronized
     override fun writeContext(context: NativeCrashContext) {
         runCatching {
             directory.mkdirs()
@@ -169,7 +171,15 @@ internal class FileNativeCrashStore(
             properties.setIfNotNull(SERVICE_VERSION, context.serviceVersion)
             properties.setIfNotNull(OS_NAME, context.osName)
             properties.setIfNotNull(OS_VERSION, context.osVersion)
-            FileOutputStream(contextPath).use { properties.store(it, null) }
+            val temporaryPath = File(directory, "${contextPath.name}.tmp")
+            try {
+                FileOutputStream(temporaryPath).use { properties.store(it, null) }
+                if (!temporaryPath.renameTo(contextPath)) {
+                    throw IOException("Failed to replace native crash context")
+                }
+            } finally {
+                temporaryPath.delete()
+            }
         }
     }
 
--- a/instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
+++ b/instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
@@ -58,7 +58,8 @@ class NativeCrashInstrumentation internal constructor(
             NativeCrashReporter(
                 store = store,
                 openTelemetryRum = openTelemetryRum,
-            ).install(crashContext)
+            ).replayPreviousCrash()
+            store.writeContext(crashContext)
 
             val sessionProvider = openTelemetryRum.sessionProvider
             if (sessionProvider is SessionPublisher) {
@@ -89,12 +90,7 @@ internal class NativeCrashReporter(
     private val store: NativeCrashStore,
     private val openTelemetryRum: OpenTelemetryRum,
 ) {
-    fun install(currentContext: NativeCrashContext) {
-        replayPreviousCrash()
-        store.writeContext(currentContext)
-    }
-
-    private fun replayPreviousCrash() {
+    fun replayPreviousCrash() {
         val crashContext = store.readContext()
         store.readCrashRecord()?.let { record -> replay(record, crashContext) }
     }
@@ -252,6 +248,8 @@ internal data class NativeCrashRecord(
             7 -> "SIGBUS"
             8 -> "SIGFPE"
             11 -> "SIGSEGV"
+            13 -> "SIGPIPE"
+            31 -> "SIGSYS"
             else -> "SIG$signalNumber"
         }
 }
--- a/instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
+++ b/instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
@@ -10,7 +10,6 @@ package io.opentelemetry.android.instrumentation.nativecrash
 import android.content.Context
 import android.os.Build
 import android.util.Log
-import androidx.core.content.pm.PackageInfoCompat
 import com.google.auto.service.AutoService
 import io.opentelemetry.android.OpenTelemetryRum
 import io.opentelemetry.android.common.RumConstants
@@ -22,7 +21,6 @@ import io.opentelemetry.android.session.SessionPublisher
 import io.opentelemetry.api.common.AttributeKey.stringKey
 import io.opentelemetry.api.common.Attributes
 import io.opentelemetry.api.common.AttributesBuilder
-import io.opentelemetry.kotlin.semconv.AppAttributes.APP_BUILD_ID
 import io.opentelemetry.kotlin.semconv.ExceptionAttributes.EXCEPTION_MESSAGE
 import io.opentelemetry.kotlin.semconv.ExceptionAttributes.EXCEPTION_TYPE
 import io.opentelemetry.kotlin.semconv.IncubatingApi
@@ -197,7 +195,6 @@ internal class FileNativeCrashStore(
         val properties = runCatching { contextPath.readProperties() }.getOrNull() ?: return null
         return NativeCrashContext(
             sessionId = properties.nonBlankProperty(SESSION_ID),
-            appBuildId = properties.nonBlankProperty(APP_BUILD_ID),
             serviceVersion = properties.nonBlankProperty(SERVICE_VERSION),
             osName = properties.nonBlankProperty(OS_NAME),
             osVersion = properties.nonBlankProperty(OS_VERSION),
@@ -210,7 +207,6 @@ internal class FileNativeCrashStore(
             directory.mkdirs()
             val properties = Properties()
             properties.setIfNotNull(SESSION_ID, context.sessionId)
-            properties.setIfNotNull(APP_BUILD_ID, context.appBuildId)
             properties.setIfNotNull(SERVICE_VERSION, context.serviceVersion)
             properties.setIfNotNull(OS_NAME, context.osName)
             properties.setIfNotNull(OS_VERSION, context.osVersion)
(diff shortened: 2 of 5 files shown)


instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
```diff
@@ -0,0 +1,281 @@
...
+        attributes.putIfNotNull(OS_NAME, osName)
+        attributes.putIfNotNull(OS_VERSION, osVersion)
+    }
+}
+
+private fun Context.currentCrashContext(openTelemetryRum: OpenTelemetryRum): NativeCrashContext {
+    val packageInfo = runCatching { packageManager.getPackageInfo(packageName, 0) }.getOrNull()
+    return NativeCrashContext(
+        sessionId = openTelemetryRum.sessionProvider.getSessionId().takeIf { it.isNotBlank() },
+        appBuildId = packageInfo?.let { PackageInfoCompat.getLongVersionCode(it).toString() },
```
aranhave: Is Android `versionCode` the right value for `app.build_id` here, or should we use a different build identifier? I couldn’t find an existing Android SDK precedent for this.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
```diff
@@ -0,0 +1,281 @@
...
+) {
+    fun install(currentContext: NativeCrashContext) {
+        replayPreviousCrash()
+        store.writeContext(currentContext)
+    }
+
+    private fun replayPreviousCrash() {
+        val record = store.readCrashRecord() ?: return
+        val attributes = Attributes.builder()
+        attributes.put(stringKey(EXCEPTION_TYPE), "signal.${record.signalName}")
```
aranhave: Should `exception.type` include the `signal.` prefix, like `signal.SIGSEGV`, or should it use the bare signal name? I want to confirm the expected shape before we establish it here.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
```diff
@@ -0,0 +1,281 @@
...
+private fun Context.currentCrashContext(openTelemetryRum: OpenTelemetryRum): NativeCrashContext {
+    val packageInfo = runCatching { packageManager.getPackageInfo(packageName, 0) }.getOrNull()
+    return NativeCrashContext(
+        sessionId = openTelemetryRum.sessionProvider.getSessionId().takeIf { it.isNotBlank() },
+        appBuildId = packageInfo?.let { PackageInfoCompat.getLongVersionCode(it).toString() },
+        serviceVersion = packageInfo?.versionName,
+        osName = "Android",
+        osVersion = Build.VERSION.RELEASE,
+    )
+}
```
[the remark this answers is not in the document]
aranhave: I tried the guarded API 33 overload, but this repo’s AnimalSniffer rejects `PackageInfoFlags` because the module targets the API 23 signature. The existing `AndroidResource` path uses the same legacy overload and lint passes, so I’m keeping it consistent here.

android-agent/src/test/kotlin/io/opentelemetry/android/OtelAndroidClockTest.kt
```diff
@@ -46,10 +46,11 @@ class OtelAndroidClockTest {
     fun `uses GNSS time when available on API 29+`() {
         every { SystemClock.currentGnssTimeClock() } returns gnssClock
 
-        val clock = OtelAndroidClock(
```
DavidGrath: Please could you re merge with main? I think @breedx-splk has resolved the lint inconsistency issues so that these unrelated changes as a result of you doing what you were meant to do by applying spotless won't be here

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
```diff
@@ -0,0 +1,291 @@
...
+            stringKey(EXCEPTION_MESSAGE),
+            "Native crash signal ${record.signalName} (${record.signalNumber})",
+        )
+        store.readContext()?.addTo(attributes)
+
+        openTelemetryRum.openTelemetry.logsBridge
+            .loggerBuilder("io.opentelemetry.native-crash")
+            .build()
+            .logRecordBuilder()
+            .setEventName(map("app.crash"))
```
DavidGrath, line 114: Oh, this also happened very recently by Jason: I think you can use the generated semantic convention event classes from the Federated semconv plan instead of building the plain logger here

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
DavidGrath: Hmm, there's actually an old issue concerning this same topic that I'd like to dig up if I remember it so that it can be aligned

DavidGrath (changes requested):
All right. That's all I have to say for this review

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
DavidGrath: Oh, his PR is still open. Don't bother for now

DavidGrath:
#1403 for the build id thing 
I kept getting a weird error when trying to reply my original comment so new thread

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
DavidGrath: Aaaand, it was just recently merged. Please check out what was done in #1888 and try to mimic it

instrumentation/native-crash/README.md
```diff
@@ -0,0 +1,44 @@
+# Native Crash Instrumentation
+
+Status: development
+
+The native crash instrumentation replays a persisted native crash as an `app.crash` event when
```
fractalwrench, line 5: How would symbolication work with `app.crash`? That's potentially more of a question around the semantic convention itself, but I'd like to know what approach we plan on taking if we're unwinding the stack in a signal handler as that won't give readable stackframes out-of-the-box.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
```diff
@@ -0,0 +1,291 @@
...
+    )
+
+    override val name: String = "native-crash"
+
+    override fun install(
+        context: Context,
+        openTelemetryRum: OpenTelemetryRum,
+    ) {
+        val applicationContext = context.applicationContext
+        val store = storeFactory(applicationContext)
```
fractalwrench: I believe this causes I/O on the main thread. Ideally creating this directory would be handed off to a background thread immediately during SDK initialization

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
fractalwrench: I'd lean towards `SIGSEGV` although tbh I'm not sure whether the existing semantic convention fits what we're trying to capture. Open to thoughts from others.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
```diff
@@ -0,0 +1,291 @@
...
+
+    fun readContext(): NativeCrashContext?
+
+    fun writeContext(context: NativeCrashContext)
+}
+
+internal class FileNativeCrashStore(
+    private val directory: File,
+) : NativeCrashStore {
+    private val crashRecordPath = File(directory, "native-crash.properties")
```
fractalwrench: This adds a constraint that only one native crash File can exist on disk at any one time. IMO it would be preferable to remove this constraint so that the instrumentation handles the case where it's not possible to load a file due to transient errors (memory, I/O, etc), _and_ so that the native crash handling can be installed as early as possible

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
```diff
@@ -0,0 +1,291 @@
...
+import io.opentelemetry.kotlin.semconv.SessionAttributes.SESSION_ID
+import java.io.File
+import java.io.FileInputStream
+import java.io.FileOutputStream
+import java.io.IOException
+import java.time.Instant
+import java.util.Properties
+
+/** Entry point for replaying native crashes captured by a previous app process. */
+@AutoService(AndroidInstrumentation::class)
```
fractalwrench, line 41: It'd probably be good to install this as one of the first instrumentations. I can't recall whether we do that for JVM crash reporting or other instrumentation, but if not then we should do

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
fractalwrench: A concat of `versionCode/versionName/packageName` would be a reasonable first attempt IMO, although really for reliability we want a UUID that's injected into the APK on each build. There is an issue for adding that somewhere in the backlog...

aranhave: I removed app.build_id from this PR and left #1403 to define it SDK-wide. versionCode alone is not a reliable build ID.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
aranhave: Changed this to SIGSEGV.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
aranhave: app.crash is not in the generated registry yet, and the JVM crash reporter still uses the logger, so I kept this path consistent.

instrumentation/native-crash/README.md
aranhave: This PR only replays crash metadata, so there are no native frames yet. I clarified that in the README; frame capture and symbolication will be follow-up work.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
aranhave: Moved store setup and replay off the main thread.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
aranhave: Switched to one marker per crash, kept transient read failures for retry, and added a multi-record test.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
aranhave: This PR only replays existing markers, so install order does not affect capture yet. I will add early priority with the signal-handler follow-up, where it matters.

DavidGrath:
Hmm, am I blocking this because I requested for an edit? Let me confirm

DavidGrath (approved):
Okay. Looks good to me

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
```diff
@@ -0,0 +1,316 @@
...
+class NativeCrashInstrumentation internal constructor(
+    private val storeFactory: (Context) -> NativeCrashStore = { context ->
+        FileNativeCrashStore(File(context.filesDir, "opentelemetry/native-crash"))
+    },
+    private val executor: Executor = Executors.newSingleThreadExecutor(),
+) : AndroidInstrumentation {
+    constructor() : this(
```
LikeTheSalad: Do we need this second constructor? It seems redundant.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
```diff
@@ -0,0 +1,316 @@
...
+            stringKey(EXCEPTION_MESSAGE),
+            "Native crash signal ${record.signalName} (${record.signalNumber})",
+        )
+        crashContext?.addTo(attributes)
+
+        openTelemetryRum.openTelemetry.logsBridge
+            .loggerBuilder("io.opentelemetry.native-crash")
+            .build()
+            .logRecordBuilder()
+            .setEventName(map("app.crash"))
```
LikeTheSalad, line 114: I understand that this `map` function is meant to map new names to old ones (like `device.crash` in this case), is just that I'm not sure we're supposed to call it directly like this. This is all fairly new, though, so it might need to stay as is for now, but I just wanted to confirm with @breedx-splk, who's been working on these conventions recently, if there's a better way to create the event without worrying about the mapping?

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
```diff
@@ -0,0 +1,316 @@
...
+    private val store: NativeCrashStore,
+    private val openTelemetryRum: OpenTelemetryRum,
+) {
+    fun install(currentContext: NativeCrashContext) {
+        replayPreviousCrashes()
+        store.writeContext(currentContext)
+    }
+
+    private fun replayPreviousCrashes() {
+        val crashContext = store.readContext()
```
LikeTheSalad, line 94: Considering the flow of:

- App starts
- There's a native crash that closes the app
- App restarts
- We call this `replayPreviousCrashes()` function, where we read the stored context using `store.readContext()`.

And considering that we rewrite the stored context's session id [here](https://github.com/open-telemetry/opentelemetry-android/pull/1887/changes#diff-cfe6778cfd114472472f5665fe15de63a3ff00b726679dc1c77d700c3cabf94fR87), in a callback that gets initialized [here](https://github.com/open-telemetry/opentelemetry-android/pull/1887/changes#diff-cfe6778cfd114472472f5665fe15de63a3ff00b726679dc1c77d700c3cabf94fR67) before we call `replayPreviousCrashes()`, I'm concerned about a potential race condition in which we rewrite the context's session id before we send the crash (effectively using a new session in the crash that happened on a previous session).

Now, the concern I mentioned shouldn't happen with the code as is, but this is because we're using `Executors.newSingleThreadExecutor()` as the default executor, which will enqueue the call to overwrite the session id from the callback, so that it happens after `replayPreviousCrashes()` is called, but it still seems a bit fragile to me, as we rely on this particular executor implementation to be used.

An idea to avoid this issue could be calling `sessionProvider.addObserver` after `NativeCrashReporter.install` inside the `NativeCrashInstrumentation.install` function.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
```diff
@@ -0,0 +1,316 @@
...
+        crashContext?.addTo(attributes)
+
+        openTelemetryRum.openTelemetry.logsBridge
+            .loggerBuilder("io.opentelemetry.native-crash")
+            .build()
+            .logRecordBuilder()
+            .setEventName(map("app.crash"))
+            .setTimestamp(record.timestamp)
+            .setAllAttributes(attributes.build())
+            .emit()
```
LikeTheSalad, line 117: I'm not familiar with NDK errors, but I'm guessing that the upcoming C signal handler would create a binary dump file to investigate the error? If that's the case, where are you planning to send it in this log event? Because right now, it seems like we're planning to send the error type and error message fields only. I'd expect we should retrieve some crash dump from the `NativeCrashStore` object.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
```diff
@@ -0,0 +1,316 @@
...
+        store.deleteCrashRecord(record)
+    }
+}
+
+internal interface NativeCrashStore {
+    fun readCrashRecords(): List<NativeCrashRecord>
```
LikeTheSalad: It seems like this design expects many crashes to be stored at a given time. I'm guessing it's because of the possibility that the latest native crash cannot be exported successfully during an app relaunch (maybe due to network connectivity issues or the like). However, I see that only one `NativeCrashContext` is stored and used for all the previously stored crashes. I think this could cause an issue where some very old crashes might wrongly get assigned the latest available context.

Having said that, **we might not need to store multiple crash records at all**. If the reason for having several "NativeCrashRecord" stored is to avoid losing them in case they cannot be successfully exported on an app restart, then I think we don't have to worry about it, because we use [disk-buffering](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/disk-buffering) to store signals on disk right after their creation. So, because of this, I think we can simplify this module by just storing a single "NativeCrashRecord" and emitting it on an app relaunch (which you already do in the `replay()` function), which will automatically get buffered on disk to send it later.

instrumentation/native-crash/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+```kotlin
+implementation("io.opentelemetry.android.instrumentation:native-crash:1.5.0-alpha")
+```
+
+The module is discovered and installed automatically when it is present on the runtime classpath.
+
+## Limitations
+
+Native signal handling, native stack unwinding, symbol upload, and symbolication are not included
```
LikeTheSalad: While I agree with this, I think `symbol upload` and `symbolication` are processes that go beyond collecting telemetry, as these are needed after the collection when users need to visualize events, and this paragraph seems to suggest that we can do so from here, so I'm a bit confused by it.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
fractalwrench: I suggested multiple crash records in [this comment](https://github.com/open-telemetry/opentelemetry-android/pull/1887#discussion_r3571264577) which gives some of the reasoning. Disk buffering is a helpful feature in general but if the data can't even be loaded for ingestion due to temporary circumstances, then a single file approach will cause data loss

LikeTheSalad: I see. Thank you for the clarification, @fractalwrench. However, I'm a bit confused on what reasons could cause a read error that happens only once. If for example, we get an IO error due to the file being corrupted somehow, then I think we should remove it in that case, and right now it seems like we're confident that we will be able to read it properly next time, but if this keeps happening, we'll just fill the disk with unreadable files. The one use case that I see more plausible is what you mentioned about not having enough memory available to read it, but how often could that happen? Is there another use case where we might give second chances to reading the same file?

fractalwrench: Another scenario is intermittent NDK crashes on startup. If two or more crashes happen in a row before the instrumentation loads, then using only a single file would lead to data loss

LikeTheSalad: I see. Thanks. I don't discard those possibilities; it's just that they seem quite edge cases to me and fall into the diminishing returns space effort-wise. However, I'm fine with adding support for those, but if we're going to do it, I think it should be done properly to avoid the following issues:

* Using a single context data for multiple crash records (as the context might be too new for the oldest ones), so if we go down the multiple crash records route, we should apply the same for the context data for each.
* Keeping files forever until we can successfully read them (as this can cause us to take up disk space unnecessarily). We should somehow tell when a read error is permanent or transient, and delete the file in the former case. If not possible, then at least we should keep a max read attempt per file and dispose of it if it's reached.

fractalwrench: It's an edge case but I've been burned by it before. In the interest of getting _some_ instrumentation in place I'd also be happy just to go with a single file to start off, and open a follow-on issue to improve the persistence approach given that it adds quite a bit of complexity?

LikeTheSalad: Sounds good to me! Thanks.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
aranhave: Removed it. The defaulted primary constructor still provides the public no-arg constructor AutoService needs.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
aranhave: I checked the generated conventions and the JVM crash reporter. There isn’t a generated `app.crash` event yet and the JVM path also uses map, so I kept this consistent for now.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
aranhave: Good catch. I moved observer registration after replay/context persistence and added a direct-executor test for this.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
aranhave: The signal-handler follow-up will only write the marker. it won’t create or attach a binary dump. Native stack capture needs a separate design, so I’m keeping it out of these first two PRs.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
aranhave: I simplified this to a single marker for now. We can handle consecutive startup crashes in a follow-up.

instrumentation/native-crash/README.md
aranhave: Updated this to make it clear that symbol processing happens downstream and isn’t handled by this instrumentation.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
```diff
@@ -0,0 +1,312 @@
...
+        openTelemetryRum: OpenTelemetryRum,
+    ) {
+        val applicationContext = context.applicationContext
+        executor.execute {
+            val store = storeFactory(applicationContext)
+            val crashContext = applicationContext.currentCrashContext(openTelemetryRum)
+            NativeCrashReporter(
+                store = store,
+                openTelemetryRum = openTelemetryRum,
+            ).install(crashContext)
```
lemnik: Assuming that `install` will (in a future PR) install the signal handler and pass the store path over to the native code, shouldn't that remain on the synchronous path, while `replay` happens on the background thread? So something more like:

```kotlin
val nativeReporter = NativeCrashReporter(store, openTelemetryRum)
// make sure the native side is configured before returning
nativeReporter.install(crashContext)
executor.execute {
  nativeReporter.replay()
```

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
```diff
@@ -0,0 +1,312 @@
...
+) {
+    val signalName: String =
+        when (signalNumber) {
+            4 -> "SIGILL"
+            5 -> "SIGTRAP"
+            6 -> "SIGABRT"
+            7 -> "SIGBUS"
+            8 -> "SIGFPE"
+            11 -> "SIGSEGV"
+            else -> "SIG$signalNumber"
```
lemnik, line 253: Probably worth including `SIGPIPE` and `SIGSYS` as well

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
aranhave: 💯 . I split replay from install so the disk work stays on the executor. The signal-handler PR can add synchronous native setup before replay is scheduled.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
aranhave: Added both and covered them in the signal mapping test.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
aranhave: Confirmed in the SIG: generated app.crash support would be useful, but it should be separate follow-up work. I’ll keep the manual `setEventName(map("app.crash"))` approach in this PR.

LikeTheSalad (approved):
Thank you! 🙏

aranhave:
Thanks for the reviews! @LikeTheSalad, could you merge this when you have a chance? I don’t have merge access, and I’m ready to start the signal-handler follow-up.

PR #1888, opened 2026-07-10
Federated semconv phase 5 - use generated event classes

breedx-splk:
Part of #1814.

The previous phase created event classes based on the semantic convention definitions, but those event classes were not yet used by production code. This now wires up the events to production code.

A couple caveats:

* the view click instrumentation still generates its own events (the old way), because those event definitions exist in upstream (not our local federation) and do not have event classes generated for them.
* the startup progress events still do their own event generation, because they need to override the timestamps, which is not something that the generated event classes allow/provide.

### DIFF
--- a/android-agent/src/test/kotlin/io/opentelemetry/android/OtelAndroidClockTest.kt
+++ b/android-agent/src/test/kotlin/io/opentelemetry/android/OtelAndroidClockTest.kt
@@ -46,11 +46,10 @@ class OtelAndroidClockTest {
     fun `uses GNSS time when available on API 29+`() {
         every { SystemClock.currentGnssTimeClock() } returns gnssClock
 
-        val clock =
-            OtelAndroidClock(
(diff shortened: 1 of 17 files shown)


DavidGrath (commented):
Ah, a spotless fail. Leaving the corrections in kinda bloats the PR, but leaving it out fails the build
How come the recent clock changes fail, though?

breedx-splk:
> Ah, a spotless fail. Leaving the corrections in kinda bloats the PR, but leaving it out fails the build How come the recent clock changes fail, though?

Some linting problems snuck into main. I will rebase this PR after the fix in #1889 lands.

DavidGrath:
> Some linting problems snuck into main. I will rebase this PR after the fix in #1889 lands.

Alright 👍

fractalwrench (approved):
LGTM

LikeTheSalad (approved):
Cheers!

PR #1895, opened 2026-07-14
Generate source code for app.jank event appropriate to use in this project

bidetofevil:
Generate the class for the `app.jank` event so that it uses the Java `Logger` interface instead. This demonstrates how we can directly import them to get these specific events into the combined manifest for this registry. We can import these differently if we want, but I feel like we should do that after the federated repo setup has been established.

### DIFF
--- a/instrumentation/slowrendering/build.gradle.kts
+++ b/instrumentation/slowrendering/build.gradle.kts
@@ -24,6 +24,7 @@ dependencies {
     implementation(project(":session"))
     implementation(project(":common"))
     implementation(project(":agent-api"))
+    implementation(project(":semconv"))
     implementation(libs.androidx.core)
     implementation(libs.opentelemetry.semconv.kotlin)
     implementation(libs.opentelemetry.sdk)
--- a/semconv/model/android/registry.yaml
+++ b/semconv/model/android/registry.yaml
@@ -1,3 +1,8 @@
+# Imports events defined in dependency registries, by event name, so event classes are
+# generated for them in this module.
+imports:
+  events:
+    - app.jank
 groups:
   - id: registry.android.semconv
     type: attribute_group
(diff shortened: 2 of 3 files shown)


semconv/model/android/registry.yaml
```diff
@@ -1,3 +1,8 @@
+# Imports events defined in dependency registries, by event name, so event classes are
+# generated for them in this module.
+imports:
+  events:
+    - app.jank
```
breedx-splk, line 5: 🪄 THIS IS MAGICAL 🪄

instrumentation/slowrendering/build.gradle.kts
```diff
@@ -24,6 +24,7 @@ dependencies {
     implementation(project(":session"))
     implementation(project(":common"))
     implementation(project(":agent-api"))
+    implementation(project(":semconv"))
```
aranhave, line 27: Do we still need `opentelemetry.semconv.kotlin` as a runtime dependency here? After this change, it looks like only the tests use its constants. the leftover `@OptIn` in `EventJankReporter` can go too.

aranhave (commented):
Nice cleanup! Just one small dependency question from me.

instrumentation/slowrendering/build.gradle.kts
bidetofevil: I thought we were still referencing the constants in OTel Kotlin -- but yeah, we don't really need to since we could be generating them locally too (and maybe already are)

Basically, if we want, we can generate all of our semantic convention constants locally, including everything upstream, IF we want to do that.

aranhave: Yep, I think the local generation already covers this one 👍

LikeTheSalad (approved):
Thank you, Hanson! Based on what we discussed in the SIG meeting yesterday, I think we can add the rest of upstream conventions, though it's fine whether you decide to do the rest in this PR or if you want to split it into multiple ones, in case the change is too big.

fractalwrench (approved):

PR #1898, opened 2026-07-15
Support enum and array based attributes in codegen

bidetofevil:
Fixed attribute templates to work with enum and array-based attribute values. Add in some codegen to create enums and then changed a few internal semconv to use them.


LikeTheSalad (approved):
Thanks! I think it's worth checking the comment from Copilot, but overall LGTM.

fractalwrench (approved):
LGTM

breedx-splk (approved):
Awesome, thanks!!

PR #1899, opened 2026-07-15
feat(native-crash): capture fatal native signals

aranhave:
## Summary

This is the second part of the native crash reporting work for #764 and builds on the persisted replay path added in #1887. It adds the native layer that catches fatal process signals and writes the marker consumed on the next app start.

- handles `SIGILL`, `SIGTRAP`, `SIGABRT`, `SIGBUS`, `SIGFPE`, `SIGSEGV`, and `SIGSYS`
- writes the signal number and crash timestamp without allocating or calling into JNI from the signal handler
- replays any previous marker and persists the current process context before enabling the signal handler
- leaves ignored signals alone and restores only the crashing signal's previous action before re-delivering it through the kernel
- preserves the previous handler's mask and flags without changing registrations for the other fatal signals
- preserves the original `siginfo` when the kernel supports it; the older-kernel fallback only raises signals that will not recur on their own, so synchronous fault details such as `si_addr` are not replaced
- prepares marker storage and uses an existing alternate signal stack, or installs one on the setup thread when needed

Native stack capture, crash dumps, unwinding, symbolication, and backend work remain outside this PR.

## Validation

- `./gradlew check koverXmlReport`
- native debug and release builds for `arm64-v8a`, `armeabi-v7a`, `x86`, and `x86_64`
- Android 15 signal-chain checks covering a previous handler's mask, flags, and `siginfo`; a handler installed after OpenTelemetry; ignored signals; an existing alternate stack; and the older-kernel fallback paths
- existing Android demo app on Android 17: after triggering `SIGSEGV`, the next launch emitted one `app.crash` using the persisted crash-time context, and a second relaunch did not emit a duplicate

The dashboard below is a local OTLP test receiver and is not part of this PR.

Native crash replay validation

Related to #764
Follow-up to #1887


instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashJni.java
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.android.instrumentation.nativecrash;
+
+final class NativeCrashJni {
```
fractalwrench: Kotlin is preferred for new code - is there a reason this needs to be in Java?

instrumentation/native-crash/src/main/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+cmake_minimum_required(VERSION 3.22.1)
```
fractalwrench, line 4: Probably worth adding a note in the instrumentation README of the min supported cmake version

instrumentation/native-crash/src/main/cpp/native_crash_handler.c
```diff
@@ -0,0 +1,319 @@
...
+#define ALTERNATE_STACK_SIZE (SIGSTKSZ * 2)
+
+static const int handled_signals[SIGNAL_COUNT] = {
+    SIGILL,
+    SIGTRAP,
+    SIGABRT,
+    SIGBUS,
+    SIGFPE,
+    SIGSEGV,
+    SIGPIPE,
```
fractalwrench: My understanding was that this signal doesn't terminate the process on Android by default. Is that correct. Also I've not run into SIGSYS before.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashJni.java
aranhave: I used Java while experimenting with the JNI boundary in this draft but there’s no reason it needs to stay that way. I’ll switch it to Kotlin soon before switching to ready for review

instrumentation/native-crash/src/main/cpp/native_crash_handler.c
aranhave: 💯  I’ll remove SIGPIPE. I included SIGSYS for seccomp violations, which Android’s debuggerd treats as fatal, so I’ll keep that one.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
```diff
@@ -52,23 +53,67 @@ class NativeCrashInstrumentation internal constructor(
         openTelemetryRum: OpenTelemetryRum,
     ) {
         val applicationContext = context.applicationContext
+        val store = storeFactory(applicationContext)
         executor.execute {
-            val store = storeFactory(applicationContext)
             val crashContext = applicationContext.currentCrashContext(openTelemetryRum)
```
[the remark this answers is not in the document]
aranhave: Good catch. I moved store creation into the executor and added a test for the ordering.

core/src/test/java/io/opentelemetry/android/SdkPreconfiguredRumBuilderTest.kt
```diff
@@ -49,10 +51,47 @@ class SdkPreconfiguredRumBuilderTest {
...
         builder.addInstrumentation(sessionInstrumentation)
 
         val result = builder.getEnabledInstrumentations()
 
-        assertThat(result[0]).isSameAs(sessionInstrumentation)
+        assertThat(result)
+            .containsExactly(
+                sessionInstrumentation,
+                nativeCrashInstrumentation,
+                fooInstrumentation,
```
breedx-splk, line 63: minor nit: the native crash was the second added, so its position doesn't change. Either move it or add another foo/fake instrumentation to verify that it moves position.

instrumentation/native-crash/src/main/cpp/native_crash_handler.c
```diff
@@ -0,0 +1,323 @@
...
+#include <stdatomic.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <string.h>
+#include <time.h>
+#include <unistd.h>
+
+#define SIGNAL_COUNT 7
+#define TEMPORARY_SUFFIX ".tmp"
+#define MARKER_BUFFER_SIZE 128
```
fractalwrench, line 23: Assuming this is just a temporary size and will be increased when the native stacktrace starts being written to the file

instrumentation/native-crash/src/main/cpp/native_crash_handler.c
```diff
@@ -0,0 +1,323 @@
...
+};
+
+static char crash_record_path[PATH_MAX];
+static char temporary_crash_record_path[PATH_MAX];
+static struct sigaction previous_actions[SIGNAL_COUNT];
+static atomic_bool handler_active[SIGNAL_COUNT];
+static unsigned char alternate_signal_stack[ALTERNATE_STACK_SIZE];
+static pthread_mutex_t install_mutex = PTHREAD_MUTEX_INITIALIZER;
+static bool handlers_installed = false;
+static bool alternate_signal_stack_installed = false;
```
fractalwrench, line 48: consider `static volatile`

fractalwrench (approved):
LGTM

instrumentation/native-crash/src/main/cpp/native_crash_handler.c
aranhave: Yes, 128 is only for the current signal/timestamp marker. If we add native stack data later, we’ll revisit the format and buffer size.

instrumentation/native-crash/src/main/cpp/native_crash_handler.c
aranhave: This is only accessed during installation while install_mutex is held and the signal handler doesn’t read it, so I don’t think volatile is needed here.

aranhave:
@breedx-splk @fractalwrench Thanks for the reviews! This is ready for merge whenever someone with merge access has time 🙇

breedx-splk (dismissed):
@aranhave Thanks so much for this PR and to help bring native crash handling to this instrumentation. It's appreciated.

I am still in the process of reviewing, both because it's kinda long and because my C is a little rusty. :)

There were a couple of things that were raised by codex that I think we should at least discuss to determine how risky/problematic they are and what we might do about it. I'm doing the "request changes" (blocking) thing only because I want to make sure that we're all in agreement about this behavior before proceeding. I'm NOT saying we shouldn't do this.

## previous handler semantics

When the previous handler is invoked, this invokes it with different semantics. It is invoked directly from the opentelemetry handler, and passes its own mask and flags (and user context?) and not necessarily the original. This could potentially cause the other handler to do something unexpected?

## potential ordering issue

When a signal is handled in `handle_signal()` it is invoked for a specific signal. When we call `invoke_prevoius_handler()`, we end up also calling `restore_previous_handlers()`. As I understand it, the risk here is that another crash library could install after OpenTelemetry and later chain into it...and then we would replace that newer library’s registrations with stale pre-OpenTelemetry actions. Please double check my understanding of this current behavior -- I want to make sure we're not breaking or nerfing other crash handling libs that might come after us in installation order.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
DavidGrath, line 237:
```suggestion
                FileOutputStream(temporaryPath).use { properties.store(it, null) }
                if (contextPath.exists() && !contextPath.delete()) {
                    throw IOException("Failed to replace native crash context")
                }
```

Apologies for the sidetrack, but I encountered what I believe is a Windows-specific issue with `renameTo` that broke my local tests and the above change lets them pass.

Image

DavidGrath: Basically, it seems to fail if the file already exists

aranhave: Thanks for checking this on Windows. I kept the normal rename first, then added a fallback that deletes the existing file and retries.

aranhave:
> @aranhave Thanks so much for this PR and to help bring native crash handling to this instrumentation. It's appreciated.
> 
> I am still in the process of reviewing, both because it's kinda long and because my C is a little rusty. :)
> 
> There were a couple of things that were raised by codex that I think we should at least discuss to determine how risky/problematic they are and what we might do about it. I'm doing the "request changes" (blocking) thing only because I want to make sure that we're all in agreement about this behavior before proceeding. I'm NOT saying we shouldn't do this.
> 
> ## previous handler semantics
> When the previous handler is invoked, this invokes it with different semantics. It is invoked directly from the opentelemetry handler, and passes its own mask and flags (and user context?) and not necessarily the original. This could potentially cause the other handler to do something unexpected?
> 
> ## potential ordering issue
> When a signal is handled in `handle_signal()` it is invoked for a specific signal. When we call `invoke_prevoius_handler()`, we end up also calling `restore_previous_handlers()`. As I understand it, the risk here is that another crash library could install after OpenTelemetry and later chain into it...and then we would replace that newer library’s registrations with stale pre-OpenTelemetry actions. Please double check my understanding of this current behavior -- I want to make sure we're not breaking or nerfing other crash handling libs that might come after us in installation order.

Thanks for digging into this. I agree with your concerns 💯 . I updated the handoff so it restores and re-delivers only the signal being handled. That lets the previous handler run with its own mask and flags without touching registrations for the other signals. I also tested the chain on Android 15 with another handler installed after OpenTelemetry.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
DavidGrath: Thanks

instrumentation/native-crash/src/test/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashReporterTest.kt
```diff
@@ -137,6 +334,26 @@ class NativeCrashReporterTest {
         assertThat(store.readContext()).isEqualTo(crashContext("current"))
     }
 
+    @Test
+    fun `reads the native marker format with nanosecond precision`() {
+        val store = FileNativeCrashStore(tempDir)
+        markerFile().apply {
+            parentFile?.mkdirs()
+            writeText(
```
LikeTheSalad, line 342: Not a blocker, but since we must rely on both the native code and the Kotlin one to agree on this file's content format/shape, I think it'd be ideal to have some sort of test that checks the e2e scenario where the file is actually written by the C code, rather than mocked here (maybe by writing an android instrumentation test, as we've done for some other instrumentations). Probably worth checking as a future addition.

core/src/main/java/io/opentelemetry/android/SdkPreconfiguredRumBuilder.kt
```diff
@@ -119,6 +119,11 @@ class SdkPreconfiguredRumBuilder internal constructor(
             result.remove(it)
             result.add(0, it)
         }
+        val nativeCrashInstrumentation = instrumentations.find { it.name == "native-crash" }
```
LikeTheSalad, line 122: I don't love that we must force some ordering here, as we need to make the core aware of stuff such as instrumentation names defined elsewhere... But I see that it's a pattern we've used in the past, so it's probably the best option we have now, but I'd be interested in finding a way to avoid this altogether.

LikeTheSalad (approved):
I added some minor comments, where really only one of them might require some future changes (not a blocker). Otherwise it looks good to me. Thank you!

instrumentation/native-crash/src/test/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashReporterTest.kt
aranhave: 💯 Agreed, that would be useful as a follow-up. An instrumentation test covering the native writer and Kotlin reader together would better protect the shared marker format for sure.

core/src/main/java/io/opentelemetry/android/SdkPreconfiguredRumBuilder.kt
LikeTheSalad: I created this issue to follow up on this separately, as it's not related to this PR: https://github.com/open-telemetry/opentelemetry-android/issues/1914

aranhave: Thanks for opening #1914. That works for me 👍

instrumentation/native-crash/src/test/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashReporterTest.kt
aranhave: Tracked this in #1915 so it doesn't get lost 👍

fractalwrench (approved):
LGTM

breedx-splk (approved):
Thanks for staying with this, @aranhave.

I think the ordering problem could still exist, but it's probably an unlikely edge case and I don't want to block this PR any longer. Let's get this merged and proceed, and I'll create a follow-up issue to track this potential ordering problem. Thanks again.

Issue #361, opened 2024-05-14
Create Compose navigation instrumentation


breedx-splk:
The existing Activity and Fragment instrumentation via [VisibleScreenTracker](https://github.com/open-telemetry/opentelemetry-android/blob/main/instrumentation/activity/src/main/java/io/opentelemetry/android/instrumentation/activity/VisibleScreenTracker.java) is nice and helpful RUM information. Some apps, however, use (Jetpack) Compose to build sophisticated UI structures, including `@Composable` components that behave (to the user) much like a Fragment or an Activity would.

Compose has a Navigation component that app authors can use to create user flows between these Compose style "screens". This navigation is not yet instrumented, and thus the current screen might appear to be reported incorrectly, if at all. 

Not entirely sure yet what this should look like, but we should build Compose Navigation instrumentation.

LikeTheSalad:
I agree. And I think it would be great if we could come up with a single api to instrument all kinds of screen types, kind of a [VisibleScreenTracker](https://github.com/open-telemetry/opentelemetry-android/blob/main/instrumentation/activity/src/main/java/io/opentelemetry/android/instrumentation/activity/VisibleScreenTracker.java) but for all use cases, Activities, Fragments, and Compose alike.

benjoseph-grafana:
@breedx-splk I'd like to pick this up. Here's a proposed shape, sketching it here for feedback before I open a PR.

### Primary API: an extension on NavController
```
// The canonical entry point — works for ANY controller, however it was obtained.
@Composable
fun NavController.withOpenTelemetry(
    getScreenNameFromDestination: (NavDestination, Bundle?) -> String
): NavController {
    DisposableEffect(this) {
        val navigationListener = NavController.OnDestinationChangedListener { _, dest, args ->
            emitNavigationEvent(
                getScreenNameFromDestination(
                    dest,
                    args
                )
            )
        }
        addOnDestinationChangedListener(navigationListener)
        onDispose { removeOnDestinationChangedListener(navigationListener) }
    }
    return this
}
```
I'd lead with the extension (on the base NavController, so nested/child controllers work too) rather than a rememberNavController wrapper, because a lot of apps don't create the controller inline — it's hoisted, DI-injected, or ViewModel-owned. An extension operates on a reference the developer already holds, so coverage is "can you get the controller?" rather than "did you create it our way?". On top of that I am hoping to add a rememberObservedNavController() convenience that just delegates to the extension.

### What it emits

[...]

PR #1901, opened 2026-07-16
feat(compose-navigation): add manual Compose Navigation instrumentation (Nav2)

benjoseph-grafana:
### Summary

Add a new `instrumentation/compose/navigation` module that emits a screen-view event on each Jetpack Compose Navigation destination change. This is the first change to address https://github.com/open-telemetry/opentelemetry-android/issues/361

Public API:
- `NavController.withOpenTelemetry(rum, screenName)` — base-class extension covering any controller you hold, including nested/child controllers.
- `NavHostController.withOpenTelemetry(...)` overload returning `NavHostController`, preserving the `rememberNavController()` type so the call chains into `NavHost`.
- `rememberObservedNavController(rum, ...)` drop-in factory.

Behavior:
- `NavigationEmitter` emits an `app.screen.view` event carrying `app.screen.name`.
- The destination-changed listener is scoped to the composition via `DisposableEffect` and reads `rum`/`screenName` through `rememberUpdatedState`, so changed values take effect without re-registering the listener.

Other changes:
- Enable the Compose compiler plugin for this module (first library module to define @Composables), working around AGP built-in Kotlin version detection
- Exclude `@Composable` functions from Kover coverage.

### Open items/ Follow up
- As is, `ScreenAttributesSpanProcessor` and `ScreenAttributesLogRecordProcessor` overwrites `app.screen.name` on every span/log from `VisibleScreenTracker.currentlyVisibleScreen`. A follow up PR will ensure that the events generated will have the Compose screen names if available. This is tracked by https://github.com/open-telemetry/opentelemetry-android/issues/1909


instrumentation/compose/navigation/src/main/kotlin/io/opentelemetry/instrumentation/compose/navigation/NavigationEmitter.kt
```diff
@@ -0,0 +1,45 @@
...
+            .loggerBuilder(INSTRUMENTATION_SCOPE_NAME)
+            .build(),
+    )
+
+    @OptIn(IncubatingApi::class)
+    fun onScreenView(screenName: String) {
+        eventLogger
+            .logRecordBuilder()
+            .setEventName(SCREEN_VIEW_EVENT_NAME)
+            .setAttribute(stringKey(map(APP_SCREEN_NAME)), screenName)
```
aranhave: Should the screen processor change land with this? In the default `OpenTelemetryRum` setup it replaces this value, so the exported event reports the Activity/Fragment screen instead of the Compose route.

instrumentation/compose/navigation/src/main/kotlin/io/opentelemetry/instrumentation/compose/navigation/NavigationEmitter.kt
```diff
@@ -0,0 +1,45 @@
...
+import io.opentelemetry.api.common.AttributeKey.stringKey
+import io.opentelemetry.api.logs.Logger
+import io.opentelemetry.kotlin.semconv.AppAttributes.APP_SCREEN_NAME
+import io.opentelemetry.kotlin.semconv.IncubatingApi
+
+internal const val INSTRUMENTATION_SCOPE_NAME =
+    "io.opentelemetry.android.instrumentation.compose.navigation"
+
+// Provisional event name pending the mobile semconv SIG agreement.
+internal const val SCREEN_VIEW_EVENT_NAME = "app.screen.view"
```
aranhave: Since `app.screen.view` is not upstream yet, should we define it in the local semconv registry instead of hard-coding it here?

instrumentation/compose/navigation/src/main/kotlin/io/opentelemetry/instrumentation/compose/navigation/NavControllerExtensions.kt
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Drop-in replacement for `rememberNavController()` that attaches OpenTelemetry navigation
+ * instrumentation; it delegates to [withOpenTelemetry].
+ */
+@Composable
+fun rememberObservedNavController(
+    rum: OpenTelemetryRum,
+    vararg navigators: Navigator<out NavDestination>,
+): NavHostController = rememberNavController(*navigators).withOpenTelemetry(rum)
```
aranhave: Could this factory take `screenName` too? Otherwise callers have to switch APIs to customize it, and chaining `withOpenTelemetry` would register a second listener.

instrumentation/compose/navigation/src/main/kotlin/io/opentelemetry/instrumentation/compose/navigation/NavControllerExtensions.kt
```diff
@@ -0,0 +1,95 @@
...
+): NavController {
+    // The listener is registered once per controller (the effect is keyed only on `this`), yet it
+    // must always use the latest `rum`/`screenName`; rememberUpdatedState lets a changed value take
+    // effect without re-registering.
+    val emitter = remember(rum) { NavigationEmitter(rum) }
+    val currentEmitter by rememberUpdatedState(emitter)
+    val currentScreenName by rememberUpdatedState(screenName)
+    DisposableEffect(this) {
+        val listener = attachOpenTelemetry({ currentEmitter }, { currentScreenName })
+        onDispose { removeOnDestinationChangedListener(listener) }
```
aranhave, line 49: Could we add a test that removes this composable and verifies later navigation no longer emits? That would cover the listener cleanup promised here.

instrumentation/compose/navigation/src/test/kotlin/io/opentelemetry/instrumentation/compose/navigation/WithOpenTelemetryComposeTest.kt
```diff
@@ -0,0 +1,87 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.annotation.GraphicsMode
+
+@OptIn(IncubatingApi::class)
+@GraphicsMode(GraphicsMode.Mode.NATIVE)
+@RunWith(AndroidJUnit4::class)
+class WithOpenTelemetryComposeTest {
+    @get:Rule
+    val composeRule = createComposeRule()
```
aranhave, line 31: Could we use the new `androidx.compose.ui.test.junit4.v2.createComposeRule` here? The current import already emits a deprecation warning.

aranhave (commented):
Thanks for putting this together! Great work! ⭐ I left a few questions, mainly about how the Compose screen name is handled, plus a couple of smaller API and test notes.

instrumentation/compose/navigation/src/test/kotlin/io/opentelemetry/instrumentation/compose/navigation/WithOpenTelemetryComposeTest.kt
benjoseph-grafana: Done

instrumentation/compose/navigation/src/main/kotlin/io/opentelemetry/instrumentation/compose/navigation/NavControllerExtensions.kt
benjoseph-grafana: Added the test.

instrumentation/compose/navigation/src/main/kotlin/io/opentelemetry/instrumentation/compose/navigation/NavControllerExtensions.kt
benjoseph-grafana: Added screenName param.

instrumentation/compose/navigation/src/main/kotlin/io/opentelemetry/instrumentation/compose/navigation/NavigationEmitter.kt
benjoseph-grafana: Done.

instrumentation/compose/navigation/src/main/kotlin/io/opentelemetry/instrumentation/compose/navigation/NavigationEmitter.kt
benjoseph-grafana: This will be part of a follow up PR. It is deliberately excluded to keep this PR scope small and not touch any existing flows.

aranhave (approved):
LGTM!

LikeTheSalad (approved):
LGTM, thank you!

instrumentation/compose/navigation/src/main/kotlin/io/opentelemetry/instrumentation/compose/navigation/NavigationEmitter.kt
bidetofevil: We may want to think about modelling this as app navigation rather than screen loading given the prior art. Those concepts related but not necessarily the same. The existing instrumentation maps activity to screen, which is also not super precise, so I'm wondering if these two should be separate - and which one should use `app.screen.*`, if any

benjoseph-grafana: I can remove the event being emitted in this PR, if that is an area that warrant's more discussion.

instrumentation/compose/navigation/src/main/kotlin/io/opentelemetry/instrumentation/compose/navigation/ScreenNames.kt
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Default screen-name resolver.
+ *
+ * Uses the route *pattern* (for example `user/{id}`, not the filled-in arguments) to avoid leaking
+ * PII. Callers can supply their own resolver (including one that reads [arguments]) via
+ * [withOpenTelemetry][NavHostController.withOpenTelemetry].
+ */
+internal fun defaultScreenName(
+    destination: NavDestination,
+    @Suppress("UNUSED_PARAMETER") arguments: Bundle?,
```
bidetofevil, line 20: Any reason why we're passing this if we're not using it? So overrides can potentially use it in the future?

If that's the case, it might be good to just remove it and add it in in the future, given that this is an API and we can't remove it once we declare it stable.

instrumentation/compose/navigation/src/main/kotlin/io/opentelemetry/instrumentation/compose/navigation/NavControllerExtensions.kt
```diff
@@ -0,0 +1,97 @@
...
+ * via [DisposableEffect] and removed when the controller leaves it.
+ *
+ * This base-class overload works for any controller you already hold, including nested/child
+ * controllers, not just the host controller returned by `rememberNavController()`.
+ *
+ * @param rum the [OpenTelemetryRum] instance to emit through.
+ * @param screenName maps a destination (and its arguments) to a screen name. Defaults to the route
+ *   pattern to avoid leaking PII.
+ */
+@Composable
```
bidetofevil, line 36: While the API is being exposed solely as `Composable`s, this instrumentation is technically not bound to Compose, but to Jetpack Navigation.  A `NavHost`, for instance, doesn't require Compose, and an app can just navigate by opening different fragments within the host.

Was it a deliberate choice to not support non-Compose usage of Jetpack Navigation?

instrumentation/compose/navigation/src/main/kotlin/io/opentelemetry/instrumentation/compose/navigation/NavControllerExtensions.kt
```diff
@@ -0,0 +1,97 @@
...
+    }
+    return this
+}
+
+/**
+ * [NavHostController] overload of [withOpenTelemetry]. Behaves identically to the [NavController]
+ * one but returns [NavHostController], preserving the type of a `rememberNavController()` result so
+ * the call can be grafted onto it (the auto-instrumentation seam) without down-typing.
+ */
+@Composable
```
bidetofevil: Any reason why you don't just merge this into the previous method and do a check on `this` and cast appropriatelly?

instrumentation/compose/navigation/README.md
```diff
@@ -0,0 +1,69 @@
...
+## Telemetry
+
+Data produced by this instrumentation will have an instrumentation scope
+name of `io.opentelemetry.android.instrumentation.compose.navigation`.
+This instrumentation produces the following telemetry:
+
+### Screen views
+
+* Type: Event
+* Name: `app.screen.view` (provisional, pending mobile semantic convention agreement)
```
bidetofevil: I think making it more abstract and talking about what precisely is happening (i.e. a navigation had completed with some destination loaded) is perhaps a better way of modelling this? When a that callback is fired, there's no guarantee that any meaningful UI has actually loaded

instrumentation/compose/navigation/src/main/kotlin/io/opentelemetry/instrumentation/compose/navigation/NavigationEmitter.kt
bidetofevil: So the thing you want most is updating the attribute that's written in the span and logs? The event is probably the more straight forward part 😅 

I think the tricky part there given the existing instrumentation is that `NavController` and `Activity` lifecycles are quite different - in fact, the former lives within the latter, but the callback firing isn't necessarily co-ordinated. But if that's your focus, you'll probably want these to play nice - and perhaps roll them up together for coordination purposes. One thing you probably don't want is let the last writer win, as you can't guarantee that onResume() will always fire before the destination load callback in all cases.

benjoseph-grafana: I understand this is a little more complicated than I anticipated 😅, thanks for the pointing out the caveats.  
My end goal is just to provide better visibility into navigation within Compose based apps and I'm not opinionated about how it should be achieved. 
Navigation events and attribute stamping on spans/logs are both parts of the end goal, and I hope we can refine this in #1909 rather than here. A few options worth putting on the table there -
- a separate attribute for Compose derived screens vs. the Activity derived one, so both can coexist
- the pattern some libraries use- an exclusion list of Activities whose names are ignored for screen attribution, so Compose screens get route based names instead of host Activity

I'll capture these on #1909. I'll drop the event emission from this PR entirely and we can settle the modelling (navigation vs. screen view, and the semconv name) in a dedicated PR.

instrumentation/compose/navigation/src/main/kotlin/io/opentelemetry/instrumentation/compose/navigation/ScreenNames.kt
benjoseph-grafana: The `Bundle?` is part of the signature so caller supplied implementations can read navigation arguments today, it's not future-proofing. Only the default resolver ignores it, so only route patterns are emitted by default to avoid PII leak.

instrumentation/compose/navigation/src/main/kotlin/io/opentelemetry/instrumentation/compose/navigation/NavControllerExtensions.kt
benjoseph-grafana: It was a deliberate choice for this PR. This has the advantage of having a lifecycle to scope listener registration/cleanup to. 

Supporting fragment-based Jetpack Navigation is mostly a matter of exposing the listener core (attachOpenTelemetry) behind a LifecycleOwner scoped entry point, which can be a follow up.

instrumentation/compose/navigation/src/main/kotlin/io/opentelemetry/instrumentation/compose/navigation/NavControllerExtensions.kt
benjoseph-grafana: The split exists for the return type. A runtime check + cast will not preserve `NavHostController` for chaining into NavHost. 
I will update this to a single generic `<T : NavController> T.withOpenTelemetry(...): T `

instrumentation/compose/navigation/README.md
benjoseph-grafana: Since I'm dropping the event emission from this PR, I'll update the README to remove the telemetry section.

instrumentation/compose/navigation/src/main/kotlin/io/opentelemetry/instrumentation/compose/navigation/NavControllerExtensions.kt
bidetofevil: The name of the package basically precludes supporting non-Compose use cases. I suppose future refactors can spin that off if someone really wants it. Cool

instrumentation/compose/navigation/src/main/kotlin/io/opentelemetry/instrumentation/compose/navigation/ScreenNames.kt
bidetofevil: Right. It's part of the callback

bidetofevil (approved):
LGTM. Thanks for taking this on!

fractalwrench (approved):
LGTM

PR #1902, opened 2026-07-16
Change session observer list type

### LINKED ISSUE #1877: Changing how the observers list in SessionManager works
In [`SessionManager`](https://github.com/open-telemetry/opentelemetry-android/blob/cba04872bf2af4631cd3fdaef4eb7c9d8cc4233d/android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionManager.kt#L28), the `observers` field uses a `synchronizedList`.

In the JavaDoc for [`synchronizedList`](https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#synchronizedList-java.util.List-), it mandates that each iteration must use `synchronized` or it will result in non-deterministic behavior, but that's currently not implemented. As suggested by Jason Morris, it might be better to use a `CopyOnWriteArrayList` to solve the issue of avoiding problems like a `ConcurrentModificationException`

breedx-splk:
Resolves #1877


fractalwrench (commented):
There's a CI failure where the new test case is failing. Otherwise this looks good

aranhave (approved):
LGTM!

fractalwrench (approved):

PR #1903, opened 2026-07-16
Add dsl config to disable all instrumentation discovery

### LINKED ISSUE #1716: Agent DSL: Add ability to disable auto-instrumentation discovery
As brought up in [this comment](https://github.com/open-telemetry/opentelemetry-android/pull/1714#pullrequestreview-4210349222), it would be helpful to expose a mechanism in the DSL for users to disable the auto-discovery of instrumentations on the classpath.

One use case is users who just want to bring their own instrumentation. It's another convenience/customization that's not easily accomplished with the agent in any other way.

breedx-splk:
Resolves #1716.


fractalwrench (approved):
LGTM

PR #1904, opened 2026-07-17
chore: Docker image housekeeping

thompson-tomo:
Explicitly pin docker version to improve info/context available to renovate.

The jaegar tracing image has been switched as existing image is eol https://github.com/jaegertracing/jaeger/issues/6321

### DIFF
--- a/demo-app/compose.yaml
+++ b/demo-app/compose.yaml
@@ -10,8 +10,5 @@ services:
       - "4318:4318" # OTLP HTTP
   jaeger:
     image: "jaegertracing/jaeger:2.19.0@sha256:ede4864215be4cd85bd8c3129a2fea6c5713c5653c7282c429dba123014bc68b"
-    environment:
-      - COLLECTOR_OTLP_ENABLED=true
-      - COLLECTOR_OTLP_HTTP_HOST_PORT=0.0.0.0:4318
     ports:
       - "16686:16686" # UI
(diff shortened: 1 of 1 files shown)


demo-app/compose.yaml
```diff
@@ -9,9 +9,9 @@ services:
       - "4317:4317" # OTLP gRPC
       - "4318:4318" # OTLP HTTP
   jaeger:
-    image: "jaegertracing/all-in-one@sha256:ab6f1a1f0fb49ea08bcd19f6b84f6081d0d44b364b6de148e1798eb5816bacac"
+    image: "jaegertracing/jaeger:2.19.0@sha256:ede4864215be4cd85bd8c3129a2fea6c5713c5653c7282c429dba123014bc68b"
```
aranhave, line 12: Could we drop the two `COLLECTOR_OTLP_*` variables here too? They’re Jaeger v1 settings, and v2 enables OTLP from its built-in config, so they’re ignored now.

aranhave (commented):
Nice cleanup! Just left one small comment about the old Jaeger env vars.

demo-app/compose.yaml
thompson-tomo: Done

aranhave (approved):
LGTM!

breedx-splk (approved):
thanks!

I tried the demo app with these versions (including the major jaeger) and it seems to work just fine.

PR #1910, opened 2026-07-21
Turn gradle tooling parallel on

breedx-splk:
I don't see any downside to turning this on for developers. The "tooling" parallelism is used to improve gradle sync performance without turning on full build parallelism, which I guess can break some builds.

https://docs.gradle.org/current/userguide/performance.html#sec:configure_tooling_api_actions_parallelism

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -26,3 +26,6 @@ android.minCompileSdk=34
 android.compileSdk=37
 version=1.6.0
 group=io.opentelemetry.android
+
+# Enabled parallel sync for Gradle 9.4+
+org.gradle.tooling.parallel=true


DavidGrath (commented):
From a brief glance at the [Tooling API docs](https://docs.gradle.org/current/userguide/tooling_api.html) page, it seems pretty niche, that is to say, I didn't see any of their examples of what the Tooling API can do that can actually be parallelized:
* Query the details of a build...
* Execute a build and listen to stdout...
* Execute a specific test class...
** This one in particular is one where I'm not sure what tooling might do 

So it's not to say I disagree, I just kinda wish there was a more concrete example in the docs of what the speedup factor was like. "IDE sync" feels a bit broad to me

DavidGrath (approved):

aranhave (approved):

benjoseph-grafana:
Does it make sense to enable this in the [demo-app](https://github.com/open-telemetry/opentelemetry-android/blob/main/demo-app/gradle.properties) as well?

breedx-splk:
> So it's not to say I disagree, I just kinda wish there was a more concrete example in the docs of what the speedup factor was like. "IDE sync" feels a bit broad to me

Yeah I agree. I think they've left it kinda vague so that it can be improved over time. The closest thing I've heard is that "gradle sync" from the IDE can be made faster. It was the latest IntelliJ that was prompting me to add this in the first place. 🤷🏻 

> Does it make sense to enable this in the [demo-app](https://github.com/open-telemetry/opentelemetry-android/blob/main/demo-app/gradle.properties) as well?

yes!

PR #1912, opened 2026-07-21
restore codeql now that kotlin 2.4.0 is supported

breedx-splk:
Kotlin 2.4.0 is supported again, so we can start using CodeQL.


breedx-splk:
oh well damn, nevermind. I missed that we're already up to 2.4.10 since 2.4.0, which codeql one again doesn't yet support. Wheeeeeeeeeeeeee!

PR #1935, opened 2026-07-29
Coroutines context propagation instrumentation

### LINKED ISSUE #239: Automatic coroutine OTel Context propagation
There's [some work](https://github.com/open-telemetry/opentelemetry-java/tree/main/extensions/kotlin) needed to be done to properly propagate an [OTel Context](https://opentelemetry.io/docs/specs/otel/context/) across a coroutine and its [coroutine-context](https://kotlinlang.org/docs/coroutine-context-and-dispatchers.html) changes. This would involve having to change the way every Kotlin coroutine is launched in the project so that it checks for an existing OTel Context (if any) and then proceeds to transform it into part of the newly launched coroutine's context.

We can automate this with bytecode instrumentation, so this issue is to work on an auto-instrumentation that takes care of it so that users wouldn't have to change the way they launch coroutines.

LikeTheSalad:
Closes #239 

Example of what this instrumentation helps with, as stated in its README.md file:

Given:

```kotlin
val span = tracer.spanBuilder("work").startSpan()
span.makeCurrent().use {
    scope.launch {
        val inner = tracer.spanBuilder("coroutine-work").startSpan()
        // ...
        inner.end()
    }
}
span.end()
```

**Without** this instrumentation, `coroutine-work` does not inherit the active context and is recorded
as a separate root span:

```
work
coroutine-work
```

**With** this instrumentation, the active context is carried into the coroutine automatically, and
`coroutine-work` is recorded as a child of `work`:

```
work
└── coroutine-work
```

### DIFF
--- a/instrumentation/coroutines/agent/src/main/kotlin/io/opentelemetry/instrumentation/agent/coroutines/CoroutinesPlugin.kt
+++ b/instrumentation/coroutines/agent/src/main/kotlin/io/opentelemetry/instrumentation/agent/coroutines/CoroutinesPlugin.kt
@@ -1,0 +1,64 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.agent.coroutines
+
+import io.opentelemetry.instrumentation.library.coroutines.internal.CoroutinesLaunchBridge
+import net.bytebuddy.asm.MemberSubstitution
+import net.bytebuddy.build.AndroidDescriptor
+import net.bytebuddy.build.Plugin
+import net.bytebuddy.description.method.MethodDescription
+import net.bytebuddy.description.type.TypeDescription
+import net.bytebuddy.dynamic.ClassFileLocator
+import net.bytebuddy.dynamic.DynamicType
+import net.bytebuddy.matcher.ElementMatchers
+
+internal class CoroutinesPlugin(
+    private val androidDescriptor: AndroidDescriptor,
+) : Plugin {
+    override fun matches(target: TypeDescription): Boolean = androidDescriptor.getTypeScope(target) == AndroidDescriptor.TypeScope.LOCAL
+
+    override fun apply(
+        builder: DynamicType.Builder<*>,
+        typeDescription: TypeDescription,
+        classFileLocator: ClassFileLocator,
+    ): DynamicType.Builder<*> =
+        builder.visit(
+            MemberSubstitution
+                .relaxed()
+                .method(
+                    ElementMatchers
+                        .named<MethodDescription>("launch")
+                        .and(ElementMatchers.isDeclaredBy(ElementMatchers.named("kotlinx.coroutines.BuildersKt"))),
+                ).replaceWith(BRIDGE_LAUNCH)
+                .method(
+                    ElementMatchers
+                        .named<MethodDescription>("launch\$default")
+                        .and(ElementMatchers.isDeclaredBy(ElementMatchers.named("kotlinx.coroutines.BuildersKt"))),
+                ).replaceWith(BRIDGE_LAUNCH_DEFAULT)
+                .on(ElementMatchers.any()),
+        )
+
+    override fun close() {}
+
+    private companion object {
+        private val BRIDGE_TYPE = TypeDescription.ForLoadedType.of(CoroutinesLaunchBridge::class.java)
+        private val BRIDGE_LAUNCH: MethodDescription =
+            BRIDGE_TYPE.declaredMethods
+                .filter(
+                    ElementMatchers
+                        .named<MethodDescription>("launch")
+                        .and(ElementMatchers.isStatic())
+                        .and(ElementMatchers.takesArguments(4)),
+                ).getOnly()
+        private val BRIDGE_LAUNCH_DEFAULT: MethodDescription =
+            BRIDGE_TYPE.declaredMethods
+                .filter(
+                    ElementMatchers
+                        .named<MethodDescription>("launch\$default")
+                        .and(ElementMatchers.isStatic()),
+                ).getOnly()
+    }
+}
--- a/instrumentation/coroutines/library/src/main/kotlin/io/opentelemetry/instrumentation/library/coroutines/CoroutinesInstrumentation.kt
+++ b/instrumentation/coroutines/library/src/main/kotlin/io/opentelemetry/instrumentation/library/coroutines/CoroutinesInstrumentation.kt
@@ -1,0 +1,31 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.library.coroutines
+
+import android.content.Context
+import com.google.auto.service.AutoService
+import io.opentelemetry.android.OpenTelemetryRum
+import io.opentelemetry.android.instrumentation.AndroidInstrumentation
+import io.opentelemetry.instrumentation.library.coroutines.internal.CoroutinesContextHelper
+
+@AutoService(AndroidInstrumentation::class)
+class CoroutinesInstrumentation : AndroidInstrumentation {
+    override val name: String = "coroutines"
+
+    override fun install(
+        context: Context,
+        openTelemetryRum: OpenTelemetryRum,
+    ) {
+        CoroutinesContextHelper.setEnabled(true)
+    }
+
+    override fun uninstall(
+        context: Context,
+        openTelemetryRum: OpenTelemetryRum,
+    ) {
+        CoroutinesContextHelper.setEnabled(false)
+    }
+}
--- a/instrumentation/coroutines/library/src/main/kotlin/io/opentelemetry/instrumentation/library/coroutines/internal/CoroutinesContextHelper.kt
+++ b/instrumentation/coroutines/library/src/main/kotlin/io/opentelemetry/instrumentation/library/coroutines/internal/CoroutinesContextHelper.kt
@@ -1,0 +1,54 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.library.coroutines.internal
+
+import io.opentelemetry.context.Context
+import io.opentelemetry.extension.kotlin.asContextElement
+import io.opentelemetry.extension.kotlin.getOpenTelemetryContext
+import kotlin.coroutines.CoroutineContext
+
+/**
+ * Decides whether the current OpenTelemetry [Context] should be added to an about-to-launch
+ * coroutine. Not intended as public API; public JVM visibility is required so that the runtime
+ * instrumentation in the parent package can configure it.
+ */
+object CoroutinesContextHelper {
+    @Volatile
+    private var enabled = false
+
+    @JvmStatic
+    fun setEnabled(value: Boolean) {
+        enabled = value
+    }
+
+    /**
+     * Returns the supplied [coroutineContext] enriched with the current OpenTelemetry context
+     * when all of the following are true:
+     *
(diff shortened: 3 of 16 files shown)


LikeTheSalad:
> ❌ Patch coverage is 52.72727% with 26 lines in your changes missing coverage. Please review.

This is because codecov doesn't currently cover Android instrumentation tests, but in reality the coverage is higher.

instrumentation/coroutines/agent/src/main/kotlin/io/opentelemetry/instrumentation/agent/coroutines/CoroutinesPlugin.kt
```diff
@@ -0,0 +1,64 @@
...
+import net.bytebuddy.asm.MemberSubstitution
+import net.bytebuddy.build.AndroidDescriptor
+import net.bytebuddy.build.Plugin
+import net.bytebuddy.description.method.MethodDescription
+import net.bytebuddy.description.type.TypeDescription
+import net.bytebuddy.dynamic.ClassFileLocator
+import net.bytebuddy.dynamic.DynamicType
+import net.bytebuddy.matcher.ElementMatchers
+
+internal class CoroutinesPlugin(
```
fractalwrench, line 18: Will folks want this included in every `scope.launch`, i.e. is there a possibility folks might want to opt-out of certain coroutines or specify which ones they want manually via a helper/extension function instead?

instrumentation/coroutines/library/src/main/kotlin/io/opentelemetry/instrumentation/library/coroutines/CoroutinesInstrumentation.kt
```diff
@@ -0,0 +1,31 @@
...
+package io.opentelemetry.instrumentation.library.coroutines
+
+import android.content.Context
+import com.google.auto.service.AutoService
+import io.opentelemetry.android.OpenTelemetryRum
+import io.opentelemetry.android.instrumentation.AndroidInstrumentation
+import io.opentelemetry.instrumentation.library.coroutines.internal.CoroutinesContextHelper
+
+@AutoService(AndroidInstrumentation::class)
+class CoroutinesInstrumentation : AndroidInstrumentation {
```
fractalwrench, line 15: I don't know whether this belongs as instrumentation. It doesn't actually record telemetry & exists solely to manage context. My instinct is that it should be in a separate opt-in module like it is now, but I'm not convinced we need to implement `AndroidInstrumentation`

fractalwrench (commented):
Thanks (and good luck) for taking this on! I wasn't sure whether this fits inside `AndroidInstrumentation` as it doesn't actually capture telemetry. I also had questions around whether it makes sense to allow folks to opt-out (presumably not)? Additionally I'd be curious how it interacts with opentelemetry-java's thread-local approach if the two are used together.

instrumentation/coroutines/agent/src/main/kotlin/io/opentelemetry/instrumentation/agent/coroutines/CoroutinesPlugin.kt
LikeTheSalad: There's no way to choose which coroutines would get the OTel context or not when using this plugin. If people wanted to do so, I think they should not use this plugin and instead directly use the [Kotlin extension](https://github.com/open-telemetry/opentelemetry-java/tree/main/extensions/kotlin) whenever they need it.

The only thing that they could do, even when applying this plugin, is to set their own coroutine OTel context manually, in case they want to use a custom one (the plugin won't override it).

instrumentation/coroutines/library/src/main/kotlin/io/opentelemetry/instrumentation/library/coroutines/CoroutinesInstrumentation.kt
LikeTheSalad: I see what you mean; it is a strange use case for calling it an "AndroidInstrumentation". I did it like this mostly to keep some consistency around opt-in tools that do code on behalf of users. If we strictly define "instrumentations" as telemetry-generating code, then it doesn't fit, but if we instead define them as "automatic code generation" in general, it could fit.

That being said, I'm open to ideas to avoid making this an "AndroidInstrumentation" implementation. If we're fine with leaving it all as is and just removing the explicit implementation of `AndroidInstrumentation`, I think it's fine. However, if we wanted to create a new category of opt-in tools for when they do bytecode weaving but don't create telemetry, I think we should wait for at least a second tool that would fit in there, because doing so for a single module seems a bit overkill.

LikeTheSalad:
> Thanks (and good luck) for taking this on! I wasn't sure whether this fits inside `AndroidInstrumentation` as it doesn't actually capture telemetry. I also had questions around whether it makes sense to allow folks to opt-out (presumably not)? Additionally I'd be curious how it interacts with opentelemetry-java's thread-local approach if the two are used together.

Thanks for the feedback, I added some responses.

Regarding

> I'd be curious how it interacts with opentelemetry-java's thread-local approach if the two are used together

I'm not sure I follow the question. This plugin actually relies on opentelemetry-java's thread-local storage to capture and then propagate an existing OTel Context in a coroutine, so they work together by design. Is there a use case that you have in mind where they could clash?

Also, we're probably going to keep this plugin simple forever, so I hope it won't become a headache in the future in case people start asking to support other means to create coroutines. The reason is that it only works because of the way OTel Java propagates its Context within threads, which I know is an OTel Java-specific feature. So when we switch to OTel Kotlin, this plugin will become obsolete. The reason we still decided to create it is that someone brought it up in a recent SIG meeting, where we noticed that [the issue for it](https://github.com/open-telemetry/opentelemetry-android/issues/239) is currently the most upvoted of all OTel Android, so we decided to address it, but I don't think it should become a maintenance burden because of our future migration plans.

instrumentation/coroutines/agent/src/main/kotlin/io/opentelemetry/instrumentation/agent/coroutines/CoroutinesPlugin.kt
fractalwrench: Using the extension manually instead sounds reasonable 👍

fractalwrench:
>I'm not sure I follow the question. This plugin actually relies on opentelemetry-java's thread-local storage to capture and then propagate an existing OTel Context in a coroutine, so they work together by design. Is there a use case that you have in mind where they could clash?

I guess I'm just asking whether there could be any edge cases where library users expect the default thread-local context storage approach, but end up receiving the coroutine context storage approach? And if so, whether we could set expectations through documenting that one approach wins over the other.

As an example, I'm unclear what the current context would be for `ctx` and `ctx2` in the following scenario:

```
val executor = Executors.newSingleThreadExecutor()
val scope = CoroutineScope(Dispatchers.Default)

scope.launch {
    tracer.createSpan("a").makeCurrent()
    val ctx = Context.current()
    
    executor.submit { 
        tracer.createSpan("b").makeCurrent()
        val ctx2 = Context.current()
    }
}
```

LikeTheSalad:
> I guess I'm just asking whether there could be any edge cases where library users expect the default thread-local context storage approach, but end up receiving the coroutine context storage approach? And if so, whether we could set expectations through documenting that one approach wins over the other.

Based on your example, it seems like the use case you're talking about is the following:

1. User launches coroutine
2. Creates span and makes it current in the coroutine backing thread
3. Launches a non-coroutine thread executor (using Java's executors)
4. Creates span within the Java-managed thread

I think we can see this as a "Coroutine to Regular thread" scenario, in which case, it's not covered by this plugin, so the fully Java-managed thread won't get the context that existed in the Kotlin coroutine-managed one, which is an inherited behavior from the Kotlin extension mechanism that the plugin uses.

I think this can be described as a feature where "Kotlin is aware of Java, but not vice versa", which I think is the ideal approach, as this is a Kotlin-first project and the test scenario can be considered quite niche. However, I agree that some people can probably expect a two-way propagation mechanism, whereas it's only one (from a regular Java thread to a Kotlin coroutine backing one), so I'll add some clarifications to the readme file on this.

LikeTheSalad:
I added a clarification here: https://github.com/open-telemetry/opentelemetry-android/pull/1935/commits/9492043b8f0822bdb2884f8707635255e6c84915 on what to expect in the scenario you mentioned, @fractalwrench. Let me know if it helps or if we could improve it a bit further.

fractalwrench:
Thanks, that's helpful!

instrumentation/coroutines/library/src/main/kotlin/io/opentelemetry/instrumentation/library/coroutines/internal/CoroutinesContextHelper.kt
```diff
@@ -0,0 +1,54 @@
...
+import io.opentelemetry.extension.kotlin.asContextElement
+import io.opentelemetry.extension.kotlin.getOpenTelemetryContext
+import kotlin.coroutines.CoroutineContext
+
+/**
+ * Decides whether the current OpenTelemetry [Context] should be added to an about-to-launch
+ * coroutine. Not intended as public API; public JVM visibility is required so that the runtime
+ * instrumentation in the parent package can configure it.
+ */
+object CoroutinesContextHelper {
```
breedx-splk: Can we not mark theseas `internal` kotlin objects instead?

instrumentation/coroutines/library/src/main/kotlin/io/opentelemetry/instrumentation/library/coroutines/internal/CoroutinesContextHelper.kt
```diff
@@ -0,0 +1,54 @@
...
+    @Volatile
+    private var enabled = false
+
+    @JvmStatic
+    fun setEnabled(value: Boolean) {
+        enabled = value
+    }
```
breedx-splk, line 24: Is this for a use case where the user has set up their build to include this instrumentation (at build time, with bytebuddy) but for some reason the application developer or other runtime configuration don't actually want it doing what it does?

instrumentation/coroutines/library/src/main/kotlin/io/opentelemetry/instrumentation/library/coroutines/CoroutinesInstrumentation.kt
breedx-splk: I definitely see both sides (pros/cons) to having this in instrumentation. I think I've convinced myself that it can be seen as instrumentation -- it's changing the bytecode and "doing otel stuff" with coroutines, even if that "stuff" is only propagating context and not actually making spans or events. I think it still counts as intstrumentation.

We definitely have prior art in other otel instrumentation projects for doing this kind of thing. Is it capital C correct? Probably not. But I think it's ok.

breedx-splk (approved):
Thanks @LikeTheSalad . I would like to see us proceed with this, so that people can start trying it out and providing feedback or opening issues.

instrumentation/coroutines/library/src/main/kotlin/io/opentelemetry/instrumentation/library/coroutines/internal/CoroutinesContextHelper.kt
LikeTheSalad: Yes. I'm not sure how likely that scenario will be, though, now that you mention it, so I think we can remove this option for now. Wdyt?

instrumentation/coroutines/library/src/main/kotlin/io/opentelemetry/instrumentation/library/coroutines/internal/CoroutinesContextHelper.kt
LikeTheSalad: Yes, I just marked `CoroutinesContextHelper` as internal.

fractalwrench (approved):
LGTM

aranhave (approved):
There's  conflicts with main but otherwise LGTM

PR #1939, opened 2026-07-29
fix(session): preserve explicit session IDs on logs

### LINKED ISSUE #1937: Native crash replay uses the current session ID instead of the persisted one
### What happened

While smoke-testing the published native-crash instrumentation, the persisted crash context contained the session ID from the process that crashed, but the replayed `app.crash` log used the new session created after relaunch.

The rest of the replay behaved as expected: the original crash timestamp and crash attributes were preserved, the marker was removed after replay, and another restart did not emit a duplicate.

### Expected behavior

The replayed `app.crash` log should keep the `session.id` stored at crash time.

### Reproduction

1. Add `android-agent:1.6.0` and `native-crash:1.6.0-alpha`.
2. Start the app and trigger a native `SIGSEGV`.
3. Inspect `native-crash-context.properties` before relaunch.
4. Relaunch the app and inspect the exported `app.crash` log.

Observed during the test:

- persisted crash context: session A
- new launch: session B
- exported `app.crash`: session B

### Environment

- Android 15 / API 35 ARM64 emulator
- OpenTelemetry Android agent `1.6.0`
- Native crash instrumentation `1.6.0-alpha`
- OpenTelemetry Android BOM `1.6.0-alpha`
- AGP `9.3.1`

### Notes

`NativeCrashReporter` adds the persisted session ID before emitting the log. `SessionIdLogRecordAppender` then sets `session.id` from the current `SessionProvider`, which appears to replace it with the new launch's session.

A regression test using the normal log processor pipeline would cover the behavior that the reporter's focused tests currently miss.

Related to #764 and the replay implementation in #1887.

aranhave:
- preserve an explicitly supplied `session.id` instead of replacing it with the current session
- keep adding the current session to logs that do not already have one
- cover the behavior through the normal logger pipeline

This keeps a recovered native crash attached to the session that actually crashed after the app restarts.

Fixes #1937.

#### Validation

- core and native-crash unit tests
- Android-agent smoke test covering session rotation and exported session attribution
- formatting, API compatibility, and Android lint checks
- Android 15 ARM64 emulator: Session A crashed, Session B started, and the replayed `app.crash` retained Session A and the original crash timestamp; another relaunch did not replay it again

### DIFF
--- a/android-agent/src/test/kotlin/io/opentelemetry/android/agent/smoke/OpenTelemetryRumSmokeTest.kt
+++ b/android-agent/src/test/kotlin/io/opentelemetry/android/agent/smoke/OpenTelemetryRumSmokeTest.kt
@@ -140,9 +140,9 @@ class OpenTelemetryRumSmokeTest {
         assertThat(currentSessionId).isNotEqualTo(endedSessionId)
         val request =
             server.awaitLogRequest {
-                findEvent(it, "session.end") != null
+                findSessionEvent(it, "session.end") != null
             }
-        val sessionEnd = checkNotNull(findEvent(request, "session.end"))
+        val sessionEnd = checkNotNull(findSessionEvent(request, "session.end"))
         val exportedSessionId =
             sessionEnd.attributesList
                 .first { it.key == SESSION_ID }
@@ -179,7 +179,7 @@ class OpenTelemetryRumSmokeTest {
                 logRecord.body.stringValue
             }.contains(logMessage)
 
-    private fun findEvent(
+    private fun findSessionEvent(
         request: ExportLogsServiceRequest,
         eventName: String,
     ): LogRecord? =
--- a/android-agent/src/test/kotlin/io/opentelemetry/android/agent/smoke/OpenTelemetryRumSmokeTest.kt
+++ b/android-agent/src/test/kotlin/io/opentelemetry/android/agent/smoke/OpenTelemetryRumSmokeTest.kt
@@ -12,12 +12,17 @@ import androidx.test.platform.app.InstrumentationRegistry
 import io.opentelemetry.android.OpenTelemetryRum
 import io.opentelemetry.android.agent.OpenTelemetryRumInitializer
 import io.opentelemetry.android.agent.dsl.OpenTelemetryConfiguration
+import io.opentelemetry.kotlin.semconv.IncubatingApi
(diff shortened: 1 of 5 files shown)


android-agent/src/test/kotlin/io/opentelemetry/android/agent/smoke/OpenTelemetryRumSmokeTest.kt
```diff
@@ -135,6 +179,16 @@ class OpenTelemetryRumSmokeTest {
                 logRecord.body.stringValue
             }.contains(logMessage)
 
+    private fun findEvent(
```
benjoseph-grafana: can we rename this method to be more clear it filters by `sessionLogScopeName`?

android-agent/src/test/kotlin/io/opentelemetry/android/agent/smoke/OpenTelemetryRumSmokeTest.kt
```diff
@@ -107,6 +113,44 @@ class OpenTelemetryRumSmokeTest {
...
+        val clock = TestClock.create()
+        lateinit var endedSessionId: String
+        lateinit var currentSessionId: String
+
+        performOpenTelemetryRumAction(
+            config = {
+                httpExport {
+                    baseUrl = server.url
+                }
+                diskBufferingConfig.enabled(false)
```
benjoseph-grafana, line 128: does enabling diskBuffering change anything?

benjoseph-grafana:
LGTM

android-agent/src/test/kotlin/io/opentelemetry/android/agent/smoke/OpenTelemetryRumSmokeTest.kt
aranhave: It shouldn’t. The session ID is attached before buffering, so the finalized log is what gets written to disk. I disabled it here to keep the smoke test deterministic, consistent with the other tests in this file.

benjoseph-grafana (approved):

LikeTheSalad (approved):
Thanks!

fractalwrench (approved):
LGTM

breedx-splk (approved):
Cool, thanks!

PR #1941, opened 2026-07-30
Fix weaver semconv schema url and allow dependency bump

fractalwrench:
Based on top of this PR that bumps dependencies with an extra fix of specifying the schema_url which was required to fix weaver's semconv generation: https://github.com/open-telemetry/opentelemetry-android/pull/1934


LikeTheSalad (approved):
Thank you!

PR #1942, opened 2026-07-30
chore: Update Renovate configuration to best practices

thompson-tomo:
Change Renovate config to extend best practices like most other repos and recommendation from security SIG.

see docs at https://docs.renovatebot.com/presets-config/


fractalwrench (approved):
LGTM - thanks!

PR #1945, opened 2026-07-30
Import the existing semconv from upstream for the session events.

breedx-splk:
Migrates away from the hard-coded semantic conventions here in favor of the existing upstream conventions.


LikeTheSalad (approved):
Thanks!

PR #1946, opened 2026-07-30
Nightly job to check for new semantic conventions 

breedx-splk:
We are already 2 minor versions behind, so it occurred to me that we should have something that updates us, given that renovate can't really handle this level of customized dependency.

This will check for a new version, update our manifest, and create a PR for it.

### DIFF
--- a/.github/workflows/auto-update-semconv-schema.yml
+++ b/.github/workflows/auto-update-semconv-schema.yml
@@ -1,0 +1,70 @@
+name: Auto-update semantic convention schema
+
+on:
+  schedule:
+    # nightly at 09:17 UTC
+    - cron: "17 9 * * *"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  pull-requests: read
+
+jobs:
+  update-semconv-schema:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Check for a new schema version
+        id: check-version
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          current=$(grep -oE 'semantic-conventions@v[0-9]+\.[0-9]+\.[0-9]+' \
+            semconv/model/manifest.yaml | sed 's/.*@v//')
+          latest=$(gh release view --repo open-telemetry/semantic-conventions \
+            --json tagName --jq .tagName | sed 's/^v//')
+          branch="otelbot/update-semconv-schema-to-${latest}"
+          open_pr=$(gh pr list --head "$branch" --state open --json number --jq '.[0].number')
+
+          echo "version=$latest" >> "$GITHUB_OUTPUT"
+          echo "branch=$branch" >> "$GITHUB_OUTPUT"
+          if [[ $current != "$latest" && -z $open_pr ]]; then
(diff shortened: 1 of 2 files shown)


.github/workflows/auto-update-semconv-schema.yml
```diff
@@ -0,0 +1,70 @@
...
+          branch="otelbot/update-semconv-schema-to-${latest}"
+          open_pr=$(gh pr list --head "$branch" --state open --json number --jq '.[0].number')
+
+          echo "version=$latest" >> "$GITHUB_OUTPUT"
+          echo "branch=$branch" >> "$GITHUB_OUTPUT"
+          if [[ $current != "$latest" && -z $open_pr ]]; then
+            echo "update-needed=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "update-needed=false" >> "$GITHUB_OUTPUT"
+          fi
```
breedx-splk, line 39: I figured that 99.9% of the time there will never be a new version, so this allows us to easily just skip all the remaining steps.

.github/workflows/auto-update-semconv-schema.yml
```diff
@@ -0,0 +1,70 @@
...
+          .github/scripts/update-semconv-schema-version.sh "$VERSION"
+          .github/scripts/use-cla-approved-github-bot.sh
+
+          message="Update the upstream semantic convention schema to $VERSION"
+          git checkout -b "$BRANCH"
+          git add semconv/model/manifest.yaml
+          git commit -m "$message"
+          auth_header=$(printf 'x-access-token:%s' "$GH_TOKEN" | base64 | tr -d '\n')
+          git -c "http.https://github.com/.extraheader=AUTHORIZATION: basic $auth_header" \
+            push --set-upstream origin "$BRANCH"
```
fractalwrench: Is this using basic auth? Is there a reason this is required rather than just using otelbot like in other workflows?

fractalwrench (commented):
This is a cool idea that I'm going to copy for opentelemetry-kotlin. I had a question about how the git branch was pushed and there was an open question from Copilot, otherwise this looks good

.github/workflows/auto-update-semconv-schema.yml
aranhave: If an update PR is closed without merging, its remote branch can remain. The next nightly run will find no open PR, create the same branch locally and then this push will be rejected as non-fast-forward, so the automation stays stuck until someone manually deletes the branch. Could we handle the existing remote branch here (or use a branch name that can be safely recreated)?

breedx-splk: Yeah, good call out. This was set up because we're trying to be secure by using `persist-credentials: false`, but we shouldn't be hacking auth basic like that, I agree.

It looks like we can use `gh auth setup-git` which can use the existing `$GH_TOKEN`, and that should be an improvement. Sorry.

breedx-splk: yeah, that does seem a little unlikely to me, but it's worth thinking about. I have made a tweak from looking at "all" pull requests to looking for "open" pull requests. The result should be that we skip the remaining workflows when a closed PR exists for the same version. Does that suffice?

aranhave: Yep, that works. Checking all PR states means a closed PR for the same version skips the push, so the stale branch won’t block the workflow. Thanks for updating it.

aranhave (approved):
LGTM

fractalwrench (approved):
LGTM - thanks!

PR #1948, opened 2026-07-31
Migrate compose click off kotlin semconv 

breedx-splk:
Instead, import into locally federated semconv and use locally generated event templates.


fractalwrench (approved):
LGTM

LikeTheSalad (approved):
Thanks!

PR #1950, opened 2026-07-31
Use upstream semconv attributes instead of hard coded literals

breedx-splk:
This builds on #1949 and will also need a rebase.

Now that we are able to generate attribute constants from upstream event definitions, we can plug those attribute definitions into the places that used literals before.


fractalwrench (approved):
LGTM

LikeTheSalad (approved):
Cheers!

PR #1951, opened 2026-07-31
Generate semantic event name constants

breedx-splk:
Builds upon #1950 and will get a rebase after that one merges.

This generates a constant for the event names, and plugs them into both the event class usages and also the tests.

### DIFF
--- a/android-agent/build.gradle.kts
+++ b/android-agent/build.gradle.kts
@@ -30,7 +30,6 @@ dependencies {
     implementation(project(":instrumentation:sessions"))
     implementation(project(":instrumentation:screen-orientation"))
 
-    testImplementation(project(":semconv"))
     testImplementation(libs.opentelemetry.semconv.kotlin)
     testImplementation(libs.robolectric)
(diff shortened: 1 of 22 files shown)


fractalwrench (approved):
LGTM - CI needs a rerun but I assume that's down to the Github actions failure yesterday.

android-agent/build.gradle.kts
```diff
@@ -30,6 +30,7 @@ dependencies {
     implementation(project(":instrumentation:sessions"))
     implementation(project(":instrumentation:screen-orientation"))
 
+    testImplementation(project(":semconv"))
```
aranhave: nit: This duplicates the existing `implementation(project(":semconv"))` above. The agent tests compile and pass without it, so I think we can drop this extra declaration.

aranhave (approved):
LGTM! Just 1 minor nit.

android-agent/build.gradle.kts
breedx-splk: good catch! was even a warning in android studio. 🫠

PR #1958, opened 2026-08-04
Optimize service loader initialization

fractalwrench:
## Goal

Allows R8 to optimize SPI by replacing reflexive instantiations during SDK initialization with a direct call in the bytecode. The previous approach to using `ServiceLoader` didn't fit the criteria that R8's rules use for optimizations so wasn't optimized.

### DIFF
--- a/core/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentationLoaderImpl.kt
+++ b/core/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentationLoaderImpl.kt
@@ -11,10 +11,26 @@ import java.util.ServiceLoader
 
 internal class AndroidInstrumentationLoaderImpl : AndroidInstrumentationLoader {
     private val instrumentations: MutableMap<Class<out AndroidInstrumentation>, AndroidInstrumentation> by lazy {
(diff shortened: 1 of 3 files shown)


LikeTheSalad (approved):
This is awesome! Cheers

aranhave (approved):

core/src/main/java/io/opentelemetry/android/instrumentation/AndroidInstrumentationLoaderImpl.kt
```diff
@@ -11,12 +11,28 @@ import java.util.ServiceLoader
...
             .associateBy { it.javaClass }
             .toMutableMap()
     }
 
+    /**
+     * Loads the [AndroidInstrumentation] implementations declared via SPI. Before making changes to
+     * this function please study R8's ServiceLoaderRewriter as it replaces
+     * reflective lookup with direct instantiation, keeping reflection off the startup path. The
+     * two-argument `load` overload, the literal class reference and the local for-each are all
+     * required for this optimization to occur.
```
breedx-splk, line 24: ❤️

breedx-splk (approved):

PR #1964, opened 2026-08-05
feat(compose-navigation): emit app.navigation.complete event on destination change

### LINKED ISSUE #1920: Generate navigation events based on Compose navigation instrumentation
Based on the SIG meeting discussion, we have decided to use the event name `app.navigation.complete` with the attribute `app.navigation.destination.name` holding the destination name.
Since the instrumentation is based on `NavController.OnDestinationChangedListener` callbacks, this event name best describes the event. The attribute name was chose similarly, appending "name" at the end following the widely used convention.

benjoseph-grafana:
The Compose Navigation instrumentation resolved a screen name on every
completed navigation but did not emit any events. This change
defines `app.navigation.complete` in the local federated semconv registry, add the
`app.navigation.destination.name` attribute it carries, and emit the
generated `AppNavigationCompleteEvent` from `NavigationEmitter`.

The attribute holds the route pattern rather than the filled-in
arguments by default, so argument values are not recorded. Callers who
want something else can still override the existing `screenName` parameter.

Also correct the 1.6.0 changelog entry, which claimed the instrumentation
emits an `app.screen.view` event. That was transcribed from the #1901
description, which had gone stale by merge time.

Resolves #1920

### DIFF
--- a/instrumentation/compose/navigation/README.md
+++ b/instrumentation/compose/navigation/README.md
@@ -19,13 +19,27 @@ This instrumentation is not currently enabled by default.
 
 ## Telemetry
 
-This instrumentation does not emit any telemetry yet. Destination changes are
-resolved to screen names (by default the route *pattern*, for example
-`user/{id}` — not the filled-in arguments — to avoid leaking PII). How they are
-reported — as an event and/or as screen attribution on other telemetry — is
-tracked in
-[#1909](https://github.com/open-telemetry/opentelemetry-android/issues/1909),
-pending the mobile semantic-conventions discussion on modelling navigation.
+This instrumentation produces the following telemetry:
+
+### Navigation
+
+* Type: Event
+* Name: `app.navigation`
+* Description: An event that fires whenever a navigation completes, that is,
+  whenever the destination of an instrumented `NavController` changes.
+* Attributes:
+
+| Attribute | Type | Description | Values | Requirement Level |
+|---|---|---|---|---|
+| `app.navigation.destination.name` | string | The name of the navigation destination reached by the completed navigation. | `home`; `user/{id}` | Required |
+
+The destination name defaults to the route *pattern*, for example `user/{id}` —
+not the filled-in arguments — to avoid leaking PII. Override it with the
+`screenName` parameter shown below.
+
+Using these Compose screen names for screen attribution on unrelated telemetry
+is tracked separately in
+[#1909](https://github.com/open-telemetry/opentelemetry-android/issues/1909).
 
 ## Installation
 
--- a/instrumentation/compose/navigation/README.md
+++ b/instrumentation/compose/navigation/README.md
@@ -24,9 +24,9 @@ This instrumentation produces the following telemetry:
 ### Navigation
 
 * Type: Event
-* Name: `app.navigation`
-* Description: An event that fires whenever a navigation completes, that is,
-  whenever the destination of an instrumented `NavController` changes.
+* Name: `app.navigation.complete`
+* Description: An event that fires on every completed navigation of an instrumented
+  `NavController`, plus on listener attach (see below).
 * Attributes:
 
 | Attribute | Type | Description | Values | Requirement Level |
@@ -36,6 +36,17 @@ This instrumentation produces the following telemetry:
 The destination name defaults to the route *pattern*, for example `user/{id}` —
 not the filled-in arguments — to avoid leaking PII. Override it with the
 `screenName` parameter shown below.
+
+The event maps 1:1 to `NavController.OnDestinationChangedListener` callbacks, so it
+also fires:
+
+* **on attach** — registering replays the current destination, so the screen visible
+  at attach time is recorded. A configuration change re-attaches and replays again
+  without a navigation. The replay supplies no `arguments`, even for a parameterised
+  start destination.
+* **on argument-only changes** — `user/1` → `user/2` is a real navigation, but the
+  default resolver names both `user/{id}`. Read `arguments` in `screenName` to tell
+  them apart, subject to the attach caveat above.
 
 Using these Compose screen names for screen attribution on unrelated telemetry
 is tracked separately in
--- a/instrumentation/compose/navigation/src/main/kotlin/io/opentelemetry/instrumentation/compose/navigation/NavControllerExtensions.kt
+++ b/instrumentation/compose/navigation/src/main/kotlin/io/opentelemetry/instrumentation/compose/navigation/NavControllerExtensions.kt
@@ -24,6 +24,11 @@ import io.opentelemetry.android.OpenTelemetryRum
  * `app.navigation.complete` event through [rum]. The listener is scoped to the composition via
  * [DisposableEffect] and removed when the controller leaves it.
  *
+ * Registering the listener replays the current destination, so attaching emits for the screen
+ * already showing, and re-attaching (for example after a configuration change) emits again without
+ * a navigation. The replay supplies no arguments, even for a parameterised start destination. See
+ * the module README for the full event contract.
+ *
  * Works for any controller you already hold, including nested/child controllers, not just the host
  * controller returned by `rememberNavController()`. The receiver is returned with its static type
  * preserved (e.g. [NavHostController]), so the call can be grafted onto a `rememberNavController()`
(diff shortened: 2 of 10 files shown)


semconv/model/android/events.yaml
```diff
@@ -1,4 +1,13 @@
 groups:
+  - id: event.app.navigation
+    type: event
+    name: app.navigation
```
aranhave: We intentionally left the event name and shape open in #1901. Can we capture why we landed on `app.navigation` and `app.navigation.destination.name`, either here or in #1920? A short rationale would help since this establishes the local semconv.

instrumentation/compose/navigation/README.md
```diff
@@ -19,13 +19,27 @@ This instrumentation is not currently enabled by default.
...
-tracked in
-[#1909](https://github.com/open-telemetry/opentelemetry-android/issues/1909),
-pending the mobile semantic-conventions discussion on modelling navigation.
+This instrumentation produces the following telemetry:
+
+### Navigation
+
+* Type: Event
+* Name: `app.navigation`
+* Description: An event that fires whenever a navigation completes, that is,
```
aranhave: The listener also fires immediately for the current destination when it is registered, and again when only the arguments change. Do we want those to count as `app.navigation` too? I think the README and a listener-level test should make that behavior explicit, especially since reattaching can emit without a new navigation.

instrumentation/compose/navigation/src/main/kotlin/io/opentelemetry/instrumentation/compose/navigation/NavigationEmitter.kt
```diff
@@ -27,8 +24,7 @@ internal class NavigationEmitter(
             .build(),
     )
 
-    @Suppress("UnusedParameter", "UNUSED_PARAMETER")
     fun onScreenView(screenName: String) {
```
aranhave: nit: Since this emits a navigation event now, `onScreenView` feels misleading. `onNavigation` or `onDestinationChanged` would match the behavior better.

instrumentation/compose/navigation/src/test/kotlin/io/opentelemetry/instrumentation/compose/navigation/NavigationEmitterTest.kt
```diff
@@ -0,0 +1,89 @@
...
+        assertThat(events).hasSize(1)
+        assertThat(events[0])
+            .hasEventName(NAVIGATION_EVENT_NAME)
+            .hasAttributesSatisfyingExactly(
+                equalTo(stringKey(DESTINATION_NAME_KEY), "home"),
+            )
+    }
+
+    @Test
+    fun `emits one event per destination change`() {
```
aranhave: nit: This calls `NavigationEmitter` directly, so it covers one event per call rather than one per destination change. Maybe rename the test to match, or move this coverage to the listener test if we want to lock down destination behavior.

instrumentation/compose/navigation/README.md
```diff
@@ -19,13 +19,27 @@ This instrumentation is not currently enabled by default.
 
 ## Telemetry
```
aranhave, line 21: `semconv/README.md` still says production use of generated event classes will come later. This PR is the first production use, so that line can be removed now.

aranhave (commented):
Nice work! 🏅 I left a few comments, mostly around making the event contract and listener behavior explicit, plus a couple of small naming and docs cleanups.

semconv/model/android/events.yaml
benjoseph-grafana: event needs to be `app.navigation.complete`, will update. Both these were agreed up on in the last SIG meeting.

instrumentation/compose/navigation/README.md
benjoseph-grafana: Yes, I think the arguments change is a valid scenario and the registration replay is something we cannot avoid unless we do some sort of state management internally. Will update the README and add new tests.

instrumentation/compose/navigation/src/main/kotlin/io/opentelemetry/instrumentation/compose/navigation/NavigationEmitter.kt
benjoseph-grafana: will update to `onNavigation`

instrumentation/compose/navigation/src/test/kotlin/io/opentelemetry/instrumentation/compose/navigation/NavigationEmitterTest.kt
benjoseph-grafana: will update to `emits one event per onNavigation invocation`

instrumentation/compose/navigation/README.md
benjoseph-grafana: will remove the line

semconv/model/android/events.yaml
aranhave: Thanks, `app.navigation.complete` makes sense. Could you add the SIG rationale to #1920 when you update it? The issue is still empty, so it would be useful to capture why this name and attribute shape were chosen.

semconv/model/android/events.yaml
```diff
@@ -1,4 +1,13 @@
 groups:
+  - id: event.app.navigation
```
aranhave: nit: Since the event is now `app.navigation.complete`, can we update this ID to `event.app.navigation.complete` too? The other local event IDs mirror their event names.

aranhave (approved):
LGTM, just one optional nit 🙂

aranhave (approved):

fractalwrench (approved):
LGTM - thanks!

instrumentation/compose/navigation/src/test/kotlin/io/opentelemetry/instrumentation/compose/navigation/NavigationEmitterTest.kt
```diff
@@ -0,0 +1,89 @@
...
+import io.opentelemetry.android.OpenTelemetryRum
+import io.opentelemetry.api.common.AttributeKey.stringKey
+import io.opentelemetry.sdk.testing.assertj.OpenTelemetryAssertions.assertThat
+import io.opentelemetry.sdk.testing.assertj.OpenTelemetryAssertions.equalTo
+import io.opentelemetry.sdk.testing.junit4.OpenTelemetryRule
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+
+// Emitted by the generated AppNavigationEvent (semconv/model/android/events.yaml).
```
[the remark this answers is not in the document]
benjoseph-grafana: updated.

instrumentation/compose/navigation/src/main/kotlin/io/opentelemetry/instrumentation/compose/navigation/NavControllerExtensions.kt
```diff
@@ -20,9 +20,9 @@ import io.opentelemetry.android.OpenTelemetryRum
 
 /**
  * Attaches OpenTelemetry navigation instrumentation to this [NavController]: on every completed
- * destination change the new destination is resolved to a screen name and handed to the
- * instrumentation backed by [rum] (telemetry emission is tracked in issue #1909). The listener is
- * scoped to the composition via [DisposableEffect] and removed when the controller leaves it.
+ * destination change, the new destination is resolved to a screen name and emitted as an
+ * `app.navigation.complete` event through [rum]. The listener is scoped to the composition via
+ * [DisposableEffect] and removed when the controller leaves it.
```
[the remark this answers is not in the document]
benjoseph-grafana: done.

instrumentation/compose/navigation/src/main/kotlin/io/opentelemetry/instrumentation/compose/navigation/NavigationEmitter.kt
```diff
@@ -6,19 +6,16 @@
...
 
 internal const val INSTRUMENTATION_SCOPE_NAME =
     "io.opentelemetry.android.instrumentation.compose.navigation"
 
 /**
- * Receives the resolved screen name each time a navigation completes (the destination changes).
- * Telemetry emission is tracked in issue #1909, pending the mobile semantic-conventions
- * discussion on modelling navigation.
+ * Emits an `app.navigation.complete` event each time a navigation completes (the destination changes).
  */
```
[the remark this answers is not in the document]
benjoseph-grafana: done.

breedx-splk (approved):
Looking good, thanks!

PR #1972, opened 2026-08-07
test: add minified emulator smoke test

### LINKED ISSUE #1955: Create smoke test that asserts the SDK launches when running on an emulator
It would be nice to create smoke test that asserts the SDK launches in a simple test harness app when running on an emulator, and exports some basic data. This would provide a higher level of confidence that changesets aren't breaking the SDK.

This would have to happen after setting up instrumented test CI job: https://github.com/open-telemetry/opentelemetry-android/issues/1954

aranhave:
Adds a small minified release app that initializes the Android agent, records a span, and exports it over OTLP/HTTP. A separate emulator test captures the request with a bounded loopback server and verifies the exported protobuf payload.

The test runs through the existing `connectedCheck` PR job and covers R8 shrinking, app startup, SDK initialization, and a real export.

## Test flow for clarity

```mermaid
flowchart TD
    A[connectedCheck] --> B[Start emulator smoke test]
    B --> C[Start bounded loopback server]
    B --> D[Launch minified release app]
    D --> E[Initialize Android agent]
    E --> F[Create one span]
    F -->|OTLP/HTTP protobuf| C
    C --> G[Decode payload and verify span]
```

Resolves #1955

### DIFF
--- a/smoke-test/build.gradle.kts
+++ b/smoke-test/build.gradle.kts
@@ -1,0 +1,64 @@
+import org.jetbrains.kotlin.gradle.dsl.JvmTarget
+import org.jetbrains.kotlin.gradle.dsl.KotlinVersion
+
+plugins {
+    id("com.android.test")
+    id("otel.spotless-conventions")
+}
+
+android {
+    namespace = "io.opentelemetry.android.smoketest"
+    compileSdk = (property("android.compileSdk") as String).toInt()
+
+    defaultConfig {
+        minSdk = (property("android.minSdk") as String).toInt()
+        testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
+    }
+
+    targetProjectPath = ":smoke-test-app"
+
+    buildTypes {
+        create("release") {
+            isMinifyEnabled = true
+            signingConfig = signingConfigs.getByName("debug")
+            testProguardFiles(
+                getDefaultProguardFile("proguard-android-optimize.txt"),
+                "proguard-rules.pro",
+            )
+        }
+    }
+
+    compileOptions {
+        sourceCompatibility(rootProject.extra["java_version"] as JavaVersion)
+        targetCompatibility(rootProject.extra["java_version"] as JavaVersion)
+    }
+
+    kotlin {
+        compilerOptions {
+            jvmTarget.set(rootProject.extra["jvm_target"] as JvmTarget)
+            apiVersion.set(rootProject.extra["kotlin_min_supported_version"] as KotlinVersion)
+            languageVersion.set(rootProject.extra["kotlin_min_supported_version"] as KotlinVersion)
+        }
+    }
+
+    packaging.resources.excludes += "META-INF/versions/9/OSGI-INF/MANIFEST.MF"
+}
+
+androidComponents {
+    beforeVariants(selector().withBuildType("debug")) {
+        it.enable = false
+    }
+}
+
+dependencies {
+    implementation(libs.androidx.junit)
+    implementation(libs.androidx.test.core)
+    implementation(libs.androidx.test.rules)
+    implementation(libs.androidx.test.runner)
+    implementation(libs.okhttp.mockwebserver)
+    implementation(libs.opentelemetry.proto)
+}
+
+configurations.implementation {
+    exclude(group = "io.opentelemetry", module = "opentelemetry-sdk-testing")
+}
--- a/smoke-test/build.gradle.kts
+++ b/smoke-test/build.gradle.kts
@@ -45,6 +45,7 @@ android {
 }
 
 androidComponents {
+    // The test must run against the target's minified release variant.
     beforeVariants(selector().withBuildType("debug")) {
         it.enable = false
     }
@@ -55,10 +56,5 @@ dependencies {
     implementation(libs.androidx.test.core)
     implementation(libs.androidx.test.rules)
     implementation(libs.androidx.test.runner)
-    implementation(libs.okhttp.mockwebserver)
     implementation(libs.opentelemetry.proto)
-}
-
-configurations.implementation {
-    exclude(group = "io.opentelemetry", module = "opentelemetry-sdk-testing")
 }
--- a/smoke-test/build.gradle.kts
+++ b/smoke-test/build.gradle.kts
@@ -40,8 +40,6 @@ android {
             languageVersion.set(rootProject.extra["kotlin_min_supported_version"] as KotlinVersion)
         }
     }
-
-    packaging.resources.excludes += "META-INF/versions/9/OSGI-INF/MANIFEST.MF"
 }
 
 androidComponents {
--- a/smoke-test/src/main/kotlin/io/opentelemetry/android/smoketest/MinifiedAppSmokeTest.kt
+++ b/smoke-test/src/main/kotlin/io/opentelemetry/android/smoketest/MinifiedAppSmokeTest.kt
@@ -9,13 +9,14 @@ import android.app.Activity
 import android.content.Intent
 import androidx.test.core.app.ActivityScenario
 import androidx.test.ext.junit.runners.AndroidJUnit4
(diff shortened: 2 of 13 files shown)


smoke-test/src/main/kotlin/io/opentelemetry/android/smoketest/MinifiedAppSmokeTest.kt
```diff
@@ -0,0 +1,159 @@
...
+
+    private fun List<KeyValue>.stringValues(): Map<String, String> = associate { it.key to it.value.stringValue }
+
+    private fun assertValidId(
+        name: String,
+        value: ByteArray,
+        expectedSize: Int,
+    ) {
+        assertSmoke(value.size == expectedSize && value.any { it != 0.toByte() }) {
+            "Expected a valid $name ID but received ${value.joinToString("") { "%02x".format(it) }}"
```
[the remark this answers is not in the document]
aranhave: Thanks, I checked the exact formatting behavior. The value is boxed as a `Byte`, so negative values already render as two hex digits (`-1` becomes `ff` and `-128` becomes `80`). The eight-digit output only happens after promotion to `Int`, so I am leaving this as-is.

smoke-test/src/main/kotlin/io/opentelemetry/android/smoketest/OtlpHttpServer.kt
```diff
@@ -0,0 +1,156 @@
...
+import java.util.concurrent.TimeUnit
+
+internal data class CapturedHttpRequest(
+    val method: String,
+    val target: String,
+    val headers: Map<String, String>,
+    val body: ByteArray,
+)
+
+internal class OtlpHttpServer : Closeable {
```
fractalwrench, line 27: I think OkHttp's [mock webserver](https://github.com/lysine-dev/okhttp/blob/main/mockwebserver/README.md) would help reduce boilerplate here. Alternatively I think @breedx-splk mentioned there were some existing test utilities for receiving OTLP export that are in opentelemetry-java?

smoke-test/build.gradle.kts
```diff
@@ -0,0 +1,60 @@
...
+
+    kotlin {
+        compilerOptions {
+            jvmTarget.set(rootProject.extra["jvm_target"] as JvmTarget)
+            apiVersion.set(rootProject.extra["kotlin_min_supported_version"] as KotlinVersion)
+            languageVersion.set(rootProject.extra["kotlin_min_supported_version"] as KotlinVersion)
+        }
+    }
+
+    packaging.resources.excludes += "META-INF/versions/9/OSGI-INF/MANIFEST.MF"
```
fractalwrench: What dependency brings this in - is it something opentelemetry-android adds?

smoke-test/src/main/kotlin/io/opentelemetry/android/smoketest/MinifiedAppSmokeTest.kt
```diff
@@ -0,0 +1,159 @@
...
+            }
+        }
+
+        throw AssertionError(
+            "Timed out waiting for span '$SMOKE_TEST_SPAN_NAME'. " +
+                "Requests: $receivedTargets; spans: $receivedSpanNames",
+        )
+    }
+
+    private fun assertExpectedTrace(
```
fractalwrench, line 75: This is fine for an MVP. One thing that might be worth considering longer-term is golden file comparisons, which would allow us to compare against a known output without needing to write individual assertions for each test case. It does require sanitization of any values that are unstable between test runs, so probably requires some discussion..

fractalwrench (commented):
Thanks for taking this on! I left a few comments inline around the test harness and assertions that follow on from discussion at the last SIG. The test case and app itself look fine to me

smoke-test/src/main/kotlin/io/opentelemetry/android/smoketest/OtlpHttpServer.kt
aranhave: I tried MockWebServer with the minified release setup. Because OkHttp and Okio are shared between the target and test APKs, the independently minified APKs fail at runtime unless the target keeps all of `okhttp3.**` and `okio.**`. That would weaken the R8 and exporter coverage this test is intended to provide, so I kept the small loopback server here.

smoke-test/build.gradle.kts
aranhave: This suppression isn’t needed. I removed it and verified that both release APKs still minify successfully and the connected API 29 smoke test passes.

smoke-test/src/main/kotlin/io/opentelemetry/android/smoketest/MinifiedAppSmokeTest.kt
aranhave: That makes sense as a follow-up. I’ll keep the focused assertions for this MVP and leave golden-file normalization and comparison for a separate discussion.

smoke-test/src/main/kotlin/io/opentelemetry/android/smoketest/OtlpHttpServer.kt
fractalwrench: Marking this as unresolved as I'd like Jason's input on this question. The test server itself _might_ be fine although my instinct is that Mockwebserver will be simpler.

There are existing assertions for asserting against telemetry (although I can't remember the exact module where they live) and we should decide whether they're a good fit here.

aranhave: Sounds good, I’ll wait for Jason’s input before changing the test harness. @breedx-splk

breedx-splk: Hey @aranhave thanks for this, I think it's awesome. I mentioned earlier that the opentelemetry-java-instrumentation (and our Splunk distro of the same) have a pretty advanced/sophisticated set of smoke tests. These run a "fake backend" via docker (image is here https://github.com/open-telemetry/opentelemetry-java-instrumentation/pkgs/container/opentelemetry-java-instrumentation%2Fsmoke-test-fake-backend)...and then they run instrumented code and can then fetch back the exported data in json format to do assertions.

I think the challenge with this right now is that it looks to me like all of the components are running on the emulate device: the (junit) test code, the fake backend app (from this PR), and the minified and instrumented app. With that approach, it's not really possible to leverage the existing docker image.

Do we think it's important/wise to have everything running in the emulator? Can we set things up so that the test and the server run off-device, and only the app runs on device?

aranhave: Only the minified app needs to run on the emulator. Moving the server and orchestration to the host seems like the cleaner long-term direction and avoids the test/target APK dependency issue. The Java fake backend currently expects OTLP/gRPC while the Android agent uses OTLP/HTTP, so I’d keep this PR focused and follow up with a host-side harness once we choose between HTTP support or a collector bridge. Does that sound reasonable?

breedx-splk: For additional discussion: https://github.com/breedx-splk/opentelemetry-android/pull/7 (draft in my fork)

aranhave: Thanks for putting this together. The smoke test passed in the emulator job, so the host-side setup looks viable. The overall failure looks like `check` also runs `smokeTest` without an emulator. I’m happy to adapt #1972 to this structure and fix the task wiring if this is the direction we want.

breedx-splk (approved):
Ok, we chatted about this in SIG today and decided that this is still helpful as-is, even if we all think that reusing the fake backend from upstream is the way to go. So we'll merge this now and I'll create a follow-up item to come back and modify the smoke test framework/architecture to reuse the fake backend.

PR #1984, opened 2026-08-13
Bump compose dependency in demo app

fractalwrench:
Supersedes #1982 by increasing the compileSdk version used for building the demo app, as the latest version of compose requires compileSdk of 37 or above.


breedx-splk (approved):
😎

Issue #1940, opened 2026-07-29
Add native stack unwinding to native crash reports


aranhave:
### Problem

The native-crash instrumentation now records the fatal signal, timestamp, and crash-time context, then replays an `app.crash` event after restart. It does not capture the native C/C++ frames, so the event cannot show where the crash occurred. This is the remaining requirement called out in #764.

### Scope

- choose a signal-safe way to capture or persist enough crash-time state to produce native frames after restart
- include a usable native stack in the replayed `app.crash` event
- cover supported Android APIs and ABIs, including malformed or partial crash data
- preserve the current handler chaining and one-time marker cleanup behavior

The signal handler must remain async-signal-safe: no allocation, locks, JNI, or logging. On-device symbolication and backend rendering can remain separate work.

Related to #764. Marker format compatibility is tracked in #1915, and signal-handler interoperability is tracked in #1928.

aranhave:
@breedx-splk @fractalwrench @LikeTheSalad Before I get started, I’d like to align on the direction. Should we keep this dependency-light or are we open to a minidump solution such as Breakpad or Crashpad? For the first version, should the replayed event contain raw frame addresses and module IDs for server-side symbolication or resolved native frames? I can compare the options and share a recommendation before implementing.

fractalwrench:
I'd lean towards stack unwinding in the signal handler using libunwindstack, mainly as that's the approach I've taken before so it's a known quantity. I'd be open to other approaches if we discussed why Breakpad/Crashpad would be a better fit for our needs.

Symbolication is something that we need to discuss and agree on. The instrumentation isn't super useful without either a symbol map, or an ID that associates against symbols that are uploaded separately. Libunwindstack does have a mode where it attempts to get useful function names, but IIRC that doesn't work in all situations..

aranhave:
> I'd lean towards stack unwinding in the signal handler using libunwindstack, mainly as that's the approach I've taken before so it's a known quantity. I'd be open to other approaches if we discussed why Breakpad/Crashpad would be a better fit for our needs.
> 
> Symbolication is something that we need to discuss and agree on. The instrumentation isn't super useful without either a symbol map, or an ID that associates against symbols that are uploaded separately. Libunwindstack does have a mode where it attempts to get useful function names, but IIRC that doesn't work in all situations..

[...]

PR #1999, opened 2026-08-21
feat(native-crash): define stack snapshot format

aranhave:
Defines the versioned native crash snapshot contract for #1940 before capture and recovery are added.

The shared C header pins the magic value, checksum constants, ABI mapping, field offsets, and total size with compile-time assertions. The format document records the writer and reader rules, crash-handler constraints, recovery ordering, and current limitations.

## Scope

This is a contract-only change. It does not change runtime crash behavior and can merge independently.

## Follow-up sequence

The implementation is prepared as small, stacked changes so each safety boundary can be reviewed separately:

1. Parse and validate the snapshot, rejecting malformed records and module entries.
2. Recover best-effort module-relative frames from the program counter, frame-pointer chain, and link register. Version 1 will not use heuristic stack scanning.
3. Add recovered frames to the replayed `app.crash` event, with bounded retries and cleanup only after successful processing. Recovery boundaries will contain ordinary exceptions and linkage failures.
4. Capture the fixed-size record in the signal handler without allocation, locks, JNI, logging, or process-map parsing, while preserving `errno`, handler chaining, and reentrancy behavior.
5. Validate the complete path across the four supported ABIs, debug and release builds, malformed and partial records, cleanup failures, and handler interoperability.

Each follow-up can be reviewed and merged independently after its prerequisite lands. Symbol upload and backend symbolication remain separate work tracked outside this stack.

Related to #1940

### DIFF
--- a/instrumentation/native-crash/SNAPSHOT_FORMAT.md
+++ b/instrumentation/native-crash/SNAPSHOT_FORMAT.md
@@ -4,6 +4,9 @@ Status: proposed version 1 design for
 [#1940](https://github.com/open-telemetry/opentelemetry-android/issues/1940). Runtime capture,
 recovery, and replay are follow-up work and are not implemented by this change.
 
+The writer and reader rules below are normative requirements for those follow-up changes. They
+describe the completed version 1 behavior even though this PR defines only the shared contract.
+
 The native crash implementation uses a fixed-size snapshot so the signal handler can copy bounded
 state without allocation, locks, JNI, logging, or process-map parsing. Capture, parsing, unwinding,
 and replay are separate changes. `native_crash_snapshot.h` is the source of truth for the binary
@@ -49,6 +52,15 @@ satisfies `loadBias <= executableStart < executableEnd`. Names are nonblank UTF-
 characters, use at most 63 bytes, and always include a NUL terminator followed by zero padding.
 Build IDs longer than 32 bytes are recorded as unavailable rather than truncated.
 
+## Files And Pairing
+
+The marker and snapshot are separate fixed internal files whose paths are supplied when the native
+handler is installed. Temporary writes use the corresponding destination path with a `.tmp` suffix.
+Exact destination names are implementation details rather than part of the binary format.
+
+Recovery pairs a snapshot with a marker only when their signal number and timestamp match. A stale
+or otherwise mismatched snapshot is invalid and is never attached to the marker's crash event.
+
 ## Writer Rules
 
 Register and module addresses are unsigned. A 32-bit writer zero-extends every address into its
@@ -63,13 +75,12 @@ outside the signal handler. This costs 20,568 bytes of process storage but does 
 alternate signal stack or allocate during a crash.
 
 The marker and snapshot use the same signal number and a timestamp from a single `clock_gettime`
-call. The snapshot is written and atomically renamed before the marker, making the marker the commit
-record for a complete crash capture. The snapshot skips `fsync` to bound work in the fatal signal
-handler; process death does not discard completed page-cache writes. The smaller marker is synced
-before rename. A marker without a snapshot remains a valid crash without native frames. A snapshot
-without a marker is an orphan and is removed during startup recovery. This ordering treats the
-marker as the commit record, but an abrupt process termination during the snapshot write can lose
-the crash before the marker is created.
+call. The marker is written, synced, and atomically renamed first. The snapshot is then written and
+atomically renamed without `fsync` to bound work in the fatal signal handler. A second fault or
+abrupt process termination during the larger snapshot write therefore still leaves the marker as a
+valid crash without native frames. A snapshot without a marker is an orphan and is removed during
+startup recovery. Signal and timestamp matching prevents an older snapshot from being attached to
+a newer marker.
 
 Stack copying is fault-tolerant; a failed or short read produces a stack size from zero through
 4,096 rather than reading through an invalid page. The handler loops over interrupted short writes,
@@ -85,10 +96,12 @@ it is not an authenticity mechanism. Malformed module entries are skipped withou
 structurally valid entries. A record with no usable module entries remains valid but produces no
 native frames.
 
-Readers reject zero program counters or stack pointers, a stack start that differs from the stack
-pointer, and a stack start that is not aligned to the recorded architecture's pointer width. A
-misaligned or otherwise corrupt stack pointer therefore produces no native frames. Readers ignore
-the link-register field on x86 and x86_64.
+Readers reject zero stack pointers, a stack start that differs from the stack pointer, and a stack
+start that is not aligned to the recorded architecture's pointer width. A zero program counter is
+valid for crashes such as an indirect call through a null function pointer. Readers omit the
+program-counter frame in that case and continue with frame-pointer and link-register recovery. A
+misaligned or otherwise corrupt stack pointer produces no native frames. Readers ignore the
+link-register field on x86 and x86_64.
 
 ARM64 readers preserve captured values and try the raw address first. If it does not resolve, they
 retry after clearing bits 56 through 63, then 52 through 63, then 48 through 63 to account for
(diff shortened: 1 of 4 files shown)


instrumentation/native-crash/SNAPSHOT_FORMAT.md
```diff
@@ -0,0 +1,154 @@
...
+| Offset | Size | Field |
+| ---: | ---: | --- |
+| 0 | 8 | Magic: `OTELNCS\0` |
+| 8 | 4 | Format version (`1`) |
+| 12 | 4 | Architecture: ARM (`1`), ARM64 (`2`), x86 (`3`), or x86_64 (`4`) |
+| 16 | 4 | Total record size (`20568`) |
+| 20 | 4 | Fatal signal number |
+| 24 | 8 | Crash timestamp in Unix epoch nanoseconds |
+| 32 | 32 | Program counter, stack pointer, frame pointer, and link register |
+| 64 | 4 | Number of populated module entries (`1..128`) |
```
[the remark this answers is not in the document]
aranhave: I kept `module_count` at `1..128` intentionally. A snapshot with no populated modules can’t produce module-relative frames, so that case falls back to the marker-only crash path. I updated the reader wording to clarify that the valid case is one or more populated entries that are all skipped during validation.

aranhave:
@fractalwrench when you have a chance, could you take a look at this? It’s the contract-only first step from the snapshot approach agreed in #1940.

instrumentation/native-crash/SNAPSHOT_FORMAT.md
```diff
@@ -0,0 +1,154 @@
...
+## Version 1
+
+Version 1 is a 20,568-byte, little-endian record. Readers require the exact version and record size.
+An incompatible record is discarded rather than interpreted using a newer layout.
+
+| Offset | Size | Field |
+| ---: | ---: | --- |
+| 0 | 8 | Magic: `OTELNCS\0` |
+| 8 | 4 | Format version (`1`) |
+| 12 | 4 | Architecture: ARM (`1`), ARM64 (`2`), x86 (`3`), or x86_64 (`4`) |
```
fractalwrench, line 28: Should we even bother supporting non-64 bit architectures? I suspect it's more effort than it's worth

instrumentation/native-crash/SNAPSHOT_FORMAT.md
```diff
@@ -0,0 +1,154 @@
...
+| 16 | 4 | Total record size (`20568`) |
+| 20 | 4 | Fatal signal number |
+| 24 | 8 | Crash timestamp in Unix epoch nanoseconds |
+| 32 | 32 | Program counter, stack pointer, frame pointer, and link register |
+| 64 | 4 | Number of populated module entries (`1..128`) |
+| 68 | 4 | Number of captured stack bytes (`0..4096`) |
+| 72 | 8 | Address of the first captured stack byte; equal to the stack pointer |
+| 80 | 16384 | 128 fixed-size module entries |
```
fractalwrench, line 36: Can you explain what these are?

instrumentation/native-crash/SNAPSHOT_FORMAT.md
```diff
@@ -0,0 +1,154 @@
...
+Each 128-byte module entry contains:
+
+| Relative offset | Size | Field |
+| ---: | ---: | --- |
+| 0 | 8 | Load bias |
+| 8 | 8 | Executable segment start |
+| 16 | 8 | Executable segment end |
+| 24 | 64 | NUL-terminated UTF-8 module basename |
+| 88 | 4 | ELF build-ID length (`0..32`; zero means unavailable) |
+| 92 | 32 | Raw ELF build-ID bytes followed by zero padding |
```
fractalwrench, line 55: I'm curious on how the ELF information will be written at crash-time. Is it definitely available with the chosen capture solution?

instrumentation/native-crash/SNAPSHOT_FORMAT.md
```diff
@@ -0,0 +1,154 @@
...
+| 124 | 4 | Reserved; must be zero |
+
+The writer zero-fills unused entries. It records app-owned modules before system modules and
+preserves loader order within each group, so the table is not address-sorted. Every populated entry
+satisfies `loadBias <= executableStart < executableEnd`. Names are nonblank UTF-8 without control
+characters, use at most 63 bytes, and always include a NUL terminator followed by zero padding.
+Build IDs longer than 32 bytes are recorded as unavailable rather than truncated.
```
fractalwrench: Is that desirable behavior vs increasing the space allocation for build ID?

fractalwrench (commented):
Took a look at the format and am broadly happy, I left a few questions. It'd definitely be good to get someone else to review given the complexity of unwinding stacks in signal handlers.

instrumentation/native-crash/SNAPSHOT_FORMAT.md
aranhave: The module currently builds all four Android NDK ABIs, including ARM32 and x86, so I kept the snapshot format aligned with that support. I added this context to the doc. If the module drops 32-bit support later, we can narrow the format too.

instrumentation/native-crash/SNAPSHOT_FORMAT.md
aranhave: It won’t be discovered during the crash. The capture implementation prepares the table from the loaded ELF headers before installing the signal handler, then the handler only copies that prepared data. If preparation fails, we still record the crash marker without a snapshot. I clarified this in the writer rules and documented late-loaded modules as a limitation.

instrumentation/native-crash/SNAPSHOT_FORMAT.md
aranhave: I kept the 32-byte limit because it covers build IDs up to 256 bits. Truncating anything longer could identify the wrong artifact, while increasing the field would only move the limit and enlarge every snapshot. Treating it as unavailable is safer, so I added that reasoning.

instrumentation/native-crash/SNAPSHOT_FORMAT.md
aranhave: Each entry represents an executable ELF segment loaded in the process. Its address range and load bias let us produce a module-relative frame, while the basename and build ID identify the binary for later symbolication. I added this above the table.

aranhave:
> Took a look at the format and am broadly happy, I left a few questions. It'd definitely be good to get someone else to review given the complexity of unwinding stacks in signal handlers.

Thanks for taking a look! 🙇  I’ve clarified each of these points in the doc. Agreed on getting another set of 👀  on this, especially for the capture and unwinding follow-ups. I’ve kept those as separate PRs so each part can be reviewed independently.
cc: @breedx-splk @LikeTheSalad

instrumentation/native-crash/src/main/cpp/native_crash_snapshot.h
```diff
@@ -0,0 +1,98 @@
...
+#if defined(__arm__)
+#define OTEL_NCS_ARCH_CURRENT OTEL_NCS_ARCH_ARM
+#elif defined(__aarch64__)
+#define OTEL_NCS_ARCH_CURRENT OTEL_NCS_ARCH_ARM64
+#elif defined(__i386__)
+#define OTEL_NCS_ARCH_CURRENT OTEL_NCS_ARCH_X86
+#elif defined(__x86_64__)
+#define OTEL_NCS_ARCH_CURRENT OTEL_NCS_ARCH_X86_64
+#else
+#error "Unsupported architecture for native crash snapshots"
```
LikeTheSalad: We don't define [abiFilters](https://developer.android.com/ndk/guides/abis#gc) for this module, and my understanding is that this line breaks the build for any other arch, so we'd lose not only crash snapshots but also the previously [added marker](https://github.com/open-telemetry/opentelemetry-android/pull/1899), which isn't arch-specific. Maybe we could define an "unknown" ABI value `0` instead, skip snapshot capture in that case, and maybe even add a `#warning` to show something in the compiler logs?

aranhave: I reproduced this with an NDK `riscv64` target. I added an internal unknown-architecture value so the marker path still builds, while unsupported architectures skip snapshot capture. I left out `#warning` because this target uses `-Werror`.

LikeTheSalad (approved):
Thank you!

fractalwrench (approved):
LGTM - thanks!

aranhave:
@fractalwrench @LikeTheSalad thanks for the review! Would you be able to merge this when you get a chance?

PR #2003, opened 2026-08-24
fix: handle unsafe call to getSystemService

### LINKED ISSUE #1998: Unsafe call to getSystemService
There are a couple of unsafe calls to `getSystemService()` in `NetworkDetectorImpl`, as the call can return null. We should use `as?` rather than `as`.

https://github.com/open-telemetry/opentelemetry-android/blob/47d391cf4ef42b2aeff409699cf206d75735f108/services/src/main/java/io/opentelemetry/android/internal/services/network/NetworkDetectorImpl.kt#L27

aaravmahajanofficial:
Fixes #1998


fractalwrench (approved):
LGTM - thanks!

LikeTheSalad (approved):
Thank you!

PR #2007, opened 2026-08-24
fix: change default dispatcher to IO for PeriodicTaskScheduler

### LINKED ISSUE #2005: PeriodicTaskSchedulerImpl should run on Dispatchers.IO
`PeriodicTaskSchedulerImpl` runs its tasks on `Dispatchers.Default` and these tasks involve I/O (disk and network). We should use `Dispatchers.IO` instead as using `Dispatchers.Default` could starve the host app's work.

aaravmahajanofficial:
Fixes #2005


fractalwrench (approved):
LGTM

aranhave (approved):

PR #2008, opened 2026-08-24
refactor: CacheStorageImpl with lazy initialization and add tests

### LINKED ISSUE #2006: Cache property read in CacheStorageImpl
CacheStorageImpl calls `cacheDir` when initializing the SDK twice, which hits the filesystem. We should cache the result and only hit the filesystem once.

aaravmahajanofficial:
Fixes #2006 

### Description
- `CacheStorageImpl` evaluated `appContext.cacheDir` on every property access via a custom getter. 
- Multiple components read `cacheDir` on startup, causing redundant disk checks

### Solution
- Updated `CacheStorageImpl.cacheDir` to use Kotlin's `by lazy` delegate.
- Added a unit test in `CacheStorageImplTest` to verify that `appContext.cacheDir` is evaluated only once on multiple reads.


fractalwrench (approved):
LGTM - thanks!

aranhave (approved):

PR #2009, opened 2026-08-24
refactor: defer executor and thread creation in `AnrInstrumentation` and `NativeCrashInstrumentation`

### LINKED ISSUE #2004: Threads are created at SPI discovery time
`AnrInstrumentation` creates `Executors.newScheduledThreadPool(1)` and `Looper.getMainLooper()` in field initialisers and `NativeCrashInstrumentation` creates
`Executors.newSingleThreadExecutor()` as a constructor default. If an app supresses these instrumentations they still are stuck with two unused threads.

Making these fields `lazy` would be a reasonable way to prevent this.

aaravmahajanofficial:
Fixes #2004

### DIFF
--- a/instrumentation/anr/src/test/java/io/opentelemetry/android/instrumentation/anr/AnrInstrumentationTest.kt
+++ b/instrumentation/anr/src/test/java/io/opentelemetry/android/instrumentation/anr/AnrInstrumentationTest.kt
@@ -1,0 +1,38 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.android.instrumentation.anr
+
+import io.mockk.every
+import io.mockk.mockk
+import io.mockk.mockkStatic
+import io.mockk.unmockkStatic
+import io.mockk.verify
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.AfterEach
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import java.util.concurrent.Executors
+import java.util.concurrent.ScheduledExecutorService
+
+class AnrInstrumentationTest {
+    @BeforeEach
+    fun setUp() {
+        mockkStatic(Executors::class)
+        every { Executors.newScheduledThreadPool(any()) } returns mockk<ScheduledExecutorService>(relaxed = true)
+    }
+
+    @AfterEach
+    fun tearDown() {
+        unmockkStatic(Executors::class)
+    }
+
+    @Test
+    fun `does not create scheduler when not installed`() {
+        assertThat(AnrInstrumentation().name).isEqualTo("anr")
+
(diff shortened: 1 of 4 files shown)


instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
```diff
@@ -43,9 +43,10 @@ class NativeCrashInstrumentation internal constructor(
     private val storeFactory: (Context) -> NativeCrashStore = { context ->
         FileNativeCrashStore(File(context.filesDir, "opentelemetry/native-crash"))
     },
-    private val executor: Executor = Executors.newSingleThreadExecutor(),
+    executor: Executor? = null,
     private val signalHandlerInstaller: NativeSignalHandlerInstaller = JniNativeSignalHandlerInstaller(),
 ) : AndroidInstrumentation {
+    private val executor: Executor by lazy { executor ?: Executors.newSingleThreadExecutor() }
```
aranhave, line 49: Android’s ThreadPoolExecutor creates workers on the first submitted task, and install() was already that point for both instrumentations. This only defers allocating the executor objects, not the threads. Can we update the issue/PR rationale to match?

instrumentation/anr/src/test/java/io/opentelemetry/android/instrumentation/anr/AnrInstrumentationTest.kt
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    @AfterEach
+    fun tearDown() {
+        unmockkStatic(Executors::class)
+    }
+
+    @Test
+    fun `does not create scheduler when not installed`() {
+        assertThat(AnrInstrumentation().name).isEqualTo("anr")
```
aranhave, line 34: Could we add install-path coverage for these defaults? The current tests only prove construction is lazy. all three new fallback branches remain partial, and codecov/patch is failing at 40%.

instrumentation/anr/src/test/java/io/opentelemetry/android/instrumentation/anr/AnrInstrumentationTest.kt
```diff
@@ -0,0 +1,38 @@
...
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import java.util.concurrent.Executors
+import java.util.concurrent.ScheduledExecutorService
+
+class AnrInstrumentationTest {
+    @BeforeEach
+    fun setUp() {
+        mockkStatic(Executors::class)
+        every { Executors.newScheduledThreadPool(any()) } returns mockk<ScheduledExecutorService>(relaxed = true)
```
breedx-splk, line 24: Is this needed if it's not ever called? Also, might it not be safer to verify that there were no other methods called? Like if there were a bug and a refactoring left it calling `Executors.newSingleThreadExecutor()` this test might still pass?

Issue #1909, opened 2026-07-21
Update VisibleScreenTracker to use Compose screen names.


benjoseph-grafana:
https://github.com/open-telemetry/opentelemetry-android/pull/1901 introduces Compose navigation instrumentation. The next step is to generate navigation events and attribute stamping on spans/logs based on the instrumentation.

For the attribute stamping, some options to consider -
- a new attribute for Compose derived screens vs. the Activity derived one, so both can coexist
- the pattern some libraries use- an exclusion list of Activities whose names are ignored for screen attribution, so Compose screens get route based names instead of host Activity

Generating navigation events is a separate issue tracked by https://github.com/open-telemetry/opentelemetry-android/issues/1920

hitkall:
Hey, I've been poking around this one and wanted to share what I found before anyone spends time on it.

I contributed the thermal status instrumentation a while back, so I've spent a bit of time in this codebase, though the Compose side is newer to me.

The thing that jumped out: VisibleScreenTracker doesn't currently have a way to hear about a screen that isn't an Activity or a Fragment. All four hooks take framework types, and the precedence is fragment, then activity, then "unknown". So neither of the two options in the description is really reachable until that changes. Might be obvious to you all already, but it wasn't to me until I went looking.

On A vs B, I lean toward the separate attribute, mostly for boring practical reasons. The exclusion list would match on javaClass.simpleName, and I couldn't find any keep rules in the repo, so a list that works fine in debug would quietly stop matching under R8 in release. That feels like the kind of thing that turns into confused bug reports rather than a clean failure. It also can't really express an Activity hosting both a Compose NavHost and fragments, and there'd be an "unknown" gap on every cold start between the activity resuming and the first onDestinationChanged.

Two things I'd genuinely like your input on before writing any code:

Which attribute to stamp. app.navigation.destination.name is already in the registry, but the brief describes the destination reached by a completed navigation, which reads like transition semantics to me. Putting it on every span and log turns it into more of a "where am I right now" attribute. Is widening that brief the right call, or would you rather have a separate one? I don't want to guess at this while the semconv work is in flight.
Staleness. A NavController destination has no pause callback, just onDispose plus a replay on config changes. When the NavHost leaves composition, or a non-Compose Activity comes forward, should the value clear or stick as last-known? I don't have a strong opinion, but it needs an answer either way.

[...]

PR #2011, opened 2026-08-25
Make VisibleScreenTracker aware of Compose Navigation destinations

hitkall:
## What this does

Gives `VisibleScreenTracker` a source-agnostic way to be told about a screen that is neither an
Activity nor a Fragment, and wires it up from the Compose Navigation instrumentation's
`NavController.OnDestinationChangedListener`.

- New `AtomicReference` slot on `VisibleScreenTrackerImpl` for a navigation destination name.
- Precedence for `currentlyVisibleScreen` becomes: **destination → fragment → activity → `"unknown"`**.
- `:instrumentation:compose:navigation` gains `implementation(project(":services"))`. This edge is
  precedented: `:instrumentation:compose:click` already depends on `:services`.
- The listener reports the resolved screen name through a new internal
  `NavigationDestinationReporter`, and the `DisposableEffect`'s `onDispose` clears it.

Clearing is **conditional**: `navigationDestinationCleared(name)` only clears when `name` is still
the recorded destination, via `compareAndSet`, matching the pattern already used by `activityPaused`
and `fragmentPaused`. Each attached controller owns its own reporter, so a nested or sibling
controller leaving the composition cannot discard a destination another controller has since
recorded. Without this, teardown ordering between nested controllers could blank a destination the
parent had just recorded and silently drop attribution back to the Activity name.

## What this deliberately does not do

- **No changes to `ScreenAttributesSpanProcessor` or `ScreenAttributesLogRecordProcessor`.**
- **No change to which attribute is stamped, and no new semantic convention entry.**
- **#1920 (navigation event generation) is untouched.** The `app.navigation.complete` event and its
  `app.navigation.destination.name` attribute behave exactly as before.
- `previouslyVisibleScreen` is untouched, so `last.screen.name` keeps its Activity/Fragment-only
  meaning. There is a test pinning this.

One consequence worth stating plainly, because it is the point of the change rather than a side
effect: no attribute *key* changes, but for apps that attach this instrumentation the *value* of
`app.screen.name` will now be the Compose destination name rather than the host Activity name. The
instrumentation is manual and opt-in — it is not auto-discovered — so only apps that call
`withOpenTelemetry` or `rememberObservedNavController` are affected.

## Public API delta — Tier 4

Two methods on `VisibleScreenTracker`, and nothing else:

```
public abstract fun navigationDestinationChanged (Ljava/lang/String;)V
public abstract fun navigationDestinationCleared (Ljava/lang/String;)V
```

`navigation.api` is unchanged — `attachOpenTelemetry` and `NavigationDestinationReporter` are
`internal`. Per CONTRIBUTING this is Tier 4: 2 approvals and a 2-day waiting period.

## Known consequence: DialogFragment

Under destination-first precedence, a `DialogFragment` shown over a Compose screen is no longer
reflected in `currentlyVisibleScreen` — the recorded destination outranks it. This follows directly
from the precedence order rather than from a defect in the implementation, and it is the tradeoff
against the alternative ordering, where a `NavHost` nested inside a Fragment would lose to its host
fragment instead.

This is documented in the class KDoc, and it relates to the staleness question still open on #1909.
Happy to reorder if maintainers prefer the other tradeoff.

## Nested-controller restore semantics

The destination slot **overwrites rather than stacks**. If a parent controller records `"a"` and a
nested controller then records `"b"`, `"a"` is already gone; disposing the child clears to the
fallback chain rather than restoring `"a"`.

I've left nested-controller restore semantics out of this PR since it needs owner/name pairs rather
than a single slot — a single displaced-value slot is not enough, because it would resurrect a stale
route from the *same* controller after an ordinary `a → b → c` navigation. That felt related to the
staleness question still open on the issue, but happy to fold it in if you'd rather.

## Tests

`:services` (JUnit 5 + MockK) — all 7 pre-existing `VisibleScreenTrackerTest` cases pass unchanged;
7 added:

- destination beats fragment and activity
- cleared falls back to fragment / to activity / to `"unknown"`
- a clear from a superseded source is ignored (the nested/sibling case above)
- the attach replay on a configuration change is idempotent
- a destination does not affect `previouslyVisibleScreen`

`:instrumentation:compose:navigation` (JUnit 4 + Robolectric) — 7 added across two new files:

- reports the resolved name on attach and on each navigation
- reports the route *pattern*, asserting `user/{id}` and never `user/2`, pinning the PII-safe default
- clears when the controller leaves the composition
- a dispose/re-attach cycle clears and then replays, in that order
- reporter unit tests: clears using only the last name reported, no-op when nothing was reported,
  idempotent on double-clear

`NavControllerExtensionsTest` was updated for the new `attachOpenTelemetry` parameter; that internal
function exists as a test seam, so its signature is pinned by that file.

## Verification

`./gradlew spotlessApply` and `./gradlew apiDump` both clean, with the updated `.api` file included
in the diff.

`./gradlew :services:check :instrumentation:compose:navigation:check :core:check` — BUILD SUCCESSFUL.
services 89 tests, compose-navigation 23 tests, core 118 tests; 0 failures. The single skipped core
test is a pre-existing `@Ignore`, and `core` is untouched by this PR.

See #1909

### DIFF
--- a/services/src/main/java/io/opentelemetry/android/internal/services/visiblescreen/VisibleScreenTracker.kt
+++ b/services/src/main/java/io/opentelemetry/android/internal/services/visiblescreen/VisibleScreenTracker.kt
@@ -13,10 +13,15 @@ import io.opentelemetry.android.internal.services.Service
  * Wherein we do our best to figure out what "screen" is visible and what was the previously visible
  * "screen".
  *
- * In general, we favor using the last fragment that was resumed, but fall back to the last
- * resumed activity in case we don't have a fragment.
+ * In general, we favor the most recently reported navigation destination, then the last fragment
+ * that was resumed, and fall back to the last resumed activity in case we have neither. While a
+ * destination is recorded it outranks both, including a DialogFragment shown over it.
  *
- * We always ignore NavHostFragment instances since they aren't ever visible to the user.
+ * Navigation destinations contribute only to the currently visible screen. The previously visible
+ * screen is still derived from fragments and activities alone.
+ *
+ * We always ignore NavHostFragment instances since they aren't ever visible to the user. That
+ * concerns the host fragment itself, not the destinations a navigation library reports to us.
  *
  * We have to treat DialogFragments slightly differently since they don't replace the launching
  * screen, and the launching screen never leaves visibility.
@@ -32,4 +37,20 @@ interface VisibleScreenTracker : Service {
     fun fragmentResumed(fragment: Fragment)
 
     fun fragmentPaused(fragment: Fragment)
+
+    /**
+     * Records the screen name of a navigation destination reached by a source that is neither an
+     * Activity nor a Fragment, such as a Compose Navigation NavController. The recorded name takes
+     * precedence over the last resumed fragment and activity until it is cleared.
+     */
+    fun navigationDestinationChanged(destinationName: String)
+
+    /**
+     * Clears the recorded navigation destination, so that the visible screen falls back to the last
+     * resumed fragment, then the last resumed activity.
+     *
+     * The clear only applies if [destinationName] is still the recorded destination. A source that
+     * has since been superseded by another one therefore cannot discard the newer destination.
+     */
+    fun navigationDestinationCleared(destinationName: String)
 }
--- a/services/src/main/java/io/opentelemetry/android/internal/services/visiblescreen/VisibleScreenTrackerImpl.kt
+++ b/services/src/main/java/io/opentelemetry/android/internal/services/visiblescreen/VisibleScreenTrackerImpl.kt
@@ -22,10 +22,15 @@ import java.util.concurrent.atomic.AtomicReference
  * Wherein we do our best to figure out what "screen" is visible and what was the previously visible
  * "screen".
  *
- * In general, we favor using the last fragment that was resumed, but fall back to the last
- * resumed activity in case we don't have a fragment.
+ * In general, we favor the most recently reported navigation destination, then the last fragment
+ * that was resumed, and fall back to the last resumed activity in case we have neither. While a
+ * destination is recorded it outranks both, including a DialogFragment shown over it.
  *
- * We always ignore NavHostFragment instances since they aren't ever visible to the user.
+ * Navigation destinations contribute only to the currently visible screen. The previously visible
+ * screen is still derived from fragments and activities alone.
+ *
+ * We always ignore NavHostFragment instances since they aren't ever visible to the user. That
+ * concerns the host fragment itself, not the destinations a navigation library reports to us.
  *
  * We have to treat DialogFragments slightly differently since they don't replace the launching
  * screen, and the launching screen never leaves visibility.
@@ -37,6 +42,7 @@ internal class VisibleScreenTrackerImpl internal constructor(
     private val previouslyLastResumedActivity = AtomicReference<String>()
     private val lastResumedFragment = AtomicReference<String>()
     private val previouslyLastResumedFragment = AtomicReference<String?>()
+    private val currentNavigationDestination = AtomicReference<String?>()
     private val activityLifecycleTracker by lazy { buildActivitiesTracker() }
     private val fragmentLifecycleTrackerRegisterer by lazy { buildFragmentsTrackerRegisterer() }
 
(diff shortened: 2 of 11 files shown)


services/src/main/java/io/opentelemetry/android/internal/services/visiblescreen/VisibleScreenTrackerImpl.kt
```diff
@@ -76,6 +82,10 @@ internal class VisibleScreenTrackerImpl internal constructor(
 
     override val currentlyVisibleScreen: String
         get() {
+            val destination = currentNavigationDestination.get()
```
aranhave, line 85: What clears this destination when its host Activity pauses and another Activity comes to the foreground? The first Activity composition can remain alive, so `onDispose` may not run; because the destination always wins here, spans and logs from Activity B can still get Activity A route. Could we scope it to the host Activity lifecycle and add an A route -> B Activity test?

services/src/main/java/io/opentelemetry/android/internal/services/visiblescreen/VisibleScreenTracker.kt
```diff
@@ -32,4 +37,20 @@ interface VisibleScreenTracker : Service {
...
+    fun navigationDestinationChanged(destinationName: String)
+
+    /**
+     * Clears the recorded navigation destination, so that the visible screen falls back to the last
+     * resumed fragment, then the last resumed activity.
+     *
+     * The clear only applies if [destinationName] is still the recorded destination. A source that
+     * has since been superseded by another one therefore cannot discard the newer destination.
+     */
+    fun navigationDestinationCleared(destinationName: String)
```
aranhave, line 55: Could this clear identify the reporting controller rather than only its route text? Two controllers can both report `home`; if the newer one is active and the older one disposes, `clear("home")` can remove the newer controller value. An opaque source token or registration handle would preserve the conditional-clear guarantee.

services/src/main/java/io/opentelemetry/android/internal/services/visiblescreen/VisibleScreenTrackerImpl.kt
hitkall: Confirmed, and thanks — this is a real hole. When Activity A is stopped rather than
destroyed its composition stays alive, so onDispose never runs and the slot keeps A's
route. activityResumed(B) updates lastResumedActivity, but the destination outranks it
in currentlyVisibleScreen, so every span and log on Activity B is attributed to A's
route until the user navigates back.

I'll scope it to the host Activity. navigationDestinationChanged can capture the
resumed Activity at report time, and currentlyVisibleScreen returns the destination
only while that host is still the resumed Activity. That has a nice side effect: the
route is restored when the user returns to A, without the composition needing to
re-report.

One pre-existing limitation worth naming rather than hiding: the tracker identifies
Activities by javaClass.simpleName, so two instances of the same Activity class are
indistinguishable. I don't propose to change that here.

Adding the A route -> B Activity test.

services/src/main/java/io/opentelemetry/android/internal/services/visiblescreen/VisibleScreenTracker.kt
hitkall: Good catch — name equality isn't identity, so this defeats the point of the
conditional clear. Two controllers reporting "home" means the older one's
clear("home") satisfies the compareAndSet and removes the newer controller's value.

Switching to an opaque owner token. The compose module already creates one
NavigationDestinationReporter per attached controller, so it can pass itself:
navigationDestinationChanged(owner, name) records the pair, and
navigationDestinationCleared(owner) clears only when the recorded owner is
identical. Route text stops participating in the clear decision entirely.

This composes with your other comment — the slot ends up holding (owner, host
activity, destination name), where the owner governs clears and the host governs
reads. Adding a test for two controllers reporting the same route.

hitkall:
@breedx-splk please review

fractalwrench:
@hitkall there are quite a few open comments from Copilot and another project member - if you can address these first then one of the project maintainers/approvers will take a look.

breedx-splk:
@hitkall are you able to revisit this one? There are conflicts now and still some unresolved comments. Thanks!

hitkall:
@breedx-splk rebased on main and addressed all open comments, ready for another look

PR #1995, opened 2026-08-20
fix lychee link check failure

breedx-splk:
It [looks like](https://github.com/open-telemetry/opentelemetry-android/actions/runs/32415870458/job/96576756417#step:3:303) another site is blocking github, so let's omit it from the link check.

PR #2013, opened 2026-08-25
Fix lychee toml escapes

breedx-splk:
Bah, I [broke the build](https://github.com/open-telemetry/opentelemetry-android/actions/runs/32884381003/job/97921251494) because I didn't realize the escaping needed extra backslashes or whatever. Let's just remove them, even if it's slightly incorrect.

I really should have listened to @benjoseph-grafana over [here](https://github.com/open-telemetry/opentelemetry-android/pull/1995#discussion_r3826378790). Sorry sorry.


LikeTheSalad (approved):
Thanks!

PR #2014, opened 2026-08-25
Obsolete bespoke network.status in favor of network.connection.type

breedx-splk:
`network.connection.type` comes from [upstream](https://github.com/open-telemetry/semantic-conventions/blob/3c3ffc3b01cdda4cabbc82e9584696ec1e63306a/model/network/registry.yaml#L118) and we should use it instead of bespoke/custom conventions.

This is a breaking change, so we use the `SemconvCompat` opt-in to keep the attribute for users who still need it.

The semantics of "available" or "lost" were previously specified in the `network.status` attribute, which is redundant with the other values of `network.connection.type`. That is, a type of `unavailable` means `lost`, whereas every other value means "available".

So this makes things simpler while also aligning better with upstream.

### DIFF
--- a/instrumentation/network/src/main/java/io/opentelemetry/android/instrumentation/network/NetworkChangeAttributesExtractor.kt
+++ b/instrumentation/network/src/main/java/io/opentelemetry/android/instrumentation/network/NetworkChangeAttributesExtractor.kt
@@ -9,6 +9,7 @@ import io.opentelemetry.android.common.internal.features.networkattributes.Curre
 import io.opentelemetry.android.common.internal.features.networkattributes.data.CurrentNetwork
 import io.opentelemetry.android.common.internal.features.networkattributes.data.NetworkState
 import io.opentelemetry.android.semconv.NetworkAttributes.NETWORK_STATUS_KEY
+import io.opentelemetry.android.semconv.internal.SemconvCompat
 import io.opentelemetry.api.common.AttributesBuilder
 import io.opentelemetry.kotlin.semconv.IncubatingApi
 import io.opentelemetry.kotlin.semconv.NetworkAttributes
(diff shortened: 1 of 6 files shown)


instrumentation/network/src/main/java/io/opentelemetry/android/instrumentation/network/NetworkChangeAttributesExtractor.kt
```diff
@@ -17,17 +18,20 @@ internal class NetworkChangeAttributesExtractor : NetworkAttributesExtractor {
...
         currentNetwork: CurrentNetwork,
     ) {
-        val status =
-            if (currentNetwork.state == NetworkState.NO_NETWORK_AVAILABLE) {
-                "lost"
-            } else {
-                "available"
-            }
-        attributesBuilder.put(NETWORK_STATUS_KEY, status)
+        if (!SemconvCompat.useLatestExperimental) {
```
aranhave, line 26: Would you update the network README and add an Unreleased migration note with this change? The README still says `network.status` is emitted on every event, but after this it is only present when `useLatestExperimental = false`.

instrumentation/network/src/test/java/io/opentelemetry/android/instrumentation/network/NetworkChangeInstrumentationTest.kt
```diff
@@ -190,6 +189,29 @@ class NetworkChangeInstrumentationTest {
...
+            .hasEventName(NETWORK_CHANGE_EVENT_NAME)
+            .hasAttributesSatisfyingExactly(
+                equalTo(stringKey(NETWORK_CONNECTION_TYPE), "unavailable"),
+            )
+    }
+
+    @Test
+    @Suppress("DEPRECATION")
+    fun legacySemconv_networkLost() {
+        SemconvCompat.useLatestExperimental = false
```
aranhave, line 201: Can this compatibility test cover an available connection as well? `available` is the other new branch, and right now only the `lost` value is pinned.

semconv/model/android/events.yaml
```diff
@@ -190,7 +190,7 @@ groups:
     brief: >
       This event indicates that Android network connectivity changed.
     attributes:
-      - ref: network.status
+      - ref: network.connection.type
```
aranhave, line 195: Could we keep the deprecated `network.status` parameter in the generated event during the compatibility window? This changes the published `NetworkChangeEvent(String?)` constructor in place: old binaries still link, but their `available`/`lost` argument is now interpreted as `network.connection.type`, while source using `networkStatus =` no longer compiles.

aranhave (commented):
The direction looks good but the generated event compatibility gap needs to be addressed before merging. The docs and remaining legacy test should be quick to cover too.

semconv/model/android/events.yaml
breedx-splk: yeah good idea I suppose. thanks.

LikeTheSalad (approved):
Thanks!

fractalwrench (approved):
LGTM

aranhave (approved):
Thanks for addressing all comments! LGTM!

PR #2016, opened 2026-08-26
fix: flush telemetry before delegating on crash

### LINKED ISSUE #1997: CrashFlushHandler does not flush telemetry before process termination
[`CrashFlushHandler` calls](https://github.com/open-telemetry/opentelemetry-android/blob/1aeced3e363347dee229c4b4780f03bb1a040cda/core/src/main/java/io/opentelemetry/android/CrashFlushHandler.kt#L25) the previous handler before attempting to flush. This results in `CrashFlushHandler` never awaiting the telemetry flush because the handler chain goes all the way to the default handler, which terminates the Android process.

We should flush before delegating to the previous handler.

benjoseph-grafana:
CrashFlushHandler invoked the previous uncaught exception handler
before force-flushing. The handler chain reaches the platform default,
which terminates the process, so the flush never ran.

Flush first, then delegate. Install the handler before instrumentations
so crash reporting wraps it and crash events are included in the flush.

Resolves https://github.com/open-telemetry/opentelemetry-android/issues/1997

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@
 
 ### 🛠️ Bug fixes
 
+- Flush telemetry on crash before delegating to the previous uncaught exception handler, so the
+  process is not terminated before the flush completes.
+  ([#1997](https://github.com/open-telemetry/opentelemetry-android/issues/1997))
+
 - Fix R8 failures caused by optional Error Prone annotations in minified applications.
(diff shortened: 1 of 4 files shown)


fractalwrench (approved):
LGTM - thanks!

aranhave (approved):

LikeTheSalad (approved):
Thank you!

PR #2021, opened 2026-08-27
Leverage fake backend on host for smoke tests.

breedx-splk:
Follows up on https://github.com/breedx-splk/opentelemetry-android/pull/7 and #1972. We no longer run the test code and telemetry service on the device, instead we can run the upstream smoke test fake backend on the host and the test. The improves the architecture of the smoke tests and allows us to reuse existing otel test components.


fractalwrench (commented):
One thing to note with this approach is that it doesn't assert that it's possible to run a minified app on an Android device. That might be the way we want to go for the majority of tests as it's faster & simpler, but IMO we should retain at least one test case that runs on a device as a backstop for discovering R8 + other unanticipated issues.

smoke-test/build.gradle.kts
```diff
@@ -1,58 +1,62 @@
...
+    group = "verification"
+    enabled = smokeTestRequested
+    useJUnitPlatform()
+    testClassesDirs =
+        sourceSets.test
+            .get()
+            .output.classesDirs
+    classpath = sourceSets.test.get().runtimeClasspath
+    dependsOn(":smoke-test-app:assembleRelease")
+    systemProperty("smoke-test.apk", smokeTestAppApk.get().asFile.absolutePath)
```
aranhave, line 61: Could we add the APK itself as an input here or disable up-to-date/build-cache reuse for this task? `dependsOn` only orders the APK build, while the system property fingerprints the unchanged path rather than the APK contents. I was able to rebuild a different APK and have `smokeTest` remain `UP-TO-DATE`, so SDK/app changes can skip testing the new artifact.

breedx-splk: Wow great catch. Yes.

aranhave (approved):
LGTM 👍

breedx-splk:
> One thing to note with this approach is that it doesn't assert that it's possible to run a minified app on an Android device. That might be the way we want to go for the majority of tests as it's faster & simpler, but IMO we should retain at least one test case that runs on a device as a backstop for discovering R8 + other unanticipated issues.

Wait, why is that? The intent is to definitely run the minified app on a(n emulated) device.

```mermaid
flowchart TB
      subgraph HOST["HOST / CI RUNNER"]
          CI["GitHub Actions"]
          GRADLE["Gradle<br/>connectedCheck + smokeTest"]
          BUILD["assembleRelease"]
          APK["Release APK<br/>R8 + resource shrinking"]
          TEST["MinifiedAppSmokeTest<br/>JUnit 5 on host"]
          ADB["adb"]
          RETRIEVER["TelemetryRetriever"]
          BACKEND["Testcontainers<br/>fake backend"]

          CI --> GRADLE
          GRADLE --> BUILD
          BUILD --> APK
          GRADLE --> TEST
          TEST --> ADB
          TEST --> RETRIEVER
          TEST --> BACKEND
          RETRIEVER -->|"GET traces"| BACKEND
          BACKEND -->|"trace data"| RETRIEVER
      end

      subgraph DEVICE["ANDROID EMULATOR"]
          APP["smoke-test-app process"]
          ACTIVITY["SmokeTestActivity"]
          OTEL["OpenTelemetry Android<br/>agent + SDK"]

          APP --> ACTIVITY
          ACTIVITY --> OTEL
      end

      APK -->|"APK path"| TEST
      ADB -->|"install + launch"| APP
      OTEL -->|"OTLP HTTP<br/>10.0.2.2:port"| BACKEND
```

Have I got something wrong?

fractalwrench:
>We no longer run the test code and telemetry service on the device,

@breedx-splk my bad - I read that and assumed that you'd shifted the test code over to run on the JVM rather than an android device.

fractalwrench (approved):
LGTM

LikeTheSalad (approved):
Thanks!

PR #2023, opened 2026-08-28
Consolidate network code in the network instrumentation module

fractalwrench:
This changeset consolidates network-related code that was previously in `services` and `core` and moves it to `services-network` instead.

The majority of this diff is removal of imports/changing of package names for consistency, along with the movement of API diff dumps.


core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.kt
```diff
@@ -23,15 +23,15 @@ import io.opentelemetry.android.features.diskbuffering.scheduler.ScheduleEnablem
 import io.opentelemetry.android.instrumentation.AndroidInstrumentation
 import io.opentelemetry.android.instrumentation.AndroidInstrumentationLoader
 import io.opentelemetry.android.instrumentation.AndroidInstrumentationLoaderImpl
+import io.opentelemetry.android.instrumentation.network.NetworkAttributesLogRecordAppender
```
aranhave: Does this also make direct `core` users auto-install `NetworkChangeInstrumentation`? The runtime dependency brings its `@AutoService` entry, and the builder installs every discovered instrumentation. Should the shared appender code live below `core` instead?

breedx-splk:
Do we really want core to depend on any instrumentations?

fractalwrench:
That's a fair point. One option would be to create a `network-common` module?

breedx-splk:
> That's a fair point. One option would be to create a `network-common` module?

Yeah we like this idea, maybe `services/network` as a module.

fractalwrench:
Addressed review comments and moved code to a new `services-network` module instead.

services-network/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
```
aranhave, line 3: Now that the network code and permission live in this module, should `ACCESS_NETWORK_STATE` come out of `services/src/main/AndroidManifest.xml`? Otherwise the non-network instrumentations that depend on `services` still add it.

aranhave (approved):
Looks good! 1 minor thing added. The description still describes the old core -> instrumentation.network approach. Mind updating it for services-network?

LikeTheSalad (approved):
Thank you!

services-network/src/main/java/io/opentelemetry/android/internal/services/network/NetworkAttributesSpanAppender.kt
```diff
@@ -3,17 +3,15 @@
...
 import io.opentelemetry.context.Context
 import io.opentelemetry.sdk.trace.ReadWriteSpan
 import io.opentelemetry.sdk.trace.ReadableSpan
 import io.opentelemetry.sdk.trace.SpanProcessor
 
 /**
- * A [io.opentelemetry.sdk.trace.SpanProcessor] implementation that appends a set of [attributes][io.opentelemetry.api.common.Attributes]
+ * A [SpanProcessor] implementation that appends a set of [attributes][io.opentelemetry.api.common.Attributes]
  * describing the
  * [current network][io.opentelemetry.android.common.internal.features.networkattributes.data.CurrentNetwork]
```
benjoseph-grafana: update the package here for CurrentNetwork

benjoseph-grafana (approved):

PR #2024, opened 2026-08-28
Move thread utils

fractalwrench:
Moves `threadIdCompat` to the more specific `instrumentation/common-api` module which is a better fit than `core` given that the symbol is only used from instrumentation.


aranhave (approved):
LGTM 👍 I’m okay with the internal API move here since common is not stable yet.

LikeTheSalad (approved):
Thanks!

Issue #2020, opened 2026-08-27
Auto generate instrumentations semconv README entries


LikeTheSalad:
This is a follow-up to this issue's comment: https://github.com/open-telemetry/opentelemetry-android/issues/1814#issuecomment-5417705210

We currently have to manually specify in each instrumentation's README what are the spans/events generated by it, along with their attributes and values in general. Ideally, this work should be done automatically as part of a CI process.

fractalwrench:
Would be curious what approach you plan on taking for this (and also would love to see this happen as it'd be really useful). My first idea would be to use ASM in the gradle script to identify any calls to `Tracer/Logger`, then walk the relevant bytecode for uses of semantic conventions (or string literals)

LikeTheSalad:
> Would be curious what approach you plan on taking for this (and also would love to see this happen as it'd be really useful). My first idea would be to use ASM in the gradle script to identify any calls to `Tracer/Logger`, then walk the relevant bytecode for uses of semantic conventions (or string literals)

That sounds good, but it would miss stuff created outside of this repo, as is the case for the OkHttp instrumentation, where a lot of the telemetry is generated upstream. My current plan is to rely on the tests from each instrumentation and capture the telemetry in-memory from them, and then filter it by the instrumentation scope and write it into the yaml file. This depends on our instrumentation tests being thorough, which I think they should be regardless anyway.

PR #2028, opened 2026-08-31
[proposal] Generating instrumentation telemetry doc file when running tests

LikeTheSalad:
Related to https://github.com/open-telemetry/opentelemetry-android/issues/2020

## Summary
Adds the plumbing for generating per-instrumentation `telemetry.yaml` files from telemetry emitted by tests and enriched with resolved semantic convention metadata.

The intended usage is:
- Apply `otel.android.telemetry-docs` to every Gradle project containing tests that should contribute telemetry documentation.
- Configure the instrumentation scope names through `telemetryDocs { scopeNames(...) }`.
- Use `RobolectricOpenTelemetryRumRule` in contributing Robolectric tests and `OpenTelemetryRumRule` in connected Android tests.
- Run the contributing test tasks with `-PcollectTelemetryDocs=true` so they write raw observation files under their project build outputs.
- Run `./gradlew mergeAllTelemetryDocs` to merge all submodules' captured observations.
- Review the generated file at `instrumentation/<instrumentation-name>/telemetry.yaml`.

## Schema example

```yaml
schema_version: 1
module: "view-click"
scopes:
  - "io.opentelemetry.android.instrumentation.view.click"
signals:
  - type: event
    name: "app.screen.click"
    scope: "io.opentelemetry.android.instrumentation.view.click"
    registry_id: "event.app.screen.click"
    attributes:
      - name: "app.screen.coordinate.x"
        type: int
        registry: upstream
```
Take a look at [TELEMETRY_SCHEMA.md](https://github.com/open-telemetry/opentelemetry-android/pull/2028/changes#diff-135a7fae7cbdf397bf33b53a77b1edac47f16c4e1b8f58348e3d154441e6dfab) for more details.

## Changes
- Adds shared capture rules for Robolectric and connected Android tests, recording spans and log events with their scopes, kinds, attributes, types, and values.
- Adds Gradle plugins that publish captured observations, discover their logical instrumentation directory, resolve the semantic convention registry, and merge observations into a yaml file.
- Adds exact Weaver registry lookup for events and ambiguity-safe registry inference for spans. Unresolved spans are marked `unidentified` with diagnostics.
- Defines the initial telemetry schema for spans and log events. Metrics remain deferred.
- Adds merger, plugin, capture, schema, and configuration-cache test coverage.


buildSrc/src/main/kotlin/io/opentelemetry/android/instrumentation/TelemetryDocsMerger.kt
```diff
@@ -0,0 +1,381 @@
...
+                            signals,
+                            registryGroup,
+                            registry.attributeProvenance,
+                        )
+                    validateCoverage(moduleName, key, attributes, registryGroup, localGroupIds)
+                    MergedSignal(
+                        key = key,
+                        attributes = attributes,
+                    )
+                }.sortedWith(compareBy({ it.key.type }, { it.key.registryId }, { it.key.scope }))
```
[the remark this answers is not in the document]
LikeTheSalad: Fixed in 0675f4b4. The final comparator now includes every signal identity field, and the merger test covers two unregistered same-scope events encountered in reverse lexical order.

docs/TELEMETRY_SCHEMA.md
```diff
@@ -0,0 +1,47 @@
...
+A span-only instrumentation has a separate file:
+
+```yaml
+schema_version: 1
+module: "okhttp3"
+scopes:
+  - "io.opentelemetry.okhttp-3.0"
+signals:
+  - type: span
+    scope: "io.opentelemetry.okhttp-3.0"
```
jaydeluca, line 34: awesome PR!

in the java instrumentation we also use a `when` condition to indicate if something is emitted by default, or if it requires some a configuration option. [We add a flag to our test suites ](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/888f78b4b3eaf776bc0e3310e1602d85cacfded6/instrumentation/grpc-1.6/javaagent/build.gradle.kts#L36) that we then use to populate it.

Not sure if thats applicable to any of these, but figured i'd mention it in case its helpful

LikeTheSalad: Thanks for your feedback, @jaydeluca 🙏 - I'm not aware right now of a similar need here, but if there is one in the future, I'll keep this approach in mind to add support for it.

fractalwrench (commented):
This is awesome - I like the general approach. My (non-blocking) questions would be where we plan to use this in the development lifecycle, i.e. is it a manual check run locally, or a CI check to prevent telemetry changing?

An additional point I can think of is the ability to edit/merge the generated telemetry. For example, a HTTP request will have different attributes depending on whether it succeeded or failed, so if a test only covers one scenario then it will miss certain attributes.

jaydeluca:
> My (non-blocking) questions would be where we plan to use this in the development lifecycle, i.e. is it a manual check run locally, or a CI check to prevent telemetry changing?

Some other info about how we have done this in the java instrumentation repo in case it's useful inspiration for you, we run a nightly job that runs all the tests, collects the data, aggregates it into a [file](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/instrumentation-list.yaml), and then opens a PR if there have been changes ([example](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/19942)).

In our case it takes a long time to run, so in it's current form it's not as convenient to incorporate into single PRs, but I have intentions of reworking it so that things can run a bit more distributed and be faster and it's easier to attribute changes to the exact pr that caused the changes. Makes reviewing the output of instrumentation changes a bit easier.

As another note, once this is setup, we would be excited to incorporate this into the [ecosystem explorer](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer) like we do with the [java instrumentation](https://explorer.opentelemetry.io/java-agent/instrumentation). We would basically start scraping your generated data and then we can build UI elements for it, as well as post this information in an AI agent friendly markdown format so it's discoverable and gives agent a better source of truth.

Another thing we did in the the instrumentation project was to go through every library and add some additional manual metadata for components to include things like categorizing the type of instrumentation, configuration options etc. Might be outside the scope of this PR, but would be really great to have. We have some docs about it [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/contributing/documenting-instrumentation.md#metadatayaml-files)

Let me know if any of this interests you and I'd be happy to help in any way i can.

LikeTheSalad:
Thank you for your feedback, @fractalwrench @jaydeluca 

> where we plan to use this in the development lifecycle, i.e. is it a manual check run locally, or a CI check to prevent telemetry changing?

I haven't decided yet, but I was pondering 2 options, either nightly as @jaydeluca mentions, or as part of the "prepare release" step during a release. Since these tests don't take as long as in upstream (that project is a beast compared to this one), it probably would be fine to run them by then so that the automated PR to prepare the new release contains the changes, if any. The benefit I see with this approach is that the `telemetry.yaml` file will match the latest OTel release, even though we might have added attribute changes in `main` afterwards. I'd like to discuss it in the SIG meeting to get a better idea of how people feel about those options, and/or if there could be other ones we haven't considered.

> For example, a HTTP request will have different attributes depending on whether it succeeded or failed, so if a test only covers one scenario then it will miss certain attributes.

Good point. This sounds related to the [`when` condition](https://github.com/open-telemetry/opentelemetry-android/pull/2028#discussion_r3895479675) that @jaydeluca mentioned earlier, so I'm up with using that approach here too if it works for this project, but I haven't checked it yet because I didn't think of an existing use case for it now in here, but if this is one of those, then we should make sure to have a feature like this in the first version of this tool.

> As another note, once this is setup, we would be excited to incorporate this into the [ecosystem explorer](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer) like we do with the [java instrumentation](https://explorer.opentelemetry.io/java-agent/instrumentation). We would basically start scraping your generated data and then we can build UI elements for it, as well as post this information in an AI agent friendly markdown format so it's discoverable and gives agent a better source of truth.

This is great! @breedx-splk showed this to us in the last SIG meeting, and it does make things so much easier for many use cases. I'm looking forward to it.

PR #2029, opened 2026-08-31
feat(native-crash): parse stack snapshots

aranhave:
This PR adds the internal parser and data model for the version 1 native crash snapshot defined in #1999.

- validates the exact size, magic, version, checksum, architecture, and marker pairing
- enforces ABI address widths, stack bounds, alignment, reserved fields, module ranges, names, and build IDs
- supports ARM, ARM64, x86, and x86_64
- skips malformed module entries while preserving valid entries
- includes boundary, corruption, and mutation coverage

This PR is intentionally limited to parsing and validation. It does not change the signal handler or emit native frames yet.

## Follow-up work

The remaining work for #1940 will be split into focused follow-ups:

1. Recover module-relative frames and build IDs from validated snapshots.
2. Integrate snapshots with restart-time crash replay, bounded retries, and cleanup.
3. Add async-signal-safe register, module, and bounded-stack capture to the native handler.
4. Run end-to-end device validation across the supported Android ABIs and malformed or partial data cases.

Symbolication and backend rendering remain separate work.

Merge tier: Tier 3

### DIFF
--- a/instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashSnapshot.kt
+++ b/instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashSnapshot.kt
@@ -1,0 +1,261 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.android.instrumentation.nativecrash
+
+import java.nio.ByteBuffer
+import java.nio.ByteOrder
+import java.nio.charset.CodingErrorAction
+
+// Mirrors native_crash_snapshot.h, which remains the binary layout source of truth.
+internal object NativeCrashSnapshotLayout {
+    const val MAGIC_SIZE = 8
+    const val VERSION_OFFSET = 8
+    const val ARCHITECTURE_OFFSET = 12
+    const val RECORD_SIZE_OFFSET = 16
+    const val SIGNAL_NUMBER_OFFSET = 20
+    const val TIMESTAMP_OFFSET = 24
+    const val PROGRAM_COUNTER_OFFSET = 32
+    const val STACK_POINTER_OFFSET = 40
+    const val FRAME_POINTER_OFFSET = 48
+    const val LINK_REGISTER_OFFSET = 56
+    const val MODULE_COUNT_OFFSET = 64
+    const val STACK_SIZE_OFFSET = 68
+    const val STACK_START_OFFSET = 72
+    const val MODULES_OFFSET = 80
+
+    const val MAX_MODULES = 128
+    const val MODULE_ENTRY_SIZE = 128
+    const val MODULE_LOAD_BIAS_OFFSET = 0
+    const val MODULE_EXECUTABLE_START_OFFSET = 8
+    const val MODULE_EXECUTABLE_END_OFFSET = 16
+    const val MODULE_NAME_OFFSET = 24
+    const val MODULE_NAME_SIZE = 64
+    const val MODULE_BUILD_ID_SIZE_OFFSET = 88
+    const val MODULE_BUILD_ID_OFFSET = 92
+    const val MODULE_BUILD_ID_CAPACITY = 32
+    const val MODULE_RESERVED_OFFSET = 124
+
+    const val STACK_OFFSET = MODULES_OFFSET + MAX_MODULES * MODULE_ENTRY_SIZE
+    const val STACK_CAPACITY = 4_096
+    const val RESERVED_OFFSET = STACK_OFFSET + STACK_CAPACITY
+    const val CHECKSUM_OFFSET = RESERVED_OFFSET + Int.SIZE_BYTES
+    const val RECORD_SIZE = CHECKSUM_OFFSET + Int.SIZE_BYTES
+}
+
+internal enum class NativeCrashArchitecture(
+    val id: Int,
+    val pointerSize: Int,
+    val hasLinkRegister: Boolean,
+) {
+    ARM(1, Int.SIZE_BYTES, true),
+    ARM64(2, Long.SIZE_BYTES, true),
+    X86(3, Int.SIZE_BYTES, false),
+    X86_64(4, Long.SIZE_BYTES, false),
+    ;
+
+    companion object {
+        fun fromId(id: Int): NativeCrashArchitecture? = entries.firstOrNull { it.id == id }
+    }
+}
+
+internal class NativeCrashSnapshot(
+    val architecture: NativeCrashArchitecture,
+    val programCounter: ULong,
+    val stackPointer: ULong,
+    val framePointer: ULong,
+    val linkRegister: ULong,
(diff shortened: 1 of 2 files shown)


instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashSnapshot.kt
```diff
@@ -0,0 +1,261 @@
...
+            framePointer = framePointer,
+            linkRegister = linkRegister,
+            modules = modules,
+            stackStart = stackStart,
+            stack = stack,
+        )
+    }
+
+    internal fun checksum(bytes: ByteArray): UInt {
+        require(bytes.size >= NativeCrashSnapshotLayout.CHECKSUM_OFFSET)
```
fractalwrench, line 144: Will this be caught elsewhere? IMO it'd be good to document any throws as it can be easy to forget about them and propagate to the library consumer..

aranhave: Thanks for flagging this. The parser already checks the full record size before calling this helper. I’ve documented the exception and added a test for truncated input.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashSnapshot.kt
```diff
@@ -0,0 +1,266 @@
...
+        val linkRegister = buffer.linkRegister(architecture) ?: return null
+        val stackStart = buffer.addressAt(NativeCrashSnapshotLayout.STACK_START_OFFSET, architecture) ?: return null
+        val moduleCount = buffer.getInt(NativeCrashSnapshotLayout.MODULE_COUNT_OFFSET)
+        val stackSize = buffer.getInt(NativeCrashSnapshotLayout.STACK_SIZE_OFFSET)
+
+        if (stackPointer == 0UL || stackStart != stackPointer) return null
+        if (stackStart % architecture.pointerSize.toULong() != 0UL) return null
+        if (moduleCount !in 1..NativeCrashSnapshotLayout.MAX_MODULES) return null
+        if (stackSize !in 0..NativeCrashSnapshotLayout.STACK_CAPACITY) return null
+        if (buffer.getInt(NativeCrashSnapshotLayout.RESERVED_OFFSET) != 0) return null
```
breedx-splk: This block is pure validation. I think that factoring it out to a method would improve readability. There's also a lot of visual duplication because of the `NativeCrashSnapshotLayout` prefix, so maybe `with(NativeCrashSnapshotLayout) { .... }` could be used to help there?

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashSnapshot.kt
breedx-splk: The test is good, and it confirms that the exception is thrown.

I think what @fractalwrench means is that if/when this is encountered in the wild on a device, will the app continue to start? Does the otel agent throw this exception and crash the app? We can't have that! It's mandatory that we consider the context that we're running in and fail safe.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashSnapshot.kt
```diff
@@ -0,0 +1,266 @@
...
+    X86(3, Int.SIZE_BYTES, false),
+    X86_64(4, Long.SIZE_BYTES, false),
+    ;
+
+    companion object {
+        fun fromId(id: Int): NativeCrashArchitecture? = entries.firstOrNull { it.id == id }
+    }
+}
+
+internal class NativeCrashSnapshot(
```
breedx-splk, line 64: Can this also be made a data class? I'm wondering if that cleans up the test a bit, because you could construct the full snapshot and do equals, instead of having to compare each field?

breedx-splk:
Thanks for putting this together @aranhave ! Had a few comments, but mostly looking good. Appreciate it.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashSnapshot.kt
aranhave: Done. I moved the checks into a helper and used `with(NativeCrashSnapshotLayout)` there.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashSnapshot.kt
aranhave: I kept this as a regular class because `ByteArray` uses reference equality in a data class, so comparing the whole snapshot would be misleading.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashSnapshot.kt
aranhave: `parse()` checks the full record size before calling `checksum()` and returns `null` for malformed snapshots. I made that clearer and added a no-throw test for truncated input.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashSnapshot.kt
breedx-splk: That's a very good reason indeed! 😅  Thanks for the clarification.

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashSnapshot.kt
breedx-splk: Maybe we're a little confused because it's basically checking the same thing in two places. I guess what I'm saying is that if `parse()` is the only caller of `checksum()`, then the `require()` is unnecessary here. If you want the `checksum()` method to be safer, I suppose it could also do a check and return a nullable type (for the caller to handle) or even possibly just some hardcoded/invalid checksum (I know, I know 😁 )

That being said, I think it's probably fine the way it is.

breedx-splk (approved):

instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashSnapshot.kt
aranhave: Got it, thanks for clarifying 😄 I’ll leave it as-is.

fractalwrench (approved):
LGTM - thanks!

PR #2032, opened 2026-09-01
feat(native-crash): unwind stack snapshots

aranhave:
Builds on #2029 to recover module-relative native frames and build IDs from validated snapshots.

It handles PC and frame-pointer recovery, ARM/ARM64 link-register fallback, address normalization, and malformed or truncated chains with a 64-frame limit. Crash replay integration will follow separately.

Part of #1940.

Merge tier: Tier 3


LikeTheSalad (approved):
Thank you!

fractalwrench (approved):
LGTM

aranhave:
@fractalwrench @LikeTheSalad Thanks for reviewing! could you help merge this when you get a chance. Thanks!

PR #2036, opened 2026-09-01
Standardize junit6

breedx-splk:
Supersedes #2031.

Not entirely sure why renovate went crazy over there, but it was confusing to have the junit5 and 6 deps there. Originally, when the smoke tests were running on device, they were compiling the same way that the smoke test app was, so they needed to support java 8, which junit6 doesn't.

Now that the junit portion of the smoke tests are running on the host, we can run them with java 17 while the app is still compiled down to java 8 bytecode.


LikeTheSalad (approved):
Thanks!

Issue #910, opened 2025-03-26
Session user inactivity


LikeTheSalad:
As discussed in this week's SIG meeting, we should better define the logic/litmus test to tell when a session should expire.

### What we're doing

We split the app's usage into 2 states:

1. The app is in the foreground.
2. The app is in the background.

Based on that information, we make decisions on when to make a session expire by following these rules:

* When the app is in the foreground, the session can only expire after 4 hours.
* When the app is in the background, the session has a timeout of 15 minutes to expire, unless the state goes back to "foreground" before that.

### How we're doing it

The underlying mechanism we use to tell each state is by registering an observer into the [ProcessLifecycleOwner](https://developer.android.com/reference/androidx/lifecycle/ProcessLifecycleOwner) and deciding that the state is the **number 1** when the [ON_START](https://developer.android.com/reference/androidx/lifecycle/Lifecycle.Event#ON_START) event is received, and then switching to state **number 2** when the [ON_STOP](https://developer.android.com/reference/androidx/lifecycle/Lifecycle.Event#ON_STOP) one is received.

### Why we're doing it

When discussed in the SIG meeting, it seemed like the idea behind this behavior is to make sure that we're grouping telemetry in a session **only when a user is interacting with the app**, so that sessions can only track **real user activity**.

### State names changes

Replace the FOREGROUND/BACKGROUND states with something that better conveys what we want to achieve, such as:

1. ~The app is in the foreground~ -> The user is active.
2. ~The app is in the background~ -> The user is inactive.

### Configurable state litmus test with simple default behavior

One of the issues with the current approach is that it assumes that the user might not be interacting with the application because the app is in the background (i.e. there are no visible activities). This might not always be true for all apps, such as those that provide notifications with action buttons, commonly used in media streaming apps.

Because of that, one idea proposed during the SIG meeting was to provide a way for our users to "signal the session manager" that the user is active or not, so that we don't have to make that decision on behalf of every app use case. For convenience, we'd come up with a default behavior to tell when a user is active or not, that should work for most apps.

Some ideas for the default behavior are:

- Keep using the FOREGROUND/BACKGROUND callbacks as a way to tell whether a user is active or not.
- Always have a timeout that gets reset when telemetry is created, while keeping the 4 hour max time to live for a session.

aranhave:
I’m planning to pick this up as the first Android session-alignment step. I’ll also account for the timeout behavior in #794 and track explicit linked reset separately.

[...]

PR #2037, opened 2026-09-02
[proposal] ApiExtension PoC

LikeTheSalad:
> PoC. Not to be merged.

A mechanism that allows extending `OpenTelemetryRum`'s APIs without having to change `OpenTelemetryRum` or make it aware of feature-specific use cases.

## Why

`OpenTelemetryRum` is the core of what this project configures and hands to users. `AndroidInstrumentation` covers the "generate telemetry automatically on my behalf" use case. There's a second use case it doesn't cover: things that cannot be inferred reliably for every app, where we should give users a manual API instead of guessing for them. Two examples:

- #910: apps need to tell the session manager whether the user is active or not. Foreground/background is a reasonable default, but it's wrong for apps that are still in use from a notification (media players, for instance).
- The "current screen" attribute. Today `ScreenAttributesSpanProcessor` takes the last resumed activity or fragment as the current screen, via `VisibleScreenTracker`. That's a guess, and it doesn't hold for every app (see also #1909 for Compose).

These APIs are not core to `OpenTelemetryRum`, so they don't belong in the interface. They are also owned by different modules, some of them optional. This PR adds a small mechanism so each module can expose its own manual API through the `OpenTelemetryRum` instance, the same way `AndroidInstrumentation` lets each module contribute automatic telemetry.

This is a scaffolding PR. It's meant to show the key parts and how they connect. It doesn't compile yet and tests haven't been updated.

### `agent-api`

- `ApiExtension`: the marker every extension implements. It has a single member, `type`, which is the public interface the extension is registered under and looked up by. The public contract of an extension is an interface extending `ApiExtension`. Implementations stay internal to the module that owns them.
- `OpenTelemetryRum.getExtension(type)`: returns the extension registered under `type`, or null when this instance doesn't have it. It has a default body returning null, so existing implementations of `OpenTelemetryRum` keep compiling (`agent-api` is stable).
- `OpenTelemetryRum.extension<T>()`: reified Kotlin shorthand for the above.

### `core`

- `ApiExtensionLoader`: discovers extensions on the classpath through `ServiceLoader`, same shape as `AndroidInstrumentationLoaderImpl` so R8 can still rewrite the lookup.
- `ApiExtensionRegistry`: the map of extensions available to one `OpenTelemetryRum` instance. Built from discovered extensions first, then explicitly registered ones. An explicit registration replaces a discovered extension of the same type, with a warning in the log.
- `OpenTelemetryRumBuilder.addApiExtension()` and `SdkPreconfiguredRumBuilder.addApiExtension()`: explicit registration. This is for extensions that wrap internals created during initialization, which can't be discovered from the classpath.
- `SdkPreconfiguredRumBuilder.build()` resolves the registry before instrumentations are installed, so any `AndroidInstrumentation` can grab an extension from the `OpenTelemetryRum` it receives in `install()`.

`core` doesn't know about any concrete extension, just like it doesn't know about concrete instrumentations.

### `android-agent` (proof of concept for #910)

- `SessionActivityApi`: public interface with `userActive()` and `userInactive()`, plus a `fun OpenTelemetryRum.sessions()` convenience accessor for Kotlin callers.
- `SessionActivityApiImpl`: internal facade over `SessionIdTimeoutHandler`. Registered explicitly in `OpenTelemetryRumInitializer` because the handler is created there.
- `SessionIdTimeoutHandler` no longer implements `ApplicationStateListener`. It exposes `onUserActive()` / `onUserInactive()` and its states are renamed to active/inactive. `onUserActive()` only leaves the inactive state, so the first telemetry item after the user becomes active can still start a new session if the timeout elapsed.
- The initializer keeps the default lifecycle based inference, but it now goes through the extension: a lifecycle listener calls `userActive()` on foreground and `userInactive()` on background.

## How it's expected to be used

From app code:

```kotlin
val rum = OpenTelemetryRumInitializer.initialize(this)

// Kotlin convenience accessor. Null when the extension isn't available in this instance.
rum.sessions()?.userInactive()

// Generic lookup, works for any extension.
rum.extension<SessionActivityApi>()?.userActive()
```

From an instrumentation:

```kotlin
override fun install(context: Context, openTelemetryRum: OpenTelemetryRum) {
    val sessions = openTelemetryRum.extension<SessionActivityApi>()
    // ...call sessions?.userActive() when this instrumentation infers activity.
}
```

Defining an extension in a module:

```kotlin
// Public contract.
interface ScreenApi : ApiExtension {
    override val type: Class<out ApiExtension> get() = ScreenApi::class.java
    fun setCurrentScreen(name: String)
}

// Internal implementation. Either registered explicitly through the builder, or
// annotated with @AutoService(ApiExtension::class) if it doesn't need anything from initialization.
internal class ScreenApiImpl : ScreenApi { ... }

// Optional Kotlin convenience accessor.
fun OpenTelemetryRum.screens(): ScreenApi? = getExtension(ScreenApi::class.java)
```

## Rules the mechanism relies on

- An extension is a pure facade over its module's internals. No lifecycle, nothing to shut down.
- An extension is the single source of truth for what it controls. Automatic instrumentations that infer the same information write through the extension too. Last write wins. The sessions PoC follows this: the lifecycle listener and the app both write through `SessionActivityApi`.
- When both a discovered and an explicitly registered extension exist for the same type, the explicit one wins.
- Lookups return null when the extension isn't available, for example when using `RumBuilder` directly without the module that provides it.

## Not in this PR

- A config toggle to disable classpath discovery of extensions (separate from the instrumentation one).
- The screen API. Where it lives is still to be decided, possibly a new shared module.
- Test updates. `SessionIdTimeoutHandler` tests still reference the old lifecycle callbacks, and any test constructing `OpenTelemetryRumImpl` directly needs the new registry argument.
- `apiDump` for `agent-api`, `core` and `android-agent`.

## Open for discussion

- Naming: `ApiExtension` and `getExtension()` vs something that doesn't collide with Kotlin extension functions.
- Whether `SessionActivityApi` belongs in `android-agent` or in `session`.
- Anything else I might've missed.

### DIFF
--- a/agent-api/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
+++ b/agent-api/src/main/java/io/opentelemetry/android/OpenTelemetryRum.kt
@@ -5,6 +5,7 @@
 
 package io.opentelemetry.android
 
+import io.opentelemetry.android.extensions.ApiExtension
 import io.opentelemetry.android.session.SessionProvider
 import io.opentelemetry.api.OpenTelemetry
 import io.opentelemetry.api.common.Attributes
@@ -44,6 +45,17 @@ interface OpenTelemetryRum {
         body: String = "",
         attributes: Attributes = Attributes.empty(),
     )
+
+    /**
+     * Returns the [ApiExtension] registered under [type], or null when this instance doesn't
+     * provide it. Extensions are manual APIs contributed by optional modules, see [ApiExtension].
+     *
+     * Kotlin callers can use the reified [io.opentelemetry.android.extensions.extension] helper
+     * instead.
+     *
+     * The default returns null so that existing implementations of this interface keep working.
+     */
+    fun <T : ApiExtension> getExtension(type: Class<T>): T? = null
 
     /**
      * Initiates orderly shutdown of this OpenTelemetryRum instance. After this method completes,
--- a/agent-api/src/main/java/io/opentelemetry/android/extensions/ApiExtension.kt
+++ b/agent-api/src/main/java/io/opentelemetry/android/extensions/ApiExtension.kt
@@ -1,0 +1,43 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.android.extensions
+
+/**
+ * A manual API that optional modules make available through
+ * [io.opentelemetry.android.OpenTelemetryRum.getExtension].
+ *
+ * [io.opentelemetry.android.OpenTelemetryRum] only exposes what every setup needs. Some modules
+ * need to let users provide information that cannot be reliably inferred on their behalf, such as
+ * whether the user is currently active, or what the current screen is. Those modules define an
+ * [ApiExtension] instead of adding members to [io.opentelemetry.android.OpenTelemetryRum].
+ *
+ * Conventions:
+ * - The public contract is an interface that extends [ApiExtension] and provides [type].
+ *   Implementations stay internal to the module that owns them.
+ * - An extension is a facade over the internals of the module that owns it. It has no lifecycle.
+ * - The extension is the single source of truth for what it controls. Automatic instrumentations
+ *   that infer the same information write through the extension too, so the last write wins.
+ * - Modules may expose a Kotlin extension function on [io.opentelemetry.android.OpenTelemetryRum]
+ *   for convenience, e.g. `fun OpenTelemetryRum.sessions(): SessionActivityApi?`.
+ *
+ * Extensions become available to an [io.opentelemetry.android.OpenTelemetryRum] instance in two
+ * ways:
+ * - Discovered on the classpath via `ServiceLoader` (`@AutoService(ApiExtension::class)`), for
+ *   extensions that don't need anything created during initialization.
+ * - Registered explicitly through the RUM builder, for extensions that wrap internals created
+ *   during initialization.
+ *
+ * Extensions are resolved before instrumentations are installed, so an
+ * [io.opentelemetry.android.instrumentation.AndroidInstrumentation] can obtain them from the
+ * [io.opentelemetry.android.OpenTelemetryRum] it receives in `install()`.
+ */
+interface ApiExtension {
+    /**
+     * The public type this extension is registered under. Lookups through
+     * [io.opentelemetry.android.OpenTelemetryRum.getExtension] must use this exact type.
+     */
+    val type: Class<out ApiExtension>
+}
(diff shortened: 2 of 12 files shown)


bidetofevil:
So it feels like this can be a way to isolate dependencies in the API that aren't always needed (i.e. exists in a module that the agent/API module doesn't have access too), which keeps the agent leaner and not forcing dependencies on users via the API. 

That would be useful!

fractalwrench (commented):
I like the idea of avoiding singleton state and making APIs discoverable on `OpenTelemetryRum`. I'm not immediately sure the benefit `ApiExtension` brings as a marker interface versus just creating an extension function in instrumentation modules, e.g. something like below:

```kotlin
// defined in instrumentation module
fun OpenTelemetryRum.screens(): ScreenApi

interface ScreenApi {
    fun setCurrentScreen(screen: String)
}

fun example(otel: OpenTelemetryRum) {
    otel.screens().setCurrentScreen("my_screen")
}
```

LikeTheSalad:
> I'm not immediately sure the benefit `ApiExtension` brings as a marker interface versus just creating an extension function in instrumentation modules

That's fair, and at first I was thinking about this too. I have a couple of reasons why I chose to have our own API type, though, which are up for debate:

- Structure: there's a clear contract, which provides an entry point and can also make these APIs easier to discover. With kotlin extensions we essentially have to hope for modules to follow similar rules (like making sure that the extensions target `OpenTelemetryRum`), and it's also a bit awkward to find all the ones that are available, at least last time I checked it was not straightforward to make AS show all the available extensions for a type, rather than just searching for all the `ApiExtension` impls.
- Flexibility: Kotlin extensions usually rely on their access to `this` (`OpenTelemetryRum` in this case) to complement their actions; however, some state isn't public and so cannot be gotten from `OpenTelemetryRum`, such as the SessionIdTimeoutHandler used in [this example](https://github.com/open-telemetry/opentelemetry-android/pull/2037/changes#diff-b3152162cda19d17c3d31ad2bc3165c2d171b140e46086f465ba88f3361d14bfR164). It's probably possible to work around this with Kotlin extensions, but I think we'd sooner or later fall into using statics for that case, so this kind of approach seems not only more flexible but also safer in general.
- Control: Kotlin extensions are essentially statics handled by the Kotlin compiler; we can't "disable" them if we wanted to, or probably even ensure that there will be no conflict across similar names chosen in different modules, especially the optional ones. Having this structure also helps us debug them more easily, e.g., by logging when an extension is found/installed at runtime.
- Friendliness: `ApiExtension`s will ensure a similar user experience regardless of whether they call them from Kotlin or Java.

PR #2046, opened 2026-09-03
Add a session provider API to the DSL

### LINKED ISSUE #2044: DSL should allow setting SessionIdProvider
The DSL does not expose a way to set a `SessionIdProvider` but core does. We should add an equivalent API: https://github.com/open-telemetry/opentelemetry-android/blob/47d391cf4ef42b2aeff409699cf206d75735f108/core/src/main/java/io/opentelemetry/android/OpenTelemetryRumBuilder.kt#L269

fractalwrench:
Fixes #2044 by adding an API to the DSL that allows a `SessionProvider` to be specified. This grants parity with the session APIs that are available in the core module.

### DIFF
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/SessionConfiguration.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/SessionConfiguration.kt
@@ -6,6 +6,8 @@
 package io.opentelemetry.android.agent.dsl
 
 import io.opentelemetry.android.session.SessionObserver
+import io.opentelemetry.android.session.SessionProvider
+import io.opentelemetry.android.session.SessionPublisher
 import kotlin.time.Duration
 import kotlin.time.Duration.Companion.hours
 import kotlin.time.Duration.Companion.minutes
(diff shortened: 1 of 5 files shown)


android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/SessionConfiguration.kt
```diff
@@ -26,6 +28,14 @@ class SessionConfiguration internal constructor() {
      */
     var maxLifetime: Duration = 4.hours
 
+    /**
+     * A custom [SessionProvider] that controls the session ID. When this is set
+     * [backgroundInactivityTimeout] and [maxLifetime] have no effect
+     * and the supplied provider owns the session lifecycle entirely. Any [observers] are registered
+     * on the supplied provider only if it also implements [SessionPublisher].
+     */
+    var provider: SessionProvider? = null
```
LikeTheSalad, line 37: I like the idea, but I believe it could be confusing right now because essentially this new setter is overriding everything that can be set within the same `SessionConfiguration` type. I guess ideally a setter like this should be in an upper layer of the DSL, but that's probably not possible without introducing breaking changes, so maybe we can at least make the name a bit more explicit? (i.e. something like `providerOverride`)

PR #2047, opened 2026-09-03
Add DSL for configuring disk buffering

### LINKED ISSUE #2041: Expose equivalent DiskBuffering config on the DSL
The DSL API currently only allows setting the `enabled` flag on the disk buffering configuration, whereas the core module allows specifying various other options. We should make the DSL equivalent with the core API, and evaluate whether each API does make sense during this process.

https://github.com/open-telemetry/opentelemetry-android/blob/77df81e5acb96031fd50532a90ea18f6c5381123/core/src/main/java/io/opentelemetry/android/features/diskbuffering/DiskBufferingConfig.kt#L20

https://github.com/open-telemetry/opentelemetry-android/blob/ad76f8937c49527134926b2e43d37cfcfdb1f253/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/DiskBufferingConfigurationSpec.kt#L15

fractalwrench:
Closes #2041 by adding a DSL that allows library consumers to configure disk buffering. This adds important functionality that is currently available in the core module (specifying the export period, storage location, and max cache size).

### DIFF
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/DiskBufferingConfigurationSpec.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/DiskBufferingConfigurationSpec.kt
@@ -9,10 +9,8 @@ import io.opentelemetry.android.agent.dsl.instrumentation.CanBeEnabledAndDisable
 import io.opentelemetry.android.config.OtelRumConfig
 import io.opentelemetry.android.features.diskbuffering.DEFAULT_MAX_CACHE_SIZE
 import io.opentelemetry.android.features.diskbuffering.DiskBufferingConfig
-import io.opentelemetry.android.features.diskbuffering.MAX_CACHE_FILE_SIZE
 import java.io.File
 import kotlin.time.Duration
-import kotlin.time.Duration.Companion.hours
 import kotlin.time.Duration.Companion.seconds
 
 /**
@@ -28,27 +26,6 @@ class DiskBufferingConfigurationSpec internal constructor(
      * The maximum amount of disk space, in bytes, that buffered telemetry may occupy.
      */
     var maxCacheSize: Int = DEFAULT_MAX_CACHE_SIZE
-
-    /**
-     * The maximum size, in bytes, of a single buffer file.
-     */
-    var maxCacheFileSize: Int = MAX_CACHE_FILE_SIZE
-
-    /**
-     * The age at which a buffer file stops being appended to and a new one is started. Must be
-     * shorter than [minFileAgeForRead].
-     */
-    var maxFileAgeForWrite: Duration = 30.seconds
-
-    /**
-     * The minimum age a buffer file must reach before it’s eligible for export.
-     */
-    var minFileAgeForRead: Duration = 33.seconds
-
-    /**
-     * The age at which a buffer file is considered stale and is discarded without being exported.
-     */
-    var maxFileAgeForRead: Duration = 18.hours
 
     /**
      * How often buffered telemetry is read back from disk and exported. Must be greater than zero.
@@ -77,10 +54,6 @@ class DiskBufferingConfigurationSpec internal constructor(
             DiskBufferingConfig.create(
                 enabled = enabled,
                 maxCacheSize = maxCacheSize,
-                maxFileAgeForWriteMillis = maxFileAgeForWrite.inWholeMilliseconds,
-                minFileAgeForReadMillis = minFileAgeForRead.inWholeMilliseconds,
-                maxFileAgeForReadMillis = maxFileAgeForRead.inWholeMilliseconds,
-                maxCacheFileSize = maxCacheFileSize,
                 signalsBufferDir = signalsBufferDir,
                 exportPeriodMillis = exportPeriod.inWholeMilliseconds,
             ),
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/DiskBufferingConfigurationSpec.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/DiskBufferingConfigurationSpec.kt
@@ -41,8 +41,7 @@ class DiskBufferingConfigurationSpec internal constructor(
     var maxFileAgeForWrite: Duration = 30.seconds
 
     /**
-     * The minimum age a buffer file must reach before its eligible for
-     * export.
+     * The minimum age a buffer file must reach before it’s eligible for export.
      */
     var minFileAgeForRead: Duration = 33.seconds
 
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/DiskBufferingConfigurationSpec.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/DiskBufferingConfigurationSpec.kt
@@ -7,7 +7,13 @@ package io.opentelemetry.android.agent.dsl
 
 import io.opentelemetry.android.agent.dsl.instrumentation.CanBeEnabledAndDisabled
 import io.opentelemetry.android.config.OtelRumConfig
+import io.opentelemetry.android.features.diskbuffering.DEFAULT_MAX_CACHE_SIZE
 import io.opentelemetry.android.features.diskbuffering.DiskBufferingConfig
+import io.opentelemetry.android.features.diskbuffering.MAX_CACHE_FILE_SIZE
+import java.io.File
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.hours
+import kotlin.time.Duration.Companion.seconds
 
 /**
  * Type-safe config DSL that controls how disk buffering of exported telemetry should behave.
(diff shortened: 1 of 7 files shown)


android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/DiskBufferingConfigurationSpec.kt
```diff
@@ -17,16 +23,68 @@ class DiskBufferingConfigurationSpec internal constructor(
...
+
+    /**
+     * The maximum amount of disk space, in bytes, that buffered telemetry may occupy.
+     */
+    var maxCacheSize: Int = DEFAULT_MAX_CACHE_SIZE
+
+    /**
+     * The maximum size, in bytes, of a single buffer file.
+     */
+    var maxCacheFileSize: Int = MAX_CACHE_FILE_SIZE
```
aranhave, line 35: `DiskManager` splits `maxCacheSize` across three signal folders, but a file can still be larger than its folder’s share. The re-entry test ends up with 2 KiB total and a 1 MiB file limit. Do we want to reject that combination so the total cap is actually enforced?

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/DiskBufferingConfigurationSpec.kt
```diff
@@ -17,16 +23,68 @@ class DiskBufferingConfigurationSpec internal constructor(
...
+    /**
+     * The minimum age a buffer file must reach before its eligible for
+     * export.
+     */
+    var minFileAgeForRead: Duration = 33.seconds
+
+    /**
+     * The age at which a buffer file is considered stale and is discarded without being exported.
+     */
+    var maxFileAgeForRead: Duration = 18.hours
```
aranhave, line 51: One edge case: `maxFileAgeForRead` can be less than or equal to `minFileAgeForRead`. That leaves no usable window where a file is eligible for export before it expires. Does the config need to reject that combination?

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/DiskBufferingConfigurationSpec.kt
```diff
@@ -17,16 +23,68 @@ class DiskBufferingConfigurationSpec internal constructor(
...
+    /**
+     * The maximum size, in bytes, of a single buffer file.
+     */
+    var maxCacheFileSize: Int = MAX_CACHE_FILE_SIZE
+
+    /**
+     * The age at which a buffer file stops being appended to and a new one is started. Must be
+     * shorter than [minFileAgeForRead].
+     */
+    var maxFileAgeForWrite: Duration = 30.seconds
```
aranhave, line 41: The [#1821](https://github.com/open-telemetry/opentelemetry-android/issues/1821) discussion leaned toward keeping the DSL to `exportPeriod` and `discardAfter`, without exposing the three storage ages. Did we decide to go with the lower-level API instead?

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/DiskBufferingConfigurationSpec.kt
fractalwrench: Feels reasonable to reject that combination.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/DiskBufferingConfigurationSpec.kt
fractalwrench: Deferring to @LikeTheSalad as he's much more familiar with this part of the codebase. It sounds reasonable to reject this combination at face value

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/DiskBufferingConfigurationSpec.kt
fractalwrench: I wasn't aware of that discussion and would be happy to move towards that sort of API (assuming it's not massively complex retrofitting that onto the disk buffering config implementation). The changeset as it stands is just a 1:1 replication of what's in the `OtelRumConfig`.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/DiskBufferingConfigurationSpec.kt
LikeTheSalad: Generally speaking, I think [disk buffering's config](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/3f41f17c5f472c0109f192bc97a9aabb82cbc146/disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/storage/impl/FileStorageConfiguration.java) is too low-level and implementation-specific for the use case we want to provide with the DSL. Following [this other discussion](https://github.com/open-telemetry/opentelemetry-android/pull/2047#discussion_r3925308798), I agree we should provide more knobs from the DSL, but in a generic way, which we can always extend as needed. I'd propose the options mentioned there, plus a per-signal max cache size (for example, `tracesMaxCacheSize`, `logsMaxCacheSize`) maybe with better names, but I hope it shows the idea. What @aranhave mentions about us exposing a "total cache size" and then dividing it by 3 to provide each signal's max size is true, but in hindsight I think it was a mistake of ours, as it assumes users will always want to use the 3 types of signals and so, if they wanted to only use logs for example, then they'd have to set the value they want times 3 to get the actual logs to use that, which is silly.

I'm fine with introducing breaking changes or deprecating stuff from `core` in case that's needed to provide a more sensible set of configs for persistence in general.

fractalwrench:
@LikeTheSalad @aranhave based on your feedback I thought for the purposes of this PR it would be good to reduce the scope of new DSL APIs to the following:

1. `exportPeriod` - to configure the frequency at which buffered telemetry is exported
2. `maxCacheSize` - limit the amount of data that can be buffered on disk
3. `signalsBufferDir` - the location on the file system where data should be persisted

To me that felt like the critical functionality I'd expect from configuration, and I think that more granular APIs such as limits for individual signal file size could be added on later.

aranhave:
> @LikeTheSalad @aranhave based on your feedback I thought for the purposes of this PR it would be good to reduce the scope of new DSL APIs to the following:
> 
> 1. `exportPeriod` - to configure the frequency at which buffered telemetry is exported
> 2. `maxCacheSize` - limit the amount of data that can be buffered on disk
> 3. `signalsBufferDir` - the location on the file system where data should be persisted
> 
> To me that felt like the critical functionality I'd expect from configuration, and I think that more granular APIs such as limits for individual signal file size could be added on later.

👍 The smaller scope works for me. The cache-limit issue still reproduces: setting 2 KB allowed over 8 KB on disk. Can we enforce that limit internally without adding another public setting?

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/DiskBufferingConfigurationSpec.kt
```diff
@@ -17,16 +21,42 @@ class DiskBufferingConfigurationSpec internal constructor(
...
+    /**
+     * How often buffered telemetry is read back from disk and exported. Must be greater than zero.
+     */
+    var exportPeriod: Duration = 10.seconds
+
+    /**
+     * The directory in which buffered telemetry is stored until it is exported. By default
+     * the application's cache directory is used.
+     */
+    var signalsBufferDir: File? = null
```
LikeTheSalad: Just wanted to note that, by design, disk buffering must only store same-type signals in the same directory, meaning that a single dir can only have either logs, spans, or metrics, but not a combination of those in any way. So, based on that, the dir path users will put in this variable will effectively be the "root" dir, where each signal's subdir shall be created to store each signal's data files.

I'm mentioning this because I assume that this might not be clear from an end-user perspective. If you agree, then I think we should either mention this in the kdoc or rename the variable to be more explicit.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/DiskBufferingConfigurationSpec.kt
```diff
@@ -17,16 +21,42 @@ class DiskBufferingConfigurationSpec internal constructor(
     private val rumConfig: OtelRumConfig,
 ) : CanBeEnabledAndDisabled {
     internal var enabled: Boolean = true
-        set(value) {
-            field = value
-            rumConfig.setDiskBufferingConfig(DiskBufferingConfig.create(enabled = value))
-        }
+
+    /**
+     * The maximum amount of disk space, in bytes, that buffered telemetry may occupy.
```
benjoseph-grafana, line 26: Can we mention this is split three ways and each type gets 1/3rd of the size defined here?

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/DiskBufferingConfigurationSpec.kt
fractalwrench: I've renamed this variable to `telemetryStorageDir`. I'd be happy with other suggestions

fractalwrench:
>The cache-limit issue still reproduces: setting 2 KB allowed over 8 KB on disk. Can we enforce that limit internally without adding another public setting?

I'd suggest we continue the discussion for that in #1821 or start a new issue

LikeTheSalad (approved):
Thank you!

benjoseph-grafana (approved):

aranhave (approved):

PR #2050, opened 2026-09-03
feat(native-crash): replay stack snapshots

aranhave:
Builds on #2032 to attach validated native frames and build IDs to restart-time `app.crash` events.

A valid marker still replays without native frames when its snapshot is missing or invalid, and marker/snapshot files are cleaned up together. Durable delivery claims and bounded retries will follow in the next stacked PR. Native signal-handler capture remains separate.

Part of #1940.

Merge tier: Tier 3


instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt
```diff
@@ -190,32 +230,50 @@ internal class FileNativeCrashStore(
...
 
-    override fun deleteCrashRecord() {
-        runCatching {
-            if (crashRecordPath.isFile && !crashRecordPath.delete()) {
-                throw IOException("Failed to delete native crash marker")
+    override fun readCrashSnapshot(record: NativeCrashRecord): NativeCrashSnapshot? {
+        if (!crashSnapshotPath.isFile) return null
+        val snapshot =
+            try {
+                DataInputStream(FileInputStream(crashSnapshotPath)).use { input ->
```
fractalwrench, line 255: nit: should this validate that the size is the expected number of bytes, or is that done elsewhere?

aranhave: It’s checked just below this: `readFully()` rejects short files, and `input.read() == -1` checks there are no extra bytes. Either case returns null safely.

fractalwrench (approved):
LGTM

LikeTheSalad (approved):
Thank you!

aranhave:
@fractalwrench @LikeTheSalad thanks for reviewing! Could one of you help get this merged? 🙇

PR #2052, opened 2026-09-03
feat(native-crash): bound crash recovery

aranhave:
Stacked on #2051.

Adds a durable delivery claim and bounded retries for marker reads, snapshot reads, and cleanup. Recovery pauses signal-handler installation while work is pending, and a claimed crash is never emitted twice.

Remaining work is ordered and will follow separately:

1. Prepare immutable module metadata outside the signal handler.
2. Add async-signal-safe register and bounded-stack capture, then write the snapshot.
3. Run end-to-end device validation across the supported Android ABIs and malformed or partial data cases.

Each step builds on the previous one and this recovery stack.

Part of #1940.

Merge tier: Tier 3


aranhave:
Refreshed at [`9b35342d`](https://github.com/open-telemetry/opentelemetry-android/commit/9b35342d5d7b3073871942173a05c4dda4ef0438), now including #2051 at `2e9981d3` and the latest #2050 fixes.

The 141 native-crash tests passed, covering retries, delivery claims, and cleanup failures. Repository checks, API checks, lint, coverage verification, and debug/release builds passed for all four ABIs. The existing native-writer/Kotlin-reader compatibility test passed on an Android 15 ARM64 emulator; a separate macOS two-JVM test checked lock release after process death and persistence of the saved claim.

Still a draft. The diff against #2051 is 608 changed lines, so size needs another look before review. Delivery claims prevent repeat handoff to OTel, not guarantee export. Native snapshot capture and end-to-end device testing across all ABIs remain follow-ups.

Issue #2048, opened 2026-09-03
Define versioned session persistence across process recreation


aranhave:
## Problem

The built-in Android session manager only keeps its state in memory, so normal process recreation loses the current session record.

At [`feaf191f`](https://github.com/open-telemetry/opentelemetry-android/commit/feaf191fe130ee458cfb0ca13f827b3a23e01971), [`SessionManager`](https://github.com/open-telemetry/opentelemetry-android/blob/feaf191fe130ee458cfb0ca13f827b3a23e01971/android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionManager.kt) uses [`InMemorySessionStorage`](https://github.com/open-telemetry/opentelemetry-android/blob/feaf191fe130ee458cfb0ca13f827b3a23e01971/android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/InMemorySessionStorage.kt). [`OpenTelemetryRumInitializer`](https://github.com/open-telemetry/opentelemetry-android/blob/feaf191fe130ee458cfb0ca13f827b3a23e01971/android-agent/src/main/kotlin/io/opentelemetry/android/agent/OpenTelemetryRumInitializer.kt#L147-L161) therefore starts each new process with an empty session manager.

This issue is limited to sequential recreation of an app that initializes OpenTelemetry in one process. General support for apps running the SDK in several processes at the same time is out of scope.

## Proposed API

Build on the pluggable [`SessionProvider` configuration proposed in #2046](https://github.com/open-telemetry/opentelemetry-android/pull/2046) instead of adding a separate persistence flag. A built-in persistent implementation could be selected through the session DSL:

```kotlin
session {
    provider {
        persistentSessionProvider(...)
    }
}
```

This is an example rather than a settled API. #2046 is still discussing the exact shape and naming of the provider configuration.

When no provider is selected, applications keep the current built-in in-memory behavior. Persistence is opt-in by selecting the built-in persistent provider. The storage type and record remain private.

### Provider responsibilities

The provider boundary must be agreed before implementation. In the current #2046 proposal, supplying a provider bypasses the built-in `SessionManager` and its lifecycle timeout handler. We need to choose whether:

1. the persistent provider owns ID generation, expiry, lifecycle input, storage, and observer publication; or
2. the public hook selects storage beneath the built-in manager, which continues to own lifecycle and session transitions.

[...]

PR #2061, opened 2026-09-11
feat(session): allow custom storage in the DSL

aranhave:
Part of #2048, following the [SIG discussion](https://github.com/open-telemetry/opentelemetry-android/issues/2048#issuecomment-5622292142).

Adds `session { storage(myStorage) }` using the existing session manager. The in-memory default, expiry, activity tracking and observers stay unchanged. The existing storage interface and new setter are marked incubating.

This is the configuration hook, not session restoration. The manager still writes an empty session at startup and never reads storage. Storage implementations handle their own failures; disk storage and restart policy stay separate.

 Tier 4: public API addition.

### DIFF
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/SessionConfiguration.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/SessionConfiguration.kt
@@ -5,6 +5,9 @@
 
 package io.opentelemetry.android.agent.dsl
 
+import io.opentelemetry.android.Incubating
+import io.opentelemetry.android.agent.session.InMemorySessionStorage
+import io.opentelemetry.android.agent.session.SessionStorage
 import io.opentelemetry.android.session.SessionObserver
 import kotlin.time.Duration
 import kotlin.time.Duration.Companion.hours
@@ -14,6 +17,7 @@ import kotlin.time.Duration.Companion.minutes
  * Type-safe config DSL that controls how sessions should behave.
  */
 @OpenTelemetryDslMarker
+@OptIn(Incubating::class)
 class SessionConfiguration internal constructor() {
     /**
      * The maximum duration which a session can remain open in the background before it
@@ -25,6 +29,19 @@ class SessionConfiguration internal constructor() {
      * The maximum duration which a session can remain open before it automatically expires.
      */
     var maxLifetime: Duration = 4.hours
+
+    internal var sessionStorage: SessionStorage = InMemorySessionStorage()
+        private set
+
+    /**
+     * Replaces the default in-memory storage without changing session generation, expiry, or
+     * observers. This does not restore sessions across launches; see [SessionStorage] for the
+     * startup and failure contract. If called more than once, the last storage supplied is used.
+     */
+    @Incubating
+    fun storage(storage: SessionStorage) {
+        sessionStorage = storage
+    }
 
     private var observersList: MutableList<SessionObserver> = mutableListOf()
 
--- a/android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionManager.kt
+++ b/android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionManager.kt
@@ -12,7 +12,6 @@ import io.opentelemetry.android.session.SessionProvider
 import io.opentelemetry.android.session.SessionPublisher
 import io.opentelemetry.sdk.common.Clock
 import java.util.concurrent.CopyOnWriteArrayList
-import java.util.concurrent.atomic.AtomicReference
 import kotlin.random.Random
 import kotlin.time.Duration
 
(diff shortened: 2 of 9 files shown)


android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionManager.kt
```diff
@@ -86,11 +87,13 @@ internal class SessionManager(
             timeoutHandler: SessionIdTimeoutHandler,
             sessionConfig: SessionConfig,
             clock: Clock,
+            sessionStorage: SessionStorage = InMemorySessionStorage(),
         ): SessionManager =
             SessionManager(
                 timeoutHandler = timeoutHandler,
                 maxSessionLifetime = sessionConfig.maxLifetime,
                 clock = clock,
+                sessionStorage = sessionStorage,
```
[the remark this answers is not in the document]
aranhave: Fixed in 221745ed. Session changes now stay serialized through the save and observer callbacks, with the timeout reset before either runs. Added regression tests for slow storage, notification ordering and reentrant session reads.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionManager.kt
```diff
@@ -24,45 +24,35 @@ internal class SessionManager(
...
-        // Check if we need to create a new session.
-        return if (sessionHasExpired(currentSession) || timeoutHandler.hasTimedOut()) {
-            val newId = idGenerator.generateSessionId()
-            val newSession = SessionImpl(newId, clock.now())
-
-            // Atomically update the session only if it hasn't been changed by another thread.
-            if (session.compareAndSet(currentSession, newSession)) {
-                sessionStorage.save(newSession)
+    override fun getSessionId(): String =
+        synchronized(lock) {
```
benjoseph-grafana: Can we use a double check here and not lock unless necessary?

android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionManager.kt
```diff
@@ -24,45 +24,35 @@ internal class SessionManager(
...
 ) : SessionProvider,
     SessionPublisher {
-    private val session: AtomicReference<Session> = AtomicReference(invalidSession)
+    private val lock = Any()
+    private var session: Session = invalidSession
     private val observers = CopyOnWriteArrayList<SessionObserver>()
 
     init {
-        sessionStorage.save(session.get())
+        sessionStorage.save(session)
```
benjoseph-grafana: I think we should drop this save in init, with persisted storage, this will overwrite previous session info.

benjoseph-grafana (commented):
@aranhave do you think this is worth a mention in the changelog?

android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionManager.kt
aranhave: Added the double-check so normal reads avoid the lock, while storage and observer updates stay ordered.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionManager.kt
aranhave: Removed it. Construction now leaves storage untouched. the first new session still replaces the stored value.

aranhave:
> @aranhave do you think this is worth a mention in the changelog?

Yes 💯 , added an entry under Unreleased for the storage hook.

benjoseph-grafana (approved):

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/SessionConfiguration.kt
```diff
@@ -26,6 +30,19 @@ class SessionConfiguration internal constructor() {
...
+    internal var sessionStorage: SessionStorage = InMemorySessionStorage()
+        private set
+
+    /**
+     * Replaces the default in-memory storage without changing session generation, expiry, or
+     * observers. This does not restore sessions across launches; see [SessionStorage] for the
+     * startup and failure contract. If called more than once, the last storage supplied is used.
+     */
+    @Incubating
+    fun storage(storage: SessionStorage) {
```
fractalwrench, line 42: This module is declared stable right now so I think it's worth discussing in the SIG how we handle adding new APIs. I would be ok with this approach, but I can't speak for everyone else.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionManager.kt
```diff
@@ -24,43 +24,47 @@ internal class SessionManager(
...
+                val newId = idGenerator.generateSessionId()
+                val newSession = SessionImpl(newId, clock.now())
+                // Readers must wait until storage and observers finish the transition.
+                transitionInProgress = true
+                try {
+                    session = newSession
+                    // Storage and observers may record telemetry that reads the session again.
+                    timeoutHandler.bump()
+                    sessionStorage.save(newSession)
+                    notifyObserversOfSessionUpdate(latestSession, newSession)
```
fractalwrench, line 59: I'm a little wary of invoking user code whilst under a lock, particularly if it could take a long time to perform an operation like writing to disk.

LikeTheSalad: +1

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/SessionConfiguration.kt
LikeTheSalad: I'm also ok with this approach. Though if we go with it, I think it'd be useful to have some policy about how long things can stay "incubating" to avoid leaving them like that forever.

aranhave: @fractalwrench @LikeTheSalad @breedx-splk I was out for the last SIG, so I may have missed an update. Did this come up? Happy to discuss async here: are we comfortable adding this as `@Incubating`, and when should we revisit it for stabilization? We can take any remaining questions to the next SIG.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionManager.kt
aranhave: Storage and observer callbacks now run outside the lock, so concurrent reads don’t wait for them. Saves and notifications still run in order. Added regression tests for slow callbacks, expiry and failures.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/SessionConfiguration.kt
LikeTheSalad: We didn't talk about it in the previous SIG, but I think this is a sensible approach that we can always remove later, so I'm fine with keeping this incubating annotation.

LikeTheSalad (approved):
Thank you!

fractalwrench (approved):
LGTM

PR #2062, opened 2026-09-11
Add module and instructions for using opentelemetry-kotlin API

fractalwrench:
## Goal

Adds a new module named `agent-api-ktx` which provides the ability to use opentelemetry-kotlin's API in opentelemetry-android.

### How to use it
This module is opt-in and should be added as a dependency to the build.gradle:

```kotlin
dependencies {
    implementation("io.opentelemetry.android:agent-api-ktx")
}
```

The module exposes an extension property on `OpenTelemetryRum` named `openTelemetryKotlin` which exposes opentelemetry-kotlin's API. An example of what this might look like is shown below:

```kotlin
@OptIn(ExperimentalApi::class)
fun example(rum: OpenTelemetryRum) {
    val otel = rum.openTelemetryKotlin

    val logger = otel.loggerProvider.getLogger("my_logger")
    logger.log("Hello, World!")

    val tracer = otel.tracerProvider.getTracer("my_tracer")
    tracer.startSpan("my_span").end()
}
```

### How it works

opentelemetry-kotlin can run in regular mode or ['compat' mode](https://opentelemetry.io/docs/languages/kotlin/getting-started/#use-compatibility-mode). Compat mode decorates an instance of the opentelemetry-java SDK with the Kotlin API which means there should be no change in underlying behavior or where telemetry is sent, and makes it easy to adopt the API piecemeal.

opentelemetry-kotlin achieves this by declaring its API in a separate module, which is then implemented in two separate modules for compat/regular mode.

In future this could allow for the implementation to be switched behind the scenes without imposing any breaking API changes on the library consumer. 

### Why ktx?

Modules with the `-ktx` suffix are widely used in the [Android ecosystem](https://developer.android.com/kotlin/ktx) to retrofit Kotlin APIs onto APIs that were originally designed for Java. They tend to provide syntactic sugar.

An opt-in extension property felt like the right balance as it avoids us committing to adding to the `OpenTelemetryRum` interface ourselves right now, and leaves the option in future while longer-term we decide what happens to Java types on that interface.

### Does anything need to change first?

The `Clock` parameter is inaccurate in the returned Kotlin API because opentelemetry-android uses a custom implementation. https://github.com/open-telemetry/opentelemetry-kotlin/pull/1034 will allow us to address the problem.

### DIFF
--- a/agent-api-ktx/README.md
+++ b/agent-api-ktx/README.md
@@ -1,0 +1,49 @@
+# Agent API Kotlin extensions
+
+Optional module that exposes the
+[opentelemetry-kotlin](https://github.com/open-telemetry/opentelemetry-kotlin) API from an existing
+`OpenTelemetryRum` instance.
+
+> **Not stable.** This module may change its API or
+> be removed in any release, with no deprecation cycle. opentelemetry-kotlin is itself pre-1.0.
+
+## Installation
+
+```kotlin
+dependencies {
+    api(platform("io.opentelemetry.android:opentelemetry-android-bom:<version>-alpha"))
+    implementation("io.opentelemetry.android:agent-api-ktx") // Version is resolved through the BOM
+}
+```
+
+## Usage
+
+```kotlin
+import io.opentelemetry.android.OpenTelemetryRum
+import io.opentelemetry.android.openTelemetryKotlin
+import io.opentelemetry.kotlin.ExperimentalApi
+
+@OptIn(ExperimentalApi::class)
+fun example(rum: OpenTelemetryRum) {
+    val otel = rum.openTelemetryKotlin
+
+    val logger = otel.loggerProvider.getLogger("my_logger")
+    logger.log("Hello, World!")
+
+    val tracer = otel.tracerProvider.getTracer("my_tracer")
+    tracer.startSpan("my_span").end()
+}
+```
+
+`@OptIn` is required for most calls as opentelemetry-kotlin APIs are mostly marked `@ExperimentalApi`.
+See the [upstream getting started guide](https://opentelemetry.io/docs/languages/kotlin/getting-started/)
+for an alternative approach that opts in globally through Kotlin's compiler options.
+
+## What you get
+
+`openTelemetryKotlin` returns the *compat* implementation of opentelemetry-kotlin: a Kotlin API that
+delegates to the OpenTelemetry Java SDK the agent already configures. Spans,
+logs and metrics recorded through it flow through the same processors and exporters as telemetry
+recorded through `OpenTelemetryRum.openTelemetry`, and the two APIs can be used side by side.
+
+The same instance is returned on every access.
--- a/agent-api-ktx/src/main/kotlin/io/opentelemetry/android/OpenTelemetryRumKtx.kt
+++ b/agent-api-ktx/src/main/kotlin/io/opentelemetry/android/OpenTelemetryRumKtx.kt
@@ -1,0 +1,41 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+@file:OptIn(ExperimentalApi::class)
+
+package io.opentelemetry.android
+
+import io.opentelemetry.kotlin.ExperimentalApi
+import io.opentelemetry.kotlin.OpenTelemetry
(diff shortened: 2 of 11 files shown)


agent-api-ktx/src/main/kotlin/io/opentelemetry/android/OpenTelemetryRumKtx.kt
```diff
@@ -0,0 +1,41 @@
...
+ * with an `-alpha` version suffix. opentelemetry-kotlin is itself pre-1.0 and every type reachable
+ * from [OpenTelemetry] is marked [ExperimentalApi], so both this property and the API it exposes
+ * may change or be removed in any release, with no deprecation cycle. Opt in with
+ * `@OptIn(ExperimentalApi::class)` to acknowledge that.
+ */
+@ExperimentalApi
+val OpenTelemetryRum.openTelemetryKotlin: OpenTelemetry
+    get() =
+        synchronized(compatLock) {
+            compatInstances.getOrPut(this) { openTelemetry.toOtelKotlinApi() }
```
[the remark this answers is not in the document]
fractalwrench: Addressed in https://github.com/open-telemetry/opentelemetry-kotlin/pull/1034, will update this PR once that PR ships

bidetofevil (approved):
Very very nice

gradle/libs.versions.toml
```diff
@@ -3,6 +3,7 @@ opentelemetry-instrumentation-alpha = "2.31.1-alpha"
 #opentelemetry-instrumentation = "2.9.0" // alpha bom includes non-alpha bom
 opentelemetry-contrib = "1.60.0-alpha"
 opentelemetry-semconv-kotlin = "0.7.0"
+opentelemetry-kotlin = "0.7.0"
```
aranhave, line 6: The 0.7.0 adapter loses the active parent: a Kotlin `startSpan("child")` inside a Java span exports a new root trace. I reproduced this locally; passing `parentContext = otel.context.implicit()` keeps it connected. Could we pick up an upstream fix or document that workaround before recommending side-by-side use?

agent-api-ktx/src/main/kotlin/io/opentelemetry/android/OpenTelemetryRumKtx.kt
aranhave, line 40: The adapter applies `startSpan { setStringAttribute(...) }` attributes after Java has already sampled the span. With an attribute-based sampler, my Java span was exported but the equivalent Kotlin span was dropped. Could we include this in the upstream compat fixes?

agent-api-ktx/README.md
```diff
@@ -0,0 +1,49 @@
...
+import io.opentelemetry.android.OpenTelemetryRum
+import io.opentelemetry.android.openTelemetryKotlin
+import io.opentelemetry.kotlin.ExperimentalApi
+
+@OptIn(ExperimentalApi::class)
+fun example(rum: OpenTelemetryRum) {
+    val otel = rum.openTelemetryKotlin
+
+    val logger = otel.loggerProvider.getLogger("my_logger")
+    logger.log("Hello, World!")
```
aranhave: `logger.log(...)` doesn't exist in 0.7.0, so this example fails to compile. Could we use `logger.emit(...)` instead? That compiles locally. The PR description has the same call.

aranhave (commented):
Thanks, Jamie! This looks great. I’ve left a few things to address before merging, with the details inline.

gradle/libs.versions.toml
fractalwrench: Upstream fix here: https://github.com/open-telemetry/opentelemetry-kotlin/pull/1047

agent-api-ktx/src/main/kotlin/io/opentelemetry/android/OpenTelemetryRumKtx.kt
fractalwrench: Upstream fix here: https://github.com/open-telemetry/opentelemetry-kotlin/pull/1051

aranhave (approved):
Few conflicts, but otherwise LGTM!

LikeTheSalad (approved):
Cheers! 🚀

aranhave (approved):
LGTM! 🎖️

PR #2063, opened 2026-09-11
Convert thermal state instrumentation to use opentelemetry-kotlin Logger API

fractalwrench:
## Goal

This changeset converts the thermal state instrumentation to use opentelemetry-kotlin's Logger API. This change does not affect library consumers of the library and is an internal-only change. It builds on #2062 so only the second commit contains the details relevant to instrumentation, which [can be viewed in isolation here](https://github.com/open-telemetry/opentelemetry-android/pull/2063/changes/0a660d8a2b4f7dc1870554e0c4b8dee5977b978b).

As a recap, opentelemetry-kotlin can run in regular mode or ['compat' mode](https://opentelemetry.io/docs/languages/kotlin/getting-started/#use-compatibility-mode). Compat mode decorates an instance of the opentelemetry-java SDK with the Kotlin API which means there should be no change in underlying behavior or where telemetry is sent.

For context, the generated code in the semconv directory now looks like this:

```kotlin
/** Emits this event through the opentelemetry-kotlin [logger], including any [additionalAttributes]. */
fun emit(
    logger: OtelKotlinLogger,
    additionalAttributes: (AttributesMutator.() -> Unit)? = null,
) {
    logger.emit(eventName = DEVICE_THERMAL_STATUS_CHANGE_EVENT_NAME) {
        setStringAttribute(ANDROID_THERMAL_THROTTLING_STATUS, androidThermalThrottlingStatus)
        additionalAttributes?.invoke(this)
    }
}
```


bidetofevil (approved):
LGTM

semconv/templates/registry/kotlin/kotlin_event_template.kt.j2
```diff
@@ -51,4 +63,25 @@ class {{ my_class_name }}(
             .setAllAttributes(attributes.build())
             .emit()
     }
+{%- if ctx.kotlin_logger_emit %}
+
+    /** Emits this event through the opentelemetry-kotlin [logger], including any [additionalAttributes]. */
+    fun emit(
```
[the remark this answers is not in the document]
bidetofevil: Seems reasonable

agent-api-ktx/README.md
```diff
@@ -0,0 +1,49 @@
...
+import io.opentelemetry.android.OpenTelemetryRum
+import io.opentelemetry.android.openTelemetryKotlin
+import io.opentelemetry.kotlin.ExperimentalApi
+
+@OptIn(ExperimentalApi::class)
+fun example(rum: OpenTelemetryRum) {
+    val otel = rum.openTelemetryKotlin
+
+    val logger = otel.loggerProvider.getLogger("my_logger")
+    logger.log("Hello, World!")
```
benjoseph-grafana: shouldn't this be `emit` instead of `log`?

fractalwrench: Yes. In an earlier iteration of the API `log` was used - I'll update to use `emit` instead.

aranhave (approved):
Looks good overall! Just a few conflicts to fix.

fractalwrench:
Once everyone is happy that #2062 should land I'll resolve the conflicts as this is stacked on top of that PR.

PR #2067, opened 2026-09-15
ci: fix self-repository references

opentelemetrybot:
Use GitHub's dedicated `$/...` syntax for same-repository actions and reusable workflows so zizmor no longer reports `self-repository` findings.

Update the shared Scorecard workflow to v0.14.0, which filters the corresponding Scorecard false positives.

Strict all-level enforcement is staged in [open-telemetry/admin#890](https://github.com/open-telemetry/admin/pull/890).


fractalwrench (approved):
LGTM

Issue #1686, opened 2026-04-14
Define ignoreURLs generic use case and weigh different technical approaches


surbhiia:
### Use case

- Ignore URLs matching a pattern from being traced by HTTP instrumentations.
- If URL is ignored on the client side, server should also get the knowledge that this was ignored on client. 

#### 1. Sampler (Exists)

```
val urlBlocklistSampler = object : Sampler {
    override fun shouldSample(
        parentContext: Context,
        traceId: String,
        name: String,
        spanKind: SpanKind,
        attributes: Attributes,
        parentLinks: List<LinkData>,
    ): SamplingResult {
        val url = attributes.get(AttributeKey.stringKey("url.full"))
        if (url != null && shouldIgnoreUrl(url)) { //shouldIgnoreUrl function uses ignoreURL pattern config provided by customer
            return SamplingResult.drop()
        }
        return Sampler.parentBased(Sampler.alwaysOn()).shouldSample(
            parentContext, traceId, name, spanKind, attributes, parentLinks,
        )
    }
    override fun getDescription(): String = "UrlBlocklistSampler"
}

// When building Opentelemetry Tracer provider, add the sampler
SdkTracerProvider.builder().setSampler(urlBlocklistSampler)
```
#### 2. Instrumentation level configuration (Needs Implementation)

Upgrade the `TracingInterceptor` and `ConnectionErrorInterceptor`

2.1  `ConnectionErrorInterceptor`

```
@Override
public Response intercept(Chain chain) throws IOException {
    Request request = chain.request();
   
    //Add this if statement to support ignoreURLs use case
    if (shouldIgnoreUrl(request.url())) {
        return chain.proceed(request);
    }

    Context parentContext = Context.current();
    Response response = null;
    Throwable error = null;
    Instant startTime = Instant.now();
    try {
        response = chain.proceed(request);
        return response;
    } catch (Throwable t) {
        error = t;
        throw t;
    } finally {
        if (HttpClientRequestResendCount.get(parentContext) == 0) {
            if (instrumenter.shouldStart(parentContext, chain)) {
                InstrumenterUtil.startAndEnd(
                    instrumenter, parentContext, chain, response, error, startTime, Instant.now());
            }
        }
    }
}
```

2.2 `TracingInterceptor`
```
@Override
public Response intercept(Chain chain) throws IOException {
    Request request = chain.request();
    Context parentContext = Context.current();

   //Add this if statement to support ignoreURLs use case
[...]

PR #2074, opened 2026-09-18
Add DSL for filtering HTTP spans by host

benjoseph-grafana:
Every HTTP call an app makes is instrumented today, so third-party SDKs'
traffic populate dashboards with requests the app author does not care 
about. This adds an allowlist:

    instrumentations {
        httpTelemetry {
            onlyHosts("api.example.com")
        }
    }

Host names are matched in full, ignoring case. Naming no host keeps every
host, so behaviour is unchanged by default.

Filtering happens at export, through the existing FilteringSpanExporter, by
reading `server.address` off each span. No instrumentation is modified, which
keeps this within existing OpenTelemetry constructs as discussed at the
Android SIG.

Both sides of the comparison are converted to punycode. OkHttp canonicalizes
hosts with IDN.toASCII, while HttpURLConnection reports whatever host the
caller wrote, so the two clients spell the same internationalized host
differently.

Spans only. HTTP client metrics, websocket events and trace context
propagation are unaffected.

Refs #1686


android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/instrumentation/HttpTelemetryConfiguration.kt
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * Telemetry that records no host is always kept, so the allowlist only decides about
+     * spans carrying `server.address`.
+     */
+    internal fun rejects(span: SpanData): Boolean {
+        if (keepsEveryHost()) {
+            return false
+        }
+        val host = span.attributes.get(SERVER_ADDRESS) ?: return false
```
aranhave: Can we apply this only to HTTP client spans? Right now it also drops gRPC and database spans when their host isn’t on the list.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/instrumentation/HttpTelemetryConfiguration.kt
```diff
@@ -0,0 +1,104 @@
...
+         * this normalizes spelling without also imposing IDN label rules on it. Returns null
+         * for a host with no punycode form, which callers treat according to whether the
+         * host came from configuration or from recorded telemetry.
+         */
+        fun punycode(host: String): String? =
+            if (host.all { it.code <= MAX_ASCII }) {
+                host
+            } else {
+                try {
+                    IDN.toASCII(host).lowercase()
```
aranhave: I tried allowing `faß.de`, but its spans still get dropped because OkHttp and `IDN.toASCII` normalize it differently. Could we handle this case and add a test for it?

aranhave (commented):
Thanks, Ben! This fits well with the existing DSL and reusing the exporter makes sense :)  I left a couple of comments on the filtering behavior to work through before this lands.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/instrumentation/HttpTelemetryConfiguration.kt
benjoseph-grafana: I'm changing the API to accept on ASCII hostname to avoid additional complexity here.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/instrumentation/HttpTelemetryConfiguration.kt
benjoseph-grafana: Updated to filter only HTTP client spans.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/instrumentation/HttpTelemetryConfiguration.kt
```diff
@@ -0,0 +1,104 @@
...
+     * Telemetry that records no host is always kept, so the allowlist only decides about
+     * spans carrying `server.address`.
+     */
+    internal fun rejects(span: SpanData): Boolean {
+        if (keepsEveryHost()) {
+            return false
+        }
+        val host = span.attributes.get(SERVER_ADDRESS) ?: return false
+        val candidate = punycode(host.lowercase()) ?: return true
+        return candidate !in allowedHosts
```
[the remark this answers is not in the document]
benjoseph-grafana: I’ve checked the Kotlin behavior here: String.lowercase() in Kotlin is locale-insensitive because it uses the invariant locale, so it is not affected by Turkish-locale casing rules the way String.toLowerCase()/toLowerCase(Locale) can be. This is the correct API for host normalization in this case.

aranhave (approved):
LGTM!

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/instrumentation/HttpTelemetryConfiguration.kt
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    /** True while no host has been named, so that filtering can be skipped entirely. */
+    internal fun keepsEveryHost(): Boolean = allowedHosts.isEmpty()
+
+    /**
+     * Only HTTP client spans are considered. gRPC and database spans also record
+     * `server.address`, and a span that records no host at all is always kept.
+     */
+    internal fun rejects(span: SpanData): Boolean {
```
fractalwrench: IMO this should be extracted out of this class. `allowedHosts` is a mutable collection without synchronization so this could throw `ConcurrentModificationException` if a user retained the DSL reference and edited afterwards. But more importantly I think it'd be good to separate configuration from the behavior the SDK implements

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/instrumentation/HttpTelemetryConfiguration.kt
```diff
@@ -0,0 +1,122 @@
...
+        /**
+         * Rejects a host that could never match a recorded one.
+         */
+        fun validatedHost(value: String): String {
+            val host = value.trim().lowercase()
+            val bare =
+                host.isNotEmpty() &&
+                    host.all { it.code <= MAX_ASCII } &&
+                    host.none { it.isWhitespace() || it in ILLEGAL_HOST_CHARS }
+            require(bare) {
```
fractalwrench: I would lean towards warning with a log rather than throwing here.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/instrumentation/HttpTelemetryConfiguration.kt
```diff
@@ -0,0 +1,122 @@
...
+ *
+ * Every host is kept by default. Naming a host turns this into an allowlist: HTTP client
+ * spans recording any other host are dropped before export.
+ *
+ * Only spans are filtered. HTTP client metrics, websocket events and trace context
+ * propagation are unaffected.
+ *
+ * ```kotlin
+ * instrumentations {
+ *     httpTelemetry {
```
fractalwrench, line 23: This name doesn't automatically make me think of filtering out HTTP network request spans - I wonder if there's a better name? I don't have a better suggestion so we don't need to block on this..

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/instrumentation/HttpTelemetryConfiguration.kt
benjoseph-grafana: Agreed, done.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/instrumentation/HttpTelemetryConfiguration.kt
benjoseph-grafana: agree throwing is aggressive, switched to logs.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/dsl/instrumentation/HttpTelemetryConfiguration.kt
benjoseph-grafana: I'm open to feedback, the only other candidate I have is `httpSpans` itself

DavidGrath (approved):
Looks great as a first phase of the discussion from the linked issue.

fractalwrench (approved):
LGTM

PR #2077, opened 2026-09-21
Verify the minimum supported versions in CI and correct them

### LINKED ISSUE #1526: Compatibility issues when using opentelemetry-android with minimum required versions
### TL;DR

Testing opentelemetry-android with our documented minimum requirements (Kotlin 2.0, AGP 7.4) revealed compatibility issues:

1. **AGP 7.4 is not viable** - fails with modern AndroidX libraries and D8 dexing errors
2. **Minimum AGP could be 8.0.2** - resolves the tooling issues
3. **OkHttp 5.3.2 forces Kotlin stdlib 2.2.21** - incompatible with our stated Kotlin 2.0 minimum
4. **Action needed:** Either bump minimum Kotlin to 2.2 OR downgrade OkHttp dependency to v4 (how?)
5. **Missing requirement:** We should define a minimum `compileSdk` version in VERSIONING.md. 34 is a good starting point, as it works with AGP 8.0.2.
6. **Testing**: We could add something like my test app to the CI, and verify that it builds with the minimum versions, at least to get some basic verification that the stated minimum requirements are correct.

### Context

Following the discussion in #1489 about bumping the minimum Kotlin version to 2.0, I created a test app to verify all the compatibility claims made in our [VERSIONING.md](https://github.com/open-telemetry/opentelemetry-android/blob/main/VERSIONING.md):

- Kotlin 2.0
- API 26
- AGP 7.4 and Gradle 7.5
- JDK 11 (build-time)
- Java language level 8

**Test app repository:** https://github.com/priettt/small-android-app
- `main` branch: works without opentelemetry-android
- `otel-android-issue` branch: fails with opentelemetry-android

#### 1. AGP 7.4 incompatibility with modern AndroidX libraries

**Error:** `checkDebugAarMetadata` fails with lifecycle-runtime-ktx:2.9.x

```
Could not resolve androidx.lifecycle:lifecycle-runtime-ktx:2.6.1.
  > The consumer was configured to find a runtime of a component, preferably optimized for Android, 
    as well as attribute 'com.android.build.api.attributes.AgpVersionAttr' with value '7.4.2'
```

**Root cause:** `androidx.lifecycle:lifecycle-process:2.9.x` uses multi-variant KMP metadata that AGP 7.4 cannot resolve.

**Workaround:** Downgrade to `androidx.lifecycle:lifecycle-process:2.8.7`

#### 2. AGP 7.4 incompatibility with androidx.core

**Error:** `checkDebugAarMetadata` fails

```
Dependency 'androidx.core:core-ktx:1.17.0' requires Android Gradle plugin 8.9.1 or higher.
This build currently uses Android Gradle plugin 7.4.2.
```

**Workaround:** Downgrade to `androidx.core:core:1.13.1`

#### 3. D8 dexing failures with AGP 7.4

**Error:** Multiple `NullPointerException` during dexing

```
ERROR:/Users/.../opentelemetry-exporter-sender-okhttp-1.57.0.jar: D8: java.lang.NullPointerException
ERROR:/Users/.../opentelemetry-api-1.57.0.jar: D8: java.lang.NullPointerException
[... many more similar errors]
```

**Analysis:** D8 in AGP 7.4 fails to process multiple OpenTelemetry artifacts. This appears to be a tooling limitation rather than a dependency issue.

**Recommendation:** **Bump minimum AGP to 8.0.2** instead of continuing to downgrade dependencies.

#### 4. Kotlin stdlib version conflict (Critical)

After bumping to AGP 8.0.2, a new issue appears:

**Error:** Kotlin metadata version incompatibility

```
e: Module was compiled with an incompatible version of Kotlin. 
   The binary version of its metadata is 2.2.0, expected version is 2.0.0.
```

**Root cause:** A transitive dependency is exposing `kotlin-stdlib:2.2.21`

**Investigation:** Running `./gradlew :app:dependencyInsight --dependency kotlin-stdlib --configuration releaseRuntimeClasspath` reveals:

```
org.jetbrains.kotlin:kotlin-stdlib:2.2.21
+--- com.squareup.okhttp3:okhttp-android:5.3.2
     \--- com.squareup.okhttp3:okhttp:5.3.2
          \--- io.opentelemetry:opentelemetry-exporter-sender-okhttp:1.57.0
               +--- io.opentelemetry.instrumentation:opentelemetry-instrumentation-bom-alpha:2.23.0-alpha
```

**The culprit:** `okhttp3-android:5.3.2` (brought in by `opentelemetry-instrumentation-bom-alpha:2.23.0-alpha`)

**Verification:** Downgrading to `opentelemetry-instrumentation-bom-alpha:2.17.1-alpha` (last version using OkHttp 4) allows the build to succeed.

### Proposed solutions

We have two options:

1. **Bump minimum Kotlin requirement to 2.2** in VERSIONING.md
2. **Downgrade OkHttp dependency** in `opentelemetry-instrumentation-bom-alpha` to OkHttp 4 (if possible)

Additionally, we should:
- Update VERSIONING.md to reflect minimum AGP 8.0.2 (AGP 7.4 is not viable)
- Consider using `coreLibrariesVersion` Kotlin compiler extension to prevent future stdlib conflicts
- Define minimum `compileSdk` version in VERSIONING.md (currently not specified). 34 may be a good starting point, as it works with AGP 8.0.2.

### Questions

1. Is it acceptable to bump the minimum Kotlin version to 2.2? I think not, it's too new.
2. Can we downgrade the OkHttp dependency in the instrumentation BOM?
3. Should we update our minimum AGP requirement to 8.0.2?

priettt:
Resolves #1526

## What this does

VERSIONING.md promises consumers Kotlin 2.0, AGP 7.4 and Gradle 7.5, but nothing checked that, and #1526 found that it does not hold. This PR adds a CI check for the documented minimums and corrects them to values that build.

### 1. Integration test (first commit)

Ports the `gradle-integration-test` harness from opentelemetry-kotlin. Snapshot artifacts are published to Maven Local, then Gradle TestKit builds real consumer apps against them with the minimum Gradle, AGP, Kotlin and `compileSdk` from `gradle/libs.versions.toml`.

Two fixtures:

- `consumer-app`: `android-agent` plus every instrumentation library, resolved through the BOM, with core library desugaring. Covers resolution, AAR metadata, compilation and dexing.
- `weaving-app`: applies the Byte Buddy Gradle plugin with the `-agent` artifacts, since weaving runs inside the consumer's own build.

It runs as a new `min-supported-versions` job on every pull request. It is a separate `minSupportedVersionsTest` task rather than `test`, so `./gradlew check` does not need published artifacts. The module is excluded from the Kover merge for the same reason.

Both fixtures fail at the currently documented minimums, which reproduces #1526.

### 2. Corrected minimums (second commit)

| | documented | now | set by |
|---|---|---|---|
| Kotlin | 2.0 | 2.2 | `kotlin-stdlib` 2.3.21, which consumers resolve through `opentelemetry-extension-kotlin` (a dependency of `coroutines-library`) |
| AGP | 7.4 | 8.13 | lowest version that accepts `compileSdk` 37 |
| Gradle | 7.5 | 8.13 | lowest version AGP 8.13 accepts |
| `compileSdk` | not stated | 37 | OkHttp 5.5.0, and `androidx.core` 1.19.0 in the released 1.7.0 |
| JDK (build time) | 11 | 17 | AGP 8.x |

Two changes keep these as low as they are:

- **`androidx.core` goes back from 1.19.0 to 1.18.0.** 1.19.0 requires consumers to use AGP 9.1.0. That update (#1788) shipped in 1.7.0, so 1.7.0 already requires AGP 9.1. This restores AGP 8, and a Renovate rule holds `androidx.core` below 1.19.0 for the published libraries.
- **`coreLibrariesVersion` is pinned to 2.0.21**, as opentelemetry-kotlin does. Our POMs used to declare `kotlin-stdlib` 2.4.10, the version of the compiler that builds them, which on its own forced consumers onto Kotlin 2.3. We already compile with `apiVersion` and `languageVersion` 2.0.

## Why not the values proposed in #1526

#1526 proposed AGP 8.0.2, `compileSdk` 34 and keeping Kotlin 2.0. Upstream dependencies have moved since January, and each of those is now blocked by AAR metadata or Kotlin metadata that we do not control:

| dependency | min AGP | `minCompileSdk` | `kotlin-stdlib` |
|---|---|---|---|
| `androidx.core` 1.18.0 | 8.9.1 | 36 | |
| `androidx.lifecycle` 2.11, `savedstate` 1.4, `tracing` 2.0 | 8.1.1 | 34 | |
| `okhttp-android` 5.4.0 (from `opentelemetry-exporter-sender-okhttp` 1.65.0) | | 36 | 2.2.21 |
| `opentelemetry-extension-kotlin` 1.65.0 | | | 2.3.21 |

- **Kotlin 2.0** is out of reach without overriding what the OTel Java artifacts request.
- **AGP 8.0.2** would need `androidx.core`, `lifecycle`, `savedstate` and `tracing` downgraded, which the discussion in #1526 wanted to avoid.
- **`compileSdk` 36** would pass today, but only while the OTel exporter stays on OkHttp 5.4.0. 37 is what 1.7.0 already requires and what the okhttp3 instrumentation is built against.

## Note on the versioning policy

VERSIONING.md says minimum versions are bumped in a major release. I don't think this is a bump in that sense: the documented values never built, and 1.7.0 already requires AGP 9.1 and `compileSdk` 37. Nobody who can build today loses support, and AGP 8 users regain it.


gradle-integration-test/build.gradle.kts
```diff
@@ -0,0 +1,74 @@
...
+                        ?.trim()
+                        .orEmpty()
+                },
+        ).orElse("")
+
+tasks.test {
+    enabled = false
+}
+
+tasks.register<Test>("minSupportedVersionsTest") {
```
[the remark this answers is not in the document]
priettt: This looks like a false positive

gradle-integration-test/src/test/resources/fixtures/weaving-app/build.gradle.kts
```diff
@@ -0,0 +1,69 @@
...
+    implementation("io.opentelemetry.android.instrumentation:coroutines-library")
+    implementation(libs.okhttp)
+    implementation(libs.kotlinx.coroutines)
+
+    byteBuddy(platform("io.opentelemetry.android:opentelemetry-android-bom:${pins.getValue("bomVersion")}"))
+    byteBuddy("io.opentelemetry.android.instrumentation:okhttp3-agent")
+    byteBuddy("io.opentelemetry.android.instrumentation:okhttp3-websocket-agent")
+    byteBuddy("io.opentelemetry.android.instrumentation:httpurlconnection-agent")
+    byteBuddy("io.opentelemetry.android.instrumentation:android-log-agent")
+    byteBuddy("io.opentelemetry.android.instrumentation:coroutines-agent")
```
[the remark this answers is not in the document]
priettt: Good catch, fixed

gradle.properties
```diff
@@ -22,7 +22,7 @@ org.gradle.configuration-cache.parallel=true
 android.useAndroidX=true
 
 android.minSdk=23
-android.minCompileSdk=34
+android.minCompileSdk=37
```
aranhave, line 25: Could we give the fixture its own SDK 37 setting? This property raises the AAR minimum for every module, including `session`, whose released minimum is 34 and which doesn't depend on OkHttp or AndroidX Core. That newly breaks standalone consumers on `compileSdk` 34-36.

VERSIONING.md
```diff
@@ -25,13 +25,17 @@ builds, but their use in production is strongly discouraged.
 
 The android-agent currently supports the following minimum versions:
 
-- Kotlin 2.0
+- Kotlin 2.2
 - API 23+ ([desugaring of the core library](https://developer.android.com/studio/write/java8-support#library-desugaring) required for API <26)
-- Android Gradle Plugin (AGP) 7.4 and Gradle 7.5
-- JDK 11 (build-time)
+- Android Gradle Plugin (AGP) 8.13 and Gradle 8.13
```
aranhave: Nit: could we say AGP `8.13.2` here to match the fixture? `8.13` suggests `8.13.0` is covered, but that version isn't tested.

gradle.properties
priettt: From the 34 published modules, 17 modules require compileSdk 36, and 1 requires compileSdk 37.

- **16 modules need nothing above compileSdk 34.** 8 have no external AARs at all (`session`, `agent-api`, `android-instrumentation`, `coroutines-library`, `coroutines-agent`, `httpurlconnection-library`, `httpurlconnection-agent`, `okhttp3-library`), and 8 only reach `semconv-android:0.8.0`, which declares minCompileSdk 34.
- **17 modules need compileSdk 36**, because they depend on `androidx.core:core:1.18.0`: `android-agent`, `core`, `common`, `services`, `android-log-agent` and 12 instrumentations (activity, anr, compose-click, compose-navigation, crash, fragment, native-crash, network, sessions, slowrendering, startup, view-click).
- **1 module needs compileSdk 37**: `okhttp3-agent`, through `okhttp-android:5.5.0`.

I think we should decide between:
- maintain a single minCompileSdk of 37
- maintain a different minCompileSdk for each module (37 for `okhttp3-agent`, 36 for the 17 listed above, 34 for the rest)

The first option would be simpler to maintain, though the second option would support more consumers.

fractalwrench: Google Play requires that apps use at least 36 so I think it's reasonable to pick that as the minCompileSdk: https://support.google.com/googleplay/android-developer/answer/11926878

I'd suggest downgrading okhttp to a version that supports 36, so that all modules are on the same version.

VERSIONING.md
```diff
@@ -25,13 +25,17 @@ builds, but their use in production is strongly discouraged.
 
 The android-agent currently supports the following minimum versions:
 
-- Kotlin 2.0
+- Kotlin 2.2
```
fractalwrench, line 28: IMO we should downgrade deps until the stated version support is achieved. We could do that in a separate PR if that helps.

gradle.properties
aranhave: I’m okay with 36 as the shared baseline and a compatible OkHttp version. Could we document it as a compatibility change? Standalone modules like `session` would still move up from 34.

PR #2078, opened 2026-09-21
fix(session): separate attribution from activity

### LINKED ISSUE #2049: Separate session ID attribution from user activity
## Problem

Reading the current session ID still refreshes Android's background inactivity timer.

On current `main` at [`feaf191f`](https://github.com/open-telemetry/opentelemetry-android/commit/feaf191fe130ee458cfb0ca13f827b3a23e01971):

- [`SessionManager.getSessionId()`](https://github.com/open-telemetry/opentelemetry-android/blob/feaf191fe130ee458cfb0ca13f827b3a23e01971/android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionManager.kt#L38-L64) calls `timeoutHandler.bump()` on every successful path;
- [`SessionIdTimeoutHandler.bump()`](https://github.com/open-telemetry/opentelemetry-android/blob/feaf191fe130ee458cfb0ca13f827b3a23e01971/android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionIdTimeoutHandler.kt#L61-L68) restarts the inactivity clock;
- the [span appender](https://github.com/open-telemetry/opentelemetry-android/blob/feaf191fe130ee458cfb0ca13f827b3a23e01971/core/src/main/java/io/opentelemetry/android/SessionIdSpanAppender.kt#L27-L32), [log appender](https://github.com/open-telemetry/opentelemetry-android/blob/feaf191fe130ee458cfb0ca13f827b3a23e01971/core/src/main/java/io/opentelemetry/android/internal/processors/SessionIdLogRecordAppender.kt#L23-L30), and [session sampler](https://github.com/open-telemetry/opentelemetry-android/blob/feaf191fe130ee458cfb0ca13f827b3a23e01971/core/src/main/java/io/opentelemetry/android/SessionIdRatioBasedSampler.kt#L28-L44) call that same method automatically;
- [native-crash context capture](https://github.com/open-telemetry/opentelemetry-android/blob/feaf191fe130ee458cfb0ca13f827b3a23e01971/instrumentation/native-crash/src/main/java/io/opentelemetry/android/instrumentation/nativecrash/NativeCrashInstrumentation.kt#L325-L332) also reads through the same provider.

The existing tests confirm both sides of the behavior: `SessionManagerTest` verifies that each `getSessionId()` call invokes `bump()`, and `SessionIdTimeoutHandlerTest` verifies that `bump()` restarts the background timeout. A local integration probe using the real manager and timeout handler advanced a background session by 14 minutes twice, reading the ID after each advance. The same ID remained active after 28 minutes, directly showing that periodic reads can prevent the 15-minute inactivity timeout.

This means background spans, logs, or other automatic telemetry can keep a session alive even when the user did nothing. Looking up an ID for attribution and recording meaningful activity are different operations and should not share this side effect.

## Desired behavior

- Automatic telemetry can obtain the applicable session ID without refreshing the user-activity timeout.
- Only behavior explicitly defined as user activity by [#910](https://github.com/open-telemetry/opentelemetry-android/issues/910) refreshes inactivity.
- Reading for attribution still follows the chosen rules for initial session creation, maximum lifetime, and an already-expired session.
- Concurrent reads and a real rotation cannot create duplicate sessions or observer events.
- Existing custom `SessionProvider` implementations remain compatible.

## API question

The `session` module was stabilized in [#1671](https://github.com/open-telemetry/opentelemetry-android/issues/1671). The September 3 SIG discussion did not settle a new public method and supported keeping `SessionProvider` as a small read-oriented contract while the manual-extension proposal in [#2037](https://github.com/open-telemetry/opentelemetry-android/pull/2037) is still under discussion.

This issue should therefore agree on how internal appenders and samplers perform a passive read before adding methods to `SessionProvider`. Possible locations include an internal agent contract or the extension mechanism from #2037. The issue should not assume that the methods currently proposed by [#2035](https://github.com/open-telemetry/opentelemetry-android/pull/2035) are the final public API.

One behavior also needs an explicit answer: if no valid session exists when passive telemetry arrives, should the read create or rotate a session, return no session ID, or use another defined fallback?

## Related work

- [#910](https://github.com/open-telemetry/opentelemetry-android/issues/910) defines meaningful activity and expiry behavior.
- [#2035](https://github.com/open-telemetry/opentelemetry-android/pull/2035) demonstrated the activity/attribution split, but currently combines the passive read, public API changes, manual activity, and click instrumentation.
- [#970](https://github.com/open-telemetry/opentelemetry-android/issues/970) covers applying session sampling to logs. This issue changes only how the sampler reads session state; it does not define sampling policy.

## Out of scope

- Choosing the public manual activity or reset API.
- Deciding which gestures or lifecycle events count as meaningful activity.
- Versioned persistence and multi-process ownership.
- Changing sampling ratios or which signals are sampled.

## Acceptance criteria

- [ ] Span, log, sampler, and native-crash attribution paths do not refresh inactivity merely by reading a session ID.
- [ ] Explicit activity still refreshes the configured inactivity window.
- [ ] The behavior for a missing or expired session during passive attribution is documented and tested.
- [ ] A background test proves that periodic automatic telemetry cannot indefinitely prevent expiry.
- [ ] A concurrency test proves that passive reads racing with expiry or rotation produce at most one transition and one observer notification sequence.
- [ ] Existing custom `SessionProvider` implementations remain source and binary compatible unless a separate API decision explicitly accepts a compatibility change.
- [ ] Tests exercise the real `SessionManager` and timeout handler, not only mocked provider calls.
- [ ] No sampling-policy behavior changes as part of this issue.

Both focused test runs and the local integration probe passed.

aranhave:
Fixes #2049.

Session lookup no longer refreshes background inactivity. It still creates a session on first access and rotates after expiry. Background spans, logs and dropped sampling attempts can no longer keep the session alive.

Activity recording stays internal and checks expiry before extending the timeout. It is not wired to activity sources yet. Activity-source wiring, public APIs and persistence stay outside this fix.

Added regression coverage for passive telemetry, explicit activity, foreground/background transitions and concurrent reads with ordered observer callbacks. Formatting, repository checks and API compatibility checks passed locally.


android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionManager.kt
```diff
@@ -23,44 +22,66 @@ internal class SessionManager(
...
-                notifyObserversOfSessionUpdate(currentSession, newSession)
-                newSession.id
-            } else {
-                // Another thread accessed this function prior to creating a new session. Use the
-                // current session.
+    private fun getSessionId(recordActivity: Boolean): String {
+        val currentSession = session
+        if (!recordActivity && !sessionHasExpired(currentSession) && session === currentSession) {
+            return currentSession.id
+        }
```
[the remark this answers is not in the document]
aranhave: The identity check already catches a rotation during the expiry check. A rotation after that can overlap this read, which returns the earlier snapshot. Locking wouldn’t prevent another rotation once the lock is released. I reran the existing rotation-race tests, and they pass.

android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionUserActivityRecorder.kt
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.android.agent.session
+
+/** Records user activity after checking whether the current session has expired. */
+internal fun interface SessionUserActivityRecorder {
```
fractalwrench: Perhaps something along the lines of `SessionUserInteractionRecorder` would work better for disambiguating from `android.app.Activity`?

android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionIdTimeoutHandler.kt
```diff
@@ -41,37 +30,37 @@ internal class SessionIdTimeoutHandler(
         sessionConfig.backgroundInactivityTimeout,
     )
 
+    @Synchronized
```
fractalwrench: I'd prefer an explicit lock backed by a private property, to make it more obvious that synchronization is happening and prevent outside observers from accidentally synchronizing on this object too

android-agent/src/main/kotlin/io/opentelemetry/android/agent/session/SessionIdTimeoutHandler.kt
```diff
@@ -41,37 +30,37 @@ internal class SessionIdTimeoutHandler(
...
-        if (state == State.FOREGROUND) {
-            return false
+        val current = state
+        if (current.expiredOnForeground) {
+            return true
         }
-        val elapsedTime = clock.nanoTime() - timeoutStartNanos
-        return elapsedTime >= sessionBackgroundInactivityTimeout.inWholeNanoseconds
+        return !current.foreground &&
+            clock.nanoTime() - current.timeoutStartNanos >= sessionBackgroundInactivityTimeout.inWholeNanoseconds
```
fractalwrench, line 58: We should check the clock is definitely monotonic as Android's wall clock can sleep when entering the background

aranhave: Checked the default clock: it uses `elapsedRealtimeNanos()`, which includes deep sleep. Added a Robolectric test confirming the session expires after sleeping in the background.

fractalwrench (approved):
LGTM

